PR #6, opened 2014-01-30
Fix bug 5914

edgarrodsil:


sevein:
Hey Edgar, I know that this is nitpicking but can you update that commit with a proper description on how you are trying to fix the problem? We try to use the following guidelines: http://git-scm.com/book/en/Distributed-Git-Contributing-to-a-Project#Commit-Guidelines

You could edit that commit using rebase --interactive or:
$ git reset --soft HEAD^ # After this, all your files go back to the staging area (review with git status -s)
$ git commit # The editor should show up and you can try maybe to write a proper description of what you're trying to do?

And you'll have to run git push -f to push changes, -f is necessary because it won't be a fost-farward merge with the remote branch because you just rewrote a commit!

Thank you in advance! :)

PR #7, opened 2014-01-31
Solve accessions concurrence in different culture, reference #5914

edgarrodsil:
Use the accession_counter for the context culture incrementing and saving
it. Add the culture code into identifier to verify that it never is the
same as another culture.


sevein:
Edgar, I'll look into this as soon as possible. Thank you so much.

sevein:
@edgarrodsil, this branch was automatically closed because we have removed our 2.x branch in favour of qa/2.1.x. We are still interested and we'll process your request soon!

PR #9, opened 2014-02-10
Add Authority description_identifier to search index and include it in the browsing - issue #6302

kehh:
We had a use case for including the authority record description_identifier in the search index and including it in the browse results. As this matches how the archival descriptions are indexed and browsed it was straightforward to implement. Corresponding feature request:  https://projects.artefactual.com/issues/6302


sevein:
Very useful! Thank you. See 9f58cd47330e090a5e895e8845eb360b338487c5.

PR #11, opened 2014-02-21
Treeview enhancements

somazx:
Preview of using the treeview from sidebar in a row across the top of the page. Work in Progress.

PR #12, opened 2014-02-27
Adding Browse Collection Button

somazx:
Previously you could only do this by doing a search and then choosing "Browse Collection" at the bottom - which was awkward and hidden from users.


sevein:
Closing this, see https://github.com/artefactual/atom/pull/11.

PR #13, opened 2014-02-28
add padding-right to site-name, refs #6292

timhutch:
This improves the spacing between the site name and the search box. It doesn't address the scenario of both site name and logo suppressed, but this seems less likely. Spacing already seems OK if there is a logo but no site name.


sevein:
Merged! See f4e525a89eb200cabd6bbff16bc88abdd7e44d0c.

PR #15, opened 2014-03-09
Autocomplete delay override

seamus-oconnor:
The autocomplete.js code now checks for a data-autocomplete-delay attribute on the input element to override the default 1 second delay.

### DIFF
--- a/apps/qubit/modules/informationobject/templates/editPhysicalObjectsSuccess.php
+++ b/apps/qubit/modules/informationobject/templates/editPhysicalObjectsSuccess.php
@@ -49,7 +49,7 @@
 
         <div class="form-item">
           <?php echo $form->containers->renderLabel() ?>
-          <?php echo $form->containers->render(array('class' => 'form-autocomplete')) ?>
+          <?php echo $form->containers->render(array('class' => 'form-autocomplete', 'data-autocomplete-delay' => 0.3)) ?>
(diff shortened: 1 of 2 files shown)


seamus-oconnor:
The '+' casts the string type to an int @sevein. It could still be useful to allow a per autocomplete box override as needed. The current 1s delay would help with server load but allowing an specific override could be beneficial in some cases.

sevein:
Nice trick. Can you be more explicit and use (int) casting?
Why do you want to decrease the delay for that specific field?

seamus-oconnor:
@sevein I will update the PR.

Reducing the delay was a requirement from my client. They found the delay to be hurting usability for that particular field.

sevein:
Oh, that's funny! When I asked you to cast it I was thinking really in the PHP (int) syntax. Thank you so much, will merge soon.

seamus-oconnor:
It has been about a month @sevein did you forget about this PR?

sevein:
Sorry for the delay. It's now in qa/2.1.x (old 2.x), fc05fe001f15d5be3a19eb6f478216b2fae51c91.

sevein:
@seamus-oconnor btw @fiver-watson is waiting for you to send him the [contributor agreement](https://www.qubit-toolkit.org/wiki/Contribute_code): ([pdf form](https://www.qubit-toolkit.org/accesstomemory-contributor-agreement-2014.pdf))

PR #18, opened 2014-04-13
add MODS import config, refs #4303

timhutch:
This should enable basic MODS import - focusing on roundtripping files created in AtoM (no elements added beyond existing template).

The exception is language of material - I couldn't figure out how to import language based on the name rather than code.


MikeFE:
Thanks Tim, merged.

PR #20, opened 2014-04-15
Fix link to external assets in showDownloadComponent, fixes #6548

mistydemeo:
This moves the construction of a public URL into an instance method on QubitDigitalObject, so consumers like showDownloadComponent don't need to actually care whether the target is local.

This fixes an issue that came up when displaying download components for components of an unknown type; the showDownloadComponent class is no longer aware of whether it's local, but it shouldn't need to care since the digital object itself retains that knowledge.


sevein:
Thanks!

PR #21, opened 2014-04-15
CSV import: add digital object columns, refs #6525

mistydemeo:
This adds two new columns to the CSV import: digital object URI, and digital object path. These two fields are mutually exclusive, and URI will be preferred over path if both are provided.

To maintain a 1:1 mapping between input rows and information objects, only a single digital object can be imported per item using this method.


sevein:
Thanks!

PR #29, opened 2014-06-02
QubitDigitalObject: make timeout configurable, refs #6549

mistydemeo:
This adds a new config parameter for the download timeout for external digital object imports; previously that was hardcoded to 10 seconds.


MikeFE:
Thanks Misty, I merged it in.

PR #37, opened 2014-07-14
TinyMCE integration for static pages

jasonhildebrand:
We are doing some consulting work on AtoM on behalf of a client.  We want to contribute general changes upstream.

One of the feature requests was integration of a wysiwyg editor to allow editing these pages with needing to know HTML (which this branch implements).  

I realize this has been rejected in the past (https://projects.artefactual.com/issues/2039) because it would allow the user to use a variety of (inconsistent) fonts and sizes.  However the current implementation does not allow free choice of fonts - it allows the user to use the header levels defined in the current CSS (see screenshot).

tinymce

Please let me know if I've missed anything.  I'm an experienced developer, but am new to AtoM and git.

### DIFF
--- a/apps/qubit/modules/search/actions/advancedAction.class.php
+++ b/apps/qubit/modules/search/actions/advancedAction.class.php
@@ -333,7 +333,7 @@ class SearchAdvancedAction extends DefaultBrowseAction
 
         case 'name':
           $queryField = new \Elastica\Query\QueryString($query);
-          $queryField->setDefaultField('names.i18n.'.$culture.'.name');
+          $queryField->setDefaultField('names.i18n.'.$culture.'.authorizedFormOfName');
           $queryField->setDefaultOperator('OR');
 
           break;
--- a/apps/qubit/modules/staticpage/templates/editSuccess.php
+++ b/apps/qubit/modules/staticpage/templates/editSuccess.php
@@ -1,3 +1,5 @@
+<?php use_javascript('//cdn.jsdelivr.net/tinymce/4.0.18/tinymce.min.js'); ?>
+<?php use_javascript('/js/staticpage_tinymce.js'); ?>
 <?php decorate_with('layout_1col.php') ?>
 
 <?php slot('title') ?>
--- a/js/staticpage_tinymce.js
+++ b/js/staticpage_tinymce.js
@@ -1,0 +1,27 @@
+(function ($)
+    {
+        Drupal.behaviors.staticpage_tinymce = {
+            attach: function (context)
+                {
+                    // make a list of all css files on the current page.
+                    // we need to tell tinymce about these so that it can 
+                    // load the same files within the editor (for better wysiwyg).
+                    var tinymce_css = '';
+                    jQuery('head').find('link').each(function() {
+                        var href = jQuery(this).attr('href')
+                        if (/.css$/.test(href)) {
+                            if (tinymce_css != '') {
+                                tinymce_css += ',' 
+                            }
+                            tinymce_css += href;
+                        }
+                    });
+                    tinymce.init({
+                          selector: 'textarea', 
+                          height: 250, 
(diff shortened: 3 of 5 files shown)


fiver-watson:
Hi Jason, 

Wow thanks for sharing this! It's true that our Director of Technical Services felt this would add unnecessary complexity to the app when #2039 was closed several years ago - however, both the application and our user community has grown significantly since then, and I know there is interest in our user community to see something like this. I will talk it over with our dev team, but personally I'd certainly like to see something like this incorporated into our public release. One thing we will have to test thoroughly: the DC and EAD XML exports. EAD includes some styling elements and in the future it could be really interesting to have our EAD export convert text styled via TinyMCE to EAD-styled tags, so that it can roundtrip. If I can gain internal interest for supporting this pull request in our public releases, then we will happily make a development wiki page in the hopes of interesting others in furthering this work. 

In the meantime, will you please sign and submit to us one of our contributor's agreements? The agreement and further information can be found here: https://www.qubit-toolkit.org/wiki/Contribute_code

You can sign as an individual, or on behalf of PeaceWorks if that is more practical - once you have signed a contributor's agreement, it will also cover any future code you may choose to contribute back to the project. Feel free to send it to me directly at dan (at) artefactual (dot) com. 

Thanks again! 

Dan Gillean
AtoM Product Manager
www.artefactual.com

fiver-watson:
Artefactual has had some discussions with Jason at PeaceWorks, the development company responsible for this pull request. At this time, we are not going to accept this pull request, as we feel it requires further development to meet our user needs and be easily maintained over time. Below is a copy of some of our thoughts on development. Ideally, in the future, someone (either PeaceWorks, Artefactual, or another community contributor) might implement at least the basic suggestions included below, so we can merge a revised pull request and share this with our community. Notes: 

Our Director of Technical Services feels strongly that this addition is something that should be packaged as a plugin, which can then be disabled by users who do not wish to use a WYSIWYG editor for their static pages. From an architecture perspective, this is also more consistent with how other modules and features in AtoM have been built and maintained.

As for the implementation itself, we feel this is a valuable tool and would definitely love to see this in AtoM - many of our users would benefit greatly from the ability to style pages via something like TinyMCE. However, we noted that there are some changes we would want to see to the editor's implementation for it to be supported in a public release, as the editor's implementation currently does not support the two most common uses - adding images, and links.

When testing internally, because the link button was not available, I noted that adding a link manually was rendered buggy - If I added an html < a > link, when I saved the page the first time, the plain code would be shown and no link would be present. When re-entering edit mode, it was then rendered as a link in TinyMCE's editor - but there was no option to edit the link, meaning the user would have to delete it, create a new one manually, save the page (and have it show up wrong again), re-enter edit mode, and then save once more. I did not test adding images via HTML tags, but I expect that the behavior might be similar.

Since TinyMCE does include edit button options for adding both links and images, we would like to see these supported for this to be truly useful to a broad section of our users. For more advanced users who are comfortable with HTML and CSS, the option to view and edit the source directly (something I saw you can do on the TinyMCE homepage) would also be included in our ideal implementation.

Since the current implementation actually makes it harder to manage links, we consider the above points blocking issues for us to accept the pull request as-is, unfortunately.

There are further improvements we would also like to aim for. Ideally, the insert video option (or at least the embed part of it) would be included as well. If it were possible, we might suggest removing the File menu, as the "New document" option will inadvertently wipe the current page - while either Ctrl+Z or using TinyMCE's back button will undo this, we anticipate a good degree of user error arising from this (though good documentation might be a fine way to get around this, if removing the option were challenging, or impractical for a reason I haven't considered).

Finally, our ideal solution would be to see a separate stylesheet implemented for static pages, which TinyMCE would use - so that the user would not have to employ the default CSS classes used throughout the application. This is something that clearly doesn't exist currently, but would improve the app greatly - and developers could then customize those CSS classes without affecting the overall theme, if desired.

PR #40, opened 2014-07-23
Added CLI arg for site URL to EAD bulk export.

mcantelon:
CLI tasks have no knowledge of hostname, etc., so can't automatically
generate a full base URL for exported XML. Added a CLI arg to the
EAD bulk export tool to specify a URL. Integrated this into the existing
EAD template.


MikeFE:
Merged into AtoM, thanks Mike.

fb98a5c770b7076e328261d3f5d0121f9736d91c

I also made the base-url optional instead of mandatory:

45bff5eae33b67384bb3aed956dd94b9c6e0a034

PR #41, opened 2014-08-08
Remove executable bit from normal files

gioele:
Normal PHP files with class definitions don't need to be executable.


sevein:
Good catch! Thanks. I'm surprised there were only five of them :)

PR #42, opened 2014-08-22
#7122 - Add links to other cultures in info obj templates

jraddaoui:
Added in index pages of ISAD, RAD, MODS, DC and DACS plugins

Fixes #7122

### DIFF
--- a/apps/qubit/modules/informationobject/actions/indexAction.class.php
+++ b/apps/qubit/modules/informationobject/actions/indexAction.class.php
@@ -130,16 +130,5 @@ class InformationObjectIndexAction extends sfAction
         $this->digitalObjectLink = $request->getUriPrefix().$request->getRelativeUrlRoot().$this->resource->digitalObjects[0]->getFullPath();
       }
     }
-
-    // Get other cultures available
-    $currentCulture = $this->getUser()->getCulture();
-    $this->translations = array();
-    foreach ($this->resource->informationObjectI18ns as $i18n)
-    {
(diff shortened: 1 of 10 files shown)


sevein:
I would suggest to implement translationsLinks as a component and let its controller do things like populate the list of languages available for a given resource, return sfView::NONE if needed, etc...

Please use .translations-links instead of div#translations-links. It's faster as a selector (specially when used as a namespace).

I'm not sure about the label "Other languages available", can you check that with Dan? Also, not sure if you should call it .translation-links instead of .translations-links, and the same thing with the partial/component. But I'm definitely not the right person to talk about grammar, please check with Dan!

Can you include a screenshot so we can see how the new widget looks?

fiver-watson:
I can always check with Sarah, but I think the "Other languages available" label is fine - if the user is looking at a translation (and not the original culture description), they should still see this option to view other cultures, so saying "Translations available" might not be as accurate? @djjuhasz thoughts?

.translation-links sounds better than .translations-links . Ditto: translationLinks = better than translationsLinks, IMHO. 

I'd love to see a screenshot too!

jraddaoui:
I've made the changes and here is a screenshot.

translation-links

If the title is not populated in one of the other languages it will show the source culture title.

fiver-watson:
Nice, Radda, looks good. Have you made sure that "Other languages available" is a translatable string? e.g. when we push the update, will it be translatable in Transifex? Thanks!

PR #49, opened 2014-09-09
Make translation links a dropdown, refs #7122

jraddaoui:
And show language instead of the ISO code in the list


sevein:
See https://projects.artefactual.com/issues/7122#note-11.

PR #52, opened 2014-09-18
get MODS repository name correctly, refs #6969

timhutch:
When there is an unencoded ampersand in the repository name, the MODS XML file is invalid (can't be parsed in browser). This is the same fix as in https://github.com/artefactual/atom/commit/4a0a18501c43b1fa43738c65a1f17a8cca86009a


sevein:
I was hoping to merge this into qa/2.1.x but it looks like you picked a dev branch, was that on purpose? I also tried to cherry-pick 11938bd but there is a conflict.

timhutch:
Yes - this came up as part of QA for the current U of S contract focusing on bulk export, so I used the dev branch in which I'm testing. In the same section of code I believe the order of the repository elements was switched. David will be able to clarify further.

PR #54, opened 2014-09-25
#7329 - Do not output spaces around subfields' titles

gioele:
Having spaces inside the `<h3>` element creates artifacts in combination
with CSS rules like `h3:after { content: ...; }`.

BTW, for the sake of i18n, the whole sentence (text + colon) should be
generated, not just the field name.

PS: the patch can be backported without any changes to `stable/2.1.x` and `qa/2.1.x`.

### DIFF
--- a/apps/qubit/modules/informationobject/templates/_creatorDetail.php
+++ b/apps/qubit/modules/informationobject/templates/_creatorDetail.php
@@ -15,13 +15,13 @@
 
       <?php if (0 < count($resource->getCreators())): ?>
         <div class="field">
-          <h3>
-            <?php if (QubitTerm::CORPORATE_BODY_ID == $item->entityTypeId): ?>
-              <?php echo __('Administrative history') ?>
-            <?php else: ?>
-              <?php echo __('Biographical history') ?>
-            <?php endif; ?>
-          </h3><div>
+          <?php if (QubitTerm::CORPORATE_BODY_ID == $item->entityTypeId): ?>
+            <?php $history_kind = __('Administrative history'); ?>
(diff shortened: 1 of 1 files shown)


jraddaoui:
Hi Gioele, thanks for the PR!

I've tried to add your fix but it doesn't seem to be working. It looks like it was more a problem with the header alignment. I've added the folowing fix to our qa/2.2.x branch:

https://github.com/artefactual/atom/commit/977a4461929d76444b9b19ad4d5d25a62ebc9a26

Please, let us know if that works for you.

Regards.

gioele:
@jraddaoui the fix you linked does not seem to be related to this PR.

The problem this PR fixes is that the presence of spaces and newlines in H3 elements, combined with the `:after { content: ':'; }` CSS rules, disables the HTML whitespace normalization process.

With this fix every browser shows

```
Administrative history: foo
```

while without this fix Firefox (correctly) shows

```
Administrative history
:
foo
```

Saving the text and outputting it in one fell swoop removes this aesthetic problem in every browser.

jraddaoui:
@gioele, looking at the screenshot in the related Redmine ticket I thought it was a different problem.

Still, I haven't been able to reproduce that exact behaviour using the latest versions of Chrome, Firefox and IE. I could only see the whitespace before the two points instead of after:

```
Administrative history :foo
```

With this fix the two points are now in the correct place:

```
Administrative history: foo
```

I've fixed the conflicts and added your fix to the qa/2.2.x branch. Sorry for the confusion, but in the end we fixed two things ;)

Thanks!

PR #58, opened 2014-09-30
#7328 - Update EAD mappings

mistydemeo:
- Find `<origination>` values when `<origination>` is not parented inside the `<did>` element
- Map `<publicationstmt><date>` to revisionHistory. This field describes the date of publication of the _description_, but it was being mapped to a field that actually refers to the content being described.

The latter change orphaned the `setPublicationDate` method on `QubitInformationObject`, so it was removed.

### DIFF
--- a/apps/qubit/modules/object/config/import/ead.yml
+++ b/apps/qubit/modules/object/config/import/ead.yml
@@ -59,22 +59,22 @@ information_object:
       Method:  setRepositoryByName
 
     creator:
-      XPath:  "did/origination/name"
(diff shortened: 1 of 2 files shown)


sevein:
We need to file a ticket for the public project in Redmine.

mistydemeo:
All right, will do.

mistydemeo:
Filed https://projects.artefactual.com/issues/7328 with summaries of the issues, and updated the commits to reference that issue.

PR #59, opened 2014-10-10
#5094 - Recreate RAD events from EAD on import

mistydemeo:
This updates the EAD import process to correctly recreate RAD events.

RAD events can have both places and actors associated with them. These places and actors are associated with the _event_ and not with the information object itself, which affects how they're displayed in the user interface. AtoM currently doesn't support exporting these to EAD or reimporting them. Prior to 61fb9c63f700a3efd5eccea2557ef09c0155431d, they weren't exported in EAD at all; now they are, with their role intact, but they will be imported as standard authorities instead of in association with their events.

This adds a new `id` attribute to events and to the places and actors associated with those events on export to EAD. On import, any actor or place with an ID will not have a relationship built to the information object and will instead be associated to its associated event.

This required a few extra bits of plumbing to support:
- `QubitInformationObject::setActorByName` and `::setTermRelationByName` now support the new `createRelation` parameter in the `$options` array. If set to false, it overrides the usual logic determining whether a relationship should be created and creates no relationship.
- `QubitInformationObject::setActorByName`, `::setTermRelationByName` and `::setDates`, which previously returned void, now return the created actor or term object if one is created. This shouldn't affect existing code.

### DIFF
--- a/apps/qubit/modules/object/config/import/ead.yml
+++ b/apps/qubit/modules/object/config/import/ead.yml
@@ -350,27 +350,27 @@ information_object:
     controlaccess_name:
       XPath:  "(controlaccess/name | controlaccess/controlaccess/name)"
       Method: setActorByName
-      Parameters: [$nodeValue, "$options = array('relation_type_id' => QubitTerm::NAME_ACCESS_POINT_ID)"]
+      Parameters: [$nodeValue, "$options = array('relation_type_id' => QubitTerm::NAME_ACCESS_POINT_ID, 'createRelation' => !$importDOM->xpath->query('@id', $domNode2)->length)"]
 
     controlaccess_corpname:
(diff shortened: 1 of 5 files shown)


mistydemeo:
Note that RAD event _notes_ are not preserved in export and reimport; per @fiver-watson's suggestion, they are simply omitted.

sevein:
LGTM. @MikeFE, can you take a look to this?

MikeFE:
Looks good. I guess we'll discuss again once we double check we can use ID like that

mistydemeo:
Updated - ID tags are now in the format of `NNN_type`, e.g. `384_place`, which ensures that they're unique across the document.

MikeFE:
All merged, thanks

PR #65, opened 2014-10-29
Fixed issues with site base URL config.

mcantelon:
Added explanation for base URL setting field on site configuration
and site settings pages. Fixed issue with base URL saving during
installation. Added notification is site base URL not set.


mcantelon:
Merged.

PR #66, opened 2014-10-30
Hide elements in print view, refs #7445

jraddaoui:
Adds print.less file in arDominionPlugin less folder
Used in Dominion and ArchivesCanada themes
Adds the css/less file with media=all in the header to allow print rules


MikeFE:
Hey, looks good to me

PR #69, opened 2014-11-13
Dev/issue 7472 pageview counter slowness

mcantelon:
Added SQL pager class and used it, with Dave' query, to speed up repository description view stats page.

### DIFF
--- a/lib/QubitPdoPager.class.php
+++ b/lib/QubitPdoPager.class.php
@@ -1,0 +1,95 @@
+<?php
+
+/*
+ * This file is part of the Access to Memory (AtoM) software.
+ *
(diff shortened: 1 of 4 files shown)


lib/QubitPdoPager.class.php
```diff
@@ -0,0 +1,97 @@
...
+  }
+
+  /**
+   * @see sfPager
+   */
+  public function getResults()
+  {
+    $query = new QubitPdo;
+
+    $this->setNbResults($query->fetchColumn($this->countSql, $this->countSqlParams));
```
MikeFE: Should probably use QubitPdo::fetchColumn() and QubitPdo::fetchAll() here instead since they're static methods

PR #71, opened 2014-11-18
Create terms for certain columns if they are missing upon CSV import

MikeFE:
For the columns radGeneralMaterialDesignation and descriptionStatus, create terms on CSV import if they aren't already in the taxonomy in AtoM

### DIFF
--- a/lib/task/import/csvImportTask.class.php
+++ b/lib/task/import/csvImportTask.class.php
@@ -463,10 +463,8 @@ EOF;
         if (isset($self->rowStatusVars['descriptionStatus'])
           && 0 < strlen($self->rowStatusVars['descriptionStatus']))
         {
-          $statusTermId = array_search(
-            trim($self->rowStatusVars['descriptionStatus']),
-            $self->status['descriptionStatusTypes']
-          );
+          $descStatus = trim($self->rowStatusVars['descriptionStatus']);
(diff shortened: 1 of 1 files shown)


lib/task/import/csvImportTask.class.php
```diff
@@ -474,9 +472,17 @@ public function execute($arguments = array(), $options = array())
           }
           else
           {
-            printf("Row %s: Invalid description status '%s', using null instead\n",
-              $self->getStatus('rows') + 1,
-              trim($self->rowStatusVars['descriptionStatus']));
+            print "\nTerm $descStatus not found in description status taxonomy, creating it...\n";
+
+            $newTerm = new QubitTerm;
```
mcantelon: Looks good in general!

For the term creation you could use the helper in QubitFlatfileImport:

  $culture = isset($self->object->culture) ? $self->object->culture : 'en';
  QubitFlatfileImport::createTerm($QubitTaxonomy::DESCRIPTION_STATUS_ID, $descStatus, $culture);

(There's another term creation later in the code this could work for too. )

MikeFE: Ah good call... I actually think that later point in the code was what I was going off of so I missed the helper function :)

MikeFE:
Merged

PR #72, opened 2014-11-21
EAD import: fix importing records with non-event ID tags

mistydemeo:
- Prefix atom_ to every event ID tag, to distinguish from non-AtoM tags
- Check suffix of the tag to make sure it's the correct event type, to avoid catching other kinds of ID tags

Fixes #7498.


sevein:
@MikeFE ^

MikeFE:
It looks good to me Misty, go ahead and merge. Thanks!

mistydemeo:
Merged!

PR #73, opened 2014-11-21
#7443 - EAD...?

mcantelon:
Some EAD fixes.

### DIFF
--- a/lib/QubitXmlImport.class.php
+++ b/lib/QubitXmlImport.class.php
@@ -217,7 +217,7 @@ class QubitXmlImport
         $langCodeConvertor = new fbISO639_Map;
         foreach ($langusage as $language)
         {
-          $isocode = trim(preg_replace('/[\n\r\s]+/', ' ', $language->nodeValue));
+          $isocode = trim(preg_replace('[\n\r\s]+', ' ', $language->nodeValue));
           // convert to Symfony culture code
           if (!$twoCharCode = strtolower($langCodeConvertor->getID1($isocode, false)))
           {
@@ -574,7 +574,7 @@ class QubitXmlImport
 
               // If an actor/event object was returned, track that
               // in the events cache for later cleanup
-              if($obj !== NULL)
+              if(!empty($obj))
               {
                 $this->trackEvent($obj, $domNode2);
               }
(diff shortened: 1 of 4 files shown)


lib/QubitXmlImport.class.php
```diff
@@ -855,12 +855,12 @@ private function trackEvent($object, $node)
     }
 
     $id = $node->getAttribute('id');
-    if ($id !== NULL)
+    if (!empty($id))
```
mistydemeo, line 858: Is this right? I believe `NULL` is actually the return value of `getAttribute()` if the requested attribute isn't present.

lib/QubitXmlImport.class.php
```diff
@@ -855,12 +855,12 @@ private function trackEvent($object, $node)
...
     {
       // The ID value is suffixed with its category, e.g. 384_place
       // This is because `id` is required to be unique within the entire
       // document.
-      $id = preg_replace('_' . $key . '$', '', $id);
+      $id = preg_replace('/_' . $key . '$/', '', $id);
```
mistydemeo, line 863: You can remove this bit - I reworked this in #72, and it's pretty different now!

plugins/sfEadPlugin/modules/sfEadPlugin/templates/indexSuccessBodyDidElement.xml.php
```diff
@@ -88,7 +88,7 @@
...
+    <unitdate <?php if ($date->getActor() !== NULL || $date->getPlace() !== NULL): ?> <?php echo 'id="'.$date->id.'_event"' ?> <?php endif; ?> <?php if ($date->typeId != QubitTerm::CREATION_ID): ?><?php if ($type = $date->getType()->__toString()): ?><?php echo 'datechar="'.strtolower($type).'" ' ?><?php endif; ?><?php endif; ?><?php if ($startdate = $date->getStartDate()): ?><?php echo 'normal="'?><?php echo Qubit::renderDate($startdate) ?><?php if (0 < strlen($enddate = $date->getEndDate())): ?><?php echo '/'?><?php echo Qubit::renderDate($enddate) ?><?php endif; ?><?php echo '"' ?><?php endif; ?> encodinganalog="<?php echo $ead->getMetadataParameter('unitdate') ?>"><?php echo escape_dc(esc_specialchars(Qubit::renderDateStartEnd($date->getDate(array('cultureFallback' => true)), $date->startDate, $date->endDate))) ?></unitdate>
```
mistydemeo: I think this clashes with a change made in #72.

lib/QubitXmlImport.class.php
mistydemeo, line 858: Ah nope, I'm wrong - it _is_ an empty string!! Good catch.

mistydemeo:
I've merged #72, so this should get rebased on top.

plugins/sfEadPlugin/modules/sfEadPlugin/templates/indexSuccessBodyDidElement.xml.php
mcantelon: Cool... I'll remove it.

mcantelon:
I think I merged this.

PR #75, opened 2014-11-26
Add top-level description filter, refs #7570

MikeFE:
The user can now use a radio button to specify whether to
show top level archival descriptions when searching/browsing,
or to display all descriptions.

When the user is browsing archival descriptions, the default
setting is now to display top level descriptions only (as it
was in ICA-AtoM).

### DIFF
--- a/apps/qubit/modules/informationobject/actions/browseAction.class.php
+++ b/apps/qubit/modules/informationobject/actions/browseAction.class.php
@@ -132,6 +132,36 @@ class InformationObjectBrowseAction extends DefaultBrowseAction
     }
   }
 
+  /**
+   * Determine url parameters based on if the "top-level descriptions" or "all descriptions"
+   * radio buttons are selected. Also determine which radio button is 'checked'.
+   *
+   * Filter out non-top level descriptions from the ES query if the user has selected
+   * "top-level descriptions."
+   */
+  private function handleTopLevelDescriptionsOnlyFilter()
+  {
+    $this->topLvlDescUrl = $this->context->routing->generate(null, array('topLod' => true) +
+                           $this->request->getParameterHolder()->getAll());
+
+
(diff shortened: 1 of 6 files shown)


sevein:
I think that this could be more elegant if you implement a new self-contained facet component just for the purpose of filtering top-level descriptions. That component could have its own action class with the logic of `handleTopLevelDescriptionsOnlyFilter`, etc... plus you don't have to pass request parameters around since from component you can access to the request object via the sfContext singleton.

_facet.php is a generic partial for regular term facets, while your case is very particular. For example, we created _datesFacet or _facetLanguage (funny naming) instead of crowding _facet.php. I guess what I'm asking is to create a new _facet***_.php and, ideally, converted into a component so in actions/ you can have a facet**_*Component.class.php with some logic in there.

One defect of our current facets is that at the end the query filtering is done in apps/qubit/modules/default/actions/browseAction.class.php. I'd love to see trying to do the filtering from the component. Please see http://symfony.com/legacy/doc/gentle-introduction/1_4/en/07-Inside-the-View-Layer#chapter_07_sub_components

MikeFE:
Hey, I considered having a specialized _facet***.php for the level of description facet, and even started to code it that way. However, after a certain point I realized that the only difference between _facet.php and _facetLevelOfDescription.php would be that one if statement (https://github.com/artefactual/atom/commit/becab72662293a2657d56a8248e4f4df56c25d79#diff-35e05e7eb7cf58be318c2596fd1bcb16R15).

Given there's already a really big problem with code duplication when it comes to these areas of code (search vs. browse), I felt inserting the 12 line if statement for _facet.php was more maintainable than duplicating lines of _facet code across multiple files, even if it isn't very elegant. Maybe there's a way to have the best of both worlds?

Please let me know if I'm just misunderstanding your suggestion.

PR #78, opened 2014-12-05
Add 'Part of' info to search/browse results #7651

MikeFE:
Results for browse and search will now shot "Part of: <top level description title" next to them if they aren't top levels of description. The Part of: title should also link to the parent's archival description.

### DIFF
--- a/apps/qubit/modules/informationobject/templates/browseSuccess.php
+++ b/apps/qubit/modules/informationobject/templates/browseSuccess.php
@@ -158,7 +158,7 @@
 
 <?php foreach ($pager->getResults() as $hit): ?>
   <?php $doc = $hit->getData() ?>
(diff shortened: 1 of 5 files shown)


jraddaoui:
Hi Mike, the ES index part looks great, but I don't think that we need a component for this matter. Also, it's assuming that the partOf title is going to be in English.

You could do something like this directly in the partial:

<?php if (isset($doc['partOf'])): ?>
  <p><?php echo __('Part of '), link_to(render_title(get_search_i18n($doc['partOf'], 'title', array('allowEmpty' => false, 'culture' => $culture))), array('slug' => $doc['partOf']['slug'], 'module' => 'informationobject')) ?></p>
<?php endif; ?>

MikeFE:
Thanks Radda, I totally missed that 'en' issue!

PR #79, opened 2014-12-06
#7469 - OAI-PMH API key authentication and embedding of digital objects

mcantelon:
Added OAI-PMH API key authentication and embedding of digital objects, as Atom XML, in record XML.


sevein:
Nicely done! My only rant is that the oai_enabled QubitSetting makes the code harder to read. Is the setting object null? Then, we change the type of value to bool and... is true? is false? I'm not sure if you have any time left for this, but isPluginEnabled() would be a nice thing to see!

mcantelon:
Merged.

PR #80, opened 2014-12-06
#7658 - Fix DC template HTML escaping

mcantelon:
Explicity cast terms as strings because when term objects get returned, you can echo them and they'll print as a string but esc_specialcharacters won't actually escape the special characters because it checks if the data its given is a string.


sevein:
Whoa, this is related to #76 where `$item` will be a special escaping decorator. I think that this is fine and you should go ahead and merge. Just adding this as a reminder.

mcantelon:
Merged!

PR #81, opened 2014-12-07
Default values for CLI add superuser. refs #7666

mcantelon:
On some AtoM installs (taking client data and importing additional data) I need to quickly add a superuser whenever I redo the install.

Added "--demo" option to "tools:add-superuser" task to set options so user is "demo", email is "demo@example.com", and password is "demo".


MikeFE:
Right you are. Looks good to me in that case

PR #82, opened 2014-12-10
Changes in setActorByName, refs #7327

jraddaoui:
- Stop saving IO, relations and events inside the function
- Chech populated events and relations first
- Check relation and events for other IOs in the same repository after

### DIFF
--- a/lib/model/QubitInformationObject.php
+++ b/lib/model/QubitInformationObject.php
@@ -1323,23 +1323,83 @@ class QubitInformationObject extends BaseInformationObject
 
   public function setActorByName($name, $options)
   {
-    // Only create an link actor if the event or relation type is indicated
+    // Only create and link actor if the event or relation type is indicated
     if (!isset($options['event_type_id']) && !isset($options['relation_type_id']))
     {
       return;
     }
 
-    // Information object must be saved before.
-    // The id is needed to save the events and relations when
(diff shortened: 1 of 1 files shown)


MikeFE:
Hey Radda, thanks so much for finding a solution to this! I ran into another bug related to it in a migration I'm doing just yesterday, so this will help me a lot.

I had some suggestions for the PR, but it seemed kind of overkill to put them all in inline comments, so I just made a gist with what I think should change: https://gist.github.com/MikeFE/bfd705298ecf0ac51ffe

Summary of changes I recommend:
- The function is getting too bloated and hard to understand, so I broke out a couple of private helper functions to make things more simple (also we can avoid break 2;)
- I refactored the checking of object/subject relations into a single function to reduce code duplication
- I improved the comment documentation for the function, since this function's scope and functionality have expanded quite a bit beyond what someone might initially expect 'setActorByName' to do. Now it should be clear to anyone reviewing the code what the function is up to.
- Couple other small tweaks to reduce complexity in the code

Let me know what you think. And again, thanks for addressing this--it came at a very good time.

MikeFE:
PS: I didn't test this gist code at all, so there may very well be some minor mistakes in it.

jraddaoui:
Awesome Mike, a lot more cleaner and clearer.

Thanks!

PR #83, opened 2014-12-11
Fixes in EAD bioghist import/export, refs #7124 (updates #7 and #8)

jraddaoui:
- Don't add empty bioghist elements in EAD export
- Don't create new 'Untitled' actors if bioghist value match
  the history value of the creator in the same position
- Improve comments showing different options on import

### DIFF
--- a/lib/model/QubitInformationObject.php
+++ b/lib/model/QubitInformationObject.php
@@ -1676,6 +1676,7 @@ class QubitInformationObject extends BaseInformationObject
    * Import creation-related data from an <bioghist> tag in EAD2002
    *
    * @param $biogHistNode  DOMNode  EAD bioghist DOM node
+   * @param $key Position of the current bioghist node
    */
   public function importBioghistEadData($biogHistNode, $key)
   {
--- a/lib/model/QubitInformationObject.php
+++ b/lib/model/QubitInformationObject.php
@@ -1800,26 +1800,37 @@ class QubitInformationObject extends BaseInformationObject
         }
       }
 
-      // Add bioghist if there is a creator in the position and it doesn't have history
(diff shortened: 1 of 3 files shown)


lib/model/QubitInformationObject.php
```diff
@@ -1800,26 +1800,37 @@ public function importBioghistEadData($biogHistNode, $key)
...
+        // Options:
+        // 1. If there isn't an actor in the current position:
+        //   - Create new 'Untitled' actor with bioghist value as history and new event
+        // 2. If the actor in the current position doesn't have history:
+        //   - Add bioghist value to the actor's history
+        // 3. If the actor in the current position has history and it's different to the bioghist value:
+        //   - Create new 'Untitled' actor with bioghist value as history and new event
+        // 4. If the actor in the current position has history and it's equal to the bioghist value:
+        //   - Do nothing
+        if (!isset($creators[$key]) ||
```
MikeFE, line 1816: Are we checking somewhere for when the actor name, history, and repository match? Re: Dan's comment: " if the actor and repo AND bioghist is the same, it will just link, and no duplicate will be created"

jraddaoui, line 1816: Yes, we're checking the actor name/repository in setActorByName, wich is called before the actors are added to the information object events. And the history is being checked in those conditions.

MikeFE, line 1816: Ah okay, makes sense. The only other suggestion I have is to add @param $key to the top of the function comments, because it wasn't very clear when reviewing this function what that parameter was used for until I did a test!

Looks good to me otherwise

jraddaoui:
Thanks Mike! Merged

PR #84, opened 2014-12-12
Changes related to DIP upload process, refs #7641

jraddaoui:
Include qtSwordPluginWorker in new jobs process:
- Add ability if qtSwordPlugin is enabled
- Use new job classes

Fixes in DIP upload process:
- Do not save target IO at the end, it was creating nested set problems, but update it in the ES index
- Register namespaces and use Xpath queries to allow namespaced and not namespaced METS versions

### DIFF
--- a/plugins/qtSwordPlugin/lib/qtSwordPluginWorker.class.php
+++ b/plugins/qtSwordPlugin/lib/qtSwordPluginWorker.class.php
@@ -17,7 +17,7 @@
  * along with Access to Memory (AtoM).  If not, see <http://www.gnu.org/licenses/>.
  */
 
-class qtSwordPluginWorker extends Net_Gearman_Job_Common
+class qtSwordPluginWorker extends arBaseJob
 {
   protected $dispatcher = null;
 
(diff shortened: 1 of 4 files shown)


plugins/qtSwordPlugin/lib/qtSwordPluginWorker.class.php
```diff
@@ -17,7 +17,7 @@
  * along with Access to Memory (AtoM).  If not, see <http://www.gnu.org/licenses/>.
  */
 
-class qtSwordPluginWorker extends Net_Gearman_Job_Common
```
MikeFE, line 20: It's great that this now uses arBaseJob. Would it be difficult to move this file into lib/jobs with the rest of them, so we could avoid that extra if (...) in lib/task/jobs/jobWorkerTask.class.php? Don't worry about it if it'd be a lot of work (I don't know if it'd break anything), but if it's something as easy as moving the file, I think we should keep all the jobs in one place.

jraddaoui, line 20: I had a conflict thinking about that. I'd like to see all jobs in one place too, but this one is closely related with the work that plugin does. And we'll still need an 'if' to check in the plugin is enabled (it's disabled by default) before adding the ability.

MikeFE, line 20: Ah I see. All right, if we need to do an if (...) anyway, let's keep the job together with the plugin. The rest looks good to me.

sevein, line 20: I think it makes sense to keep it inside the plugin tree so if the plugin is not enabled the whole thing is ignored. We could have a jobs directory inside the plugin too for organizational purposes if that makes sense to you.

The exception "Ability not defined" could maybe be extended for this particular case of plugin workers?

jraddaoui:
Thanks! Merged

PR #87, opened 2014-12-15
CSS fixes for print view, refs #7548

jraddaoui:
- Fix padding/margin problems in field headers for IE and FF
- Give the content full width
- Don't show links href value
- Fix imageflow appearance


sevein:
Merged, thanks!

PR #88, opened 2014-12-17
Global search bar - add search icon, reverse order of Browse and Search menu #6376

sbreker:
USASK development to address issue #6376.
- Reverses the order of the Browse button and the Search Menu
- Changes made to the arDominion theme - add a search icon/button on the right hand side of the search text box.


sevein:
Awesome. Thank you so much! Dan and I changed the background of the search button to a darker colour, other than that we love it! See 677b7f0.

PR #89, opened 2014-12-17
Changes in settings page, refs #7722

jraddaoui:
- Make setting list-menu stick on scroll
- Split setting list in multiples pages
- Ghange settings menu path to 'settings/global' (migration/fixtures)
- Move Finding Aid format setting to Finding Aid settings page
- Change redirections to corresponding pages
- Few style changes


lib/form/SettingsPermissionsForm.class.php
```diff
@@ -17,6 +17,7 @@
  * along with Access to Memory (AtoM).  If not, see <http://www.gnu.org/licenses/>.
  */
 
+ProjectConfiguration::getActive()->loadHelpers('I18N');
```
sevein: Can we stop using the template helper and do the following instead for consistency?

``` php
$i18n = sfContext::getInstance()->i18n;
echo $i18n->__('Title');
```

PR #90, opened 2014-12-17
Fix partial call in advancedSearch, fixes #7726

jraddaoui:
The searchResult partial uses now the ES hit instead of the doc data,
and the selected/current culture. Other calls to the same partial have been
normalized


sevein:
Thanks, merged.

PR #94, opened 2014-12-20
Remove rights section when editing or creating accessions and archival descriptions #7716

MikeFE:
This commit removes the rights section while creating or
editing either archival descriptions or accessions. From now
on, the user will create rights separately via the
"More" -> "Create new rights" button. And can edit existing
rights by clicking 'edit' when viewing archival descriptions.

Also added is the "More" -> "Create new rights" button for
accessions, which wasn't there previously.

Style fix


jraddaoui:
Hi Mike, some details:
- The 'right/edit' partial and the related component could be removed from the right module
- We're still calling the old rights form (that component) in the edit actions of those plugins. I don't think we need that anymore, see for example:
  
  https://github.com/artefactual/atom/blob/qa/2.2.x/plugins/sfIsadPlugin/modules/sfIsadPlugin/actions/editAction.class.php#L99-L101
  https://github.com/artefactual/atom/blob/qa/2.2.x/plugins/sfIsadPlugin/modules/sfIsadPlugin/actions/editAction.class.php#L215
- In the index page of those plugins there is a link to the deleted fieldset in the edit actions, in the rights section header. We may want to remove that link too.

MikeFE:
Hey Radda, can you clarify your third point above? Did you mean, like for instance when visiting an ISAD archival description, where it says under rights "Edit | Delete" and uses Javascript to delete when you click the delete link. You mean remove that? Or did you mean something else?

(Dan says we should probably have a confirmation page when deleting rights--but that is a bit outside the scope of this PR and we can do that another time)

jraddaoui:
Hi Mike, no, I mean the link in the section header:

https://github.com/artefactual/atom/blob/qa/2.2.x/plugins/sfIsadPlugin/modules/sfIsadPlugin/templates/indexSuccess.php#L328

PR #95, opened 2014-12-20
Add option to export EAD like public user in CLI, refs #7743

jraddaoui:
If the '--public' option is added when executing the exportBulk task
draft descriptions (and their descendants) are not added in EAD export.
It won't add physloc either if physical storage is hidden for public user

Refs #7743

### DIFF
--- a/lib/helper/QubitHelper.php
+++ b/lib/helper/QubitHelper.php
@@ -276,8 +276,14 @@ function is_using_cli()
   return php_sapi_name() === 'cli';
 }
 
-function check_field_visibility($fieldName)
+function check_field_visibility($fieldName, $options = array())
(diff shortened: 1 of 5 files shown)


MikeFE:
Does the physical location visible / not visible work for you? I can't for the life of me get sfConfig::get('app_element_visibility_physical_storage', false) to return anything other than false (sfConfig::getAll() doesn't even seem to include that setting). Whereas I can confirm in the database it's set to true. I think there's something really weird with my dev box...

Otherwise, looks good to me, heh.

jraddaoui:
Thanks Mike!

I's missing the step where the settings are added to sfConfig for tasks.

PR #98, opened 2015-01-02
Dev/issue 7727 bulk export cleanup

mcantelon:
Refactored bulk export code, added criteria specification to EAC export, fixed minor EAC issues.


plugins/sfEacPlugin/modules/sfEacPlugin/templates/indexSuccess.xml.php
```diff
@@ -168,7 +168,7 @@
 
       <?php foreach ($resource->getActorRelations() as $item): ?>
         <cpfRelation cpfRelationType="<?php echo sfEacPlugin::toCpfRelationType($item->type->id) ?>" xlink:href="<?php echo url_for(array($item->getOpposedObject($resource), 'module' => 'actor'), true) ?>" xlink:type="simple">
-          <relationEntry><?php echo render_title($item->getOpposedObject($resource)) ?></relationEntry>
```
MikeFE, line 171: Is this related to refactoring the CLI tasks? Also, I think Jesús was doing some fancy stuff with auto-escaping in Symfony so I'm not even sure if we need these helper functions anymore (might want to double check with him).

PR #99, opened 2015-01-15
Removed unneeded line of code. refs #7824

mcantelon:


MikeFE:
Looks good Mike, thanks for fixing this!

The only thing I can point out is I think we should keep the includeClassesAndHelpers method as protected, since exportBulkTask inherits from exportBulkBaseTask and therefore should still have access to includeClassesAndHelpers. Not a big deal though.

PR #100, opened 2015-01-16
DIP upload integration, refs #7754

jraddaoui:
Adds arRestApiPlugin:
- Used to fetch LOD from Archivematica for DIP uploads to AtoM
- Only with browse taxonomies endpoint
- Not enabled by default but available in the plugin list
- With authoritation fix for Apache

Adds hierarchical DIP upload method:
- Used if a logical structMap exists in the METS file

Changes non-hieararchical DIP upload method:
- It now uses the fileGrp element to create the IO instead of looking at the objects folder
- The IO are now saved only once (in both process)
- Helper function moved to Qubit class

Changes metsData ES mapping:
- Adds Mediainfo tracks and other data that was added for the DRMC project

Adds QubitMetsParser:
- Moved most of the METS related code in the DIP upload process to it
- Used to get the data needed from the METS file for the IO search index
- Code clean-ups in the ES related code
- Allow METS with and without the namespace changes


plugins/arElasticSearchPlugin/lib/model/arElasticSearchInformationObjectPdo.class.php
```diff
@@ -945,286 +945,31 @@ protected function getPhysicalObjects()
...
-              $agent['type'] = (string)$value[0];
-            }
+      return;
+    }
 
-            $metsData['agent'][] = $agent;
-          }
+    // Get METS filepath
+    $metsFilepath = sfConfig::get('sf_web_dir').
+      DIRECTORY_SEPARATOR.'uploads'.
```
sevein: Use sf_uploads_dir, because uploads/ doesn't have to be necessarily located under the root dir.

sevein:
Just one more thing. Do you mind to take a look at the headers when the client doesn't send cookies and it's hitting the API? For some reason back when we wrote this Symfony was sending Set-Cookie twice and I'm not sure why, but it's probably because one of the filters is maybe calling a function that we shouldn't be calling at that point.

jraddaoui:
Merged!

PR #101, opened 2015-01-27
Improve exception handling in jobs, refs #7822

jraddaoui:
- Catch all possible exceptions in job execution and throw Net_Gearman_Job_Exception
  to avoid breaking the worker
- Change fail function in worker to acept Exception (inheritance in
  Net_Gearman exception was changed to avoid using PEAR_Exception)
- Change exceptions manually thrown to Net_Gearman_Job_Exception
- Remove unneeded try/catch in jobs and jobWorker class
- Add JOB_FAIL callback to the worker to log the exception message
- Remove batchDocs from the queue even if the dump to the index fails

### DIFF
--- a/lib/job/arBaseJob.class.php
+++ b/lib/job/arBaseJob.class.php
@@ -27,12 +27,14 @@
 
 class arBaseJob extends Net_Gearman_Job_Common
 {
-  // Required parameters for all jobs:
-  // - Parameters that are mandatory for all jobs
-  // - They are checked at the begining of the job in checkRequiredParameters() function
-  // - Required parameters for each job can be declared in the subclases
-  // - The job will fail if any of the required paramaters are missing
-  private $requiredParametersForAllJobs = array('id', 'name');
+  /*
+   * Required parameters:
+   *
+   * Declares parameters that are mandatory for the jobs execution. They can be
+   * extended on each job subclasse using the $extraRequiredParameters property.
+   * If any of the required paramareters is missing the job will fail.
+   */
+  private $requiredParameters = array('id', 'name');
 
   public function run($parameters)
   {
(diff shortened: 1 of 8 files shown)


lib/job/arGenerateFindingAidJob.class.php
```diff
@@ -24,102 +24,99 @@
 
 class arGenerateFindingAidJob extends arBaseJob
 {
+  // Required parameters:
+  // - Parameters that are mandatory for the job
+  // - They are checked at the begining of the job from arBaseJob run() function
+  // - 'id' and 'name' are required parameters for all jobs and they are added in arBaseJob
+  // - The job will fail if any of the required paramaters are missing
+  protected $requiredParameters = array('objectId');
```
sevein: This comment repeated each time you redefine the member is a bit too verbose. You could just use the phpdoc tag @see, e.g. `@see arBaseJob::$requiredParametersForAllJobs`.

lib/job/arBaseJob.class.php
```diff
@@ -27,11 +27,15 @@
 
 class arBaseJob extends Net_Gearman_Job_Common
 {
-  private $requiredParameters = array();
+  // Required parameters for all jobs:
+  // - Parameters that are mandatory for all jobs
+  // - They are checked at the begining of the job in checkRequiredParameters() function
+  // - Required parameters for each job can be declared in the subclases
+  // - The job will fail if any of the required paramaters are missing
+  private $requiredParametersForAllJobs = array('id', 'name');
```
sevein: Do you mind to use the standard docblock format like we do with methods?

lib/job/arBaseJob.class.php
```diff
@@ -27,11 +27,15 @@
 
 class arBaseJob extends Net_Gearman_Job_Common
 {
-  private $requiredParameters = array();
+  // Required parameters for all jobs:
```
sevein: It's not like you are introducing a list of items here. I'd just reword this with a simple paragraph.

lib/job/arTestJob.class.php
```diff
@@ -27,14 +27,15 @@
 
 class arTestJob extends arBaseJob
 {
-  public function run($parameters)
+  // Required parameters:
+  // - Parameters that are mandatory for the job
+  // - They are checked at the begining of the job from arBaseJob run() function
+  // - 'id' and 'name' are required parameters for all jobs and they are added in arBaseJob
+  // - The job will fail if any of the required paramaters are missing
+  protected $requiredParameters = array();
```
sevein: Looking at checkRequiredParameters() it seems that removing this is ok as isset() is used, right?

Also, I find the name a bit confusing because it makes you think that it's final while it's being merged with the parent list. I'd use maybe $extraRequiredParameters or just $extraParameters.

sevein:
I've suggested a few changes but they are not very important.

jraddaoui:
Thanks Sevein, I've made those changes. Please, let me know if I should merge.

jraddaoui:
Merged.

PR #104, opened 2015-02-02
Fix accession counter in different cultures, refs #5914

jraddaoui:
Save accession counter setting value always in source culture


jraddaoui:
Merged with --squash

PR #105, opened 2015-02-02
Add full metadata model for Finding Aid generation and other fixes, refs #7723

jraddaoui:
- Adds other XSL file for FA generation with full metadata in all levels of description
- Fix repository name duplication in both models
- Fixes in helper-functions file
- Adds FA settings: model and export like public user
- Fix disable autorefresh in browse jobs
- Changes FA filename tu use the slug of the description
- Adds QubitSlug::getByObjectId static function
- Open/download FA in a new tab
- Move FA section logic to a component
- Hide FA status for public view if no download available
- Hide FA section for draft descriptions if export l ike public user is set to 'yes'
- Remove FA txt format option


lib/job/arGenerateFindingAidJob.class.php
```diff
@@ -74,10 +81,17 @@ public function run($parameters)
         return false;
       }
 
-      // Crank the XML through XSL stylesheet and fix header / fonds URL
-      $eadXslFilePath = $appRoot . '/lib/task/pdf/ead-pdf.xsl';
+      // Use XSL file selected in Finding Aid model setting
+      $findingAidModel = 'bombardier';
```
MikeFE: Can we call this something more generic? This is meant to go into mainline AtoM right?

jraddaoui:
Oh, right! I forgot to change that default value (in case the setting is missing) when the real setting values were added.

Thanks Mike, it's fixed now.

jraddaoui:
Merged

PR #109, opened 2015-02-17
Remove login options in read only mode, refs #7913

jraddaoui:
Checks if the app is in read only mode looking at the env. variable
ATOM_READ_ONLY and the setting read_only in app.yml,
giving priority to the env. variable.

If the app is in read only mode, the login button will be removed
and the '.../user/login' url will redirect to the homepage

### DIFF
--- a/apps/qubit/modules/user/actions/loginAction.class.php
+++ b/apps/qubit/modules/user/actions/loginAction.class.php
@@ -25,7 +25,8 @@ class UserLoginAction extends sfAction
     $this->form->getValidatorSchema()->setOption('allow_extra_fields', true);
 
     // Redirect to @homepage if the user is already authenticated
-    if ($this->context->user->isAuthenticated())
+    // or the read only mode is enabled
+    if ($this->context->user->isAuthenticated() || Qubit::isReadOnly())
     {
       $this->redirect('@homepage');
     }
--- a/apps/qubit/modules/user/actions/loginAction.class.php
+++ b/apps/qubit/modules/user/actions/loginAction.class.php
@@ -26,7 +26,7 @@ class UserLoginAction extends sfAction
 
     // Redirect to @homepage if the user is already authenticated
     // or the read only mode is enabled
-    if ($this->context->user->isAuthenticated() || Qubit::isReadOnly())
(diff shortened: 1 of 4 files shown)


apps/qubit/modules/user/actions/loginAction.class.php
```diff
@@ -25,7 +25,8 @@ public function execute($request)
     $this->form->getValidatorSchema()->setOption('allow_extra_fields', true);
 
     // Redirect to @homepage if the user is already authenticated
-    if ($this->context->user->isAuthenticated())
+    // or the read only mode is enabled
+    if ($this->context->user->isAuthenticated() || Qubit::isReadOnly())
```
sevein: Wouldn't make more sense to forward the user to the unauthorized page if the app is in read only mode?

lib/Qubit.class.php
```diff
@@ -214,4 +214,33 @@ public static function likeISO8601Date($date)
...
+    {
+      return true;
+    }
+
+    // Otherwise:
+    // - The env. variable and the setting don't exist
+    // - The env. variable is set to anything different than 'true', '1', 'on' or 'yes'
+    // - The env. varibale doesn't exist and the setting is set to anything different than 'true'
+    return false;
+  }
```
sevein: I think that it would be more elegant to have QubitSettingsFilter looking at $_ENV, so we can keep interfacing with sfConfig from our code, i.e. I'd rather keep interfacing with sfConfig::get('app_whatever') that introducing new methods to retrieve individual settings like your suggested Qubit::isReadOnly().

QubitSettingsFilter avoids repetitive access to the database with cache. Once that's done (#L38), you can iterate over a list of known environment variables. I think that it makes sense to define an internal hashmap where the key is the environment variable and the value is its type, e.g. "ATOM_READ_ONLY" => "boolean", "ATOM_FOOBAR" => "string", etc...

Thoughts?

sevein:
Hey, and don't forget [isAuthenticated()](https://github.com/artefactual/atom/blob/qa/2.2.x/lib/myUser.class.php#L213).

jraddaoui:
Merged with --squash

PR #110, opened 2015-02-18
Prevent duplicate creation of new terms #8006

MikeFE:
Make sure to reload the status variable when creating new terms, so subsequent rows can attach to the newly created terms. Currently, they'll fail to detect new terms in subsequent rows and create new, duplicate terms indefinitely whenever one is encountered.


jraddaoui:
Looks good to me!

PR #114, opened 2015-03-06
#8063 - Catch render_field() helper exception

sevein:
Redmine ticket: https://projects.artefactual.com/issues/8063

Don't let the "Unknown record property" bubble up as it's a possible that the
developer is using render_field to generate form fields that don't map with ORM
objects.


mcantelon:
This looks good to me... I don't see any logic or style issues!

PR #121, opened 2015-03-18
Fix translation lang. in user creation, refs #8115

jraddaoui:
Allowed languages for translation must be saved after the user is created


mcantelon:
Woops, qubot was actually me. Was logged in via lastpass. Sorry for the confusion.

PR #122, opened 2015-03-18
Fix edit actor/repository permisions for users, refs #8116

jraddaoui:
Allowed languages for translation is creating a row with null object_id that is
creating an All table in those pages. In that context permissions for all objects (null)
and for the root actor/repository object are equivalent.


mcantelon:
Woops, qubot was actually me. Was logged in via lastpass. Sorry for the confusion.

PR #131, opened 2015-03-26
Fix DO extractText function, refs #8147

jraddaoui:
Use localPath instead of absolutePath for external resources


sevein:
Thanks, @jraddaoui. It happened that I was hanging around qa/2.2.x so I took the chance and merged it.

PR #133, opened 2015-03-28
Add OAI admin emails setting, refs #8156

jraddaoui:
- Add migration and fixtures for new setting
- Add field to OAI settings form
- Use setting value in OAI identify component

### DIFF
--- a/data/fixtures/settings.yml
+++ b/data/fixtures/settings.yml
@@ -3,7 +3,7 @@ QubitSetting:
     name: version
     editable: 0
     deleteable: 0
-    value: 119
+    value: 120
   milestone:
     name: milestone
     editable: 0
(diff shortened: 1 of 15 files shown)


data/fixtures/settings.yml
```diff
@@ -515,6 +515,13 @@ QubitSetting:
     editable: 1
     deleteable: 0
     value: 100
+  QubitSetting_oai_repository_code:
```
mcantelon: You might have forgot to change this key in the YAML.

plugins/arOaiPlugin/modules/arOaiPlugin/templates/_identify.xml.php
```diff
@@ -3,7 +3,7 @@
     <baseURL><?php echo QubitOai::getBaseUrl() ?></baseURL>
     <protocolVersion><?php echo $protocolVersion ?></protocolVersion>
     <?php
-    foreach($adminEmail as $email){
+    foreach($adminEmails as $email){
```
mcantelon: Minor style nitpick: the "{" should be moved to the following line (I probably put the incorrect code style in there originally).

mcantelon: (I think there's supposed to be a space between "foreach" and "(" too.)

jraddaoui:
Thanks Mike! Merged.

PR #135, opened 2015-03-30
Dev/issue 8164 update pub status improvements

MikeFE:


lib/task/tools/updatePublicationStatusTask.class.php
```diff
@@ -62,8 +63,7 @@ protected function execute($arguments = array(), $options = array())
       $resource = QubitRepository::get($criteria)->__get(0);
     }
 
-    $publicationStatus = QubitTerm::getById($arguments['publicationStatusId']);
-
+    $publicationStatus = QubitTerm::getById($this->getPublicationStatusIdByName($arguments));
```
mcantelon, line 66: Super minor nitpick: only the 'publicStatus' element of $arguments gets used in getPublicationStatusIdByName so you could just pass that. But that's pretty subjective so it doesn't really matter. :)

MikeFE, line 66: good call

PR #136, opened 2015-03-31
Oai sets for collections

marktriggs:
Hi all,

I wrote to the users list about some OAI changes here:

https://groups.google.com/d/msgid/ica-atom-users/871tkl26oy.fsf%40teaspoon-consulting.com

and it was suggested that I send my proposed changes as a PR for discussion.

My goal is to generalise the way OAI sets work so that there doesn't have to be a 1:1 mapping between OAI sets and collections.  In my case, I'd like to be able to define an OAI set containing all top-level (collection) records.

The current OAI implementation defines an OAI set with three pieces of information:
- a setSpec
- a `lft` value
- a `rgt` value

This is oriented around the "one set per collection" that is currently implemented.  These three values are used throughout the OAI code in a couple of ways:
- Searching the list of OAI sets by `lft` value to find which set a given Information Object belongs to.
- Searching the list of OAI sets by `setSpec` to find the definition of a requested OAI set.
- Applying the `lft` and `rgt` values of a given OAI set as `Criteria` to find matching Information Objects

To generalise this, I've pulled the concept of an OAI set into its own class.  An OAI set satisfies the above use cases pretty directly:
- You can ask an OAI set for its setSpec
- You can ask an OAI set whether it contains a given Information Object
- You can pass it a `Criteria` object and have it apply its own set restrictions

I've refactored the existing places that use the old OAI set format to use the new object interface instead.

I realise there's a lot to review here, so I've tried to break the work into smaller logical pieces:
- 98944ad Generalise the OAI set implementation
  
  This makes the fundamental changes described above.  This is really where the action is.  This commit also addresses this issue: https://projects.artefactual.com/issues/6436.
- 2754416 Change OAI terminology from "collection" to "OAI set"
  
  This changes method names and comments to reflect the fact that we're not collection-centric anymore.  It switches the terminology to "OAI Set" where relevant.  It's a messy commit, but isn't intended to change behaviour.
- 72a8026 Add a new "virtual" OAI set matching top-level records
  
  Add a new OAI Set for collection-level records (the one I wanted originally).
- 53041a7 Add new configuration option for additional OAI sets
  
  Add a configuration option to turn these new OAI sets off.  This includes a database migration to add the new option, and it defaults to "disabled" to preserve backwards compatibility.

Sorry for the essay, and please let me know if I can clarify or change anything.  I'll add some notes regarding my testing as a separate comment.

### DIFF
--- a/apps/qubit/modules/settings/actions/oaiAction.class.php
+++ b/apps/qubit/modules/settings/actions/oaiAction.class.php
@@ -72,6 +72,7 @@ class SettingsOaiAction extends sfAction
     $oaiRepositoryIdentifier = QubitOai::getRepositoryIdentifier();
     $sampleOaiIdentifier = QubitOai::getSampleIdentifier();
     $resumptionTokenLimit = QubitSetting::getByName('resumption_token_limit');
+    $oaiAdditionalSetsEnabled = QubitSetting::getByName('oai_additional_sets_enabled');
 
     // Set defaults for global form
     $this->oaiRepositoryForm->setDefaults(array(
@@ -80,7 +81,8 @@ class SettingsOaiAction extends sfAction
       'oai_repository_identifier' => $oaiRepositoryIdentifier,
       'oai_admin_emails' => $oaiAdminEmails,
       'sample_oai_identifier' => $sampleOaiIdentifier,
-      'resumption_token_limit' => (isset($resumptionTokenLimit)) ? $resumptionTokenLimit->getValue(array('sourceCulture'=>true)) : null
+      'resumption_token_limit' => (isset($resumptionTokenLimit)) ? $resumptionTokenLimit->getValue(array('sourceCulture'=>true)) : null,
+      'oai_additional_sets_enabled' => (isset($oaiAdditionalSetsEnabled)) ? intval($oaiAdditionalSetsEnabled->getValue(array('sourceCulture'=>true))) : 0
     ));
   }
 
@@ -118,6 +120,12 @@ class SettingsOaiAction extends sfAction
       $setting->setValue($resumptionTokenLimit, array('sourceCulture' => true));
       $setting->save();
     }
+
+    // OAI additional sets enabled radio button
+    $oaiAdditionalSetsEnabledValue = $thisForm->getValue('oai_additional_sets_enabled');
+    $setting = QubitSetting::getByName('oai_additional_sets_enabled');
+    $setting->setValue($oaiAdditionalSetsEnabledValue, array('sourceCulture' => true));
+    $setting->save();
 
     return $this;
   }
--- a/data/fixtures/settings.yml
+++ b/data/fixtures/settings.yml
@@ -3,7 +3,7 @@ QubitSetting:
     name: version
     editable: 0
     deleteable: 0
-    value: 120
+    value: 122
   milestone:
     name: milestone
     editable: 0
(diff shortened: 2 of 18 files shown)


marktriggs:
Regarding testing, I've hit the following URLs both before and after applying my commits:
- http://localhost/index.php/;oai?verb=ListSets
- http://localhost/index.php/;oai?verb=ListRecords&metadataPrefix=oai_dc
- http://localhost/index.php/;oai?verb=ListRecords&metadataPrefix=oai_dc&set=oai:localhost:anuarchives_10044
- http://localhost/index.php/;oai?verb=ListIdentifiers&metadataPrefix=oai_dc
- http://localhost/index.php/;oai?verb=GetRecord&metadataPrefix=oai_dc&identifier=oai:localhost:anuarchives_10044
- http://localhost/index.php/;oai?verb=ListRecords&metadataPrefix=oai_dc&set=oai:virtual:top-level-records

Then I diff'd the two sets of XML.  The notable changes:
- The setSpec element isn't blank anymore (due to https://projects.artefactual.com/issues/6436 being fixed)
- The third call to `ListRecords` with a set specified now seems to be doing the right thing, where I'm not sure it was before.  Prior to my change, I got the following response where the set it returned doesn't seem to match the one I requested:
  
  ```
   <OAI-PMH ...>
     ...
     <request verb="ListRecords" metadataPrefix="oai_dc" set="oai:localhost:anuarchives_10044">http://localhost/;oai</request>
     <ListRecords>
      <record>
       <header>
         ...
         <setSpec>oai:localhost:anuarchives_10086</setSpec>
       </header>
  ```
  
  but after applying my commits:
  
  ```
   <OAI-PMH ...>
     ...
     <request verb="ListRecords" metadataPrefix="oai_dc" set="oai:localhost:anuarchives_10044">http://localhost/;oai</request>
     <ListRecords>
      <record>
       <header>
         ...
         <setSpec>oai:localhost:anuarchives_10044</setSpec>
       </header>
  ```
  
  the setSpec returned matches the one requested.  I'm not sure if there was an outstanding bug report for this, but the new behaviour seems right to me, even though it's different.

mcantelon:
Hi Mark... thanks so much! I'm checking this out today!

mcantelon:
This works well and the code's elegant . The only issue I found is that if a non-existent set is specified it returns all records. What do you think about making it so if `QubitOai::getMatchingOaiSet($this->set, $this->oaiSets)` (in arOaiPluginComponent.class.php) returns false we return a bad argument response to the harvester?

marktriggs:
Thanks so much for reviewing this, Mike.  I think the way you describe handling the non-existent set request is a good way to go.  I'll update this pull request early next week with a change to do that.

Cheers,
Mark

marktriggs:
Hi Mike,

I've just added a commit to implement the check on set name.  Would you mind taking a look?  Hopefully fairly minor!

https://github.com/marktriggs/atom/commit/23225f4f45077bad33741fc2162155a8dfc05285

Cheers,

Mark

mcantelon:
Hi Mark,

It works like a charm... thanks very much for this work! I've merged the changes into qa/2.2.x (https://github.com/artefactual/atom/commit/8d72f57ad5b6e4988016a49c35ee56c8180ebc60) so they'll be included in the upcoming release.

marktriggs:
Thanks Mike!  That's great news.  I'll be doing some more work with OAI over the coming weeks, so I'll send more pull requests if I come across any issues.  I think that's the big one, though, so thanks again for your review.

Cheers,
Mark

fiver-watson:
Hi Mark, 

I've added some basic documentation here, which will be available in our 2.2 docs when the release is ready: 

https://github.com/artefactual/atom-docs/commit/e4651d8ae2bc848ce85083ea9460e2cd348e3f6b

Thanks again for your contributions!

marktriggs:
Nice!  Thanks Dan :)

PR #137, opened 2015-03-31
Fix/add RAD encodinganalog values, refs #4505

jraddaoui:
It also fix the controlaccess section for lower level descriptions, using the same logic that we're using for the top-level description.


jraddaoui:
Merged.

PR #149, opened 2015-04-14
Oai improvements

marktriggs:
Hi there,

I have a few more enhancements to the OAI implementation to share.  Happy to split these into multiple pull requests--I wasn't sure what would be easier to review.  Here's a breakdown of the changes:
- d949214486b6579aeffeb4316183e2f4a51032d4 -- XML encode the strings returned in OAI responses to avoid problems with special characters.  This issue was documented in https://projects.artefactual.com/issues/8159.
- 1c41118743bb25e6ecca22951a8d309c175f72db -- Change the resumption token implementation to avoid the need for escaping, and extend it to support OAI sets.  The resumption token is now represented as a base64-encoded JSON string, which isn't visually beautiful but saved inventing a custom serialisation format :)  Addresses https://projects.artefactual.com/issues/8158.
- 3fdc8dd4a4071f1550117cdfb2ee0d001a8c5173 and 250bea356eb508d41318cd818af51beda9594e92 are a small modification to my previous OAI sets pull request (which has since been merged).  While testing, I realised that I wasn't always returning all top-level records, so I fixed that up.

With those changes, I can harvest all of my records using jOAI and it looks pretty good.  As always, feedback is greatly appreciated, and let me know if anything here needs changing.

Cheers,
Mark


mcantelon:
Hi Mark! Awesome... I'm going to take a look today (and having them be in one pull request is totally fine)!

mcantelon:
Merged... thanks again Mark!

PR #152, opened 2015-04-21
Dev/csv export 8122

mcantelon:
This PR adds bulk CSV export of descriptions, either RAD or ISAD, and does a little bit of cleanup/refactoring of the bulk export class. A parent CSV export class exists upon which other CSV export child classes can be created.


lib/flatfile/QubitFlatfileExport.class.php
```diff
@@ -0,0 +1,421 @@
...
+      {
+        $this->fileIndex++;
+        unset($this->currentFileHandle);
+      }
+
+      // Generate filename
+      $filename = exportBulkBaseTask::generateSortableFilename($this->fileIndex, 'csv', $this->standard);
+      $filePath = $this->path .'/'. $filename;
+    }
+    else {
```
MikeFE: minor style issue, braces on their own lines always

lib/flatfile/csvInformationObjectExport.class.php
```diff
@@ -0,0 +1,390 @@
...
+  protected function setAlternativeIdentifierColumns()
+  {
+    $properties = $this->resource->getProperties(null, 'alternativeIdentifiers');
+
+    $alternativeIdentifiers      = array();
+    $alternativeIdentifierLabels = array();
+
+    foreach($properties as $property)
+    {
+      array_push($alternativeIdentifierLabels, $property->name);
```
MikeFE: http://php.net/manual/en/function.array-push.php seems to indicate in the note if you're just pushing a single value to an array, $ary[] = 123 is slightly better than array_push(). But I don't think it's a big deal.

lib/job/arBaseJob.class.php
```diff
@@ -135,7 +135,7 @@ private function formatLogMsg($message)
   /**
    * Clear various Qubit classes' caches.
    */
-  private function clearCache()
+  static function clearCache()
```
MikeFE: I think this method makes more sense at this point put into lib/helper/QubitHelper.php as something named clear_all_orm_cache() or similar. It seems kind of weird to be in arBaseJob if it's being used way elsewhere in the code base.

mcantelon: Agreed. I'll look into making it a helper.

PR #157, opened 2015-04-30
Add <repository> info for more cases in EAD #6547

MikeFE:
If you export EAD on a non-top level description's page, previously it wouldn't
include repository information if that record inherited its repository. Now it will
always include <repository> info in the <archdesc>...<did> section if there is any.

### DIFF
--- a/plugins/sfEadPlugin/modules/sfEadPlugin/templates/indexSuccessBodyDidElement.xml.php
+++ b/plugins/sfEadPlugin/modules/sfEadPlugin/templates/indexSuccessBodyDidElement.xml.php
@@ -97,7 +97,7 @@
     </physdesc>
   <?php endif; ?>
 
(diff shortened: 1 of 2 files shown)


plugins/sfEadPlugin/modules/sfEadPlugin/templates/indexSuccessBodyDidElement.xml.php
```diff
@@ -97,7 +97,7 @@
     </physdesc>
   <?php endif; ?>
 
-  <?php if ($value = $$resourceVar->getRepository()): ?>
+  <?php if ($value = $$resourceVar->getRepository(array('inherit' => $topLevelDid ? true : false))): ?>
```
jraddaoui: Hi Mike,

$topLevelDid is not interpreted as a boolean in getRepository()?

PR #158, opened 2015-04-30
Add hyperlinks back to records in job status page #8228

MikeFE:
Add a hyper link next to the job status on the job management page. These links will (if the job has any), lead the user back to the information object / actor / etc. the job was running for.

### DIFF
--- a/lib/model/QubitJob.php
+++ b/lib/model/QubitJob.php
@@ -136,6 +136,34 @@ class QubitJob extends BaseJob
   }
 
   /**
+   * Get the module type for this job's corresponding object.
+   * e.g., informationobject, actor, etc.
+   *
+   * @return mixed  A string indicating the module type of the object for this job,
+   * or else null.
+   */
+  public function getObjectModule()
+  {
+    $className = QubitPdo::fetchColumn('SELECT class_name FROM object WHERE id = ?', array($this->objectId));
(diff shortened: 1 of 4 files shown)


lib/model/QubitJob.php
```diff
@@ -136,6 +136,34 @@ public function getStatusString()
   }
 
   /**
+   * Get the module type for this job's corresponding object.
+   * e.g., informationobject, actor, etc.
+   *
+   * @return mixed  A string indicating the module type of the object for this job,
+   * or else null.
```
mcantelon, line 143: Ultra-minor nitpick... maybe line up "or else null" vertically with "A string".

lib/model/QubitJob.php
```diff
@@ -136,6 +136,34 @@ public function getStatusString()
...
+  }
+
+  /**
+   * Get the associated object's slug for this job.
+   *
+   * @return mixed  A string indicating the object's slug. If none, return null.
+   */
+  public function getObjectSlug()
+  {
+    return QubitPdo::fetchColumn('SELECT slug FROM slug WHERE object_id = ?', array($this->objectId));
```
mcantelon, line 163: You could also do the ORM way, but slower:

```
return QubitSlug::getByObjectId($this->objectId)->slug;
```

PR #159, opened 2015-05-05
Inventory page, refs #8384

sevein:
https://projects.artefactual.com/issues/8384

### DIFF
--- a/apps/qubit/modules/informationobject/actions/inventoryAction.class.php
+++ b/apps/qubit/modules/informationobject/actions/inventoryAction.class.php
@@ -1,0 +1,186 @@
+<?php
+
+/*
+ * This file is part of the Access to Memory (AtoM) software.
+ *
+ * Access to Memory (AtoM) is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * Access to Memory (AtoM) is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with Access to Memory (AtoM).  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+class InformationObjectInventoryAction extends DefaultBrowseAction
+{
+  private static $levels;
+
+  public function execute($request)
+  {
+    $this->resource = $this->getRoute()->resource;
+
+    // Check that this isn't the root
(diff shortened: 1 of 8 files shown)


apps/qubit/modules/informationobject/actions/inventoryAction.class.php
```diff
@@ -0,0 +1,186 @@
...
+      $query->setFrom(($page - 1) * $limit);
+    }
+
+    $q = new \Elastica\Query\Bool;
+
+    $q1 = new \Elastica\Query\Term;
+    $q1->setTerm('ancestors', $resource->id);
+    $q->addMust($q1);
+    $q2 = new \Elastica\Query\Terms;
+    $q2->setTerms('levelOfDescriptionId', self::getLevels());
```
MikeFE: What happens here if self::getLevel() returns null? (apparently /"return;" and "return null;" are equivalent in PHP)

apps/qubit/modules/informationobject/actions/inventoryAction.class.php
```diff
@@ -0,0 +1,186 @@
...
+    {
+      return;
+    }
+
+    self::$levels = $value;
+
+    return $value;
+  }
+
+  private static function getResults(&$resource, &$limit = 10, &$page = 1, &$sort = 'identifierUp')
```
MikeFE, line 90: Is there a reason the helper functions in this action are all static? Not a big deal though.

apps/qubit/modules/settings/actions/inventoryAction.class.php
```diff
@@ -0,0 +1,109 @@
...
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with Access to Memory (AtoM).  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+class SettingsInventoryAction extends DefaultEditAction
+{
+  // Arrays not allowed in class constants
+  public static
+    $NAMES = array(
```
MikeFE, line 24: minor space issue

apps/qubit/modules/settings/actions/inventoryAction.class.php
```diff
@@ -0,0 +1,109 @@
...
+
+        $this->form->setValidator('levels', new sfValidatorPass);
+
+        $choices = array();
+        foreach (QubitTerm::getLevelsOfDescription() as $item)
+        {
+          $choices[$item->id] = $item->__toString();
+        }
+
+        $size = count($choices); if ($size === 0) $size = 4;
```
MikeFE: should break this up and use braces for the if()

apps/qubit/modules/informationobject/actions/inventoryAction.class.php
sevein, line 90: That was made like that on purpose. I had to make that method static because one of the two paces where it's being used is also a static method.

sevein:
Thanks!

PR #162, opened 2015-05-08
Use the dc prefix in ListRecords responses (addresses issue 8159)

marktriggs:
Hi there,

This commit addresses issue https://projects.artefactual.com/issues/8159 and implements the last of the changes suggested by João Pereira.

I've run test harvests with these modifications using jOAI, and the people from http://trove.nla.gov.au/ also kindly ran a test harvest into their system, so I think we're looking pretty good.

As always, let me know if anything needs changing.

Cheers,
Mark


sevein:
Thank you, @marktriggs! f129fb01c8d80dcb4ffa6fcc33416d6645a0563b

PR #171, opened 2015-05-20
Add full width treeview, refs #6375

jraddaoui:
- Includes Andy Koch's treeview created for WBG without their restrictions
- Removes treeview toggle options
- Adds treeview type setting in global settings form
- Fixes culture problems


sevein:
Can you try to minimize the number of files that we are adding to vendor/ to what's strictly necessary? e.g. I don't think we are going to need vendor/jstree/libs. I'm not sure about jstreegrid.js, are we using it? Also, we already have a copy of jquery-ui in vendor/jquery-ui.js, can we update that one instead? Do we need their css files, images etc? If not, just avoid them.

sevein:
What exactly js/history.js does? Do we need it? It looks like excepting IE9 the history API is available pretty much everywhere. Can we just use interact with it without unnecessary compatibility layers? https://developer.mozilla.org/en-US/docs/Web/API/History | http://caniuse.com/#feat=history

apps/qubit/config/view.yml
```diff
@@ -25,6 +25,8 @@ default:
     # graphic:
     # main:
     # print: { media: print, position: last }
+    /vendor/jquery-ui/css/ui-lightness/jquery-ui-1.10.4.custom.min.css:
+    //netdna.bootstrapcdn.com/font-awesome/4.0.3/css/font-awesome.css:
```
sevein: We have a local copy of FA, right?

apps/qubit/modules/informationobject/actions/fullWidthTreeViewAction.class.php
```diff
@@ -0,0 +1,127 @@
...
+ *
+ * @package AccesstoMemory
+ * @subpackage model
+ * @author Andy Koch <koch.andy@gmail.com>
+ */
+class InformationObjectFullWidthTreeViewAction extends sfAction
+{
+  public function execute($request)
+  {
+    $start_time = microtime(TRUE);
```
sevein: Why are we doing this?

jraddaoui:
Merged: https://github.com/artefactual/atom/commit/95185345cf944c1e97c2ab5608a482ed0022bc08

PR #174, opened 2015-05-22
Dev/issue 8348 search result export v2

mcantelon:
Added CSV export of search results.


lib/job/arSearchResultExportCsvJob.class.php
```diff
@@ -0,0 +1,235 @@
...
+   *     'sf0' => ''
+   *   );
+   *
+   * @return object  \Elastica\Query\Bool instance
+   */
+  protected function parseQuery()
+  {
+    $queryBool = new \Elastica\Query\Bool();
+
+    $count = -1;
```
MikeFE: could we do count = 0 here then just count++ at the very end of the while loop, instead of -1 and ++count at the start?

lib/job/arSearchResultExportCsvJob.class.php
```diff
@@ -0,0 +1,235 @@
...
+    // Add criteria for main search fields
+    if (null !== $criterias = $this->parseQuery())
+    {
+      $this->search->queryBool->addMust($criterias);
+    }
+
+    // Add criteria fo secondary search fields
+    foreach (SearchAdvancedAction::$NAMES as $name)
+    {
+      if (
```
MikeFE: Minor nitpick but in most of AtoM we don't do newlines after the (, and the ) usually isn't on its own line either.

lib/job/arSearchResultExportCsvJob.class.php
```diff
@@ -0,0 +1,235 @@
...
+   * @param string  Path of file to write CSV data to
+   *
+   * @return int  Number of descriptions exported
+   */
+  protected function exportResults($path)
+  {
+    $itemsExported = 0;
+
+    // Exporter will create a new file each 10,000 rows
+    $writer = new csvInformationObjectExport(
```
MikeFE: I think it's fine to put this all in one line as long as the line is around < 120 chars

plugins/arElasticSearchPlugin/lib/arElasticSearchPluginQuery.class.php
```diff
@@ -0,0 +1,182 @@
...
+      $filter = new \Elastica\Filter\Bool;
+
+      // Sets a filter for this facet
+      if (isset($item['filter']))
+      {
+        switch ($item['filter'])
+        {
+          case 'hideDrafts':
+            QubitAclSearch::filterDrafts($filter);
+          break;
```
MikeFE: indent break

plugins/arElasticSearchPlugin/lib/arElasticSearchPluginQuery.class.php
```diff
@@ -0,0 +1,182 @@
...
+      $this->filters['languages'] = $this->request->languages;
+      $term = new \Elastica\Filter\Term(array($this->facets['languages']['field'] => $this->request->languages));
+
+      $this->filterBool->addMust($term);
+    }
+
+    // Add facet selections as search criteria
+    foreach ($params as $param => $value)
+    {
+      if (
```
MikeFE: see above about if parentheses

apps/qubit/modules/default/actions/browseAction.class.php
```diff
@@ -288,27 +177,20 @@ public function execute($request)
...
     }
 
-    $this->query = new \Elastica\Query();
-    $this->query->setLimit($request->limit);
-
-    if (!empty($request->page))
-    {
-      $this->query->setFrom(($request->page - 1) * $request->limit);
-    }
+    $indexType = (property_exists($this, 'INDEX_TYPE')) ? $this::INDEX_TYPE : null;
```
MikeFE: I don't think you need extra () here

lib/job/arSearchResultExportCsvJob.class.php
```diff
@@ -0,0 +1,235 @@
...
+    // configuration
+    $writer->loadResourceSpecificConfiguration('QubitInformationObject');
+    array_unshift($writer->columnNames, 'referenceCode');
+    array_unshift($writer->standardColumns, 'referenceCode');
+
+    $resultSet = QubitSearch::getInstance()->index->getType('QubitInformationObject')->search($this->search->query);
+
+    foreach ($resultSet as $hit)
+    {
+      $resource = QubitInformationObject::getById($hit->getId());
```
MikeFE, line 203: I know it's an edge case, but what if the ES index and MySQL get out of sync? We could get null returned by QubitInformationObject::getById() if our ES index thinks an info object is there when it isn't (I actually get this often when developing, when dealing with sql dumps and the like, but forgetting to search:populate again after). I just mention it since I ran into this when testing this feature and exportResource() crashes if you pass it null.

Maybe we could add here:

```
if (!$resource)
{
  $this->info('ES search result information object not found in MySQL, skipping... (you may need to re-run search:populate)');
  continue;
}
```

jraddaoui:
Hi @mcantelon , do we have a public ticket in Redmine for this PR? We should refer to that ticket when it's merged.

MikeFE:
Can we close this PR, Mike?

mcantelon:
Okay, I'll make a public issue and will merge this today.

mcantelon:
OK, I've merge this, referring  to public issue #8563

PR #175, opened 2015-05-23
Removed unneeded CSV export columns. refs #8384

mcantelon:
Removed CSV export columns that interfere with roundtripping.


mcantelon:
8385 is the issue actually.

mcantelon:
8335 actually agh.

PR #176, opened 2015-05-25
Respect UI-configured max image width generally

zacho:
When importing descriptions with digital objects either on the command-line or through the UI, the ui-configured maximum image width setting is not respected. This is likewise true when issuing the digitalobject:regen-derivatives command. It seems the wrong (legacy?) setting is checked. FWIW, changing reference_image_maxwidth in data/fixtures/settings.yml doesn't help. This change checks the UI value for maxwidth and uses it.

video: https://youtu.be/LEvv6N_mlnM

This little change works for me. the double casting is ugly, but I don't understand the object structure well enough to go directly for the integer value of the setting.

### DIFF
--- a/lib/model/QubitDigitalObject.php
+++ b/lib/model/QubitDigitalObject.php
@@ -2329,7 +2329,7 @@ class QubitDigitalObject extends BaseDigitalObject
   {
     if ($usageId == QubitTerm::REFERENCE_ID)
     {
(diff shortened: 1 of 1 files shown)


jraddaoui:
Good catch @zacho !

The fix looks great, but we have this problem in other tasks, where the settings are not added to sfConfig. I've proposed a fix in https://projects.artefactual.com/issues/7652#note-6 that will probably fix this too.

Thanks!

zacho:
Thanks, I expect you're right.

PR #186, opened 2015-06-05
Dev/issue 8335 csv export fixes

mcantelon:
Fixes to CSV export to fix roundtripping issues:
- Added "event" column series to CSV export
- Changing "creator" column series (input and export) so these columns just add creator and creator history (in ISAD creators and creation events aren't linked)
- Added migration to add missing RAD note terms
- Added genre access point to RAD CSV import and output
- Changed RAD CSV import and output so normal general notes, not old RAD-specific ones, roundtrip
- Added export of digital object URLs
- Added method to information object class to return digital object URL


lib/model/QubitInformationObject.php
```diff
@@ -1236,6 +1241,40 @@ function($row)
...
   }
 
+  /**
+   * Get the digital object's public URL
+   *
+   * @return string  digital object URL or null
+   */
+  public function getDigitalObjectPublicUrl()
+  {
+    require_once(dirname(__FILE__).'/../../vendor/symfony/lib/helper/UrlHelper.php');
```
SymianFE: Is it possible to use something like: sfContext::getInstance()->getConfiguration()->loadHelpers('Url'); here?
Or alternatively if for some reason we can't do that in CLI tasks, can we try something like: require_once(sfConfig::get('sf_root_dir').DIRECTORY_SEPARATOR.'vendor/symfony/lib/helper/UrlHelper.php')

Just to avoid hard coding folder paths. In the off chance we ever moved tasks around under the tasks/ folder, it could break here.

MikeFE: Is it possible to use something like: sfContext::getInstance()->getConfiguration()->loadHelpers('Url'); here?
Or alternatively if for some reason we can't do that in CLI tasks, can we try something like: require_once(sfConfig::get('sf_root_dir').'vendor/symfony/lib/helper/UrlHelper.php') ?

Just to avoid hard coding folder paths. In the off chance we ever moved tasks around under the tasks/ folder, it could break here. Ideally we'd use the DIRECTORY_SEPARATOR constant too instead of '/'s, but I don't think it's a big deal since / works on both Linux and Windows.

lib/model/QubitInformationObject.php
```diff
@@ -1236,6 +1241,40 @@ function($row)
...
+
+    if (isset($do))
+    {
+      $path = $do->getFullPath();
+
+      // if path is external, it's absolute so return it
+      if (QubitTerm::EXTERNAL_URI_ID == $do->usageId)
+      {
+        return $path;
+      } else
```
MikeFE: minor tweak, else on a new line here

PR #187, opened 2015-06-08
Generate longer and more readable random slugs

gioele:
The new code generates fixed-length random slugs, divided in small blocks and separated by a hyphen (configurable). Characters such as `0`, `o`, `1` or `l` are avoided to prevent reading errors. All these measures greatly improve the readability of the URLs based on the random slugs.

These new slugs are also much more likely to be unique.

### DIFF
--- a/lib/model/QubitSlug.php
+++ b/lib/model/QubitSlug.php
@@ -19,26 +19,26 @@
 
 class QubitSlug extends BaseSlug
 {
-  public static function random()
+  public static function random($length = 12)
   {
-    $slug = null;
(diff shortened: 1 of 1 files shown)


mcantelon:
This code seems to work well... I'm able to generate 1,000,000 slugs repeatedly with no collisions.

MikeFE:
Merged (https://github.com/artefactual/atom/commit/b176ee00ed6fe5a046fbe7026e184e7d02788d66).

Thanks for the enhancement Gioele :)

PR #192, opened 2015-06-17
Made export task class methods static. refs #8534

mcantelon:
Made a few export task class methods static so ad-hoc export scripts
can more easily be created using them (for exports where the selection is more complicated than simple SQL criteria).


mcantelon:
Merged!

PR #193, opened 2015-06-17
Dev/issue 8563 search export fixes

mcantelon:
Fixed a couple of issues with CSV export of advanced search results:
- Removed filtering out of drafts
- Set search result limit to 1 billion (as there doesn't seem to be a way to return all in Elastica: https://groups.google.com/forum/#!topic/elastica-php-client/us512h6gv-s)


mcantelon:
Merged.

PR #194, opened 2015-06-22
Avoid printing '|'s when blank column value #8571

MikeFE:
If we had an array full of NULLs (can happen in some i18n situations), AtoM would just
output separators ('|'s) into the CSV column instead of leaving it blank. This fix
will prevent us from inserting useless NULLs into our column value arrays.


mcantelon:
Looks good to me.

PR #197, opened 2015-06-26
Set language with model functionality. refs #8611

mcantelon:
When setting language-related properties during CSV import of
information objects, the import code was duplicating logic already
handled in the model.


mcantelon:
Yeah, seems like it might not work.

PR #199, opened 2015-06-30
Improve worker/jobs logging, refs #8620

jraddaoui:
- Include timestamp in log messages
- Redirect all logs to worker log file
- Remove logged output from exec() calls in arGenerateFindingAidJob


jraddaoui:
Merged in qa/2.2.x

PR #202, opened 2015-07-09
Add date range query in adv. search, refs #8648

jraddaoui:
- Uses Drupal jQuery datepicker and Symfony validation
- Make events nested type objects in the ES index to allow
  querying over the start and end dates from the same event
- Fix datepicker position to avoid overlapping the input field
- Included date range query in arSearchResultExportCsvJob
- Normalize event dates without month or day in the ES index
- Don't throw exception if adv. search form is not valid

### DIFF
--- a/apps/qubit/modules/search/actions/advancedAction.class.php
+++ b/apps/qubit/modules/search/actions/advancedAction.class.php
@@ -28,7 +28,9 @@ class SearchAdvancedAction extends DefaultBrowseAction
       't', // 'mediaType',
       'r', // 'repository',
       'f', // 'fonds/collection'
-      's' // 'searchFields'
+      's', // 'searchFields'
+      'sd', // 'startDate'
+      'ed' // 'endDate'
     );
 
   protected function addField($name)
@@ -173,6 +175,16 @@ class SearchAdvancedAction extends DefaultBrowseAction
         $this->form->setWidget($name, new sfWidgetFormSelect(array('choices' => $choices)));
 
         break;
+
+      case 'sd':
+      case 'ed':
+        $this->form->setValidator($name, new sfValidatorString);
+        $this->form->setWidget($name, new sfWidgetFormInput(array(), array('placeholder' => 'YYYY-MM-DD')));
+        $this->form->setValidator($name, new sfValidatorDate(array(
+          'date_format' => '/^(?P<year>\d{4})-(?P<month>\d{2})-(?P<day>\d{2})$/',
+          'date_format_error' => 'YYYY-MM-DD')));
+
+        break;
     }
   }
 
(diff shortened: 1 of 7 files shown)


apps/qubit/modules/search/actions/advancedAction.class.php
```diff
@@ -450,6 +462,65 @@ static function addToQueryBool(&$queryBool, $operator, $queryField)
...
+  {
+    if (empty($start) && empty($end))
+    {
+      return;
+    }
+
+    $query = new \Elastica\Query\Bool();
+    $range = array();
+
+    if (!empty($start))
```
MikeFE: Maybe we could add && empty($end) to the condition here, and likewise && empty($start) in the second if() below, that way it isn't assigning all these variables when it doesn't need to, if I've understood the code correctly

MikeFE, line 475: Could we turn these 3 ifs into an if/else if/else? like:

if (!empty($start) && empty($end)) { ... } else if (!empty($end) && empty($start)) { ... } else { ...}

That way all those variables aren't assigned when I think they don't need to. I could be wrong with understanding what you're trying to do though.

apps/qubit/modules/search/actions/advancedAction.class.php
```diff
@@ -501,6 +572,12 @@ public function execute($request)
       }
     }
 
+    // Process date range
+    if (null !== $criterias = $this->getDateRangeQuery($this->request['sd'], $this->request['ed']))
```
MikeFE: you can just name this variable 'criteria', since it's already plural (criterion is for singular)

lib/job/arSearchResultExportCsvJob.class.php
```diff
@@ -111,6 +111,12 @@ protected function addCriteriaBasedOnSearchParameters()
       }
     }
 
+    // Add criteria for date range
+    if (null !== $criterias = SearchAdvancedAction::getDateRangeQuery($this->searchParams['sd'], $this->searchParams['ed']))
```
MikeFE: same here as above

apps/qubit/modules/search/actions/advancedAction.class.php
jraddaoui, line 475: Hi Mike, it must be that way. If only one date is populated it will enter in only one if and add one query, but if both dates are populated it must enter in the three ifs and add the three queries. I did it like you're suggesting at the begining to sepparate the three cases more clearly, but in the end, this is the best way to avoid duplicated code/queries inside de ifs.

MikeFE, line 475: Ah okay, yeah I wasn't sure if there was a reason. Makes sense then.

PR #209, opened 2015-07-20
Dev/issue 8551 treeview optimize

mcantelon:
Treeview and digital object viewer optimizations. A number of tweaks that speed the treeview up a lot and prevents out-of-memory errors when attempting to count a description and descendent's digital objects.


mcantelon:
Merged.

PR #211, opened 2015-07-23
Allow slugs to be gen'd by title or ref code #8638

MikeFE:
Add option for user to set whether we generated slugs for information objects based on their title or reference code.

Also take this new setting into account when using the propel:generate-slugs task.

### DIFF
--- a/lib/model/QubitInformationObject.php
+++ b/lib/model/QubitInformationObject.php
@@ -115,42 +115,7 @@ class QubitInformationObject extends BaseInformationObject
 
         if (sfConfig::get('app_inherit_code_informationobject'))
         {
-          if (!isset($this->identifier))
-          {
-            return;
-          }
-
-          $identifier = array();
-          $repository = null;
-          foreach ($this->ancestors->andSelf()->orderBy('lft') as $item)
-          {
-            if (isset($item->identifier))
-            {
-              $identifier[] = $item->identifier;
-            }
-
-            if (isset($item->repository))
-            {
-              $repository = $item->repository;
-            }
-          }
-          $identifier = implode(sfConfig::get('app_separator_character', '-'), $identifier);
-
-          if (isset($repository->identifier))
-          {
-            $identifier = "$repository->identifier $identifier";
-          }
-
-          if (isset($repository))
-          {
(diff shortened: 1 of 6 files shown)


lib/model/QubitInformationObject.php
```diff
@@ -2913,4 +2913,23 @@ public function getCleanExtentAndMedium($options = array())
...
+  public function getSlugBasisValue()
+  {
+    switch ((int)QubitSetting::getByName('slug_basis_informationobject'))
+    {
+      case QubitSlug::SLUG_BASIS_REFERENCE_CODE:
+        return $this->getIdentifier();
+
+      case QubitSlug::SLUG_BASIS_TITLE:
+      default:
+        return $this->getTitle(array('cultureFallback' => true));
```
jraddaoui: We were using the source culture to create the slug.

lib/model/QubitInformationObject.php
```diff
@@ -2913,4 +2913,23 @@ public function getCleanExtentAndMedium($options = array())
...
+   * and return the value of it.
+   *
+   * @return string  the value of the property
+   */
+  public function getSlugBasisValue()
+  {
+    switch ((int)QubitSetting::getByName('slug_basis_informationobject'))
+    {
+      case QubitSlug::SLUG_BASIS_REFERENCE_CODE:
+        return $this->getIdentifier();
```
jraddaoui: Shouldn't we use the full reference code?

lib/task/propel/propelGenerateSlugsTask.class.php
```diff
@@ -161,6 +162,24 @@ public function execute($arguments = array(), $options = array())
...
+
     $this->logSection('propel', 'Done!');
   }
+
+  private function getInformationObjectSlugBasis()
+  {
+    switch ((int)QubitSetting::getByName('slug_basis_informationobject'))
+    {
+      case QubitSlug::SLUG_BASIS_REFERENCE_CODE:
+        return 'base.identifier';
```
jraddaoui: Shouldn't we use the full reference code?

jraddaoui:
Hi Mike, the commit should refer to a public ticket ;)

PR #217, opened 2015-07-30
Dev/issue 8629 dip upload

mcantelon:
Add optional, based on a setting that can be changed on a new DIP Upload settings page, stripping of extensions during DIP upload.


mcantelon:
Merged.

PR #220, opened 2015-08-08
Dev/issue 8629 rename dialog fixes

mcantelon:
Fixes and improvements to rename modal:
- Added ability to add tooltips to Bootstrap modals
- Added tooltips to rename modal
- Added notification if submitted slug had to be modified
- Added removal of special characters from filenames

### DIFF
--- a/apps/qubit/modules/informationobject/actions/renameAction.class.php
+++ b/apps/qubit/modules/informationobject/actions/renameAction.class.php
@@ -89,8 +89,20 @@ class InformationObjectRenameAction extends sfAction
       $filename = QubitSlug::slugify($fileParts['filename']) .'.'. QubitSlug::slugify($fileParts['extension']);
 
       $digitalObject = $resource->digitalObjects[0];
+
+      // Rename master file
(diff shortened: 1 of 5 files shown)


apps/qubit/modules/informationobject/actions/renameAction.class.php
```diff
@@ -57,20 +68,41 @@ private function updateFields()
...
+      $newFilePath = $basePath . DIRECTORY_SEPARATOR . $filename;
+      rename($oldFilePath, $newFilePath);
+      chmod($newFilePath, 0644);
+
+      // Change name in database
+      $digitalObject->name = $filename;
       $digitalObject->save();
+
+      // Regeneraate derivatives
+      digitalObjectRegenDerivativesTask::regenerateDerivatives($digitalObject);
```
jraddaoui, line 105: I think we should check if they exist and delete the old derivates before. I guess it's not done in the regenerateDerivatives task when the filename is changed.

jraddaoui, line 105: Never mind, I think they're actually being deleted.

mcantelon:
Merged.

PR #222, opened 2015-08-11
Fixed issue with EAD export. refs #8794

mcantelon:
Fixed issue with EAD export where if a name access point is a term,
rather than an actor, the export would crash.


mcantelon:
Merged.

PR #225, opened 2015-08-17
Dev/issue 8629 rename dialog v3

mcantelon:
- Changed rename modal to use the Symfony form API
- Fix a CSS z-index conflict between Imageflow and Bootstrap dropdown menus
- Fix issue with deleting digital object derivatives (the "descendents" property of the digital object model was recently removed so the logic to find descendents had to be updated)


sevein:
@mcantelon note that we abandoned the `lib/form/*` pattern to build forms, we only keep doing that in the settings page. The way to build forms these days is done by subclassing DefaultEditAction.

apps/qubit/modules/informationobject/templates/_renameModal.php
```diff
@@ -13,29 +13,27 @@
...
-      <div style="float:right"><input id="renameModalEnableTitle" type="checkbox" checked="checked" /> <?php echo __('Update title') ?></div>
-      <label><?php echo __('Description title') ?></label>
-      <input id="renameModalTitle" name="title" type="text" value="<?php echo $resource->title ?>" />
-      <div class="description">
-        <?php echo __('Editing the description title will automatically update the slug field if the "Update slug" checkbox is selected - you can still edit it after.') ?>
-      </div>
+      <div style="float:right"><input id="rename_enable_title" type="checkbox" checked="checked" /> <?php echo __('Update title') ?></div>
+      <label><?php echo $renameForm->title->label ?></label>
+      <?php echo $renameForm->title ?>
+      <div class="description"><?php echo $renameForm->title->help ?></div>
```
jraddaoui: Hi Mike.

The label must be related to the input using the "for" and "id" attributes. There is a couple of easy ways to do it, in this three fields I think you can replace the three lines of each field for one of this two options (depending on how translations are handled in this feature):

<?php echo render_field($renameForm->title), $resource) ?> (with translations)
<?php echo $renameForm->title->renderRow() ?> (without translations)

apps/qubit/modules/informationobject/actions/renameAction.class.php
```diff
@@ -37,55 +37,75 @@ public function execute($request)
...
 
-    $resource = $this->updateFields();
-
-    // Let user know description was updated (and if slug had to be adjusted)
+    // Internationalization needed for flash messages
     ProjectConfiguration::getActive()->loadHelpers('I18N');
 
-    $message = __('Description updated.');
+    // Handle rename form submission
+    if (null !== $request->rename)
```
sevein: You probably wanted to use here something like `if ($request->isMethod('post')) { ... }`, e.g.: https://github.com/artefactual/atom/blob/qa/2.3.x/apps/qubit/modules/informationobject/actions/editAction.class.php#L617-L628

sevein:
I wish we have designed this feature as a regular page instead of a dialog, this is such a waste of energy solving unnecessary complexities for very little usability benefit. I'm going to write @fiver-watson to see if we are still in time to revert this decision.

mcantelon:
I ended up changing the form to a component (as it needs to be included in another action).

apps/qubit/modules/informationobject/actions/renameAction.class.php
```diff
@@ -22,39 +22,28 @@ class InformationObjectRenameAction extends sfAction
...
   public function execute($request)
   {
-    // Return 401 if unauthorized
-    if (!sfContext::getInstance()->user->isAuthenticated()
-      || !QubitAcl::check($this->resource, 'update'))
-    {
-      $this->response->setStatusCode(401);
-      return sfView::NONE;
-    }
-
```
jraddaoui, line 32: I think it's a good idea to keep the ACL check, to avoid manually modified requests.

apps/qubit/modules/informationobject/actions/renameAction.class.php
```diff
@@ -22,39 +22,28 @@ class InformationObjectRenameAction extends sfAction
...
-    ProjectConfiguration::getActive()->loadHelpers('I18N');
+      $renameComponent = new InformationObjectRenameComponent($this->context, 'informationobject', 'index');
+      $renameComponent->resource = $resource;
+      $renameComponent->execute($request);
 
-    // Handle rename form submission
-    if (null !== $request->rename)
-    {
-      $this->renameForm = new InformationObjectRenameForm;
+      $this->form = $renameComponent->form;
```
jraddaoui, line 34: I'd move the flash add and the redirect (L33-L55) to the 'updateResource' method in the 'renameComponent', where the form is already binded and valid.

apps/qubit/modules/informationobject/actions/renameComponent.class.php
```diff
@@ -0,0 +1,129 @@
...
+    foreach(array('title', 'slug', 'filename') as $name)
+    {
+      $this->form->setValidator($name, new sfValidatorString(array('required'=>false)));
+      $this->form->setWidget($name, new sfWidgetFormInput);
+    }
+
+    $i18n = sfContext::getInstance()->i18n;
+
+    // Add labels
+    $this->form->getWidgetSchema()->setLabels(array(
```
sevein, line 36: We normally set the labels in the template, e.g. https://github.com/artefactual/atom/blob/qa/2.3.x/apps/qubit/modules/menu/templates/editSuccess.php#L39-L42.

apps/qubit/modules/informationobject/actions/renameComponent.class.php
```diff
@@ -0,0 +1,129 @@
...
+      'title' => $i18n->__('Description title'),
+      'slug' => $i18n->__('Slug'),
+      'filename' => $i18n->__('File name')
+    ));
+
+    // Add helper text
+    $this->form->getWidgetSchema()->setHelps(array(
```
sevein, line 43: We normally set the help messages in the template, e.g. https://github.com/artefactual/atom/blob/qa/2.3.x/apps/qubit/modules/menu/templates/editSuccess.php#L39-L42.

plugins/arDominionPlugin/css/less/forms.less
```diff
@@ -355,6 +355,25 @@ button.delete-small {
...
+#renameModalForm {
+  div.modal-body {
+    div.form-item {
+      label {
+        display: inline;
+      }
+    }
+  }
+}
+
```
sevein, line 376: I'm not sure why you need both elements to be inline, but just regarding on the syntax, could you convert that snippet into something equivalent like the following?

``` css
#renameModalForm .modal-body {
  .form-item, label {
    display: inline;
  }
}
```

apps/qubit/modules/informationobject/templates/_renameModal.php
```diff
@@ -9,45 +9,32 @@
 
   <form id="renameModalForm" action="<?php echo url_for(array('module' => 'informationobject', 'action' => 'rename', 'slug' => $resource->slug)) ?>" method="POST">
```
sevein, line 10: Can you use renderFormTag() here?

apps/qubit/modules/informationobject/templates/_renameModal.php
```diff
@@ -9,45 +9,32 @@
...
 
-  <div class="modal-body">
-    <div class="alert"><?php echo __('Use this interface to update the description title, slug (permalink), and/or digital object filename.') ?></div>
+    <div class="modal-body">
+      <div class="alert"><?php echo __('Use this interface to update the description title, slug (permalink), and/or digital object filename.') ?></div>
 
-    <div>
       <div style="float:right"><input id="rename_enable_title" type="checkbox" checked="checked" /> <?php echo __('Update title') ?></div>
```
sevein, line 15: Can you avoid inline CSS if possible? Also, for id and class attributes we normally use dashes (e.g. rename-enable-title). You may find some places where camelCase is used, but that's being changed.

mcantelon:
It was decided to implement this as a page, not a modal.

PR #226, opened 2015-08-18
[WIP] Add Basis to Admin » Rights setting page, refs #8769

sevein:
Redmine ticket: https://projects.artefactual.com/issues/8769

See the commit `6db1d9a` for a full description. And the following is a screenshot of the new settings/permission page, so you have a better idea of how this is going to look like.

Screenshot

### DIFF
--- a/js/permissionsSettings.js
+++ b/js/permissionsSettings.js
@@ -1,0 +1,65 @@
+(function ($)
+  {
+    $(document).ready(function() {
+
+      var $area = $('#premisAccessPermissionsArea');
+
+      $area.find('.cbx')
+        .hover(function () {
+          $(this).addClass('hover');
+        }, function () {
+          $(this).removeClass('hover');
+        })
+        .on('click', function (event) {
+          if (event.target.tagName !== 'INPUT') {
+            $(event.target).find('input').trigger('click');
+          }
+        });
+
+      $area.find('.all').on('click', function (event) {
+        event.preventDefault();
+        $area.find('input').prop('checked', true);
+      });
+
+      $area.find('.none').on('click', function (event) {
+        event.preventDefault();
+        $area.find('input').prop('checked', false);
+      });
+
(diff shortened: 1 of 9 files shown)


js/permissionsSettings.js
```diff
@@ -0,0 +1,65 @@
...
+        var index;
+        $area.find('th.premis-permissions-mrt li').each(function (i) {
+          if ($li.is(this)) {
+            index = i;
+            return false;
+          }
+        });
+
+        // Create a jQuery object with all the checkboxes of the column in it
+        var cbs = jQuery();
```
mcantelon: Should "cbs" be changed to "$cbs" (because it's a jQuery object)?

lib/form/SettingsPermissionsForm.class.php
```diff
@@ -26,85 +26,69 @@
...
+    $defaults = QubitSetting::$premisAccessRightValueDefaults;
+
+    $form = new sfForm;
+    $form->getValidatorSchema()->setOption('allow_extra_fields', true);
+
+    foreach (QubitTaxonomy::getTermsById(QubitTaxonomy::RIGHT_BASIS_ID) as $item)
     {
-      $this->premisAccessRight = QubitSetting::createNewSetting('premisAccessRight', 'delete', array('deleteable' => false));
-      $this->premisAccessRight->save();
+      $f = new sfForm;
```
mcantelon: I think longer variable names make things clearer, but it's a matter of taste as they also add more text which could be considered visual clutter.

sevein: I think you are right, I'll fix it. I've reading too much Go code recently.

js/permissionsSettings.js
sevein: Right! It's a good practice so remember that we are dealing with jQuery objects. I'll fix it.

PR #230, opened 2015-08-25
Add missing alt tags, refs #8855

jraddaoui:
Alt values from discussion with DRMC:

https://docs.google.com/spreadsheets/d/1Y0ofG6SVkYrBKXp6pPjWKFSVB-VZ_bveOI7rWl4nztk

### DIFF
--- a/apps/qubit/modules/contactinformation/templates/_edit.php
+++ b/apps/qubit/modules/contactinformation/templates/_edit.php
@@ -19,7 +19,6 @@
         </th><th style="width: 15%">
           <?php echo __('Primary') ?>
         </th><th style="text-align: center; width: 10%">
-          <?php echo image_tag('delete', array('align' => 'top', 'class' => 'deleteIcon')) ?>
         </th>
       </tr>
     </thead><tbody>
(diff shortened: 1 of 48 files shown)


apps/qubit/modules/contactinformation/templates/_edit.php
```diff
@@ -19,7 +19,6 @@
         </th><th style="width: 15%">
           <?php echo __('Primary') ?>
         </th><th style="text-align: center; width: 10%">
-          <?php echo image_tag('delete', array('align' => 'top', 'class' => 'deleteIcon')) ?>
```
MikeFE, line 22: Was this intentionally removed and not replaced with something?

jraddaoui, line 22: They are being removed with JS, so I just removed the images, I'll remove the following line in the JS too:

https://github.com/artefactual/atom/blob/qa/2.3.x/js/multiDelete.js#L46

jraddaoui:
Merged: https://github.com/artefactual/atom/commit/2c70ef8cdcfcb2cf838f195b1ad4708a39d2bf23

PR #238, opened 2015-09-10
Add clipboard functionality, refs #8941

jraddaoui:
- Add QubitClipboard class to manage selected IO slugs in the
  current Symfony user storage
- Add toogle buttons in the IO index page and in the search results
- Add header menu (migration and fixtures)
- Add actions and templates to manage the clipboard
- Add clipboard jQuery prototype to manage Ajax requests
- Add clipboard unit test

### DIFF
--- a/apps/qubit/config/view.yml
+++ b/apps/qubit/config/view.yml
@@ -47,6 +47,7 @@ default:
     /vendor/meetselva/attrchange:
     qubit:
     treeView:
+    clipboard:
     dominion: { position: last }
 
   has_layout:     true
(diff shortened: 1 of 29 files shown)


jraddaoui:
Sorry @sevein , it looks like I deleted your comments after fixing the same commit with your suggestions and doing a 'push -f'.

I made all your suggestions except moving the search code to QubitClipboard (or a new QubitClipboardQuery). In UserClipboardAction we are using sorting and pagination, and we may want to use filters in the future, so I thought it was better to extend DefaultBrowseAction and use arElasticSearchPluginQuery. In UserClipboardClearAction we are getting the clipboard slugs and redirecting first if the clipboard is empty; if the clipboard is not empty, we are doing a simple query with the slugs already extracted.

But let me know if you really think it needs to be changed ;)

sevein:
LGTM but let's wait for @mcantelon, lots of new things here.

jraddaoui:
Merged in https://github.com/artefactual/atom/commit/3c1a0fff186cedf19bca75e55c96cc22bad991cd

PR #239, opened 2015-09-13
Dev/issue 8629 rename dialog v4

mcantelon:
I've redone the rename functionality so the edit form is a separate page instead of a modal.

On the page template (renameSuccess.php) there are some <BR />. They are in there because if the field enable/disable checkboxes are shown directly to the right of the field labels there's some CSS weirdness where the field labels somehow prevent clicking on the checkboxes. In my earlier PR I'd used some CSS tweaks to fix this (making some elements inline) but thought maybe the workaround approach is least hack-y.


mcantelon:
Merged.

PR #243, opened 2015-09-29
[Atom 2.2] Fixed PHP 5.3 issue with CSV import. refs #8981

mcantelon:
Fixed issue that occurs in older version of PHP 5.


mcantelon:
Merged.

PR #245, opened 2015-10-01
Fixed issues with rename page. refs #8759

mcantelon:
Fixed a number of issues with the rename page:
- Added alert if new slug is already used
- Prevent the slug field from being auto-updated if disabled
- Added availability checking to slug field
- Fix issue with repository quota setting in sidebar
- Fixed issue with background AJAX operation bookkeeping


mcantelon:
Merged but forgot to close

PR #268, opened 2016-01-26
Fix OAI Identify response

drbauder:
Redmine: https://projects.artefactual.com/issues/9395

I noticed that the XML returned by the OAI Identify response does not validate. Validation would fail because of the format of the repositoryIdentifer and sampleIdentifier elements. This patch fixes the output of those elements and ensures validation succeeds.

This is my first pull request here, so any and all feedback is welcome.

Refer to the following:
- http://www.openarchives.org/OAI/2.0/openarchivesprotocol.htm#Identify, for information on the Identify response
- http://www.openarchives.org/OAI/2.0/guidelines.htm, section 3.1, for information on the description container
- http://www.openarchives.org/OAI/2.0/guidelines-oai-identifier.htm, for specifics on the oai-identifier format

### DIFF
--- a/apps/qubit/modules/settings/actions/oaiAction.class.php
+++ b/apps/qubit/modules/settings/actions/oaiAction.class.php
@@ -25,6 +25,7 @@
  * @author     Peter Van Garderen <peter@artefactual.com>
  * @author     Jack Bates <jack@nottheoilrig.com>
  * @author     David Juhasz <david@artefactual.com>
+ * @author     Damian Bauder <drbauder@ucalgary.ca>
  */
 
 class SettingsOaiAction extends sfAction
--- a/apps/qubit/modules/settings/actions/oaiAction.class.php
+++ b/apps/qubit/modules/settings/actions/oaiAction.class.php
@@ -70,7 +70,7 @@ class SettingsOaiAction extends sfAction
     $oaiRepositoryCode = QubitSetting::getByName('oai_repository_code');
     $oaiAdminEmails = QubitSetting::getByName('oai_admin_emails');
     $oaiRepositoryIdentifier = QubitOai::getRepositoryIdentifier();
-    $sampleOaiIdentifier = QubitOai::getSampleIdentifier();
+    $sampleOaiIdentifier = QubitOai::getOaiSampleIdentifier();
     $resumptionTokenLimit = QubitSetting::getByName('resumption_token_limit');
     $oaiAdditionalSetsEnabled = QubitSetting::getByName('oai_additional_sets_enabled');
 
--- a/lib/QubitOai.class.php
+++ b/lib/QubitOai.class.php
@@ -24,6 +24,7 @@
  * @subpackage lib
  * @author     Mathieu Fortin Library and Archives Canada <mathieu.fortin@lac-bac.gc.ca>
  * @author     Peter Van Garderen <peter@artefactual.com>
+ * @author     Damian Bauder <drbauder@ucalgary.ca>
  */
 
 class QubitOai
--- a/lib/QubitOai.class.php
+++ b/lib/QubitOai.class.php
@@ -277,24 +277,31 @@ class QubitOai
     }
     return false;
   }
+  
+  public static function getOaiNamespaceIdentifier()
+  {
+	$oaiNamespaceIdentifier = sfContext::getInstance()->request->getHost();
+    
+    return $oaiNamespaceIdentifier; 
+  }
 
   public static function getRepositoryIdentifier()
   {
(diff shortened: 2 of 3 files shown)


sevein:
@drbauder, can you get rid of e8ff5dc? You should probably use `git-rebase` instead to keep your branch up to date. Thank you!

drbauder:
@sevein, there you go! Thanks for the tip.

I'm working on getting a Contributor's Agreement signed for our institution, but hopefully that won't hold up any other review you want to do.

sevein:
That looks much better, thank you.

drbauder:
The only one that hurts my soul is the lack of whitespace around an operator ;-)

sevein:
On Wed, Jan 27, 2016 at 8:26 AM, Damian Bauder notifications@github.com
wrote:

> The only one that hurts my soul is the lack of whitespace around an
> operator ;-)
> 
> We borrowed that habit from Symfony 1.x, see https://git.io/vz9r7.

## 

Jesús García Crespo,

sevein:
@drbauder, thank you so much! I've asked @fiver-watson to test it.

sevein:
@fiver-watson and I have been testing this locally and it seems to work as expected. Something that we've noticed though is that `<baseURL>` is populated using [a different function in the QubitOai class (getBaseUrl)](https://github.com/drbauder/atom/blob/fix-oai-identify-response/lib/QubitOai.class.php#L168-L181). How do you feel about updating it to take advantage of the `siteBaseUrl` setting? That function might as well be used by `getOaiNamespaceIdentifier`.

drbauder:
@sevein, I'm not sure I'm clear. Are you suggesting that we update `getBaseUrl` to make use of the `siteBaseUrl` setting, and then call `getBaseUrl` in `getOaiNamespaceIdentifier` as well as in `<baseUrl>`? That makes sense to me for a number of reasons, but before I get too far into it, I want to make sure I'm understanding your comment!

fiver-watson:
Hi @drbauder , I'll let @sevein respond re: the technical aspects, but from my perspective: 

Your changes looked great! While I was testing I happened to notice that the <baseURL> value in the Indentify response was not populating correctly. I asked @sevein to check if this was a previously unnoticed bug or due to a change that you'd made. It's nothing you did, but he noticed that the way the Base URL is being fetched for the <baseURL> output in the response is using a different function than the one you used in your changes to the repositoryidentifier... which works! As far as I understand, the proposal was to see if you were willing to address that issue as well, and populate the <baseURL> value using the `siteBaseUrl` method/function/whatever... beyond this, I wasn't clear if @sevein was also suggesting changes to `getOaiNamespaceIdentifier`; I will let him clarify. 

If you don't have the time to take this on, I think your PR looks fine - hopefully we can fix the <baseURL> issue I noticed at a future date. If you were willing though, we'd be grateful!

drbauder:
Thanks, @fiver-Watson. I don't mind adding to the changes. It is probably a pretty easy fix to throw in. The one thing I'm not clear on, that maybe @sevein can fill me in on, is whether `SERVER['SERVER_NAME']` will contain the proper domain in a situation with a proxy, which was why he suggested using `siteBaseUrl` in the first place. If `SERVER['SERVER_NAME']` actually plays nice with proxies, it might make more sense to use `getBaseUrl` (almost) as-is. It returns a nicely formatted URL that parse_url should get along with quite well.

I say "almost", above, because I noticed a tiny bug in `getBaseUrl` that will return http as the scheme while adding 443 as a port, which will result in a 400 response code should anything try to use it.

Either way, I can drop in whatever fixes are necessary once @sevein fills me in.

sevein:
I'd prefer to relay in `siteBaseUrl` because it'd be more predictable for AtoM users that are already familiar with that setting. IIRC, `siteBaseUrl` had to be added because `$_SERVER` is not available in the command line environment, which is frequently used by AtoM users for import/export purposes, e.g. `siteBaseUrl` is used in EAD export, which is a functionality also exposed in the command line.
It'd be probably simpler if we rely on `siteBaseUrl` everywhere rather than implementing two different methods through the application. Also, because `siteBaseUrl` is manually defined by the user, proxies won't ever be a problem. Does that make sense?

drbauder:
@sevein and @fiver-watson, I made a few changes. Please take a look!

sevein:
LGTM! Thank you so much.
We'll merge soon. @fiver-watson, do you want to give it another test to verify?

drbauder:
You are very welcome! Thanks to you both for the review. Hopefully everything works this time through.

sevein:
Merged and squashed @ 712ce7b

PR #270, opened 2016-01-30
Update pub. status via job scheduler, refs #9304

jraddaoui:
- Remove pub. status field from IO edit forms
- Use default pub. status setting on IO creation
- Add update pub. status link in IO actions more dropdown,
  only added for user with publish permissions over the IO
- Add form action and template, only accessible for users
  with publish permission
- Update current resource synchronously
- Update descendants (if required) using job scheduler
- Create arUpdatePublicationStatusJob and add it to the worker


jraddaoui:
Merged with --squash

PR #271, opened 2016-01-31
Change fields affected by expander, refs #9266

jraddaoui:
Apply expander to all fields except those with only one child,
if that child is a list


sevein:
:+1:

PR #276, opened 2016-02-11
Validate base URL for tools:purge, refs #9431

MikeFE:
Previously users could specify any base url they wanted, which was causing a bug in XML generation when they had "s in the URL.


sevein:
LGTM

PR #277, opened 2016-02-11
Use objectId in publication status job, refs #9304

jraddaoui:
Using the objectId parameter adds a link back to the description
in the browse jobs page.


sevein:
:+1: if you change the commit log message as agreed in the chat!

PR #280, opened 2016-02-19
Fix genre term creation issue, refs #9300

MikeFE:
Redmine ticket: https://projects.artefactual.com/issues/9300

When editing an archival description in either ISAD or DACS, adding a new genre form term in the access
points section would cause a JS error and prevent the user from saving the record changes.


sevein:
LGTM

MikeFE:
merged

PR #282, opened 2016-02-25
Make REST API token authenticated. refs #8631

mcantelon:
Add display and ability to generate/delete REST API keys via the
user management interface. Require token to be sent in HTTP
headers to access API. Remove user/password API authentication.


mcantelon:
Merged.

PR #283, opened 2016-03-01
Fix QubitTransactionFilter, refs #9401

jraddaoui:
This fix will wrap almost the entire request process and the related Propel
connection in the same transaction. The changes in the database will be
rolled back in case of a fatal error, exception or timeout in the process,
or they will be committed in successful requests at the end of the process.
PropelPDO class takes care to prevent errors due to already in progress
transactions, so other calls to beginTransaction, commit and rollback
functions in the same request will be ignored.

Remove PropelException catches in save methods that avoid the rollback.


jraddaoui:
More info in the Redmine ticket:

https://projects.artefactual.com/issues/9401#note-4

PR #285, opened 2016-03-05
Transactions process improvements, refs #9533

jraddaoui:
- Only use transactions in QubitTransactionFilter in HTTP requests.
- Commit current database transaction and start a new one before we
  dispatch the task to gearmand so the resources modified are persisted
  before the assigned worker starts processing the task.
- Use Propel to get the connection in model classes.


sevein:
Can we stop using `QubitTransactionFilter::getConnection` from our models? It's unnecessarily confusing, shouldn't we use `Propel::getConnection()` instead. That will probably require a change in the generator and commit our new models but I think it's worth it.

jraddaoui:
Merged with -squash

PR #287, opened 2016-03-10
Skip BOM if present on CSV import, refs #9543

MikeFE:


sevein:
Nice! It's really weird though that we have to print the value returned by logError, but I guess that's not the scope of this pull request :)

sevein:
:+1:

MikeFE:
Haha yeah, the way we do logging / stdout printing is kind of strange. I was just fixing the spacing a bit, I dislike all these newlines all over the place where it's under 120 chars and breaking the line just makes it look ugly.

PR #288, opened 2016-03-14
Fix EAD export of child levels. refs #9560

mcantelon:
Fixed inability of single slug export to export child EAD levels.


mcantelon:
Merged.

PR #296, opened 2016-03-21
Dev/issue 9602

MikeFE:
Include EAC and MODS plugins for export:bulk. Previously they weren't included and when the user wanted to export these, it'd fail.


sevein:
:+1:

PR #301, opened 2016-03-24
Handle Propel exception in filter, refs #9533

sevein:
We need to handle the possibility of Propel not delivering a connection where
there is none available, e.g. in the AtoM installer before the database is
configured (missing `config/config.php`). There might be other scenarios where
this is possible too and we want to handle it gracefully.

### DIFF
--- a/lib/filter/QubitTransactionFilter.class.php
+++ b/lib/filter/QubitTransactionFilter.class.php
@@ -34,31 +34,37 @@ class QubitTransactionFilter extends sfFilter
 
   public function execute($filterChain)
   {
-    self::getConnection();
-    self::$connection->beginTransaction();
+    try
+    {
+      $conn = self::getConnection();
+      $conn->beginTransaction();
+    }
+    catch (PropelException $e)
+    {
+    }
 
     try
     {
       $filterChain->execute();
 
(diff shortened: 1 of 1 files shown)


lib/filter/QubitTransactionFilter.class.php
```diff
@@ -34,31 +34,37 @@ public static function getConnection()
...
   {
-    self::getConnection();
-    self::$connection->beginTransaction();
+    try
+    {
+      $conn = self::getConnection();
+      $conn->beginTransaction();
+    }
+    catch (PropelException $e)
+    {
```
MikeFE, line 43: Should we log something here or rollback?

MikeFE:
LGTM if you don't think we need logging/rollback in that one catch block. :+1:

lib/filter/QubitTransactionFilter.class.php
sevein, line 43: Good thinking! I thought so too, but it's too early to rollback as the transaction hasn't been initialized yet - as the exception is raised before completion of `beginTransaction()`. I think logging is fine as it is because Propel is already sending an event to the logger when something goes wrong. Thanks!

PR #307, opened 2016-04-01
Fix notes source culture, refs #9578

jraddaoui:
Use current culture instead of source culture of the related
information object as it was creating notes without source
culture and notes that can't make culture fallback and can't
be translated


sevein:
👍

PR #310, opened 2016-04-07
Finding aid addons, refs: #9627 #9655 #9699 #9700

jraddaoui:
- Add option to delete or upload a finding aid using the AtoM worker.
- Check ACL update permission to generate, upload and delete.
- Changes in finding aid work to allow the different workflows.
- Add finding aid transcript and status to a DB property and to the ES index.
- Changes in the finding aid sidebar section.
- Rename finding aid job and generate action.
- Add partial update function for ES documents.
- Add missing translations in jobs.
- Show finding aid option from top-level in descendants.
- Add link in information object index page content.
- Allow download and delete after finding aid format changes.
- Add filter and boolean option in adv. search panel.
- Add button for FA download and use PDF icon..
- Small fixes and cosmetic changes.


apps/qubit/modules/informationobject/actions/uploadFindingAidAction.class.php
```diff
@@ -49,7 +49,15 @@ public function execute($request)
 
       if ($this->form->isValid() && isset($file))
       {
-        $tempPath = $file->getTempName();
+        $i18n = $this->context->i18n;
+
+        $params = array(
+          'objectId'    => $this->resource->id,
+          'description' => $i18n->__('Uploading finding aid for: ').$this->resource->getTitle(array('cultureFallback' => true)),
```
sevein: I think this is more correct:

```
'description' => $i18n->__('Uploading finding aid for: %1%', array('%1%' => $this->resource->getTitle(array('cultureFallback' => true)))),
```

It enables the translator to arrange the sentence in the way it's needed.

lib/job/arGenerateFindingAidJob.class.php
```diff
@@ -29,25 +29,45 @@ class arGenerateFindingAidJob extends arBaseJob
...
+
   public function runJob($parameters)
   {
-    $appRoot = rtrim(sfConfig::get('sf_root_dir'), '/');
+    $this->resource = QubitInformationObject::getById($parameters['objectId']);
 
-    $resource = QubitInformationObject::getById($parameters['objectId']);
-
-    if (!$resource)
+    if (!$this->resource)
```
sevein: `if (null === $this->resource)`?

apps/qubit/modules/informationobject/actions/uploadFindingAidAction.class.php
```diff
@@ -0,0 +1,66 @@
...
+
+    // Check user authorization
+    if (!$this->context->user->isAuthenticated())
+    {
+      QubitAcl::forwardUnauthorized();
+    }
+
+    // Create form for file upload
+    $this->form = new sfForm;
+    $this->form->getValidatorSchema()->setOption('allow_extra_fields', true);
```
sevein: Do we really need `allow_extra_fields`? Maybe not?

apps/qubit/modules/informationobject/actions/uploadFindingAidAction.class.php
```diff
@@ -0,0 +1,66 @@
...
+    $this->form->setWidget('file', new sfWidgetFormInputFile);
```
sevein: This could be a good candidate for the `accept="application/pdf"` attribute? I think you can pass examples to the constructor or to `render()`.
- http://caniuse.com/#feat=input-file-accept
- http://www.wufoo.com/html5/attributes/07-accept.html

apps/qubit/modules/informationobject/actions/uploadFindingAidAction.class.php
```diff
@@ -0,0 +1,66 @@
...
+    $this->form->setValidator('file', new sfValidatorFile);
+
+    // Process form
+    if ($request->isMethod('post'))
+    {
+      $this->form->bind($request->getPostParameters(), $request->getFiles());
+      $file = $this->form->getValue('file');
+
+      if ($this->form->isValid() && isset($file))
```
sevein: Can `isset($file)` ever be `False` considering that sfValidatorFile won't validate the field as `required` defaults to `True` in `sfValidatorBase`? Or maybe we need to pass `array('required' => true)` to `sfValidatorFile`?

apps/qubit/modules/informationobject/actions/generateFindingAidAction.class.php
```diff
@@ -0,0 +1,49 @@
...
+    if (!$this->context->user->isAuthenticated())
+    {
+      QubitAcl::forwardUnauthorized();
+    }
+
+    $i18n = $this->context->i18n;
+
+    $params = array(
+      'objectId' => $this->resource->id,
+      'description' => $i18n->__('Generating finding aid for: ').$this->resource->getTitle(array('cultureFallback' => true))
```
sevein: Use gettext param replacement? e.g. `%1%`

apps/qubit/modules/informationobject/templates/deleteFindingAidSuccess.php
```diff
@@ -0,0 +1,24 @@
...
+    <section class="actions">
+      <ul>
+        <li><?php echo link_to(__('Cancel'), array($resource, 'module' => 'informationobject'), array('class' => 'c-btn')) ?></li>
+        <li><input class="c-btn c-btn-delete" type="submit" value="<?php echo __('Delete') ?>"/></li>
+      </ul>
+    </section>
+
+  </form>
+
+<?php end_slot() ?>
```
sevein, line 33: Could you show the title of the corresponding information object and perhaps the filename or public path? The idea is to provide some context so we can make easier for the user to decide if it really should be removed.

apps/qubit/modules/informationobject/actions/deleteFindingAidAction.class.php
```diff
@@ -17,24 +17,39 @@
...
-        'objectId' => $resource->id,
-        'description' => $i18n->__('Generating finding aid for: ').$resource->getTitle(array('cultureFallback' => true))
+        'objectId'    => $this->resource->id,
+        'description' => $i18n->__('Deleting finding aid for: ').$this->resource->getTitle(array('cultureFallback' => true)),
+        'delete'      => true
       );
 
-      $job = QubitJob::runJob('arGenerateFindingAidJob', $params);
-    }
+      $job = QubitJob::runJob('arFindingAidJob', $params);
```
sevein: What do you use `$job` for? I guess because this is running asynchronously you can just fire and forget? But if it's posible to receive some kind of error you should probably check it.

apps/qubit/modules/informationobject/actions/generateFindingAidAction.class.php
```diff
@@ -0,0 +1,49 @@
...
+    );
+
+    $job = QubitJob::runJob('arFindingAidJob', $params);
```
sevein: Is `$job` unsed?

apps/qubit/modules/informationobject/actions/uploadFindingAidAction.class.php
```diff
@@ -0,0 +1,66 @@
...
+    // Create form for file upload
+    $this->form = new sfForm;
+    $this->form->getValidatorSchema()->setOption('allow_extra_fields', true);
+    $this->form->setWidget('file', new sfWidgetFormInputFile);
+    $this->form->setValidator('file', new sfValidatorFile);
+
+    // Process form
+    if ($request->isMethod('post'))
+    {
+      $this->form->bind($request->getPostParameters(), $request->getFiles());
```
sevein: Is there any post field that you need to bind to the form? I wonder if just binding getFiles() will do it?

apps/qubit/modules/informationobject/templates/uploadFindingAidSuccess.php
```diff
@@ -0,0 +1,44 @@
...
+
+    <?php echo $form->renderHiddenFields() ?>
+
+    <div id="content">
+
+      <fieldset class="collapsible">
+
+        <legend><?php echo __('Upload finding aid') ?></legend>
+
+        <?php echo $form->file->label(__('PDF file'))->renderRow() ?>
```
sevein: Not sure if you can pass HTML attributes to `renderRow()` but it'd be nice to use `accept` as mentioned earlier.

lib/job/arFindingAidJob.class.php
```diff
@@ -22,41 +22,67 @@
...
+    {
+      $this->upload($parameters['tempPath']);
+    }
+    else
+    {
+      $this->generate();
+    }
+
+    // Update resource in ES to add finding aid transcript
+    QubitSearch::getInstance()->update($this->resource);
```
sevein: If `true === $parameters['delete']` could we just delete the stuff from the document that needs to be deleted instead of repopulate? I know that we normally take the brute-force path but this could be an opportunity to start working on granular updates and perhaps implement some function helpers that enable us to do that from different places in the application?

I've found this in the ES Update API:

```
curl -XPOST 'localhost:9200/test/type1/1/_update' -d '{
    "script" : "ctx._source.remove(\"name_of_field\")"
}'
```

Is that what we need? Can we do that from Elastica?

sevein:
The reason why I was thinking that changing the name of `arGenerateFindingAidJob` may not be a good idea is because that name is used from configuration files not tracked by git, e.g. we track `config/gearman.yml` but the user may override that from `apps/qubit/config/gearman.yml`. If you want to go ahead with the new name make sure that the name change is documented in the release notes (put a note on the ticket and mark as requires documentation?). Thanks!

apps/qubit/modules/informationobject/actions/deleteFindingAidAction.class.php
```diff
@@ -17,24 +17,39 @@
...
+    if ($request->isMethod('delete'))
     {
       $i18n = $this->context->i18n;
-      $resource = $this->getRoute()->resource;
 
       $params = array(
-        'objectId' => $resource->id,
-        'description' => $i18n->__('Generating finding aid for: ').$resource->getTitle(array('cultureFallback' => true))
+        'objectId'    => $this->resource->id,
+        'description' => $i18n->__('Deleting finding aid for: ').$this->resource->getTitle(array('cultureFallback' => true)),
```
sevein: gettext!

apps/qubit/modules/informationobject/actions/uploadFindingAidAction.class.php
```diff
@@ -0,0 +1,84 @@
...
+
+    // Process form
+    if ($request->isMethod('post'))
+    {
+      $this->form->bind(array(), $request->getFiles());
+
+      if (!$this->form->isValid())
+      {
+        return;
+      }
```
sevein, line 68: I like that. Nesting is ugly!

lib/job/arFindingAidJob.class.php
```diff
@@ -125,20 +152,116 @@ public function runJob($parameters)
...
     fclose($foFileHandle);
 
-    $this->job->setStatusCompleted();
-    $this->job->save();
+    return true;
+  }
+
+  private function upload($tempPath)
+  {
+    $this->info($this->i18n->__('Uploading finding aid (%1)...', array('%1' => $this->resource->slug)));
```
sevein, line 198: Does it work when only using `%1`? Interesting, I always though we had to type `%1%` :)

lib/job/arFindingAidJob.class.php
```diff
@@ -125,20 +152,116 @@ public function runJob($parameters)
...
+    }
+
+    $this->info($this->i18n->__('Obtaining finding aid transcript...'));
+
+    $command = sprintf('pdftotext %s - 2> /dev/null', $pdfPath);
+    exec($command, $output, $status);
+
+    if ($status != 0)
+    {
+      $message = $this->i18n->__('Obtaining the transcript has failed.');
```
sevein: I think there may be a better way to say that but I'm ESL too, you may want to ask Dan...

sevein: CC @fiver-watson

fiver-watson: Looks fine to me! Really it's "transcript" that is the unclear part (the grammar is fine), but I can't think of a better term.

lib/job/arFindingAidJob.class.php
```diff
@@ -203,22 +327,59 @@ public static function getStatus($id)
...
+
       case QubitTerm::JOB_STATUS_ERROR_ID:
-        return 'error';
+        return $i18n->__('error');
+
       default:
-        return 'unknown';
+        return $i18n->__('unknown');
+    }
+  }
```
sevein: I wonder if we should do something like the following instead:

``` php
class Foo
{
  private static
    $JOB_STATUSES = array(
      QubitTerm::JOB_STATUS_COMPLETED_ID => 'completed',
      QubitTerm::JOB_STATUS_IN_PROGRESS_ID => 'in progress',
      QubitTerm::JOB_STATUS_ERROR_ID => 'error'
    );

  public static function getStatusString($id)
  {
    $i18n = sfContext::getInstance()->i18n;

    if (null === $status = self::$JOB_STATUSES[$id])
    {
      return $i18n->__('unknown');
    }

    return $i18n->__($id);
  }
}
```

plugins/arElasticSearchPlugin/lib/arElasticSearchPlugin.class.php
```diff
@@ -346,6 +346,38 @@ public function addDocument($data, $type)
...
+
+    $document = new \Elastica\Document($object->id, $data);
+
+    try
+    {
+      $this->index->getType(get_class($object))->updateDocument($document);
+    }
+    catch (\Elastica\Exception\NotFoundException $e)
+    {
+      // Create document if it's not found
```
sevein, line 375: Are we sure we want to do this instead of throwing an exception or ignore? What would be the benefit of having a document in that state?

sevein:
Great! :+1:  Make sure that @fiver-watson is aware of the behavior of the `accept` attribute. Hopefully he can verify that it works, it's a subtle change.

plugins/arElasticSearchPlugin/lib/arElasticSearchPlugin.class.php
jraddaoui, line 375: The partial data is not used to create the document in that case, it creates the full document like we use to do.

lib/job/arFindingAidJob.class.php
jraddaoui: I have been discussing the status changes with Dan and I think I'm going to remove this function and create one in the QubitInformationObject class, as we will need now to get the status from the job or a property.

MikeFE: :+1: 

Also if you plan to use the above code in QubitInformationObject, use isset() for the if statement.

jraddaoui: Or maybe I'll do it directly in InformationObjectFindingAidComponent

apps/qubit/modules/informationobject/templates/_findingAid.php
```diff
@@ -1,6 +1,15 @@
 <li class="separator"><h4><?php echo __('Finding aid') ?></h4></li>
 
-<?php if ($sf_user->isAuthenticated() && (!isset($status) || $status === QubitTerm::JOB_STATUS_ERROR_ID)): ?>
+<?php if ($showStatus): ?>
+  <li>
+    <a>
+      <i class="icon-info-sign"></i>
+      <?php echo __('Status: ') . $status ?>
```
sevein: gettext args? :)

sevein:
👍

sevein:
👍

sevein:
👍

jraddaoui:
This branch needed a rebase from qa/2.3.x for testing, so I've done that and fixed it to remove the Fontawesome upgrade, already included in qa/2.3.x. The branch and this pull request have now all the changes included in one commit.

sevein:
Merged in `qa/2.4.x`: 317e14fe8098259eea8fcec30bf104066c2d01d7.

PR #311, opened 2016-04-08
Export unassociated authority records, refs #9681

MikeFE:
Previously, authority records wouldn't get exported if they weren't associated as a creator in an archival description. This fix makes it so all auth recs get exported, even if they are "stand alone".

### DIFF
--- a/lib/task/export/eacExportTask.class.php
+++ b/lib/task/export/eacExportTask.class.php
@@ -64,42 +64,45 @@ class eacExportTask extends exportBulkBaseTask
 
     $this->includeXmlExportClassesAndHelpers();
 
-    // Get descriptions matching optional specification
-    $rows = $conn->query($this->informationObjectQuerySql($options), PDO::FETCH_ASSOC);
+    // Using left join and comparisons to null here to make sure we exclude
+    // actor rows associated with repositories and users.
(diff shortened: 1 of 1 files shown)


lib/task/export/eacExportTask.class.php
```diff
@@ -64,42 +64,46 @@ public function execute($arguments = array(), $options = array())
...
-    // Get descriptions matching optional specification
-    $rows = $conn->query($this->informationObjectQuerySql($options), PDO::FETCH_ASSOC);
+    // Using left join and comparisons to null here to make sure we exclude
+    // actor rows associated with repositories and users.
+    $query = 'SELECT a.id FROM actor a
+              JOIN actor_i18n ai ON a.id=ai.id
+              LEFT JOIN repository r ON a.id=r.id
+              LEFT JOIN user u ON a.id=u.id
+              WHERE u.id IS NULL AND r.id IS NULL AND a.id != ?';
+
```
jraddaoui, line 73: I'd JOIN with 'object' and check the 'class_name' name for 'QubitActor'. We may add other descendant objects in the future.

MikeFE, line 73: I hadn't thought of that, but that sounds reasonable to me.

PR #313, opened 2016-04-16
Add csv:import skip-derivative option (#9556)

djjuhasz:
Allow skipping derivative generation during import process.

### DIFF
--- a/apps/qubit/modules/informationobject/actions/addDigitalObjectAction.class.php
+++ b/apps/qubit/modules/informationobject/actions/addDigitalObjectAction.class.php
@@ -118,9 +118,18 @@ class InformationObjectAddDigitalObjectAction extends sfAction
     }
     else if (null !== $this->form->getValue('url'))
     {
-      $digitalObject->importFromURI($this->form->getValue('url'));
+      // Catch errors trying to download remote resource
+      try
+      {
+        $digitalObject->importFromURI($this->form->getValue('url'));
+        $this->resource->digitalObjects[] = $digitalObject;
+      }
+      catch (sfException $e)
+      {
+        // Log download exception
+        $this->logMessage($e->getMessage, 'err');
+      }
     }
 
-    $this->resource->digitalObjects[] = $digitalObject;
   }
 }
--- a/lib/model/QubitDigitalObject.php
+++ b/lib/model/QubitDigitalObject.php
@@ -1386,24 +1386,41 @@ class QubitDigitalObject extends BaseDigitalObject
    * Download external file via sfWebBrowser and return its temporary location
    *
    * @param string URI
-   * @return mixed FALSE if error, array otherwise
+   * @return string contents
    */
   private function downloadExternalObject($uri)
   {
-    // Parse URL into components and get file/base name
-    $uriComponents = parse_url($uri);
-
     // Initialize web browser
     $timeout = sfConfig::get("app_download_timeout");
     $browser = new sfWebBrowser(array(), null, array('Timeout' => $timeout));
     $browser->get($uri);
 
-    if ($browser->get($uri)->responseIsError() || 1 > strlen(($filename = basename($uriComponents['path']))))
+    if ($browser->get($uri)->responseIsError())
     {
       return false;
     }
 
-    return array($filename, $browser->getResponseText());
+    return $browser->getResponseText();
+  }
+
+  /**
+   * Get filename from URI path
+   *
+   * @param string URI
+   * @return mixed null if error, string otherwise
+   */
+  private function getFilenameFromUri($uri)
+  {
+    // Parse URL into components and get file/base name
+    $uriComponents = parse_url($uri);
+    $filename = basename($uriComponents['path']);
+
+    if (1 > strlen($filename))
+    {
+      throw new sfException(sprintf('Couldn\'t parse filename from uri %s', $uri));
+    }
+
+    return $filename;
   }
 
   /**
@@ -1415,7 +1432,9 @@ class QubitDigitalObject extends BaseDigitalObject
    */
   public function importFromURI($uri, $options = array())
   {
-    list($filename, $contents) = $this->downloadExternalObject($uri);
+    $filename = $this->getFilenameFromUri($uri);
+    $contents = $this->downloadExternalObject($uri);
+
     if (false === $this->localPath = Qubit::saveTemporaryFile($filename, $contents))
     {
       throw new sfException('Encountered error fetching external resource.');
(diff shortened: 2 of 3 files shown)


lib/model/QubitDigitalObject.php
```diff
@@ -1386,53 +1386,121 @@ public function writeToFileSystem($asset, $options = array())
...
+    return $filename;
   }
 
   /**
    * Populate a digital object from a resource pointed to by a URI
    * This is for, eg. importing encoded digital objects from XML
    *
-   * @param string  $uri  URI pointing to the resource
-   * @return boolean  success or failure
+   * @param string $uri     URI pointing to the resource
```
mcantelon: Nitpick: extra spaces

mcantelon:
Other than one nitpick, looks good!

lib/model/QubitDigitalObject.php
```diff
@@ -1386,53 +1386,121 @@ public function writeToFileSystem($asset, $options = array())
...
    */
-  private function downloadExternalObject($uri)
+  private function downloadExternalObject($uri, $options = array())
   {
-    // Parse URL into components and get file/base name
-    $uriComponents = parse_url($uri);
-
     // Initialize web browser
     $timeout = sfConfig::get("app_download_timeout");
     $browser = new sfWebBrowser(array(), null, array('Timeout' => $timeout));
```
sevein: Should we explicitly pass the desired adapter to the constructor? e.g.

``` php
$browser = new sfWebBrowser(array(), 'sfCurlAdapter`, array('Timeout' => $timeout));
```

In AtoM, `sfCurlAdapter` will probably be the adapter being used as `php-curl` is listed as a dependency and it's the preferred adapter of `sfWebBrowser`, but I think it'd be clearer to express that constraint here as you are relying on curl-specific features a few lines below?

apps/qubit/modules/informationobject/actions/addDigitalObjectAction.class.php
```diff
@@ -118,9 +118,18 @@ public function processForm()
...
+      // Catch errors trying to download remote resource
+      try
+      {
+        $digitalObject->importFromURI($this->form->getValue('url'));
+        $this->resource->digitalObjects[] = $digitalObject;
+      }
+      catch (sfException $e)
+      {
+        // Log download exception
+        $this->logMessage($e->getMessage, 'err');
```
sevein, line 130: I had no idea that sfAction had this shortcut...

lib/task/import/csvImportTask.class.php
```diff
@@ -956,16 +962,35 @@ public function execute($arguments = array(), $options = array())
...
-              $this->log("Failed to download $uri after multiple tries. Continuing task...");
-            }
+            $do->importFromURI($uri, $options);
+            $do->save($conn);
           }
-          catch (Exception $e)
+          catch (sfException $e)
           {
-            print "\nFailed to download '$uri': ". $e->getMessage() ."\n";
+            echo $e->getMessage()."\n";
```
sevein: I think it's better to use `$this->log` or `$this->logSection`, we don't use it consistently though.

lib/task/import/csvImportTask.class.php
```diff
@@ -956,16 +962,35 @@ public function execute($arguments = array(), $options = array())
...
           {
-            if (!downloadExternalURIWithRetries($uri, $self->object))
-            {
-              $this->log("Failed to download $uri after multiple tries. Continuing task...");
-            }
+            $do->importFromURI($uri, $options);
+            $do->save($conn);
           }
-          catch (Exception $e)
+          catch (sfException $e)
```
sevein: Subtle change here but this will cause a range of exceptions unrelated to `sfException` not to be caught, e.g. the library you are consuming may change impl. and bubble up a `InvalidArgumentException` which you may also want to capture in this case... maybe `Exception` was a better idea?

sevein:
Very nice, David. I've added a couple of comments but they are far from deal breakers.

lib/task/import/csvImportTask.class.php
djjuhasz: Agreed. For some reason I thought $this->log was not in scope, but I was wrong.  It's annoying that $this->log() only echoes to STDOUT and doesn't actually write to the _log_ file.  I thought for a minute I could ditch the subsequent code for writing the error to the log. :(

djjuhasz:
Note: I'll squash all of these commits when I merge them to qa/2.3.x

lib/task/import/csvImportTask.class.php
```diff
@@ -996,18 +996,33 @@ public function execute($arguments = array(), $options = array())
...
           }
+          catch (Exception $e)
+          {
+            // Echo error to STDOUT
+            $this->log($e->getMessage());
+
+            // Log error
+            if (sfConfig::get('sf_logging_enabled'))
+            {
+              sfContext::getInstance()->getLogger()->err($e->getMessage());
```
sevein: If logging is disabled, don't you receive an instance of `sfNoLogger` so you don't have to do the check yourself? I'm not entirely sure though.

sevein: See https://git.io/vwCkL.

sevein: i.e. sfNoLogger is like a no-op logger

sevein:
Minor observation but 👍

lib/task/import/csvImportTask.class.php
djjuhasz: Hmm, I got that snippet directly from [Listing 16-6 - Adding a Custom Log Message from Anywhere](http://librosweb.es/libro/symfony_1_2_en/chapter_16/logging.html), but it's from symfony 1.2 and there's no 1.4 version of the doc.  I'll try removing the conditional though, and see what happens. :)

djjuhasz:
@sevein you were right about not having to check for sfConfig::get('sf_logging_enabled')! :)  I also overrode the sfTask::log() class to echo and log output at the same time rather than manually doing both.

djjuhasz:
I also did a rebase against qa/2.3.x, which has piled all my commits at the bottom of the page. :-/

sevein:
Cool. Some other day we could create our own logger to listen to `console.log` events by extending sfFileLogger, which [only listens](https://github.com/artefactual/atom/blob/3628689ed554df780ad552fcea21b9acad39bf1a/vendor/symfony/lib/log/sfLogger.class.php#L84) to `application.log`, and set that new logger up in `config/factories.yml` for the `cli` front-end. Probably overkill but I think that would be the idiomatic way in Symfony?

sevein:
:+1:

PR #314, opened 2016-04-18
Fix some export:bulk edge case bugs refs #8905

MikeFE:
Fixed issue when user entered an unwritable path while using --single-slug
Fixed issue when user specified non-existant slug using --single-slug


sevein:
:+1:

PR #315, opened 2016-04-20
Fix adv. search form validation, refs #9718

jraddaoui:
- Validate form after query creation but before execution to keep the
  boolean query and other params.
- Stop populating the LOD filter with the facet data as there may be
  no facet data after the form validation fails.
- Check results pager in the template to remove unnecessary sections.


MikeFE:
:+1:

PR #317, opened 2016-04-21
Change UI labels to adhere to ISAAR-CPF ref #9632

MikeFE:
Previously in the UI the authority record identifier was labelled "description identifier"
since that's what we call it in the database. This changes the labels to say
"authority record identifier" in the UI.


sevein:
:+1:

PR #319, opened 2016-04-21
Don't render empty EAC elements, refs #9606

MikeFE:
Also some minor improvements to the EAC export task's logging.

### DIFF
--- a/plugins/sfEacPlugin/lib/sfEacPlugin.class.php
+++ b/plugins/sfEacPlugin/lib/sfEacPlugin.class.php
@@ -259,7 +259,12 @@ class sfEacPlugin implements ArrayAccess
 
         // TODO <date/>, <dateRange/>, <dateSet/>, <descriptiveNote/>, simple
         // natural language parsing?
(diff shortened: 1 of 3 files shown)


plugins/sfEacPlugin/lib/sfEacPlugin.class.php
```diff
@@ -1049,4 +1058,36 @@ public static function renderDates($item)
...
+      $resource->places,
+      $resource->legalStatus,
+      $resource->functions,
+      $resource->mandates,
+      $this->structureOrGenealogy,
+      $this->generalContext,
+      $this->biogHist
+    );
+
+    $result = false;
```
mcantelon: I think you could replace lines 1085-1091 with this use of a weird PHP function (that removes empty elements):

`return (bool)array_filter($a);`

mcantelon:
Other than that, looks good!

plugins/sfEacPlugin/lib/sfEacPlugin.class.php
MikeFE: :+1:

PR #320, opened 2016-04-25
Fix file write permission check, refs #9737

MikeFE:
Apparently is_writeable will counter intuitively return FALSE if you give it a path
to a file that doesn't exist (but that you'd have create/write permissions to). We
were erroneously using this function to check if we could create/write to a file.


sevein:
:+1:

PR #322, opened 2016-04-27
Fix digital object upload, refs #9748

djjuhasz:
Fix regression introduced in commit 03c17eaf4546457e6ac00b2f4a0b517bbf96a9fb


MikeFE:
:+1:

PR #324, opened 2016-04-28
Clean up csv examples for 2.3

fiver-watson:
I found some unnecessary columns (e.g. RAD columns in the ISAD example), some unneeded NULL values, and a few old errors (e.g. example scripts) that I have tidied up to make the new CSV template examples more easy for users to understand.


sevein:
👍

PR #325, opened 2016-04-29
Remove empty name fields on save, refs #8243

MikeFE:


apps/qubit/modules/default/actions/editAction.class.php
```diff
@@ -133,7 +133,7 @@ protected function processField($field)
 
         foreach ($this[$field->getName()] as $item)
         {
-          if (isset($value[$item->id]))
+          if (isset($value[$item->id]) && $value[$item->id])
```
sevein: Maybe better to do...?

``` php
if (!empty($value[$item->id]))
```

MikeFE: ah yeah, good call. I always forget how PHP has so many ways to check these sorts of things, all with their own nuances :|

sevein:
👍

PR #326, opened 2016-04-29
Export dates of existence in EAC, refs #9755

MikeFE:
There was a regression where the dates of existence for an actor wasn't exporting in EAC anymore.
This was caused by trying to use isset() to test a property that isn't normal but instead uses
__get and __set magic, but the __isset magic wasn't defined.

### DIFF
--- a/plugins/sfEacPlugin/lib/sfEacPlugin.class.php
+++ b/plugins/sfEacPlugin/lib/sfEacPlugin.class.php
@@ -256,10 +256,9 @@ class sfEacPlugin implements ArrayAccess
         return;
 
       case 'existDates':
-
         // TODO <date/>, <dateRange/>, <dateSet/>, <descriptiveNote/>, simple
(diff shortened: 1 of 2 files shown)


sevein:
Could you try to dig into the actual problem a bit more? I took a quick look here: https://github.com/artefactual/atom/blob/qa/2.3.x/lib/model/om/BaseActor.php#L145-L156, which should be calling BaseActorI18n's `__isset()`, right?

MikeFE:
Yeah you're right, L261 of sfEacPlugin did work with isset() actually, so removing it wasn't really necessary (but having isset there isn't necessary anyway, we'll get null if it isn't set).

isset($eac->existDates) in indexSuccess.xml.php won't work though unless you want me to define __isset in sfEacPlugin and add a case for that, but that seems like extra code for no functional difference.

sevein:
👍

PR #328, opened 2016-04-30
LDAP authentication support, refs #9765

mcantelon:
Redmine: https://projects.artefactual.com/issues/9765

Added LDAP authentication support. To enable change "myUser" to "ldapUser" in config/factories.yml. Three settings need to be set on the settings/ldap page.


vendor/password_compat/password.php
```diff
@@ -0,0 +1,279 @@
...
+     * @return string The substring
+     */
+    function _substr($binary_string, $start, $length) {
+       if (function_exists('mb_substr')) {
+           return mb_substr($binary_string, $start, $length, '8bit');
+       }
+       return substr($binary_string, $start, $length);
+   }
+
+}
```
sevein, line 317: Should we add a mention to this library in `/COPYRIGHT`?

sevein, line 317: Also, should we use the most recent version?
https://github.com/ircmaxell/password_compat/blob/master/lib/password.php

apps/qubit/modules/settings/templates/ldapSuccess.php
```diff
@@ -0,0 +1,43 @@
...
+
+<?php slot('title') ?>
+
+  <h1><?php echo __('LDAP authentication settings') ?></h1>
+
+<?php end_slot() ?>
+
+<?php slot('content') ?>
+
+  <form action="<?php echo url_for('settings/ldap') ?>" method="post">
```
sevein: Can you use `renderFormTag`? e.g.

``` php
<?php echo $form->renderFormTag(url_for(array('module' => 'settings', 'action' => 'ldap'))) ?>
```

apps/qubit/modules/settings/actions/ldapAction.class.php
```diff
@@ -0,0 +1,91 @@
...
+
+    foreach($fieldToSettingMap as $field => $setting)
+    {
+      if (null !== $value = $thisForm->getValue($field))
+      {
+        $setting = QubitSetting::getByName($setting);
+        $setting->setValue($value, array('sourceCulture' => true));
+        $setting->save();
+      }
+    }
```
sevein, line 92: How would you feel about not adding these three settings in the fixtures / migration and just populate them when they are created for the first time? I've done that before with other settings page. It saves us the migration, which is always a good thing. Your code above would be almost the same, just consider the possibility that `getByName` returns `NULL` so you have to create a new setting from scratch?

It may be a dirty trick, you don't have to do it if you don't want!

lib/ldapUser.class.php
```diff
@@ -0,0 +1,165 @@
...
+ * Access to Memory (AtoM) is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with Access to Memory (AtoM).  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+include sfConfig::get('sf_root_dir') .'/vendor/password_compat/password.php';
```
sevein, line 20: Do we need this? Maybe it's been auto-loaded or you could add it to https://github.com/artefactual/atom/blob/qa/2.3.x/apps/qubit/config/autoload.yml? I can't remember how it works! I always forget but you shouldn't need to include AFAIK.

lib/ldapUser.class.php
```diff
@@ -0,0 +1,165 @@
...
+  }
+
+  protected function getLdapConnection()
+  {
+    if (isset($this->ldapConnection))
+    {
+      return $this->ldapConnection;
+    }
+    else
+    {
```
sevein: I like to avoid nesting code because I find it clearer, not sure how you feel about it, but I don't think you need that `else`?

lib/ldapUser.class.php
```diff
@@ -0,0 +1,165 @@
...
+      $host = 'ldap://'. (string)QubitSetting::getByName('ldapHost');
+      $port = (string)QubitSetting::getByName('ldapPort');
+
+      $connection = ldap_connect($host, $port);
```
sevein: As you are using full LDAP URIs you could as well add the port to it, e.g. `ldap_connect('ldap://hostname:port')` or you could `ldap_connect('localhost', 12345)`.

lib/ldapUser.class.php
```diff
@@ -0,0 +1,165 @@
...
+
+    // Fallback to non-LDAP authentication if need be and load/create user data
+    if (!$authenticated)
+    {
+      $authenticated = parent::authenticate($username, $password);
+
+      // Load user
+      $criteria = new Criteria;
+      $criteria->add(QubitUser::EMAIL, $username);
+      $user = QubitUser::getOne($criteria);
```
sevein: ...and what would happen if `$user` is `NULL`?

sevein:
Your commit `LDAP authentication support.` is missing the ref to the ticket.

sevein:
What do you think about replacing:

``` php
$port = (string)QubitSetting::getByName('ldapPort');
```

... with:

``` php
$port = QubitSetting::getByName('ldapPort')->getValue(array('sourceCulture' => true));
```

It's more verbose but I like that it doesn't rely on magic methods. Not a big deal though, up to you!

sevein:
The ticket on the branch is private. We normally create a public ticket beforehand.

sevein:
Please remember don't merge into `qa/2.3.x`. We're going to create `qa/2.4.x` once we release AtoM 2.3.0 (or maybe earlier if we start having too many topic branches waiting).

apps/qubit/modules/settings/actions/ldapAction.class.php
mcantelon, line 92: Yeah, that makes more sense to me too... easy to do and It'll simplify migrations.

lib/ldapUser.class.php
mcantelon, line 20: Hrm, I'll look into that as having it auto-load would definitely be less hack-y.

lib/ldapUser.class.php
mcantelon: True... good call.

lib/ldapUser.class.php
mcantelon: Interesting... will give that a try.

apps/qubit/modules/settings/actions/ldapAction.class.php
```diff
@@ -0,0 +1,99 @@
...
+    );
+
+    foreach($fieldToSettingMap as $field => $settingName)
+    {
+      if (null !== $value = $thisForm->getValue($field))
+      {
+        if (null === $setting = QubitSetting::getByName($settingName))
+        {
+          $setting = new QubitSetting;
+          $setting->setName($settingName, array('sourceCulture' => true));
```
sevein: `name` is not a i18n property.

apps/qubit/modules/settings/actions/ldapAction.class.php
```diff
@@ -0,0 +1,99 @@
...
+        }
+        $setting->setValue($value, array('sourceCulture' => true));
+        $setting->save();
```
sevein: Should we also set a common `scope` for the three settings? e.g. `ldap`?

apps/qubit/modules/settings/actions/ldapAction.class.php
```diff
@@ -0,0 +1,99 @@
...
+      }
+    }
```
sevein, line 92: Another observation about avoid nesting code, don't you think this may be easier to read?

``` php
  foreach ($fieldToSettingMap as $field => $settingName)
  {
    if (null === $value = $thisForm->getValue($field))
    {
      continue;
    }

    if (null === $setting = QubitSetting::getByName($settingName))
    {
      $setting = new QubitSetting;
      $setting->setName($settingName);
      $setting->setScope('ldap');
    }

    $setting->setValue($value, array('sourceCulture' => true));
    $setting->save();
  }
```

apps/qubit/modules/user/actions/loginAction.class.php
```diff
@@ -42,9 +42,19 @@ public function execute($request)
     $this->form->setValidator('next', new sfValidatorString);
     $this->form->setWidget('next', new sfWidgetFormInputHidden);
 
-    $this->form->setValidator('email', new sfValidatorEmail(array('required' => true), array(
-      'required' => $this->context->i18n->__('You must enter your email address'),
-      'invalid' => $this->context->i18n->__('This isn\'t a valid email address'))));
+    // Use string validation if LDAP authentication's used as email not used to log in
+    if ($this->context->user instanceof ldapUser)
```
sevein, line 46: You use `instanceof` here but `get_class` in `menuComponent.class.php`, is that on purpose?

lib/ldapUser.class.php
```diff
@@ -0,0 +1,166 @@
...
+      ldap_set_option($connection, LDAP_OPT_PROTOCOL_VERSION, 3);
+
+      $this->ldapConnection = $connection;
+      return $connection;
+    }
+  }
+
+  protected function ldapBind($username, $password)
+  {
+    if ($ldap = $this->getLdapConnection())
```
sevein: `$conn` or `$connection` better than `$ldap` and more consistent with the way you call it in `getLdapConnection`?

sevein:
A few more observations:
- Even though it's declared as a dependency in the docs, the LDAP extension won't be always available and we should notify the user somehow, e.g. https://github.com/artefactual/atom/blob/qa/2.3.x/vendor/symfony/lib/cache/sfSQLiteCache.class.php#L38-L41
- Something that I would seriously consider if have some time is to add some logging in the new LDAP factory. Authentication is always problematic and we want to point users somewhere where they can look at when LDAP auth is not working for them. In particular, it could be interesting to log the details stored in `LDAP_OPT_DIAGNOSTIC_MESSAGE` as explained [here](http://php.net/manual/en/function.ldap-get-option.php#112593) or [here](http://php.net/manual/en/function.ldap-bind.php#103034).
  `ldap_error` and `ldap_errno` look also really useful! Not sure right how how logging should be done in a factory, I can't remember. Perhaps something like the following:
  
  ``` php
  $this->dispatcher->notify(new sfEvent($this, 'application.log', array(sprintf('Authentication problem: "%s"', 'FATAL ERROR !!!'))));
  ```
- We should try to be gentle with the LDAP server and unbind with `ldap_unbind`. Sounds like you could extend `shutdown` (see `vendor/symfony/lib/user/sfUser.class.php`) to trigger that right in time.

lib/ldapUser.class.php
mcantelon, line 20: It looks like autoload will only handle classes, unfortunately. It parses files to determine if they're a class. We could subclass the autoloader class, though, as per:

https://capocasa.net/trait-with-symfony-1-4

I think the thing to do would be if a file, rather than a directory, is specified in autoload.yml then include it without parsing it.

sevein, line 20: Perhaps sounds too complicated and not worth the time?

On Mon, May 2, 2016 at 3:05 PM, Mike Cantelon notifications@github.com
wrote:

> In lib/ldapUser.class.php
> https://github.com/artefactual/atom/pull/328#discussion_r61813076:
> 
> > - *
> > - \* Access to Memory (AtoM) is free software: you can redistribute it and/or modify
> > - \* it under the terms of the GNU Affero General Public License as published by
> > - \* the Free Software Foundation, either version 3 of the License, or
> > - \* (at your option) any later version.
> > - *
> > - \* Access to Memory (AtoM) is distributed in the hope that it will be useful,
> > - \* but WITHOUT ANY WARRANTY; without even the implied warranty of
> > - \* MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
> > - \* GNU General Public License for more details.
> > - *
> > - \* You should have received a copy of the GNU General Public License
> > - \* along with Access to Memory (AtoM).  If not, see http://www.gnu.org/licenses/.
> > - */
> >   +
> >   +include sfConfig::get('sf_root_dir') .'/vendor/password_compat/password.php';
> 
> It looks like autoload will only handle classes, unfortunately. It parses
> files to determine if they're a class. We could subclass the autoloader
> class, though, as per:
> 
> https://capocasa.net/trait-with-symfony-1-4
> 
> I think the thing to do would be if a file, rather than a directory, is
> specified in autoload.yml then include it without parsing it.
> 
> —
> You are receiving this because you commented.
> Reply to this email directly or view it on GitHub
> https://github.com/artefactual/atom/pull/328/files/a0611c3b67f116b00509c4db63ed4b49b7bc0b03#r61813076

## 

Jesús García Crespo,

sevein, line 20: Would it work if it's added here? https://github.com/artefactual/atom/blob/qa/2.3.x/config/ProjectConfiguration.class.php#L69-L71

apps/qubit/modules/user/actions/loginAction.class.php
mcantelon, line 46: Oops. Fixed.

lib/ldapUser.class.php
mcantelon, line 20: As that's an SPL autoloader too, I don't think it'll work. It looks like there's an RFC for autoloading functions rather than classes (https://wiki.php.net/rfc/autofunc) but is has been opposed (as of PHP 5.4: http://www.serverphorums.com/read.php?7,597588). One trick people use is making functions into static methods of classes, but password_compat doesn't do that. :(

sevein, line 20: I see ok hehe I give up :)

sevein:
Are you waiting for more feedback or still working on it?
I was going to say that I don't see logging work but maybe you are planning to add it soon?

mcantelon:
I just tweaked some stuff, but am now ready for more feedback. I didn't end up adding the logging, unfortunately, as there's now only 1/2 hour left in the budget.

lib/form/SettingsLdapForm.class.php
```diff
@@ -0,0 +1,65 @@
+<?php
```
mcantelon: Oops just realized I don't need this file anymore... deleting.

apps/qubit/modules/settings/actions/ldapAction.class.php
```diff
@@ -0,0 +1,94 @@
...
+    switch ($name = $field->getName())
+    {
+      case 'ldapHost':
+      case 'ldapPort':
+      case 'ldapBaseDn':
+        if (null === $this->{$name})
+        {
+          $this->{$name} = new QubitSetting;
+          $this->{$name}->name = $name;
+          $this->scope = 'ldap';
```
sevein: ... that's not setting the scope in the right object right?

apps/qubit/modules/settings/actions/ldapAction.class.php
```diff
@@ -0,0 +1,94 @@
...
+        }
+        $this->{$name}->setValue($field->getValue(), array('culture' => 'en'));
```
sevein: The array should pass `'sourceCulture' => true` to work with installations where `default_culture != 'en'` I believe.

apps/qubit/modules/settings/actions/ldapAction.class.php
```diff
@@ -0,0 +1,94 @@
...
+
+  protected function addField($name)
+  {
+    switch ($name)
+    {
+      case 'ldapHost':
+      case 'ldapPort':
+      case 'ldapBaseDn':
+        $default = ($name == 'ldapPort') ? 389 : '';
+        if (null !== $this->{$name} = QubitSetting::getByName($name))
```
sevein, line 42: Is this an optimization to save extra queries later when POSTing? If that's the case, you may actually don't need it because `BaseSetting` is caching populated objects for you here: https://github.com/artefactual/atom/blob/qa/2.3.x/lib/model/om/BaseSetting.php#L32. You may want to double check though.

lib/ldapUser.class.php
```diff
@@ -0,0 +1,185 @@
...
+  public function authenticate($username, $password)
+  {
+    // Allow LDAP authentication to be overridden during development
+    $configuration = sfContext::getInstance()->getConfiguration();
+    if ($configuration->isDebug() || 'dev' == $configuration->getEnvironment())
+    {
+      return parent::authenticate($username, $password);
+    }
+
+    $authenticated = false;
```
sevein: Do you need this? $authenticated gets a new value in L55 anyways.

apps/qubit/modules/settings/actions/ldapAction.class.php
mcantelon: Agh, you're right. I'll fix.

apps/qubit/modules/settings/actions/ldapAction.class.php
mcantelon, line 42: It's just a way to store the setting, if it has been created, as an object property for convenience... I copied the technique from https://github.com/artefactual/atom/blob/qa/2.3.x/apps/qubit/modules/settings/actions/digitalObjectDerivativesAction.class.php#L38 . I could alternatively refer to the setting directly, though... makes the code a bit more verbose, but clearer.

lib/ldapUser.class.php
mcantelon: Oops... fixed.

apps/qubit/modules/settings/actions/ldapAction.class.php
sevein, line 42: Hahaha and it seems I wrote that :)
I was trying to understand the benefit... and I'm not sure yet. But I think it's 👍.

apps/qubit/modules/settings/actions/ldapAction.class.php
```diff
@@ -0,0 +1,94 @@
...
+class SettingsLdapAction extends DefaultEditAction
+{
+  // Arrays not allowed in class constants
+  public static
+    $NAMES = array(
+      'ldapHost',
+      'ldapPort',
+      'ldapBaseDn');
+
+  protected function earlyExecute()
```
sevein, line 29: Why empty?

sevein:
👍

apps/qubit/modules/settings/actions/ldapAction.class.php
mcantelon, line 42: Hehehe awesome. :D

apps/qubit/modules/settings/actions/ldapAction.class.php
mcantelon, line 29: I tried taking it out and it caused an error... my guess is the parent class doesn't check that the method exists before attempting to execute it.

mcantelon:
Awesome... will merge into 2.4.x!

sevein:
There is no qa/2.4.x yet, just leave it here for now. I guess we'll have to create one but I wanted to delay that as much as possible but I guess it's going to stop making sense soon.

sevein:
Merged in `qa/2.4.x`: 6c0b6682c1dfc99fb61b8d01089da7cf2a94abc2.

PR #332, opened 2016-05-05
Clean up example accessions CSV, refs #9753

fiver-watson:
This commit is a partial response to issue #9753. It does not resolve
all the issues outlined there (which is beyond my abilities), but it
does remove unnecessarily confusing columns from the example
Accessions CSV template, and includes better date examples.


fiver-watson:
I got a weird message about line endings in the CSV changing from CRLF to LF - but LF is is the unix style `/n` line endings which we want... so I'm a bit confused. can someone else please test importing with this example as part of CR to confirm that it works?

MikeFE:
It seemed to import the 2 rows correctly in my tests.

:+1:

PR #333, opened 2016-05-06
Remove blank "Use for" list items, refs #8244

MikeFE:
Also fixed this same issue for a couple other fields that had it. This is a similar
fix to #8243


sevein:
👍

PR #334, opened 2016-05-06
Tidy XML exported for EAC, refs #9606 note-4

MikeFE:
I re-arranged the EAC export structure a bit so it matches the EAD plugin's style and we can tidy the XML.
I also refactored the code a bit to use ob_get_clean for brevity.


sevein:
👍

PR #335, opened 2016-05-06
Hide upload stats when editing descs, refs #9275

MikeFE:
Hide repository upload statistics in context menu when we're not actually on a repository page.

These stats were previously appearing when editing archival descriptions, weren't useful, and led to timeout
issues when there were huge datasets.

### DIFF
--- a/apps/qubit/modules/repository/templates/_contextMenu.php
+++ b/apps/qubit/modules/repository/templates/_contextMenu.php
@@ -1,9 +1,9 @@
 <?php echo get_component('repository', 'logo') ?>
 
-<?php if (QubitAcl::check($resource, 'update')): ?>
-  <?php include_component('repository', 'uploadLimit', array('resource' => $resource)) ?>
-<?php endif; ?>
-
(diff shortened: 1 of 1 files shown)


apps/qubit/modules/repository/templates/_contextMenu.php
```diff
@@ -1,9 +1,9 @@
...
-<?php endif; ?>
-
 <?php if ($resource->getRawValue() instanceof QubitRepository): ?>
+  <?php if (QubitAcl::check($resource, 'update')): ?>
+    <?php include_component('repository', 'uploadLimit', array('resource' => $resource)) ?>
+  <?php endif; ?>
+
   <?php include_component('repository', 'holdings', array('resource' => $resource)) ?>
-<?php endif; ?>
+<?php endif; ?>
```
sevein: Misty always used to be picky about this and I kind of got used to it! EOF blank line is missing! In Sublime you can enable `ensure_newline_at_eof_on_save` to handle that automagically.

sevein:
Minor comment about EOF but 👍

PR #336, opened 2016-05-07
Import EAD address tags. refs #9156

mcantelon:
Within EAD originalsloc tags, address tags, in addressline tags within
them, were previously ignored. Added support for importing these.

### DIFF
--- a/lib/model/QubitInformationObject.php
+++ b/lib/model/QubitInformationObject.php
@@ -1582,6 +1582,37 @@ class QubitInformationObject extends BaseInformationObject
   }
 
   /**
+   * Import data from a <originalsloc> tag in EAD2002
+   *
+   * As the locations of originals field is unstructured, EAD content must
+   * be amalgamated as text. Note that when EAD exported from AtoM the content
+   * is exported in a p tag within the originalsloc tag.
+   *
+   * @param $originalsNode  DOMNode  EAD originalsloc DOM node
+   */
+  public function importOriginalsLocationEadData($originalsNode)
+  {
+    $text = '';
+
+    // Add <p> tag content to amalgamated value
+    foreach ($originalsNode->getElementsByTagName('p') as $pNode)
+    {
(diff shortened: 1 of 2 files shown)


lib/model/QubitInformationObject.php
```diff
@@ -1582,6 +1582,37 @@ public function setActorByName($name, $options)
...
   /**
+   * Import data from a <originalsloc> tag in EAD2002
+   *
+   * @param $originalsNode  DOMNode  EAD originalsloc DOM node
+   */
+  public function importOriginalsLocationEadData($originalsNode)
+  {
+    $text = '';
+
+    $pNodes = $originalsNode->getElementsByTagName('p');
```
MikeFE: Can probably just leave out the intermediate variables here and stick  $originalsNode->getElementsByTagName('p') directly in the foreach. Same for L1600 & L1604?

lib/model/QubitInformationObject.php
```diff
@@ -1582,6 +1582,37 @@ public function setActorByName($name, $options)
   }
 
   /**
+   * Import data from a <originalsloc> tag in EAD2002
+   *
```
MikeFE, line 1586: Can you explain your strategy in this function for fixing the bug in the comments? It seems like we're just removing the `<p>` tag and `<addressline>` tags and making them all just in a giant string with newlines? It wasn't immediately obvious to me.

mcantelon, line 1586: Basically, the locations of originals is a text field and the address fields are unstructured... so there's no way I can see to parse these into AtoM actors. As we don't export these address fields this is just a way to capture information from address tags so we conform to the EAD standard and don't ignore them.

MikeFE, line 1586: Sweet! I meant maybe add that to a comment in the code, hehe

mcantelon, line 1586: Ah cool, makes sense.

mcantelon:
Merged.

PR #337, opened 2016-05-10
Ensure entity type exports in EAC, refs #9828

MikeFE:
Due to us using __get magic, calling empty() on ->entityType wasn't the correct
way to check for its presence.


sevein:
d26fcd9

PR #338, opened 2016-05-11
Disable quota check if unlimited uploads, refs #9674

sevein:
Redmine: https://projects.artefactual.com/issues/9674


MikeFE:
:+1:

PR #340, opened 2016-05-12
Fixed browser compatibility issue. refs #9000

mcantelon:
Fixed an issue where, when in the search field, typing an
open parenthesis would be interpreted as pressing the down
arrow.

### DIFF
--- a/js/dominion.js
+++ b/js/dominion.js
@@ -506,8 +506,12 @@
             break;
 
           case 40: // Down arrow
-            e.preventDefault();
-            this.move(1);
+            // If charCode is 40 then, in Chrome/IE, it's an open parenthesis
+            if (e.charCode == 0)
+            {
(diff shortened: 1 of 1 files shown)


MikeFE:
:+1:

sevein:
Should we upgrade to `e.which`? https://api.jquery.com/event.which/

mcantelon:
Yeah, `e.which` might be good. I'll try it out.

mcantelon:
`e.which` is even more different between browsers, alas. http://stackoverflow.com/questions/18177818/why-jquerys-event-which-gives-different-results-in-firefox-and-chrome

sevein:
👍

sevein:
Reminder: body of the commit message doesn't have to be wrapped to 50 cols but 72-80.

PR #342, opened 2016-05-13
Fixed issue with search result display. refs #9351

mcantelon:
Fixed issue with our custom link tags (like `"Google":http://google.ca`) displaying, without being converted
to links, in search results. Fixed by converting to HTML then stripping HTML tags to convert to text. Conversion to text was done so the text in the results could be easily truncated (given I couldn't find decent code to truncate HTML without edge cases).


sevein:
Merged! e124e8461642bd246a1d0de31030da4bced7889c

PR #344, opened 2016-05-15
Escaping issues, refs #8522

mcantelon:
Fixed a couple of escaping issues:
- treeview popovers were being unneccesarily escaped
- help text in the DACS editing form was being escaped

The fix for the DACS editing escaping is a bit unelegant, involving adding PHP to the template to disable escaping while rendering the field with the problem help text. Unfortunately, when I tried to disable escaping in the action as a whole I'd run into an exception:

{sfException} Unknown record property "rawValue" on "QubitInformationObject"

### DIFF
--- a/js/treeView.js
+++ b/js/treeView.js
@@ -240,7 +240,7 @@
         }
 
         // Pass function so the placement is computed every time
-        $li.popover({ placement: function(popover, element) {
+        $li.popover({ html: true, placement: function(popover, element) {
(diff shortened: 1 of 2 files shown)


sevein:
@mcantelon, I've merged this for you. I've replaced 3eefc7e with aec8fdd90eaf2d1c0702e6e81de23f41932ec387, it's simpler.

PR #345, opened 2016-05-15
Improve full width treeview query, refs #9850

jraddaoui:
Use joins instead of mixing all the tables to query only over the required data


sevein:
I've merged it. Missing description on the branch name.

PR #346, opened 2016-05-15
Fix untyped EAD <odd> elements issue. refs #6141

mcantelon:
Untyped <odd> elements were being imported as a note type
that doesn't make sense for most use-cases.

### DIFF
--- a/lib/QubitXmlImport.class.php
+++ b/lib/QubitXmlImport.class.php
@@ -408,6 +408,12 @@ class QubitXmlImport
    */
   private function processMethods(&$domNode, &$importDOM, $methods, &$currentObject, $importSchema)
   {
+    // We want to keep track of nodes processed so we don't process one twice
+    // if multiple selectors apply to it (for example the generic "odd" tag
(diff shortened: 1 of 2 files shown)


sevein:
Reminder: only the first line of the commit has to be wrapped to 50 columns. Body to 72 or 80 cols, up to you.

lib/QubitXmlImport.class.php
```diff
@@ -507,6 +513,13 @@ private function processMethods(&$domNode, &$importDOM, $methods, &$currentObjec
           default:
             foreach ($nodeList2 as $key => $domNode2)
             {
+              // Skip this node if method path isn't "self" and node's previously been processed
+              if ($methodMap['XPath'] != '.' && isset($processed[$domNode2->getNodePath()]))
+                continue;
```
sevein, line 519: We always use curly brackets.

mcantelon, line 519: Oops... fixed!

mcantelon:
Merged.

PR #347, opened 2016-05-17
slugify: check iconv error, refs #9855

sevein:
Redmine: https://projects.artefactual.com/issues/9855

iconv() returns `FALSE` when there is a failure, which was happening
consistently in our Alpine-based Docker image as musl's iconv does not support
`//TRANSLIT`. This commit gives it a second shot without `//TRANSLIT`` or
ignore otherwise.


MikeFE:
:+1:

PR #348, opened 2016-05-17
Fix Archives Canada search issue. refs #9141

mcantelon:
Basic search, in Archives Canada theme, was returning only results from the
top level. Fixed it so it includes children.


sevein:
Merged.

PR #350, opened 2016-05-17
Fix accession edit permissions, refs #9761

MikeFE:
Previously, any user could view / edit accession records. Now, only admins &
editors can do it.


sevein:
👍

PR #352, opened 2016-05-17
Elasticsearch: handle search exceptions, refs #9851

sevein:
Handle Elastica exceptions gracefully. Redirect user to a custom error page
showing the exception name and if debug mode is enabled a more detailed error
message. The exceptions are also logged with level "err". The goal is to tell
the user that Elasticsearch is somehow involved in the error, which is a common
issue.

It reuses the QubitMeta filter so we don't add more levels in the stack.


MikeFE:
:+1:

PR #353, opened 2016-05-18
Fix migration issue. refs #8598

mcantelon:
Migration 130 is redundant in isolated cases (for reasons not yet determined).
Add logic to prevent migration from being applied if it's not needed.


MikeFE:
:+1:

mcantelon:
Merged.

PR #354, opened 2016-05-18
Don't check uploads dir in DIP upload, refs #9799

jraddaoui:
It will be created if needed before copying the METS file


sevein:
Where is it created?

sevein:
Oh I've seen the update in https://projects.artefactual.com/issues/9799. Cool, 👍 .

PR #355, opened 2016-05-18
Fix "0 More" bug, refs #8672

MikeFE:
In certain cases a "0 More" button would appear when there is actually 1 more
in the tree view. This fix corrects the issue.


sevein:
OBOE error? :) 👍

PR #357, opened 2016-05-18
Improve job worker prefix uniqueness, refs #9858

MikeFE:
Add root directory to add to the uniqueness of gearman job prefixes. I don't
think we need all the sprintf / addslashes in there previously either since
we're just producing an md5 hash in the end anyway.


sevein:
👍

PR #360, opened 2016-05-22
Decoding HTML entities in scrubber. refs #9181

mcantelon:
Added logic to decode any HTML entities to normal characters.


sevein:
👍

mcantelon:
Merged.

PR #364, opened 2016-05-25
Dev/issue 9897 accession editing

mcantelon:
- Adds a setting to enable/disable the accession number mask (on by default)
- Adds a custom validator to prevent use of an accession number that's already in use
- Allows accession numbers to be entered manually and edited


mcantelon:
Closing this as I discovered more work needs to be done.

PR #370, opened 2016-06-01
Include ability to select if slugs are based on title, identifier, or reference code for archival descriptions

MikeFE:


lib/model/QubitInformationObject.php
```diff
@@ -2959,18 +2962,21 @@ public function getInheritedReferenceCode()
 
     $identifier = implode(sfConfig::get('app_separator_character', '-'), $identifier);
 
-    if (isset($repository->identifier))
+    if ($includeRepoAndCountry)
```
jraddaoui, line 2988: We could check if the repo exists in here, to avoid issues getting the identifier and to remove the repository check before getting the country code bellow.

jraddaoui, line 2988: Oh, you made that after, great :)

PR #371, opened 2016-06-01
Merge browse digital object view into browse information object view

MikeFE:
Combines the functionality of the digital object browse page into the existing search/browse page. A new toggle, similar to that found on the repository browse page, will be added, allowing users to switch between the search results stub view, or a card-based view (as currently found on the digital objects browse page). This way, users will be able to better refine digital object searches with all available filters and facets.


apps/qubit/modules/informationobject/templates/_tableViewResults.php
```diff
@@ -0,0 +1,10 @@
+<?php if ($pager->hasResults()): ?>
+  <?php foreach ($pager->getResults() as $hit): ?>
+    <?php echo get_partial('search/searchResult', array('hit' => $hit, 'culture' => $selectedCulture)) ?>
+  <?php endforeach; ?>
+<?php else: ?>
+  <section id="no-search-results">
+    <i class="fa fa-search"></i>
+    <p class="no-results-found"><?php echo __('No results found.') ?></p>
+  </section>
+<?php endif; ?>
```
jraddaoui: We usually add end line. You can set it in Sublime to make it automatic (and I guess in other text editors)

PR #372, opened 2016-06-02
Virtual institutional database, refs #5968

sbreker:
**NOTE!  The following files are in this pull request even though they were not part of my push into sbreker/atom - I do not know why these are here but these changes are definitely not part of this pull request:
lib/form/arWidgetFormSelectRadioOaiAdditionalSetEnable.class.php
lib/form/SettingsOaiRepositoryForm.class.php
lib/oai/QubitOai.class.php**
# 

Enable Institutional Scoping - feature 5968-virtual institutional database

USASK Feature SK17.

Summary:
- new setting added to activate feature "Enable Institutional Scoping"
- new component added "Institutional Block" in the Repository Module
- Institutional Block allows searching and browsing within the current search-realm
- Top search field in header defaults to global search when "Enable Inst Scoping" is ON and search-realm is set
- search-realm user context variable is used to track repo context
- when feature is activated, and search-realm is set, new Institutional Block will be displayed
- back ported panel.less to arDominion theme so that the panel in the Institutional Block works correctly
- Repository Theming will track through modules if "Enable Inst Scoping" is ON and search-realm is set.


sevein:
We need this to be broken into smaller meaningful commits before we start the review.

sbreker:
Ok - I can do that.  I will try to break it up into pieces that make sense in terms of functional components.

sevein:
Not really work in progress, feel free to reopen when needed.

PR #373, opened 2016-06-03
Move objects using job scheduler. refs #9945

mcantelon:
Changed the move action (the ability to change an AtoM object's parent) so move
operations happen as background tasks.


mcantelon:
Merged.

PR #374, opened 2016-06-03
Archivematica DIP object import task. refs #9960

mcantelon:
Added, based on a task by David Juhasz, a task for importing digital objects
from Archivematica DIPs. The script can also run through a DIP, verifying if
the digital objects have previously been imported.

### DIFF
--- a/lib/task/import/importDipObjectsTask.class.php
+++ b/lib/task/import/importDipObjectsTask.class.php
@@ -1,0 +1,567 @@
+<?php
+
+/*
+ * This file is part of the Access to Memory (AtoM) software.
+ *
+ * Access to Memory (AtoM) is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * Access to Memory (AtoM) is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with Access to Memory (AtoM).  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+/**
+ * Process a CSV file to import digital objects from an Archivematica DIP and
+ * associate them with existing information objects in AtoM
+ *
+ * @package    symfony
+ * @subpackage task
+ * @author     David Juhasz <david@artefactual.com>
+ */
+class importDipObjectsTask extends arBaseTask
+{
+  protected
+    $conn,                    // Database connection
+    $dipDir,                  // Path to DIP
+    $columnNames = array(),   // Sequence of column names
+    $columnIndexes = array(), // Index in CSV row for each column
+    $uniqueValueColumnName;   // Column name of unique identifier ("identifier" or "slug")
+
+  /**
+   * @see sfTask
+   */
+  protected function configure()
+  {
+    $this->addArguments(array(
+      new sfCommandArgument('dip', sfCommandArgument::REQUIRED, 'The DIP directory.'),
+    ));
+
+    $this->addOptions(array(
+      new sfCommandOption('application', null, sfCommandOption::PARAMETER_OPTIONAL, 'The application name', true),
+      new sfCommandOption('env', null, sfCommandOption::PARAMETER_REQUIRED, 'The environment', 'cli'),
+      new sfCommandOption('connection', null, sfCommandOption::PARAMETER_REQUIRED, 'The connection name', 'propel'),
+      new sfCommandOption('undo-log-dir', null, sfCommandOption::PARAMETER_OPTIONAL, 'Directory to write undo logs to', false),
+      new sfCommandOption('audit', null, sfCommandOption::PARAMETER_NONE, 'Audit mode')
+    ));
+
+    $this->namespace = 'import';
+    $this->name = 'dip-objects';
+    $this->briefDescription = 'Import digital objects from Archivematica DIP using CSV file';
+
+    $this->detailedDescription = <<<EOF
+Process a CSV file to import digital objects from an Archivematica DIP to
+existing information objects in AtoM.
+
+The CSV file can be named anything, but must have the extension "csv"
+(lower-case).
+
+The CSV file must start with a header row specifying column order. A "filename"
+column must be included. Additionally, either an "identifier" or a "slug"
+column must be included (not both: one of these will be used to specify
+information objects to import to).
+
+The import script expects that Archivematica will have modified the filename
+so:
+
+a) The original object file has been converted to a derivative with a
+   correspondiing extension ("jpg" for example)
+
+b) A UUID has been pre-pended to the filename (for example:
+   "815da5cf-f49f-41f5-aa5d-c40d9d4dec3c-MARBLES.jpg")
+
+Note: Filenames must be unique (without UUID) to avoid colliding on import.
+
+The undo-log-dir option can be used to log which information objects have
+digital objects added to them. This log can be used, in event of an incomplete
+import, to either establish where the import stopped or to manually remove the
+imported digital objects. Undo logs contain two columns: the ID of the
+information object to which objects have beem imported and the DIP directory
+the objects were imported from.
+
+The audit option can be used to verify that all objects specified in a DIP's
+CSV file were imported. If any are found to be missing the object filename
+will be output.
+EOF;
+  }
+
+  /**
+   * @see sfTask
+   */
+  public function execute($arguments = array(), $options = array())
+  {
+    $this->checkArgumentsAndOptions($arguments, $options);
+
+    sfContext::createInstance($this->configuration);
+
+    QubitSearch::getInstance()->enable();
+
+    $databaseManager = new sfDatabaseManager($this->configuration);
+    $this->conn = $databaseManager->getDatabase('propel')->getConnection();
+
+    // Set dip directory
+    $this->dipDir = $arguments['dip'];
+
+    // Set undo log filename, if undo log directory is specified
+    $undoLog = null;
+    if (isset($options['undo-log-dir']))
+    {
+      $undoLog = rtrim($options['undo-log-dir'], '/') .'/'. date('Y-m-d') .'-'. basename($this->dipDir) .'.log';
+    }
+
+    // Set path to DIP objects
+    $objectsPath = rtrim($this->dipDir, '/') .'/objects';
+    $this->logSection(sprintf("Looking for objects in: %s", $this->dipDir));
+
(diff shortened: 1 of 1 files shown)


lib/task/import/importDipObjectsTask.class.php
```diff
@@ -0,0 +1,585 @@
...
+
+/**
+ * Process a CSV file to import digital objects from an Archivematica DIP and
+ * associate them with existing information objects in AtoM
+ *
+ * @package    symfony
+ * @subpackage task
+ * @author     David Juhasz <david@artefactual.com>
+ */
+class importDipObjectsTask extends sfBaseTask
```
jraddaoui: Could we extend arBaseTask?

lib/task/import/importDipObjectsTask.class.php
```diff
@@ -0,0 +1,585 @@
...
+  /**
+   * @see sfTask
+   */
+  public function execute($arguments = array(), $options = array())
+  {
+    $this->checkArgumentsAndOptions($arguments, $options);
+
+    sfContext::createInstance($this->configuration);
+
+    QubitSearch::getInstance()->disabled = false;
```
jraddaoui: That property is called 'enabled' and it's protected. Please, use 'QubitSearch::getInstance()->enable();' instead if that's really what you want, or  'QubitSearch::getInstance()->disable();' for the opposite.

lib/task/import/importDipObjectsTask.class.php
```diff
@@ -0,0 +1,585 @@
...
+
+    $databaseManager = new sfDatabaseManager($this->configuration);
+    $this->conn = $databaseManager->getDatabase('propel')->getConnection();
+
+    sfConfig::set('app_upload_dir', $this->getUploadDir());
```
jraddaoui: I think that, if you extend arBaseTask, that setting is already added to sfConfig and you could remove the 'getUploadDir' function bellow and probably this line.

lib/task/import/importDipObjectsTask.class.php
```diff
@@ -0,0 +1,585 @@
...
+    {
+      $undoLog = rtrim($options['undo-log-dir'], '/') .'/'. date('Y-m-d') .'-'. basename($this->dipDir);
+    }
+
+    // Set path to DIP objects
+    $objectsPath = rtrim($this->dipDir, '/') .'/objects';
+    $this->logSection(sprintf("Looking for objects in: %s", $this->dipDir));
+
+    // Parse CSV file and import/audit objects
+    $digitalObjects = $this->parseCsvData($this->openFirstCsvFile(), $objectsPath);
```
jraddaoui: You could sent the '$objectsPath' to the 'openFirstCsvFile()' function, as this variable already has the possible extra slash trimed, and in that function we're not doing the same.

lib/task/import/importDipObjectsTask.class.php
```diff
@@ -0,0 +1,585 @@
...
+   *
+   * @param string $objectsPath  Path to DIP objects
+   *
+   * @return array  Hash (see function description)
+   */
+  protected function createFilenameLookup($objectsPath)
+  {
+    if (!($files = scandir($objectsPath)))
+    {
+      throw new sfException("Couldn't read path '$objectsPath'");
```
jraddaoui: I think we could remove this check, at this point we should already have the csv file from the objects directory. We could get the files array directly in the foreach declaration bellow.

lib/task/import/importDipObjectsTask.class.php
```diff
@@ -0,0 +1,585 @@
...
+
+    return $filenames;
+  }
+
+  /*
+   * Parse UUID from string
+   *
+   * For example: "ad0acf68-34b4-47ea-aef7-e5e29c5b7388-bob.tif" would parse to
+   * "ad0acf68-34b4-47ea-aef7-e5e29c5b7388". If multiple UUIDs exist in a
+   * string the first found is returned. Boolean false is returned if no UUID
```
jraddaoui: It's returning null if no one is found

lib/task/import/importDipObjectsTask.class.php
```diff
@@ -0,0 +1,585 @@
...
+    {
+      throw new sfException('Error: CSV header row includes both an "identifier" column and a "slug" column. Please use only one of the two.');
+    }
+
+    // Make sure there is either an "identifier" or a "slug" column  in the CSV header row
+    if (!in_array('identifier', $this->columnNames) && !in_array('slug', $this->columnNames))
+    {
+      throw new sfException('Error: CSV header row must include either an "identifier" column or a "slug" column.');
+    }
+
```
jraddaoui, line 276: We could make both 'in_array()' checks at the begining of the function and asign the values to variables to avoid making them again other times.

lib/task/import/importDipObjectsTask.class.php
```diff
@@ -0,0 +1,585 @@
...
+    $this->processCsvHeaderRow($fh);
+
+    $digitalObjects = array();
+
+    // Build hash on information_object key, with value being an array if information
+    // object has multiple digital objects attached
+    while ($row = fgetcsv($fh, 1000))
+    {
+      $filename = $this->getRowColumnValue('filename', $row);
+      $filepath = rtrim($objectsPath, '/') .'/'. $filename;
```
jraddaoui: $objectsPath is already rtrimed.

lib/task/import/importDipObjectsTask.class.php
```diff
@@ -0,0 +1,585 @@
...
+    // information objects (slug or identifier)
+    $this->processCsvHeaderRow($fh);
+
+    $digitalObjects = array();
+
+    // Build hash on information_object key, with value being an array if information
+    // object has multiple digital objects attached
+    while ($row = fgetcsv($fh, 1000))
+    {
+      $filename = $this->getRowColumnValue('filename', $row);
```
jraddaoui, line 205: Could we get the filename column index only once instead of making am array_search() for each row?

lib/task/import/importDipObjectsTask.class.php
```diff
@@ -0,0 +1,585 @@
...
+        {
+          $filepath = str_replace($filename, $filenames[$key], $filepath);
+        }
+        else
+        {
+          throw new Exception("Error: Couldn't find file $filepath");
+        }
+      }
+
+      $uniqueValue = $this->getRowColumnValue($this->uniqueValueColumnName, $row);
```
jraddaoui, line 230: Same here, could we get the filename 'uniqueValueColumnName' index only once instead of making am array_search() for each row?

lib/task/import/importDipObjectsTask.class.php
```diff
@@ -0,0 +1,585 @@
...
+        {
+          $this->auditDigitalObject($item);
+          $count++;
+        }
+        else
+        {
+          // If more than one digital object linked to this information object
+          $parent = $this->getInformationObjectUsingUniqueId($key);
+
+          for ($i = 0; $i < count($item); $i++)
```
jraddaoui: Why not 'foreach $item as $do'? I think it's more readable.

lib/task/import/importDipObjectsTask.class.php
```diff
@@ -0,0 +1,585 @@
...
+      {
+        if (!is_array($item))
+        {
+          $this->auditDigitalObject($item);
+          $count++;
+        }
+        else
+        {
+          // If more than one digital object linked to this information object
+          $parent = $this->getInformationObjectUsingUniqueId($key);
```
jraddaoui: Do we need the parent for something in the audit case?

lib/task/import/importDipObjectsTask.class.php
```diff
@@ -0,0 +1,585 @@
...
+          }
+        }
+      }
+    }
+
+    return $count;
+  }
+
+  /**
+   * Import digital objects
```
jraddaoui: Wrong description

lib/task/import/importDipObjectsTask.class.php
```diff
@@ -0,0 +1,585 @@
...
+
+        if (!is_array($item))
+        {
+          $this->addDigitalObject($informationObject, $item, $undoLog);
+          $count++;
+        }
+        else
+        {
+          // If more than one digital object linked to this information object
+          $parent = $this->getInformationObjectUsingUniqueId($key);
```
jraddaoui: '$parent = $informationObject;'

lib/task/import/importDipObjectsTask.class.php
```diff
@@ -0,0 +1,585 @@
...
+
+          for ($i = 0; $i < count($item); $i++)
```
jraddaoui: Again, I think 'foreach' is more readable.

lib/task/import/importDipObjectsTask.class.php
```diff
@@ -0,0 +1,585 @@
...
+          }
+        }
+      }
+      else
+      {
+        $informationObject = $this->getInformationObjectUsingUniqueId($key);
+
+        if (!is_array($item))
+        {
+          $this->addDigitalObject($informationObject, $item, $undoLog);
```
jraddaoui, line 348: I think $undoLog is not here, you may need to make it global or send it to this function (importDigitalObjects).

lib/task/import/importDipObjectsTask.class.php
```diff
@@ -0,0 +1,585 @@
...
+          {
+            // Create new information objects, to maintain one-to-one
+            // relationship with digital objects
+            $informationObject = new QubitInformationObject;
+            $informationObject->parent = $parent;
+            $informationObject->title = basename($item[$i]);
+
+            $informationObject->save($this->conn);
+
+            $this->addDigitalObject($informationObject, $item[$i], $undoLog, true);
```
jraddaoui: Again, I think $undoLog is not here

lib/task/import/importDipObjectsTask.class.php
```diff
@@ -0,0 +1,585 @@
...
+    if (null !== $objectUUID = $this->getUUID(basename($filepath)))
+    {
+      $this->log('Creating property: object UUID '. $objectUUID);
+      $informationObject->addProperty('objectUUID', $objectUUID);
+    }
+
+    // Save and, optionally, log
+    $informationObject->save($this->conn);
+
+    if (isset($undoLog))
```
jraddaoui, line 487: I think it will always be null until we send it to 'importDigitalObjects()' first or we make it a global variable.

lib/task/import/importDipObjectsTask.class.php
mcantelon, line 205: I've added caching of column indexes, so should be good now.

lib/task/import/importDipObjectsTask.class.php
mcantelon, line 348: Ah, nice catch... thanks.

jraddaoui:
Awesome Mike, thanks!

I'd appreciate if you could keep the commit history intact until merge the next time. It makes easier to review the feedback changes ;)

mcantelon:
Thanks... sounds good!

mcantelon:
Merged.

PR #380, opened 2016-06-08
Metadata-only digital objects, refs #9969

sevein:
Redmine: https://projects.artefactual.com/issues/9969

Archivematica is incorporating a new type of DIP upload known as "metadata only" where the actual objects are not uploaded but its metadata. This pull request makes the necessary changes in AtoM to accommodate this new type of upload by introducing a new type of digital object where `usage` is `offline` (along with `master`, `representative` and `thumbnail`).

The HTTP API has also been updated to accept some extra fields that are relevant in this upload: format name, format version, format registry key and format registry name.

### DIFF
--- a/lib/model/QubitInformationObject.php
+++ b/lib/model/QubitInformationObject.php
@@ -65,6 +65,10 @@ class QubitInformationObject extends BaseInformationObject
 
     switch ($name)
     {
+      case 'formatName':
+      case 'formatVersion':
+      case 'formatRegistryKey':
+      case 'formatRegistryName':
       case 'objectUUID':
       case 'aipUUID':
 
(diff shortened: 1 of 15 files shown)


lib/task/migrate/migrations/arMigration0142.class.php
```diff
@@ -0,0 +1,46 @@
...
+/*
+ * Add new term to digital object usages taxonomy
+ *
+ * @package    AccesstoMemory
+ * @subpackage migration
+ */
+class arMigration0142
+{
+  const
+    VERSION = 142, // The new database version
```
mcantelon, line 29: The fixture sets the version to 141... should they be different?

lib/model/QubitInformationObject.php
```diff
@@ -1218,6 +1222,10 @@ public function getDigitalObjectPublicUrl()
       {
         return $path;
       }
+      else if (QubitTerm::OFFLINE_ID === $do->usageId)
+      {
+        throw new sfException('getDigitalOBjectPublicUrl() is not available for offline digital objects');
```
mcantelon: Typo: "OB" instead of "Ob"

lib/task/migrate/migrations/arMigration0142.class.php
sevein, line 29: Thank you, I've fixed it!

sevein:
Thank you @mcantelon, I'll merge once you confirm you're 👍 with it.

mcantelon:
Looks good to me!

PR #381, opened 2016-06-10
Fallback to slug basis title, refs #9981

MikeFE:
When first installing AtoM, some slugs are apparently generated before fixtures
are inserted into the database. This was causing the code to look for the
slug basis setting before it existed, and it was throwing an exception.


mcantelon:
Looks good to me!

PR #382, opened 2016-06-10
Add digital object text adv search, refs #9952

MikeFE:
Add the option to search digital object transcript text via the advanced search
options.


mcantelon:
Looks good!

PR #383, opened 2016-06-10
Add end date sorting to description browse #9953

MikeFE:
No other sortPickers being used in AtoM use the 'date' sort type as far as I
can tell, so removing the 'date' case and just renaming it to startDate in
apps/qubit/modules/informationobject/actions/browseAction.class.php should
be okay.


mcantelon:
Seems good to me!

PR #385, opened 2016-06-14
Fixed issue with OAI virtual sets. refs #9889

mcantelon:
Fixed issue with sets not being passed when querying for descriptions to
display.


mcantelon:
Merged.

PR #386, opened 2016-06-14
Added label for accession setting. refs #9915

mcantelon:
Added a label for the accession mask enabled setting.


mcantelon:
Merged.

PR #387, opened 2016-06-15
Enhance csv import matching behavior, refs #9995

sbreker:
Added a new checksum column for digital objects on both the import and export
csv files.  This will affect CLI imports when the -update parameter is set.
In this case, if a checksum is present in the input file, and it matches
the checksum in the database, the digital object will not be updated.  If
the checksum is missing or does not match, the digital object will be
loaded as normal.

This change will prevent digital objects from being reloaded on an -update
if they are exact copies of one another.


lib/model/QubitInformationObject.php
```diff
@@ -1168,6 +1168,28 @@ public function getDigitalObject()
...
+    // Set digital object URL
+    $digitalObjects = $this->getDigitalObjects();
+    if (count($digitalObjects) > 0)
+    {
+      $do = $this->digitalObjects[0];
+      $checksum = $do->getChecksum();
+
+      return $checksum;
+    }
+    else
```
MikeFE: Else isn't really needed here. You can just return null after the if. Also, if you don't explicitly return at all, PHP returns null by default in functions. Very minor point though, I think this is fine.

lib/model/QubitInformationObject.php
```diff
@@ -1168,6 +1168,28 @@ public function getDigitalObject()
...
+  public function getDigitalObjectChecksum()
+  {
+    // Set digital object URL
+    $digitalObjects = $this->getDigitalObjects();
+    if (count($digitalObjects) > 0)
+    {
+      $do = $this->digitalObjects[0];
+      $checksum = $do->getChecksum();
+
+      return $checksum;
```
MikeFE: maybe just return $this->digitalObjects[0]->getChecksum(); ?

lib/task/import/csvImportTask.class.php
```diff
@@ -219,6 +219,9 @@ public function execute($arguments = array(), $options = array())
       // Where to log errors to
       'errorLog' => $options['error-log'],
 
+      // Is the 'update' parameter set
+      'doUpdate' => ($options['update']) ? true : false,
```
MikeFE: This will generate a PHP notice if the 'update' key doesn't exist in the $options array (it might not show up in AtoM--we sometimes suppress warnings and notices). I think also the parentheses are optional in PHP here (though in C if I recall they are required!). I'd do 'doUpdate' => isset($options['update']) ? true : false;

MikeFE: We also have a variable (assigned to @ line 1103 looks like) called 'updateExisting' which seems to be doing the same thing as 'doUpdate' is here. I think you should remove either that or this 'doUpdate' one (I'm noticing updateExisting has a  similar undeclared variable issue as I commented on above--heh). I don't really have a preference which one we should keep.

lib/task/import/csvImportTask.class.php
```diff
@@ -439,14 +443,38 @@ public function execute($arguments = array(), $options = array())
...
+
+            if ($self->doUpdate)
+            {
+              // if   - there is a checksum in the import file
+              //      - the checksum is non-blank
+              //      - the checksum in the csv file matches what is in the database
+              // then - do not re-load the digital object from the import file on UPDATE (leave existing recs as is)
+              // else - reload the digital object in the import file (i.e. delete existing record below)
+              if ((isset($self->rowStatusVars['digitalObjectChecksum']) && $self->rowStatusVars['digitalObjectChecksum'])
+                && (strcmp($self->rowStatusVars['digitalObjectChecksum'], $results->getChecksum()) == 0 ))
```
MikeFE: the () around strcmp is not needed here I don't think

lib/task/import/csvImportTask.class.php
```diff
@@ -1007,7 +1066,7 @@ public function execute($arguments = array(), $options = array())
             $this->log($e->getMessage(), sfLogger::ERR);
```
djjuhasz: This if block looks like a copy and paste of the code at lines 450 - 474.  I would recommend setting a variable on the first check rather than doing the check twice.

lib/task/import/csvImportTask.class.php
```diff
@@ -439,14 +440,38 @@ public function execute($arguments = array(), $options = array())
...
+              //      - the checksum in the csv file matches what is in the database
+              // then - do not re-load the digital object from the import file on UPDATE (leave existing recs as is)
+              // else - reload the digital object in the import file (i.e. delete existing record below)
+              if ((isset($self->rowStatusVars['digitalObjectChecksum']) && $self->rowStatusVars['digitalObjectChecksum'])
+                && strcmp($self->rowStatusVars['digitalObjectChecksum'], $results->getChecksum()) == 0 )
+              {
+                // if the checksum matches what is stored with digital object, do not import this digital object.
+                $deleteDigitalObject = false;
+              }
+              else
```
djjuhasz: This else block is unnecessary, as you've set $deleteDigitalObject = true on line 450

lib/task/import/csvImportTask.class.php
```diff
@@ -439,14 +440,38 @@ public function execute($arguments = array(), $options = array())
...
+              {
+                // if checksum mismatch then delete digital objects so they can be reloaded.
+                $deleteDigitalObject = true;
+              }
+            }
+            if ($deleteDigitalObject)
```
djjuhasz, line 466: There should be a blank line before this "if" statement

lib/task/import/csvImportTask.class.php
sbreker: Those code blocks are in different lambda functions, and I don't think that I can set a local var and have it be avail between the two.   The first lambda is conditionally called and if a var is set in there for the next function, it will be a time bomb I think.

lib/task/import/csvImportTask.class.php
```diff
@@ -979,7 +998,19 @@ public function execute($arguments = array(), $options = array())
...
+        // exist and therefore should be created.
+
+        // When this CSV importer is run and the --update parameter is set,
+        // the function 'updatePreparationLogic' (above) will be run.
+        // 'updatePreparationLogic' has been enhanced to compare the DO
+        // checksum in the DB against the checksum in the import CSV and will
+        // only remove the DO if the checksums differ.  If they are removed
+        // because the checksums are different, the following code will
+        // recreate them from the CSV file.
+        if (($uri = $self->rowStatusVars['digitalObjectURI']) && null === $self->object->getDigitalObject())
```
djjuhasz: I think the parentheses around `($uri = $self->rowStatusVars['digitalObjectURI'])` are unnecessary.

lib/task/import/csvImportTask.class.php
```diff
@@ -1007,7 +1038,7 @@ public function execute($arguments = array(), $options = array())
             $this->log($e->getMessage(), sfLogger::ERR);
           }
         }
-        else if ($path = $self->rowStatusVars['digitalObjectPath'])
+        else if (($path = $self->rowStatusVars['digitalObjectPath']) && null === $self->object->getDigitalObject())
```
djjuhasz: I think the parentheses around `($uri = $self->rowStatusVars['digitalObjectPath'])` are unnecessary.

lib/model/QubitInformationObject.php
```diff
@@ -1168,6 +1168,22 @@ public function getDigitalObject()
...
 
   /**
+   * Get the digital object's checksum value
+   *
+   * @return string  digital object checksum or null
+   */
+  public function getDigitalObjectChecksum()
+  {
+    // Set digital object URL
+    $digitalObjects = $this->getDigitalObjects();
```
djjuhasz: Does this return only the master digital object (information_object.id = digital_object.information_object_id)?  IIRC it doesn't return derivatives.

MikeFE: I think there's a `QubitInformationObject::getDigitalObject()` that does basically what this code is doing. I'm actually not sure if getDigitalObject**s**() ever returns an array with more than one digital object.

https://github.com/artefactual/atom/blob/qa/2.3.x/lib/model/QubitInformationObject.php#L1142

MikeFE: I'd do 

```
if (null !== $do = $this->getDigitalObject())
{
  return $do->getChecksum();
}

return null;
```

lib/task/import/csvImportTask.class.php
```diff
@@ -439,14 +440,32 @@ public function execute($arguments = array(), $options = array())
...
+
+            if ($self->updateExisting)
+            {
+              // if   - there is a checksum in the import file
+              //      - the checksum is non-blank
+              //      - the checksum in the csv file matches what is in the database
+              // then - do not re-load the digital object from the import file on UPDATE (leave existing recs as is)
+              // else - reload the digital object in the import file (i.e. delete existing record below)
+              if ((isset($self->rowStatusVars['digitalObjectChecksum']) && $self->rowStatusVars['digitalObjectChecksum'])
+                && strcmp($self->rowStatusVars['digitalObjectChecksum'], $do->getChecksum()) === 0 )
```
djjuhasz: Should use yoda condition, e.g. `0 === strcmp($self->rowStatusVars['digitalObjectChecksum'], $do->getChecksum())`

lib/model/QubitInformationObject.php
```diff
@@ -1168,6 +1168,21 @@ public function getDigitalObject()
...
+   * @return string  digital object checksum or null
+   */
+  public function getDigitalObjectChecksum()
+  {
+    if (null !== $do = $this->getDigitalObject())
+    {
+      return $do->getChecksum();
+    }
+
+    return null;
```
djjuhasz: Sorry, missed this even though Mike commented on it.  We don't explicitly call `return null` in functions as it's the default behaviour in PHP

MikeFE: We don't? :)

djjuhasz: Well, we didn't in my days as a developer.  I don't know what you crazy kids get up to these days. :stuck_out_tongue_winking_eye:

MikeFE: I'm not sure, I genuinely can't remember it ever coming up in CR! haha. I
don't mind either way.

Mike Gale,

lib/task/import/csvImportTask.class.php
```diff
@@ -1038,7 +1072,7 @@ public function execute($arguments = array(), $options = array())
     $import->searchIndexingDisabled = ($options['index']) ? false : true;
 
     // Allow updating to be enabled via a CLI option
-    $import->updateExisting = ($options['update']);
+    $import->updateExisting = isset($options['update']) ? true : false;
```
MikeFE: OK last suggestion I promise if it isn't too late! This can just be `$import->updateExisting = isset($options['update']);` since it'll return true/false anyways (my bad, I didn't catch it earlier).

PR #389, opened 2016-06-17
Derivative regeneration: type option. refs #9964

mcantelon:
Added a command-line option, to the derivative regeneration task, that allows
optional selection of derivative type ("reference" or "thumbnail"). Default
is derivative type is master.


lib/task/digitalobject/digitalObjectRegenDerivativesTask.class.php
```diff
@@ -191,7 +192,21 @@ public static function regenerateDerivatives(&$digitalObject)
       }
     }
 
-    $digitalObject->createRepresentations(QubitTerm::MASTER_ID, $conn);
+    switch($type)
+    {
+      case "reference":
```
jraddaoui, line 197: We try to use single quotes when possible.

mcantelon:
Merged.

PR #391, opened 2016-06-21
Keep transcript renaming DO filename, refs #9983

jraddaoui:
It's needed to avoid an error trying to save a deleted property when the
method is called from IO rename action.
- Send 'keepTranscript' option to regenerateDerivatives
- Regenerate derivates before saving DO to reflect all changes

### DIFF
--- a/apps/qubit/modules/informationobject/actions/renameAction.class.php
+++ b/apps/qubit/modules/informationobject/actions/renameAction.class.php
@@ -140,7 +140,7 @@ class InformationObjectRenameAction extends DefaultEditAction
       rename($oldFilePath, $newFilePath);
       chmod($newFilePath, 0644);
 
-      // Regeneraate derivatives
+      // Regenerate derivatives
       digitalObjectRegenDerivativesTask::regenerateDerivatives($digitalObject, array('keepTranscript' => true));
 
       // Change name in database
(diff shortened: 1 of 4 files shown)


lib/task/digitalobject/digitalObjectRegenDerivativesTask.class.php
```diff
@@ -182,12 +182,15 @@ public static function regenerateDerivatives(&$digitalObject)
       $derivative->delete();
     }
 
-    // Delete existing transcripts
-    foreach ($digitalObject->propertys as $property)
+    // Delete existing transcripts if 'keepTranscript' option is not sent or it's false
+    if (!isset($options['keepTranscript']) || !$options['keepTranscript'])
     {
-      if ('transcript' == $property->name)
+      foreach ($digitalObject->propertys as $property)
```
sevein: `QubitInformationObject` implements `getPropertyByName`, should `QubitDigitalObject` do the same?

apps/qubit/modules/informationobject/actions/renameAction.class.php
```diff
@@ -140,12 +140,12 @@ private function updateResource()
       rename($oldFilePath, $newFilePath);
       chmod($newFilePath, 0644);
 
+      // Regeneraate derivatives
```
sevein: Typo!

sevein:
Can you add to the commit message a description of the actual problem and how you are addressing it? Code comments in the regenerateDerivatives function would be really useful too. Just trying to make sure that someone else looking at this code will understand why we're doing this.

PR #392, opened 2016-06-22
Improve job report UI consistency, refs #10049

MikeFE:
Redmine: https://projects.artefactual.com/issues/10049

### DIFF
--- a/apps/qubit/modules/jobs/templates/reportSuccess.php
+++ b/apps/qubit/modules/jobs/templates/reportSuccess.php
@@ -48,7 +48,7 @@
 
   <?php if ($job->getObjectModule() && $job->getObjectSlug()): ?>
     <div class="job-report-field">
-      <div><?php echo __('Associated item') ?></div>
+      <div><?php echo __('Associated record') ?></div>
       <div><?php echo link_to(__('Link'), array('module' => $job->getObjectModule(), 'slug' => $job->getObjectSlug())) ?></div>
     </div>
   <?php endif; ?>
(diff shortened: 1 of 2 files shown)


apps/qubit/modules/jobs/templates/reportSuccess.php
```diff
@@ -48,26 +48,30 @@
...
-    <?php echo render_value($job->output) ?>
+  <div>
+    <pre id="job-log-output"><?php echo render_value(trim($job->output)) ?></pre>
   </div>
 </section>
 
-<div id="job-return-link">
-  <?php echo link_to(__('« Return to jobs management page'), array('module' => 'jobs', 'action' => 'browse')) ?>
-</div>
+<section class="actions" id="job-return-link">
```
sevein: Almost, but you need to take the `actions` out of the main `#content`. We do that with slots, see for example: https://github.com/artefactual/atom/blob/qa/2.3.x/plugins/qtAccessionPlugin/modules/accession/templates/indexSuccess.php. Once you do that change, you shouldn't need `#job-return-link`.

plugins/arDominionPlugin/css/less/jobs.less
```diff
@@ -48,7 +48,7 @@
   margin: 5px 5px 0px 5px;
   padding: 5px;
   background-color: #f2f2f2;
-  border-radius: 10px;
+  border-radius: 4px;
   color: @black;
   font-family: Courier New;
   max-height: 400px;
```
sevein, line 53: I'd prefer no to use `max-height`, there's nothing wrong with having the user scrolling down the page, unless you expect that the contents inside the log are going to be huge?

Also, do you think we should use `white-space: nowrap;` together with `overflow: auto;` so the text is not wrapped but it can be horizontally scrolled when needed?

plugins/arDominionPlugin/css/less/jobs.less
```diff
@@ -48,7 +48,7 @@
   margin: 5px 5px 0px 5px;
   padding: 5px;
   background-color: #f2f2f2;
-  border-radius: 10px;
+  border-radius: 4px;
   color: @black;
   font-family: Courier New;
```
sevein: Don't do that, unless it's really necessary. Our theme sets `pre` to a lesscss variable called `@monoFontFamily` (see [here](https://github.com/artefactual/atom/blob/qa/2.4.x/plugins/arDominionPlugin/css/less/variables.less#L71)), which equals to `Monaco, Menlo, Consolas, "Courier New", monospace`, providing better consistency across multiple browsers and platforms.

plugins/arDominionPlugin/css/less/jobs.less
```diff
@@ -48,7 +48,7 @@
   margin: 5px 5px 0px 5px;
   padding: 5px;
   background-color: #f2f2f2;
-  border-radius: 10px;
+  border-radius: 4px;
   color: @black;
```
sevein: Try `color: #333`, I think it'll look much better.

plugins/arDominionPlugin/css/less/jobs.less
MikeFE, line 53: Yes, the job log might be huge in the future. We'll be moving all the CSV / XML import tasks to job scheduler soon, the output of these tasks might be hundreds of lines if it's a large import.

sevein, line 53: Makes sense!

sevein:
I merged! 👍

PR #393, opened 2016-06-22
Fix browse digital object link, refs #9949

MikeFE:
Previously, clicking browse digital object would only show digital objects
that belong to top level descriptions by default. Now the link will show
all digital objects regardless of the level of description by default.


jraddaoui:
Hi Mike, it looks good, but could you modify the version displayed at the end of the upgrade-sql task, so it says 2.4 instead of 2.3. Thanks!

PR #394, opened 2016-06-24
Modification of the method replaceLineBreak to find <lb/> tags nested…

pierremancini:
```
… in <physdesc> tag

When importing a xml ead file all the <lb/> tags aren't read:
A <physdesc> tag may contain an <extent> tag that containt <lb/>. Before modifaction <lb/> was hidden inside <extent>.
```

### DIFF
--- a/lib/QubitXmlImport.class.php
+++ b/lib/QubitXmlImport.class.php
@@ -303,11 +303,14 @@ class QubitXmlImport
             continue;
           }
 
+
+          //A remettre avant push, Pas de changement de langue pour pouvoir tester plus facilement
+          /*
           if ($currentCulture !== $twoCharCode)
           {
             $this->errors[] = sfContext::getInstance()->i18n->__('EAD "langmaterial" is set to').': "'.$isocode.'" ('.format_language($twoCharCode, 'en').'). '.sfContext::getInstance()->i18n->__('Your XML document has been saved in this language and your user interface has just been switched to this language.');
           }
-          $sf_user->setCulture($twoCharCode);
+          $sf_user->setCulture($twoCharCode);*/
           // can only set to one language, so have to break once the first valid language is encountered
           break;
         }
@@ -810,6 +813,7 @@ class QubitXmlImport
   public static function replaceLineBreaks($node)
   {
     $nodeValue = '';
+    $fieldsArray = array('extent', 'physfacet', 'dimensions');
 
     foreach ($node->childNodes as $child)
     {
@@ -817,29 +821,25 @@ class QubitXmlImport
       {
         $nodeValue .= "\n";
       }
-      else
+      else if (in_array($child->tagName, $fieldsArray)) 
       {
-        if($child->tagName=='extent') 
+        foreach ($child->childNodes as $childNode)
         {
-          foreach ($child->childNodes as $childChild)
+          if ($childNode->nodeName == 'lb')
           {
-            if ($childChild->nodeName == 'lb')
-            {
-              $nodeValue .= "\n";
-            } 
-            else
-            {
-              $nodeValue .= preg_replace('/[\n\r\s]+/', ' ', $childChild->nodeValue);
-            }
+            $nodeValue .= "\n";
+          } 
+          else
+          {
+            $nodeValue .= preg_replace('/[\n\r\s]+/', ' ', $childNode->nodeValue);
           }
-        } else {
-
-          $nodeValue .= preg_replace('/[\n\r\s]+/', ' ', $child->nodeValue);
-
         }
+      } 
+      else 
+      {
+        $nodeValue .= preg_replace('/[\n\r\s]+/', ' ', $child->nodeValue);
       }
     }
-
     return $nodeValue;
   }
 
--- a/lib/QubitXmlImport.class.php
+++ b/lib/QubitXmlImport.class.php
@@ -819,7 +819,24 @@ class QubitXmlImport
       }
       else
       {
-        $nodeValue .= preg_replace('/[\n\r\s]+/', ' ', $child->nodeValue);
+        if($child->tagName=='extent') 
+        {
+          foreach ($child->childNodes as $childChild)
+          {
+            if ($childChild->nodeName == 'lb')
+            {
+              $nodeValue .= "\n";
+            } 
+            else
+            {
+              $nodeValue .= preg_replace('/[\n\r\s]+/', ' ', $childChild->nodeValue);
+            }
+          }
+        } else {
+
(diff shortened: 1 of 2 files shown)


fiver-watson:
Hi @Pierrocana, thank you for your contribution to the AtoM project. Our developers will [review](https://wiki.accesstomemory.org/Development/Code_review) your pull request against our [coding standard](https://wiki.accesstomemory.org/Development/Coding_standard) soon. In the meantime, will you please review the Contributor's guidelines we have posted, and send us a signed copy of the Contributor's agreement to agreement@artefactual.com? Details and a link to the Agreement can be found here:  
- https://wiki.accesstomemory.org/Development/Contribute_code

Thanks again!

pierremancini:
Hi Dan, thank you for the documentation. I read it carefully.

To introduce my self, I am a developper experienced with one year and half
in php and currently working for the Centre National de la Recherche
Scientifique (CNRS) in Pessac (France). Our project is to use Atom to put
online digitized maps with its metadata. We especially intent to use the
XML/EAD import functionality.

I sent a signed copy of the Contributor's agreement.

Best regards,

Pierre Mancini

Centre _REGARDS_ CNRS – UMR 5319 _PASSAGES_

05.56.84.68.58

www.regards.cnrs.fr
http://imageo.hypotheses.org/

2016-06-24 18:42 GMT+02:00 Dan Gillean notifications@github.com:

lib/QubitXmlImport.class.php
```diff
@@ -819,7 +819,24 @@ public static function replaceLineBreaks($node)
       }
       else
       {
-        $nodeValue .= preg_replace('/[\n\r\s]+/', ' ', $child->nodeValue);
+        if($child->tagName=='extent')
```
jraddaoui: Some spaces missing around the operator and after the if, it should be ...

`if ($child->tagName == 'extent')`

Also, we could remove this nested 'if' by making the previous 'else' an 'else if' with this condition.

lib/QubitXmlImport.class.php
```diff
@@ -819,7 +819,24 @@ public static function replaceLineBreaks($node)
       }
       else
       {
-        $nodeValue .= preg_replace('/[\n\r\s]+/', ' ', $child->nodeValue);
+        if($child->tagName=='extent') 
+        {
+          foreach ($child->childNodes as $childChild)
```
jraddaoui: I'd prefer `$childNode` instead of `$childChild`, but nothing important.

lib/QubitXmlImport.class.php
```diff
@@ -819,7 +819,24 @@ public static function replaceLineBreaks($node)
...
+            if ($childChild->nodeName == 'lb')
+            {
+              $nodeValue .= "\n";
+            } 
+            else
+            {
+              $nodeValue .= preg_replace('/[\n\r\s]+/', ' ', $childChild->nodeValue);
+            }
+          }
+        } else {
```
jraddaoui: The `else` and the bracket should be both in new lines, like the other `else` before.

lib/QubitXmlImport.class.php
```diff
@@ -819,7 +819,24 @@ public static function replaceLineBreaks($node)
...
+
+          $nodeValue .= preg_replace('/[\n\r\s]+/', ' ', $child->nodeValue);
+
+        }
```
jraddaoui, line 833: The else and the bracket should be both in new lines, like the other else before. And please remove the empty lines in the block if there is only one line.

``` php
}
else 
{
  $nodeValue .= preg_replace('/[\n\r\s]+/', ' ', $child->nodeValue);
}
```

jraddaoui:
@fiver-watson: This will only replace 'lb' tags inside the 'extent' tag, which were not being replaced because we are targeting the 'physdesc' and we don't look in all descendants to remove the line breaks, only in the direct children. Not sure if this could happen in other elements, but it should be easy to add after this, maybe if you know of some we could add them directly.

@Pierrocana: Thank you for the PR and nice work. I've added some comments to the changes, nothing big. Also, if Dan thinks we should add more elements, we could add an array of fields and change the condition for `else if (in_array($child->tagName, $fieldsArray))`.

And one more thing, about the commit message, we try to limit the subject line to 50 characters and wrap the body at 72/80 characters, please check:

http://chris.beams.io/posts/git-commit/

Thanks!

fiver-watson:
Thanks for taking a look, @jraddaoui ! I'm happy to accept this as-is for now if that is easiest - we are trying to finalize the 2.3 release. But looking into it a bit to explore our options, the lb element in EAD can appear within many other elements: 

> abstract, addressline, archref, author, bibref, bibseries, container, corpname, creation, date, descrules, dimensions, edition, emph, entry, event, extent, extref, extrefloc, famname, function, genreform, geogname, head, head01, head02, imprint, item, label, langmaterial, language, langusage, legalstatus, materialspec, name, num, occupation, origination, p, persname, physdesc, physfacet, physloc, publisher, ref, refloc, repository, resource, runner, sponsor, subarea, subject, subtitle, title, titleproper, unitdate, unitid, unittitle

http://www.loc.gov/ead/tglib/elements/lb.html

That said, some of these elements (such as language) map to controlled values in AtoM; others (such as unitid) don't necessarily make sense to support given AtoM's UI, or are edge cases at best. Finally, many elements (such as scopecontent) require nested p tags, but adding blanket support for lb in any p element could cause issues - and I believe that lb is importing properly within scopecontent anyway. 

Of the above, if we were going to add an array, I think the elements it would make most sense to support would be the other physdesc nested elements - e.g. physfacet and  dimensions, in addition to extent, and physdesc itself since it can accept #PCDATA.

jraddaoui:
Hi @fiver-watson, most of those elements are targeted directly, so it will replace lb tags in most of them. It's only for special cases where we target the parent of the elements that need the replacement,  and I think we don't have a lot that are processed this way (I think bioghist may be one of those).

In this case the #PCDATA inside the physdesc element is being processed already as it's the target element, and after this fix the child extent element would be processed too, so I agree we should include the other physdesc childs in here.

@Pierrocana, if you agree, it should be an easy addon, as I said, we could create and array with those elements tag names (`$fieldsArray = array('extent', 'physfacet', 'dimensions');`) and check it with in_array().

Thanks!

pierremancini:
Ok, I see no problem in checking other elements tag names. Thank you for the correction.

jraddaoui:
Merged in https://github.com/artefactual/atom/commit/d45cefcb4e85a1fa483bf25a945e9f1c5cca27a6

I've created a Redmine ticket to keep track of the changes: https://projects.artefactual.com/issues/10105

@Pierrocana: I've added some clean-ups and rewritted the commit message to follow the rules I said before and to point to the new redmine ticket. Thanks again for all your work :wink:

PR #397, opened 2016-06-27
Speed up PDF page counting, refs #10062

mcantelon:
Redmine: https://projects.artefactual.com/issues/10062

Use pdfinfo, if available, to determine how many pages of content a PDF has. If pdfinfo isn't installed then the option to create thumbnails from pages, other than the front page, gets disabled.

### DIFF
--- a/lib/model/QubitDigitalObject.php
+++ b/lib/model/QubitDigitalObject.php
@@ -2045,16 +2045,21 @@ class QubitDigitalObject extends BaseDigitalObject
   {
     if ($this->canThumbnail() && self::hasImageMagick())
     {
-      if (QubitTerm::EXTERNAL_URI_ID == $this->usageId)
-      {
-        $command = 'identify '.$this->getLocalPath();
-      }
-      else
-      {
-        $command = 'identify '.$this->getAbsolutePath();
-      }
+      $filename = (QubitTerm::EXTERNAL_URI_ID == $this->usageId) ? $this->getLocalPath() : $this->getAbsolutePath();
 
-      exec($command, $output, $status);
+      $extension = pathinfo($filename, PATHINFO_EXTENSION);
+
+      // If processing a PDF, attempt to use pdfinfo as it's faster
+      if (strtolower($extension) == 'pdf' && sfImageMagickAdapter::pdfinfoToolAvailable())
+      {
+        $pages = sfImageMagickAdapter::getPdfinfoPageCount($filename);
+      }
+      else 
+      {
+        $command = 'identify '. $filename;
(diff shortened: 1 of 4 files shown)


lib/model/QubitDigitalObject.php
```diff
@@ -2045,25 +2045,30 @@ public function setPageCount($connection = null)
...
-
       if ($status == 0)
       {
         // Add "number of pages" property
         $pageCount = new QubitProperty;
         $pageCount->setObjectId($this->id);
         $pageCount->setName('page_count');
         $pageCount->setScope('digital_object');
-        $pageCount->setValue(count($output), array('sourceCulture' => true));
+        $pageCount->setValue($pages, array('sourceCulture' => true));
```
MikeFE, line 2071: This might be out of scope (a separate bug), but happens if the user regenerates their derivatives for a PDF? Will it create a duplicate property in the database?

plugins/sfThumbnailPlugin/lib/sfImageMagickAdapter.class.php
```diff
@@ -390,6 +398,43 @@ private function getCount($image)
...
+  {
+    $command = 'pdfinfo '. escapeshellarg($filename);
+    exec($command, $stdout, $retval);
+    if ($retval === 1)
+    {
+      throw new Exception('PDF could not be analyzed.');
+    }
+
+    // Parse page number from output
+    foreach($stdout as $line)
```
MikeFE: space here

plugins/sfThumbnailPlugin/lib/sfImageMagickAdapter.class.php
```diff
@@ -390,6 +398,43 @@ private function getCount($image)
...
+      return 1;
+    }
+
+    return sfImageMagickAdapter::getPdfinfoPageCount($filename);
+  }
+
+  public static function getPdfinfoPageCount($filename)
+  {
+    $command = 'pdfinfo '. escapeshellarg($filename);
+    exec($command, $stdout, $retval);
```
MikeFE: probably don't need $command as a variable if we just use it once after?

lib/model/QubitDigitalObject.php
mcantelon, line 2071: Hrm... I'll do a test to see if deleting a digital object deletes its properties.

mcantelon, line 2071: Yeah, seems like these properties get deleted when the objects deleted.

mcantelon:
Merged.

PR #399, opened 2016-06-30
Fix type issue in derivative regeneration task

mcantelon:
Redmine: https://projects.artefactual.com/issues/9964

Fixed issues related to regenerating derivatives of a specific type.

### DIFF
--- a/lib/model/QubitDigitalObject.php
+++ b/lib/model/QubitDigitalObject.php
@@ -2004,17 +2004,25 @@ class QubitDigitalObject extends BaseDigitalObject
         break;
 
       case QubitTerm::TEXT_ID:
-        if ($usageId == QubitTerm::EXTERNAL_URI_ID || $usageId == QubitTerm::MASTER_ID)
+        if ($this->canThumbnail())
         {
-          // Thumbnail PDFs (may add other formats in future)
-          if ($this->canThumbnail())
+          if ($usageId == QubitTerm::EXTERNAL_URI_ID || $usageId == QubitTerm::MASTER_ID)
           {
+            // Thumbnail PDFs (may add other formats in future)
             $this->createReferenceImage($connection);
             $this->createThumbnail($connection);
-          }
 
-          // Extract text
-          $this->extractText($connection);
+            // Extract text
+            $this->extractText($connection);
+          }
+          else if ($usageId == QubitTerm::REFERENCE_ID)
+          {
+            $this->createReferenceImage($connection);
+          }
+          else if ($usageId == QubitTerm::THUMBNAIL_ID)
+          {
+            $this->createThumbnail($connection);
+          }
         }
 
         break;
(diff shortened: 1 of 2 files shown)


lib/model/QubitDigitalObject.php
```diff
@@ -2010,6 +2010,17 @@ public function createRepresentations($usageId, $connection = null)
           // Extract text
           $this->extractText($connection);
         }
+        else if ($this->canThumbnail())
```
djjuhasz: I think it would be better to wrap the whole TEXT case body in `if ($this->canThumbnail())`, similar to the IMAGE case.  If the PDF can't be thumbnailed then none of the derivative creation methods will work.

lib/task/digitalobject/digitalObjectRegenDerivativesTask.class.php
```diff
@@ -174,9 +174,25 @@ protected function execute($arguments = array(), $options = array())
...
+        break;
+
+      default:
+        $usageId = QubitTerm::MASTER_ID;
+    }
+
     // Delete existing derivatives
     $criteria = new Criteria;
     $criteria->add(QubitDigitalObject::PARENT_ID, $digitalObject->id);
+    $criteria->add(QubitDigitalObject::USAGE_ID, $usageId);
```
djjuhasz: Hmm, there should never be a case where `$usageId == QubitTerm::MASTER_ID` and `parent_id <> NULL`.  In practice this means the that when `$usageId = QubitTerm::MASTER_ID` this query will return nothing, and I think it should return all derivatives.  You could wrap this line in an `if (QubitTerm::MASTER_ID != $usageId)` conditional to produce the desired results.

mcantelon: Ah good catch.

lib/task/digitalobject/digitalObjectRegenDerivativesTask.class.php
```diff
@@ -174,9 +174,28 @@ protected function execute($arguments = array(), $options = array())
...
+        break;
+
+      default:
+        $usageId = QubitTerm::MASTER_ID;
+    }
+
     // Delete existing derivatives
     $criteria = new Criteria;
-    $criteria->add(QubitDigitalObject::PARENT_ID, $digitalObject->id);
+    $criteria->add(QubitDigitalObject::USAGE_ID, $usageId);
```
djjuhasz: No, this will delete _ALL_ derivatives in the application of the given type when `$type == thumbnail` or `$type == reference`.  :fearful:  

When no $type is specified we want to regenerate the thumbnail and reference derivatives for _this $digitalObject only_.  I would recommend:

```
$criteria->add(QubitDigitalObject::PARENT_ID, $digitalObject->id);
if (isset($usageId))
{
  $criteria->add(QubitDigitalObject::USAGE_ID, $usageId);
}
```

And delete the `default:` clause from your switch statement (lines 187-189), so `$usageId = null` when `$type = null`.

mcantelon: Ack... sorry... I've revised.

mcantelon:
Merged.

PR #400, opened 2016-07-04
Enhance XML-EAD and CSV import matching behaviour, refs #9989

sbreker:
Added logic in the XML EAD import to try to match the information
object being loaded to a record in the database. If a record is
found, the information object will not be loaded and message
containing the record identifier will be logged. The XML EAD
process did not historically try to match to existing records.

Matching logic for XML EAD import will cascade through different
matching logic:
- first try to match the file name to the keymap table
- if no match, try to match against identifier, title, repo name
- if both fail, no match and record will not be loaded

XML EAD import process also modified to create a keymap record
when an information object is successfully matched and loaded.

### DIFF
--- a/lib/QubitFlatfileImport.class.php
+++ b/lib/QubitFlatfileImport.class.php
@@ -616,10 +616,14 @@ class QubitFlatfileImport
       {
         $this->object = $this->createOrFetchRepository($this->columnValue('authorizedFormOfName'));
       }
-      else if ($legacyId && $mapEntry = $this->fetchKeymapEntryBySourceAndTargetName($legacyId,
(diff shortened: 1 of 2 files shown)


lib/QubitFlatfileImport.class.php
```diff
@@ -638,6 +642,22 @@ public function row($row = array())
             )
           );
 
+          // Unable to match to keymap entry.
+          // Try matching via secondary method - match record in DB by
+          // identifier + title + repository.
+          if ($this->columnExists('identifier')
```
MikeFE, line 648: Feel free to do a few conditions on a single line as long as it's <=120 characters.

PR #401, opened 2016-07-05
CSV import: issue importing digital objects

sbreker:
An issue was introduced in 9995/6 where digital object
URLs could not be imported: doing so would produce the
error "could not resolve host" in the console. This
change resolves this issue.

Assignment of the url from the CSV file was not
working correctly. This same issue would have
affected loading from path as well.

### DIFF
--- a/lib/task/import/csvImportTask.class.php
+++ b/lib/task/import/csvImportTask.class.php
@@ -1011,7 +1011,7 @@ EOF;
         // only remove the DO if the checksums differ.  If they are removed
         // because the checksums are different, the following code will
         // recreate them from the CSV file.
-        if ($uri = $self->rowStatusVars['digitalObjectURI']
+        if (($uri = $self->rowStatusVars['digitalObjectURI'])
           && null === $self->object->getDigitalObject())
         {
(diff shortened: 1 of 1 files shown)


MikeFE:
:+1: looks good to me. One thing, before pushing this to the main repository can you rename the commit title message so it includes the bug ticket# + is <=50 characters? It's a convention we use.

So I'd switch the commit title to something like "Fix CSV import digitalObjectURI, refs #10085"

sbreker:
Forgot about that!  Change made.

sevein:
The body should be wrapped to 80 characters (or 72), e.g.

```
Fix CSV import digitalObjectURI, refs #10085

An issue was introduced in 9995/6 where digital object URLs could not be
imported: doing so would produce the error "could not resolve host" in the
console. This change resolves this issue.

Assignment of the url from the CSV file was not working correctly. This same
issue would have affected loading from path as well.
```

More here: http://chris.beams.io/posts/git-commit/

sbreker:
I have updated both the title as indicated by Mike, and the with of the message body.

PR #402, opened 2016-07-07
APCu compatibility fix, refs #10064

sevein:
In APC, `cache_list` items (`apc_cache_info('user')`) always had the cache items
indexed in an array item named `info`. APCu switched that to `key`, introducing
compatibility issues that were not solved until `v4.0.7`. This commit makes info
the default accessor but switches to key in the situation explained previously.


jraddaoui:
Merged in https://github.com/artefactual/atom/commit/582994892b2210d19b0cbc72efc148dc32ae2872

sevein:
Thank you!

PR #403, opened 2016-07-08
CSV/XML import matching option, refs #10054

sbreker:
CSV Import:

Modified --update option to be able to take parameters to modify the update
logic. Adding further parameters for --update can leverage work done for this
task. Parameters will be specified as: --update="match". Parameters are
verified and an exception will be thrown if an unknown parameter is specified.

Added logic to implement 'match' parameter with --update option. When set,
Information Objects must match to a record in the database or they will be
skipped and a message will be logged in the console.

### DIFF
--- a/lib/QubitFlatfileImport.class.php
+++ b/lib/QubitFlatfileImport.class.php
@@ -34,6 +34,7 @@ class QubitFlatfileImport
 
   public $searchIndexingDisabled = true;  // disable per-object search indexing by default
   public $updateExisting         = false; // if object already imported, attempt update
+  public $matchExisting          = false; // when updating, only load rows that we can match
 
   public $status          = array(); // place to store data related to overall import
   public $rowStatusVars   = array(); // place to store data related to current row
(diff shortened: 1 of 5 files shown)


lib/QubitFlatfileImport.class.php
```diff
@@ -658,8 +659,24 @@ public function row($row = array())
...
+
+            $msg = sprintf('Unable to match row. Skipping record: %s (id: %s)',
+                           $this->columnExists('title') ? trim($this->columnValue('title')) : '',
+                           $this->columnExists('identifier') ? trim($this->columnValue('identifier')) : '');
+
+            print $this->logError($msg);
+          }
+          else
+          {
+            // create new object
```
MikeFE: hehe I realize this is a legacy comment, but it's silly. I'd remove it.

lib/task/import/csvImportTask.class.php
```diff
@@ -1074,6 +1074,27 @@ public function execute($arguments = array(), $options = array())
...
     $import->updateExisting = isset($options['update']);
 
+    // Are there params set on --update flag?
+    if ($options['update'])
+    {
+      // Parameters for --update are validated in csvImportBaseTask.class.php.
+      switch ($options['update'])
+      {
+        case 'match':
+          // Save match option. If update is ON, and match is set, only updating
```
MikeFE, line 1084: good comment

lib/task/import/csvImportBaseTask.class.php
```diff
@@ -77,6 +77,21 @@ protected function validateOptions($options)
...
+    // validate params that can be used in conjunction with --update='opt'
+    if ($options['update'])
+    {
+      // array of valid params
+      $validParams = array('match');
+
+      // if supplied param is not in $validParams, throw exception
+      if (!in_array(trim($options['update']), $validParams))
+      {
+        $msg = sprintf("Parameter \"%s\" is not valid for --update option.", $options['update']);
```
MikeFE: I'd use single quotes here instead of double, so you don't need to do the backslash escaping.

lib/QubitXmlImport.class.php
```diff
@@ -414,7 +418,7 @@ private function populateObject(&$domNode, &$importDOM, &$mapping, &$currentObje
 
     $doSave = true;
     // if this is an information object in an XML EAD import, run the enhanced matching check.
-    if ($currentObject instanceof QubitInformationObject && $importSchema == 'ead')
+    if ($currentObject instanceof QubitInformationObject && $importSchema == 'ead' &&  $this->options['match'])
```
MikeFE: If the key 'match' isn't in $this->options I think it will throw a PHP notice or warning (sometimes can't see them without error_reporting(E_ALL) :/). I recommend using `isset($this->options['match']) && $this->options['match']`

MikeFE:
:+1:

PR #405, opened 2016-07-09
Delete keymap with info object delete, refs #10111

sbreker:
Add code to QubitInformationObject to ensure that any associated keymap
records are deleted when the delete() method is called on an information
object.

New private function removeKeymapEntries() added and is called
from the delete() method of the information object.


MikeFE:
:+1:

PR #408, opened 2016-07-11
Dev/issue 10098 custom import bug

mcantelon:
Fixes some issues with custom imports. Also added automatic naming for unnamed CSV columns and ability to ignore specified rows during transformations,


mcantelon:
Going to split up into multiple PRs.

PR #409, opened 2016-07-13
get <language> content

pierremancini:
when reading the langmaterial tag importLangmaterialEadData method was missing language tag content

### DIFF
--- a/lib/model/QubitInformationObject.php
+++ b/lib/model/QubitInformationObject.php
@@ -1757,6 +1757,10 @@ class QubitInformationObject extends BaseInformationObject
       {
         $noteContent .= "\n";
       }
+      else if ($child->nodeName == 'language')
+      {
+        $noteContent .= ' '.trim($child->firstChild->textContent);
+      }
     }
 
     // add language and script note, if so


jraddaoui:
Hi @Pierrocana,

I've filed [#10160](https://projects.artefactual.com/issues/10160) for this fix. But after a deeper review and a conversation with @fiver-watson, we think it's not needed.

AtoM imports those tags using the 'langcode' and 'scriptcode' attributes a little earlier in that method, they are imported as properties instead of as part of the note content, which is better for a latter export.

Thanks!

fiver-watson:
Hi @Pierrocana, 

To follow up a bit on what @jraddaoui was saying: Here's an extract from an EAD XML import for multiple languages and scripts: 

```
<langmaterial encodinganalog="3.4.3">
    <language langcode="eng">English</language>
    <language langcode="fre">French</language>
    <language scriptcode="Latn">Latin</language>
    <language scriptcode="Cyrl">Cyrillic</language>
     Language and script notes (ISAD 3.4.3)
  </langmaterial>
```

in AtoM, the language of materials field is an autocomplete, already pre-loaded from Symfony with a list of languages. During import, AtoM ignores the label, and looks at the `@langcode` and `@scriptcode` values to find matches with the existing values. It looks like this in the user interface: 

View page after import: http://i.imgur.com/fvKLC6X.png
Edit page after import: http://i.imgur.com/wJZYXEG.png

As you can see, in the display, the human-readable language label (e.g. English, French) will be shown, but during import you'll need the 2-letter ISO code as part of the `@langcode` attribute. We think it might just cause unnecessary duplication of data to also add the label to the language notes field. 

For this reason, at this time we are not going to accept this pull request. I hope you understand, and that you continue to share other PRs with the AtoM community in the future! 

Thank and apologies for the delay in reviewing this.

PR #410, opened 2016-07-13
Fix adv. search repository options, refs #10117

jraddaoui:
Remove cached options when a repository is saved or deleted


sbreker:
Looks good Radda.

PR #412, opened 2016-07-14
Allow access to text DO master, refs #10125

jraddaoui:
As we intended in QubitInformationObject getDigitalObjectLink


sevein:
👍

PR #414, opened 2016-07-15
Auto-rename blank CSV columns. refs #10115

mcantelon:
When transforming/importing CSV, give blank header columns names so column
values, etc., can more easliy be referenced.

### DIFF
--- a/lib/QubitFlatfileImport.class.php
+++ b/lib/QubitFlatfileImport.class.php
@@ -466,6 +466,38 @@ class QubitFlatfileImport
    *  ----------------------
    */
 
+  /**
+   * Assign names to unnamed columns
+   *
+   * @return void
+   */
+  protected function handleUnnamedColumns()
+  {
+    // Assign names to unnamed columns
+    $baseLabel = 'Untitled';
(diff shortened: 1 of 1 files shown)


lib/QubitFlatfileImport.class.php
```diff
@@ -468,6 +468,25 @@ private function handleByteOrderMark($fh)
...
   {
+    // Assign names to blank columns
+    $baseLabel = 'Untitled';
+    $labelNumber = 1;
+    foreach ($this->columnNames as $index => $name)
+    {
+      if (empty($name))
+      {
+        // Increment label number if column already exists
+        while(in_array($baseLabel . $labelNumber, $this->columnNames))
```
MikeFE, line 484: not sure if it's worth breaking out into a helper function since it's only several lines, but generating a new column name seems outside the scope of handleColumnRenaming?

mcantelon, line 484: Hrm, yeah true... I'll do that.

mcantelon:
Merged.

PR #415, opened 2016-07-15
Add ability to ignore rows (by row index) in CSV transformations, refs #10122

mcantelon:
Redmine: https://projects.artefactual.com/issues/10122

Add ability to ignore CSV rows either by their index relative to the header row or by a logic that will check to see if the row should be ignored.


sevein:
👍

mcantelon:
Merged.

PR #416, opened 2016-07-15
Improve ES search for archival descriptions, #10082

MikeFE:
The following PR attempts to accomplish 3 things:
1. Narrow down how many fields in nested, foreign types inside information object we search over. Now we only search for things like repository name (but not history), creator name (but not history), etc. This will ensure we don't accidentally get weird search results just because a keyword was in a creator's biog history or things like that.
2. Add weighting to search fields when searching, so certain fields yield higher scores when returning from queries. Also added a 'sort by relevance' option so we can take advantage of this way of searching.
3. Change the default search operator from OR to AND. This will result in more concise results.

### DIFF
--- a/plugins/arElasticSearchPlugin/lib/arElasticSearchPluginUtil.class.php
+++ b/plugins/arElasticSearchPlugin/lib/arElasticSearchPluginUtil.class.php
@@ -460,48 +460,6 @@ class arElasticSearchPluginUtil
   }
 
   /**
-   * Add boost values to various fields.
-   *
-   * @param array &$fields  An array of the fields to be modified with their boost values added.
-   * @param array $i18nfields  Specifies which i18n fields to boost, and which boost value to use.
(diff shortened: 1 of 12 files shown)


plugins/arElasticSearchPlugin/lib/arElasticSearchPluginUtil.class.php
```diff
@@ -439,6 +459,39 @@ protected static function getAllObjectStringFields($rootIndexType, $object, $pre
     return $fields;
   }
 
+  public static function addBoostValuesToFields(&$fields, $i18nFields, $nonI18nFields)
```
sevein: Shouldn't this method belong to `arElasticSearchModelBase`?

sevein:
Looking great! Just one small observation about the location of the `addBoostValuesToFields` method, but other than that 👍 !

MikeFE:
squashed&merged

PR #418, opened 2016-07-18
Postpone apiKey property save(), refs #10139

sevein:
We can't save a QubitProperty if its foreign key has not been populated yet.
The solution is to associate the object to the "propertys" magic member and
QubitObject will take care of it and postpone the call to save() when it's
ready.


MikeFE:
:+1:

PR #419, opened 2016-07-22
Add clipboard icons on Desc Update, refs #10007

sbreker:
Added a new column to the Description Updates page for archival descriptions,
authority records, and repository records. This new column will include the
clipboard clip icon, so users can add descriptions recently added or updated
directly to the clipboard.

Added ability to bulk add all or remove all items to/from the clipboard.
'All / None' links are located in the column header for the clipboard button
column.

### DIFF
--- a/js/clipboardToggleAll.js
+++ b/js/clipboardToggleAll.js
@@ -1,0 +1,41 @@
+(function ($)
+  {
+    // Toggle buttons
+    var clickClipButton = function (node, selectAll) {
+      var $button = $(node);
+
+      // We only want to click the button when needed
+      if (selectAll === $button.hasClass('added'))
+      {
+        return;
+      }
+
+      $button.click();
+    };
+
+    // Convenience wrapper in jQuery
+    $.fn.clickClipButton = function(clicked) {
+      this.each(function () {
+        clickClipButton.apply(null, [this, clicked]);
+      });
+    };
+
+    // get all affected clipboard buttons and select all or none based on
+    // item clicked.
+    $(document).ready(function() {
+
+      var $area = $('#clipboardButtonNode');
+      var $buttons = $area.find('button');
+
+      $area.find('.all').on('click', function (event) {
+        event.preventDefault();
+        $buttons.clickClipButton(true);
+      });
+
+      $area.find('.none').on('click', function (event) {
+        event.preventDefault();
+        $buttons.clickClipButton(false);
+      });
+    });
+  }
+)(jQuery);
--- a/apps/qubit/modules/search/actions/descriptionUpdatesAction.class.php
+++ b/apps/qubit/modules/search/actions/descriptionUpdatesAction.class.php
@@ -114,6 +114,8 @@ class SearchDescriptionUpdatesAction extends sfAction
       $this->addField($name);
     }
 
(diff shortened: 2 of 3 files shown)


js/clipboardToggleAll.js
```diff
@@ -0,0 +1,39 @@
...
+        clickClipButton.apply(null, [this, clicked]);
+      });
+    };
+
+    // get all affected clipboard buttons and select all or none based on
+    // item clicked.
+    $(document).ready(function() {
+
+      var $area = $('#clipboardButtonNode');
+
```
sevein, line 29: You could add `var $buttons = $area.find('button')` and use later in your callbacks so the browser doesn't have to find them and save the computation each time the user clicks on All / None.

js/clipboardToggleAll.js
```diff
@@ -0,0 +1,39 @@
...
+      $area.find('.all').on('click', function (event) {
+        $area.find('button').clickClipButton(true);
```
sevein: Both callbacks should probably call `event.preventDefault()` to cancel the browser's default event when you click on the `a` element (`href=#`).

js/clipboardToggleAll.js
```diff
@@ -0,0 +1,39 @@
+(function ($)
+  {
+    // Toggle buttons
+    var clickClipButton = function (node, selectAll) {
+      $button = $(node);
```
sevein: TIP: always use `var` to make sure that you are declaring a new variable within the scope of that function and not assigning a new value to an existing variable defined outside of that scope ([more here](https://github.com/getify/You-Dont-Know-JS/tree/master/scope%20%26%20closures)). It will save you some headaches if you get into that habit!

js/clipboardToggleAll.js
```diff
@@ -0,0 +1,39 @@
...
+
+      // if 'none' was chosen, click any button that has 'added' on
+      if (selectAll == false && $button.hasClass('added')) {
+        $button.click();
+      }
+      // if 'all' was chosen, don't click the button if 'added' is on
+      else if (selectAll == true && !$button.hasClass('added')) {
+        $button.click();
+      }
+    };
```
sevein, line 14: A bit of boolean minimization, what do you think?

``` js
var clickClipButton = function (node, selectAll) {
   var $button = $(node);

   // We only want to click the button when needed
   if (selectAll === $button.hasClass('added'))
   {
     return;
   }

   $button.click();
};
```

js/clipboardToggleAll.js
sbreker, line 29: Good idea!   I have added code to do find and assign the buttons inside $(document).ready. just after $area is assigned.

js/clipboardToggleAll.js
sbreker: Also a good idea!  Change made.

js/clipboardToggleAll.js
sbreker: Noted.  I have modified my code accordingly.

js/clipboardToggleAll.js
sbreker, line 14: Looks good.  Added and re-tested.

sevein:
:+1:

PR #422, opened 2016-07-26
Full DB config for CSV transform. refs #10161

mcantelon:
Removed redundant error check and added support for custom DB host, user,
and name (for storing temp data) in CSV transform class.

### DIFF
--- a/lib/QubitCsvTransform.class.php
+++ b/lib/QubitCsvTransform.class.php
@@ -85,9 +85,12 @@ class QubitCsvTransform extends QubitFlatfileImport
       throw new sfException('You must specifiy the output-file option.');
     }
 
-    if (getEnv('MYSQL_PASSWORD') === false)
+    foreach(array('MYSQL_HOST', 'MYSQL_USER', 'MYSQL_PASSWORD', 'MYSQL_DB') as $var)
     {
-      throw new sfException('You must set the MYSQL_PASSWORD environmental variable. This script will use the "root" user and a database called "import".');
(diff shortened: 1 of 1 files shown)


MikeFE:
Is there a reason we're using all the funky mysqli_connect stuff in QubitCsvTransform other than it's super old / was never updated? We should use QubitPdo if possible to keep it consistent across the app (although I still need to get rid of QubitFlatfileImport::sqlQuery)

mcantelon:
I'll check out QubitPdo... might be an easy fix.

MikeFE:
I'd be careful about doing much work on that stuff anyway, I don't think anyone has used it for years and I'm not sure how useful it is moving forward (maybe it is, I've never used it)

mcantelon:
It's useful when it works (pretty fast to rig up a transform for a migration), at least to me anyways. Can QubitPdo work with more than one DB? Part of the reason I was using a raw connection was to be able to use a different DB for throwaway data.

MikeFE:
Ah okay. If you find it useful then I guess it's worth keeping around. I don't think QubitPdo lets you specify the database manually. You'd have to add a new method to it so you can manually set its $conn protected member variable. I'm not sure if that's worth it vs. how you're doing it now or not.

mcantelon:
Yeah, not using QubitPdo's a bit hack-y, but yeah, it might not be worth modifying QubitPdo. I'll add a comment that this is something that could be done in the future.

mcantelon:
Merged.

PR #424, opened 2016-07-29
Specify csv transform script WebUI, refs #10138

sbreker:
The CSV Import page in AtoM now allows a user to choose if they would like
to run a transformation script against the csv file being uploaded.

The transformation script to be run must be configured in config/app.yml by an
administrator prior to the user being able to see the checkbox on the CSV
Import page. If a transformation script is configured, the full path and
name of the transform script will also display opposite the checkbox on the
CSV Import page.

If the transformation script returns an exit code greater than zero, this
indicates a failure in the transform script and an exception will be thrown
and the csv import will not proceed.

The transform script must take 2 parameters:
1) The file to be transformed
2) The output file name that will contain the tranformed csv

The transform script must return a non-zero exit code to indicate an error
condition. An exit code of zero indicates success and will cause the csv
import to proceed. Any error messages must be output by the script to stdout
or stderr.

### DIFF
--- a/lib/QubitCsvImport.class.php
+++ b/lib/QubitCsvImport.class.php
@@ -33,9 +33,16 @@ class QubitCsvImport
     $parent = null;
 
   public $indexDuringImport = false;
+  public $doCsvTransform = false;
 
   public function import($csvFile, $type = null)
   {
+    // perform the transformation, if requested and correctly configured
+    if ($this->doCsvTransform)
+    {
+      $transformedFile = $this->doTransform($csvFile);
+    }
+
     // Find the proper task
     switch ($type)
     {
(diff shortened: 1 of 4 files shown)


lib/QubitCsvImport.class.php
```diff
@@ -110,6 +117,63 @@ public function import($csvFile, $type = null)
...
+      throw new sfException('Transform failed. Unable to locate file: ' . $csvFile);
+    }
+
+    // build output filename and path. Take source dir and name and create a
+    // parallel temp file based on that.
+    $csvFilePath = pathinfo($csvFile);
+    $outputFileName = tempnam($csvFilePath['dirname'], $csvFilePath['filename'] . '-');
+    $logFileName = $outputFileName . '.log';
+
+    // ./transform_csv.py <source file> <destination file>
```
MikeFE: maybe make this comment more generic?

MikeFE:
:+1:

lib/QubitCsvImport.class.php
sbreker: Done.

sevein:
@sbreker, are you planning to add i18n as we discussed yesterday?

sbreker:
The mechanism of those Exception messages need to be changed when the import becomes an asynchronous job. I was thinking of altering them as part of that task.

PR #425, opened 2016-08-04
WebUI CSV XML import filename keymap, refs #10175

sbreker:
CSV and XML files uploaded via the WebUI generate keymap table records when
uploaded. Keymap records are used for matching import records to records in
the database.

Currently uploaded files are assigned a randomly generated temporary file
name and these were being assigned to the keymap record on import. This has
been changed such that the original file name is now being assigned to the
keymap record. This will ensure that matching logic based on the keymap
table will work correctly when loading update files.

### DIFF
--- a/lib/QubitCsvImport.class.php
+++ b/lib/QubitCsvImport.class.php
@@ -35,8 +35,20 @@ class QubitCsvImport
   public $indexDuringImport = false;
   public $doCsvTransform = false;
 
-  public function import($csvFile, $type = null)
+  public function import($csvFile, $type = null, $csvOrigFileName = null)
   {
+    if (null === $csvOrigFileName)
+    {
+      // WebUI passes a temp file name in $csvFile. e.g. /tmp/phpLjBIBv
+      // If $csvOrigFileName is null, save $csvFile in keymap record
+      $csvOrigFileName = basename($csvFile);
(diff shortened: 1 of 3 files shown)


lib/QubitCsvImport.class.php
```diff
@@ -35,8 +35,20 @@ class QubitCsvImport
   public $indexDuringImport = false;
   public $doCsvTransform = false;
 
-  public function import($csvFile, $type = null)
+  public function import($csvFile, $type = null, $csvOrigFileName = null)
   {
+    if (null === $csvOrigFileName)
```
MikeFE, line 40: Maybe you can use PHP's ternary operator + shorthand syntax `$csvOrigFileName = basename($csvOrigFileName ?: $csvFile);` to condense this. But maybe we want it as-is so the comments are clearer, I'll leave it up to you.

MikeFE:
:+1:

lib/QubitCsvImport.class.php
sbreker, line 40: Yes, I considered doing what you're describing, but opted for readability as there are a bunch of comments there to explain what's going on.

MikeFE, line 40: sounds reasonable

PR #426, opened 2016-08-10
Background bulk file import, refs #10137

sbreker:
Created new arFileImportJob class to create the CSV and XML import objects
and ingest the files on the server in the background.

arFileImportJob will catch any exceptions thrown during the CSV or XML import
process and display them on the job details window.

Simplified post import importSuccess page as much on there was no longer
relevant now that import occurs in the background.

If more than one import file is specified on the WebUI import page,
modules/object/actions/importAction.class.php will launch a new
arFileImportJob for each file to be imported.

Removed logic pretaining to zip files from importAction.class.php.


apps/qubit/modules/object/actions/importAction.class.php
```diff
@@ -51,103 +51,30 @@ public function execute($request)
       $this->redirect($importSelectRoute);
     }
 
-    // set indexing preference
-    if (isset($request->noindex))
+    $options = array('noIndex' => ($request->getParameter('noindex') == 'on') ? false : true,
+                     'doCsvTransform' => ($request->getParameter('doCsvTransform') == 'on') ? true : false,
```
MikeFE: can you just remove the part after ? since it'll evaluate as bool anyway? or is it a bit weird since you're doing the expression in the array definition like that? Same thing with above except changing == to !=

lib/job/arFileImportJob.class.php
```diff
@@ -0,0 +1,90 @@
...
+    try
+    {
+      switch ($parameters['importType'])
+      {
+        case 'csv':
+          $importer = new QubitCsvImport;
+          if ($parameters['doCsvTransform']) $this->info($this->i18n->__('Applying transformation to CSV file.'));
+          $importer->doCsvTransform = $parameters['doCsvTransform'];
+
+          if ($parameters['noIndex']) $this->info($this->i18n->__('Indexing imported records.'));
```
MikeFE: We as a rule always use curly braces even when it's not strictly necessary with the 1-liners like this.

apps/qubit/modules/object/actions/importAction.class.php
sbreker: I wrote it this way for clarity, as the arFileImportJob depends on it being true/false, as opposed to 'on' or ''.  I don't think there is a technical reason preventing me from shortening it.  Thoughts?

lib/job/arFileImportJob.class.php
sbreker: 👍

apps/qubit/modules/object/actions/importAction.class.php
MikeFE: It's a very minor point, if you feel this way is clearer, that's cool :+1:

apps/qubit/modules/object/actions/importSelectAction.class.php
```diff
@@ -31,29 +37,162 @@ public function execute($request)
...
+        // Get list of repositories
+        $criteria = new Criteria;
+        // Do source culture fallback
+        $criteria = QubitCultureFallback::addFallbackCriteria($criteria, 'QubitActor');
+        // Ignore root repository
+        $criteria->add(QubitActor::ID, QubitRepository::ROOT_ID, Criteria::NOT_EQUAL);
+        $criteria->addAscendingOrderByColumn('authorized_form_of_name');
+        $cache = QubitCache::getInstance();
+        $cacheKey = 'file-import:list-of-repositories:'.$this->context->user->getCulture();
+        //$cacheKey = 'advanced-search:list-of-repositories:'.$this->context->user->getCulture();
```
MikeFE: this is no longer needed, right?

sbreker: The commented out line?  :)   Nope - I will remove that.

apps/qubit/modules/object/actions/importSelectAction.class.php
```diff
@@ -31,29 +37,162 @@ public function execute($request)
...
+    }
+  }
+
+  public function execute($request)
+  {
+    parent::execute($request);
+
+    if ($request->isMethod('post'))
+    {
+      //$this->getParameters = $request->getGetParameters();
```
MikeFE: not needed?

apps/qubit/modules/object/actions/importSelectAction.class.php
```diff
@@ -31,29 +37,162 @@ public function execute($request)
...
+      if ($this->form->isValid())
+      {
+        $this->processForm();
+
+        $this->doBackgroundImport($request);
+
+        $this->setTemplate('importResults');
+      }
+    }
+
```
MikeFE: we usually don't use 2 newlines before starting elses

lib/QubitCsvImport.class.php
```diff
@@ -80,42 +90,74 @@ public function import($csvFile, $type = null, $csvOrigFileName = null)
...
+      {
+        case 'import-as-new':
+          $commandSkipMatched = ($this->skipMatched) ? '--skip-matched' : '';
+          break;
+
+        case 'match-and-update':
+        case 'delete-and-replace':
+          $commandUpdate = ($this->updateType == 'match-and-update') ? '--update="match-and-update"' : '--update="delete-and-replace"';
+          $commandSkipUnmatched = ($this->skipUnmatched) ? '--skip-unmatched' : '';
+          $commandLimit = ('' !== $this->limit) ? "--limit=\"$this->limit\"" : '';
```
MikeFE, line 108: `'--limit="$this->limit"'` ?

lib/job/arFileImportJob.class.php
```diff
@@ -0,0 +1,190 @@
...
+
+    // Mark job as complete.
+    $this->info($this->i18n->__('Import complete.'));
+    $this->job->setStatusCompleted();
+    $this->job->save();
+
+    return true;
+  }
+
+  private function setCsvImportParams(&$importer, $parameters)
```
MikeFE, line 104: it's usually good for non-trivial functions to have a comment block at the top summarizing parameters/return values and the function's purpose. You can see examples elsewhere in the codebase on how we do it if you're not certain

lib/job/arFileImportJob.class.php
```diff
@@ -0,0 +1,190 @@
...
+  {
+    foreach ($parameters as $key => $value)
+    {
+      if (empty($value)) { continue; }
```
MikeFE: curly braces should always be on their own line

lib/job/arFileImportJob.class.php
```diff
@@ -0,0 +1,190 @@
...
+
+  private function setXmlImportParams(&$importer, &$parameters)
+  {
+    $options = array();
+
+    $options['strictXmlParsing'] = false;
+
+    foreach ($parameters as $key => $value)
+    {
+      if (empty($value)) { continue; }
```
MikeFE: same as above

lib/QubitCsvImport.class.php
sbreker: Changing to single quotes there will print the literal: **--limit="$this->limit"**
as opposed to evaluating $this->limit: **--limit="museum-of-agriculture"**

MikeFE: fair enough

PR #427, opened 2016-08-19
Add various csv:import --update options

MikeFE:
See https://projects.artefactual.com/issues/10150 for details

### DIFF
--- a/lib/QubitFlatfileImport.class.php
+++ b/lib/QubitFlatfileImport.class.php
@@ -606,6 +606,66 @@ class QubitFlatfileImport
   }
 
   /**
+   * Set default culture to en if not present; ensure current culture is set to the current row's culture.
+   */
+  private function handleCulture()
+  {
+    // Add blank culture field if not present in import
+    if (!in_array('culture', $this->columnNames))
+    {
+      $this->columnNames[] = 'culture';
+      $this->addColumns[]  = 'culture';
+    }
+
+    // Default culture to English
+    if (0 == strlen($this->columnValue('culture')))
+    {
+      $this->columnValue('culture', 'en');
+    }
+
+    // Set current culture to culture specified in CSV row.
+    if (isset($this->context) && 'sfContext' == get_class($this->context))
+    {
+      $this->context->getUser()->setCulture($this->columnValue('culture'));
+    }
+  }
+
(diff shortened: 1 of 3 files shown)


lib/QubitFlatfileImport.class.php
```diff
@@ -779,17 +712,175 @@ public function row($row = array())
...
+    if ($this->object->sourceCulture == $this->columnValue('culture'))
+    {
+      $msg = sprintf('Matching description found, %s; row (id: %s, culture: %s, legacyId: %s)...',
+                      $this->getActionDescription(), $this->object->id, $this->object->sourceCulture, $legacyId);
+
+      if ($this->isUpdating())
+      {
+        $this->status['updated']++;
+
+        // execute ad-hoc row pre-update logic (remove related data, etc.)
```
mcantelon: Comment capitalization

mcantelon:
Looks good overall!

lib/QubitFlatfileImport.class.php
```diff
@@ -37,6 +37,7 @@ class QubitFlatfileImport
   public $deleteAndReplace       = false; // Delete matching records & replace them
   public $skipMatched            = false; // Skip creating new record if matching one is found
   public $skipUnmatched          = false; // Skip creating new record if matching one is not found
+  public $limitToId              = 0;     // Id of repository or TLD to limit our update matching under
```
mcantelon: Maybe filterByRepositoryId would be clearer. Limit makes me think of skip/limit in queries and I'd assume that "Id" is referencing the ID of a matched object rather than its repository.

lib/QubitFlatfileImport.class.php
```diff
@@ -838,6 +841,41 @@ private function createNewInformationObject()
...
 
   /**
+   * The user can specify a --limit option on import that makes it so --update matches only occur
+   * if the matching description is under a specified repository or top level description.
+   *
+   * This function will check to ensure if the current matching information object is within the limit,
+   * and if not, set the object back to null since it isn't a match we want.
+   *
+   */
+  private function checkInformationObjectMatchLimit()
```
mcantelon: Maybe "Filter" rather than "Limit"?

PR #428, opened 2016-08-25
Sort by relevance by default for searching #10082

MikeFE:
If the user is searching for archival descriptions, set the default sorting to
by relevance.


sevein:
👍

PR #430, opened 2016-09-08
Prevent trying to save deleted digi objs #10272

MikeFE:
When we call delete on a digital object here: https://github.com/artefactual/atom/blob/qa/2.4.x/lib/task/import/csvImportTask.class.php#L486

Sometimes when we do a subsequent call to the associated information object's save(), it hasn't removed that digital object instance from the information object's ->digitalObjects magic property, so it throws the error "this object has already been deleted". This PR creates a helper function to manually clear that property when needed.

I tried tracking down exactly why this was happening (and only sometimes--sometime it'd work fine) to no avail. If anyone has any ideas for a more elegant solution or insights into what the problem might have been, I'm open to suggestions. :)

### DIFF
--- a/lib/model/QubitDigitalObject.php
+++ b/lib/model/QubitDigitalObject.php
@@ -1264,10 +1264,23 @@ class QubitDigitalObject extends BaseDigitalObject
     if (!empty($this->informationObjectId))
     {
       QubitSearch::getInstance()->update($this->getInformationObject());
+      $this->deleteFromAssociatedInformationObject();
     }
 
     // Delete self
     parent::delete($connection);
+  }
+
+  /**
+   * If we have an associated information object, ensure this digital object is cleared
+   * from its digitalObjects property.
+   */
+  private function deleteFromAssociatedInformationObject()
+  {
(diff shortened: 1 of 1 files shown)


jraddaoui:
Hi Mike,

Could we check somehow if the DO has been deleted before it's saved here:
https://github.com/artefactual/atom/blob/qa/2.4.x/lib/model/QubitInformationObject.php#L242

If not, I think we should move that clearing to the DO delete method, without a helper function. That way we don't need to call it again if it happens in other parts and I can't see any problem having it in a normal workflow (without this issue).

Regards.

lib/model/QubitDigitalObject.php
```diff
@@ -1264,13 +1264,26 @@ public function delete($connection = null)
     if (!empty($this->informationObjectId))
     {
       QubitSearch::getInstance()->update($this->getInformationObject());
+      $this->deleteFromAssociatedInformationObject();
```
jraddaoui, line 1267: I'd do it before updating the search index, just in case it tries to get some DO data in there.

MikeFE: fair point, I'll change it before committing.

MikeFE:
merged

PR #431, opened 2016-09-08
Fix various --update issues, refs #10150

MikeFE:
- Prevent duplicate events from being imported
- Handle case where parentId isn't found in keymap
- Fix FK restraint when match-and-replacing a hierarchy

### DIFF
--- a/lib/QubitFlatfileImport.class.php
+++ b/lib/QubitFlatfileImport.class.php
@@ -855,6 +855,9 @@ class QubitFlatfileImport
   {
     $oldSlug = $this->object->slug;
 
+    // Prevent FK restraint errors; we'll rebuild the hierarchy from the csv.
+    QubitPdo::prepareAndExecute('UPDATE information_object SET parent_id=null WHERE parent_id=?',
+                                array($this->object->id));
     $this->object->delete();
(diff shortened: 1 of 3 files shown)


lib/QubitFlatfileImport.class.php
```diff
@@ -645,6 +645,45 @@ private function handleVirtualCols()
...
   }
 
+  /**
+   * Check array of event data from import, check if this exact event already exists.
+   *
+   * @return bool  True if exists, false if not
+   */
+  public function hasDuplicateEvent($event)
+  {
+    if (!isset($this->object) || !$this->object->id)
```
mcantelon, line 655: Could use empy() for check

MikeFE: hah good call, I always seem to forget to use it.

mcantelon:
Other than one nitpick, looks good to me!

lib/QubitFlatfileImport.class.php
MikeFE: hmm actually I need to check for null on that id property, but I think I can just use isset($this->object->id) instead of the 2 conditions

PR #432, opened 2016-09-08
Add EAD --update abilities, refs #10150

MikeFE:
Add ability to 'delete-and-replace' existing, matching records when doing EAD imports.


lib/QubitXmlImport.class.php
```diff
@@ -39,6 +39,10 @@ class QubitXmlImport
 
   public function import($xmlFile, $options = array(), $xmlOrigFileName = null)
   {
+    // save options so we can access from processMethods
```
mcantelon, line 42: Nitpick: comment capitalization.

mcantelon:
Other than one minor nitpick (that was there before your PR), looks good!

PR #437, opened 2016-09-16
Added new alert types, refs #10321

mcantelon:
Added ability to create "info" and "success" Bootstrap alerts.

This isn't needed for a feature, but I thought I needed it so I added it.

### DIFF
--- a/apps/qubit/templates/_alerts.php
+++ b/apps/qubit/templates/_alerts.php
@@ -1,13 +1,8 @@
-<?php if ($sf_user->hasFlash('notice')): ?>
-  <div id="notice-alerts" class="alert">
-    <button type="button" class="close" data-dismiss="alert">&times;</button>
-    <?php echo $sf_user->getFlash('notice', ESC_RAW) ?>
-  </div>
-<?php endif; ?>
-
-<?php if ($sf_user->hasFlash('error')): ?>
-  <div id="error-alerts" class="alert alert-error">
-    <button type="button" class="close" data-dismiss="alert">&times;</button>
-    <?php echo $sf_user->getFlash('error', ESC_RAW) ?>
-  </div>
-<?php endif; ?>
+<?php foreach(array('notice', 'info', 'error', 'success') as $alertType): ?>
+  <?php if ($sf_user->hasFlash($alertType)): ?>
+    <div class="alert app-alert alert-<?php echo $alertType ?>">
+      <button type="button" class="close" data-dismiss="alert">&times;</button>
+      <?php echo $sf_user->getFlash($alertType, ESC_RAW) ?>
+    </div>
+  <?php endif; ?>
+<?php endforeach; ?>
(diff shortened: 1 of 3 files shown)


apps/qubit/templates/_alerts.php
```diff
@@ -1,13 +1,8 @@
...
-  </div>
-<?php endif; ?>
-
-<?php if ($sf_user->hasFlash('error')): ?>
-  <div id="error-alerts" class="alert alert-error">
-    <button type="button" class="close" data-dismiss="alert">&times;</button>
-    <?php echo $sf_user->getFlash('error', ESC_RAW) ?>
-  </div>
-<?php endif; ?>
+<?php foreach(array('notice', 'info', 'error', 'success') as $alertType): ?>
```
sevein, line 1: Nitpick... foreach is a statement, should we write as `foreach (...)` instead of `foreach(...)`?

apps/qubit/templates/_alerts.php
```diff
@@ -1,13 +1,8 @@
...
+  <?php if ($sf_user->hasFlash($alertType)): ?>
+    <div id="notice-alerts" class="alert alert-<?php echo $alertType ?>">
+      <button type="button" class="close" data-dismiss="alert">&times;</button>
+      <?php echo $sf_user->getFlash($alertType, ESC_RAW) ?>
+    </div>
+  <?php endif; ?>
+<?php endforeach; ?>
```
sevein, line 8: Can you make sure that there is some space between the alert box and the header bar? IIRC there was no space at all between them making it visually not very pleasant.

sevein (commented):
Good idea, just a couple of suggestions!

apps/qubit/templates/_alerts.php
mcantelon: Ah right... thanks!

apps/qubit/templates/_alerts.php
mcantelon: Good idea... I'll tweak the CSS.

mcantelon:
I've fixed those two issues. Thanks for ze review and let me know if there are any others.

plugins/arDominionPlugin/css/less/bootstrap-mod.less
```diff
@@ -27,3 +27,8 @@
 .open > .dropdown-menu {
   z-index: 10004
 }
+
+// Upper padding for alerts
+.alert {
+  margin-top: 1em
```
sevein, line 33: One issue with this approach is that it's affecting all the cases where `.alert` is used, see `git grep -E "class=\".*alert.*\""`. Should you only target `#notice-alerts` instead? Additionally, not entirely sure but is it worth trying to find a better `id` for this element? e.g. `#app-alert` or `#global-alert` or similar?

mcantelon: Good call... I'll make it more specific (and another change I should make is probably changing "#notice-alerts" to "#<notification type>-alert" so the ID only gets used once).

mcantelon: (I ended up just getting rid of the ID and using app-alert as that's probably sufficient granularity for our needs.)

mcantelon:
OK, ready for another review. Thanks!

sevein:
👍

mcantelon:
Merged.

PR #438, opened 2016-09-20
Geolocation improvements, refs #10330

mcantelon:
Redmine: https://projects.artefactual.com/issues/10330

Added ability to specify latitude/longitude of digital object and show resulting map on description pages:
- Add ability for admins to toggle display of digital object map on default page elements page
- Add ability to specify latitude/longitude when editing digital object metadata

AtoM tweaks/fixes related to feature:
- Change Google Maps API setting from config file var to DB setting
- Fix issue with use of get_partial instead of get_component when displaying digital object metadata
- Darken links within info alerts

### DIFF
--- a/apps/qubit/modules/digitalobject/actions/metadataComponent.class.php
+++ b/apps/qubit/modules/digitalobject/actions/metadataComponent.class.php
@@ -40,5 +40,16 @@ class DigitalObjectMetadataComponent extends sfComponent
     {
       $this->denyFileNameByPremis = true;
     }
+
+    // Provide Google Maps API key to template
+    $this->googleMapsApiKey = sfConfig::get('app_google_maps_api_key');
+
+    // Provide latitude to template
+    $latitudeProperty = $this->infoObj->digitalObjects[0]->getPropertyByName('latitude');
+    $this->latitude = $latitudeProperty->value;
+
+    // Provide longitude to template
+    $longitudeProperty = $this->infoObj->digitalObjects[0]->getPropertyByName('longitude');
+    $this->longitude = $longitudeProperty->value;
   }
 }
--- a/apps/qubit/modules/settings/actions/globalAction.class.php
+++ b/apps/qubit/modules/settings/actions/globalAction.class.php
@@ -91,6 +91,7 @@ class SettingsGlobalAction extends sfAction
     $showTooltips = QubitSetting::getByName('show_tooltips');
     $defaultPubStatus = QubitSetting::getByName('defaultPubStatus');
     $swordDepositDir = QubitSetting::getByName('sword_deposit_dir');
+    $googleMapsApiKey = QubitSetting::getByName('google_maps_api_key');
     $slugTypeInformationObject = QubitSetting::getByName('slug_basis_informationobject');
 
 
@@ -117,7 +118,8 @@ class SettingsGlobalAction extends sfAction
       'slug_basis_informationobject' => (isset($slugTypeInformationObject)) ? intval($slugTypeInformationObject->getValue(array('sourceCulture'=>true))) : QubitSlug::SLUG_BASIS_TITLE,
       'show_tooltips' => (isset($showTooltips)) ? intval($showTooltips->getValue(array('sourceCulture'=>true))) : 1,
       'defaultPubStatus' => (isset($defaultPubStatus)) ? $defaultPubStatus->getValue(array('sourceCulture'=>true)) : QubitTerm::PUBLICATION_STATUS_DRAFT_ID,
-      'sword_deposit_dir' => (isset($swordDepositDir)) ? $swordDepositDir->getValue(array('sourceCulture'=>true)) : null
+      'sword_deposit_dir' => (isset($swordDepositDir)) ? $swordDepositDir->getValue(array('sourceCulture'=>true)) : null,
+      'google_maps_api_key' => (isset($googleMapsApiKey)) ? $googleMapsApiKey->getValue(array('sourceCulture'=>true)) : null
     ));
   }
 
@@ -361,6 +363,20 @@ class SettingsGlobalAction extends sfAction
 
       // Force sourceCulture update to prevent discrepency in settings between cultures
       $setting->setValue($swordDepositDir, array('sourceCulture' => true));
+      $setting->save();
+    }
+
+    // Google Maps Javascript API key
+    if (null !== $googleMapsApiKey = $thisForm->getValue('google_maps_api_key'))
+    {
+      if (null === $setting = QubitSetting::getByName('google_maps_api_key'))
+      {
+        $setting = new QubitSetting;
+        $setting->name = 'google_maps_api_key';
+      }
+
+      // Force sourceCulture update to prevent discrepency in settings between cultures
+      $setting->setValue($googleMapsApiKey, array('sourceCulture' => true));
       $setting->save();
     }
 
--- a/apps/qubit/modules/settings/templates/pageElementsSuccess.php
+++ b/apps/qubit/modules/settings/templates/pageElementsSuccess.php
@@ -50,6 +50,18 @@
             <td><?php echo $form->toggleIoSlider ?></td>
           </tr>
           <tr>
+            <td>
+              <?php echo $form->toggleDigitalObjectMap->label('Digital object map')->renderLabel() ?>
+              <?php if (!$googleMapsApiKeySet): ?>
+                <div class="description">
+                  <?php $securitySettingsLinkHtml = link_to(__('Security'), array('module' => 'settings', 'action' => 'security')) ?>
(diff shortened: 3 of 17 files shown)


sevein:
Ready for review or still work in progress, @mcantelon?

mcantelon:
I talked with Sara and realized that I'd missed one feature... will get PR review ready today.

mcantelon:
Alright, this one's ready to roll.

apps/qubit/modules/settings/templates/pageElementsSuccess.php
```diff
@@ -50,6 +50,18 @@
             <td><?php echo $form->toggleIoSlider ?></td>
           </tr>
           <tr>
+            <td>
+              <?php echo $form->toggleDigitalObjectMap->label('Digital object map')->renderLabel() ?>
+              <?php if (!$googleMapsApiKeySet): ?>
+                <div id="notice-alerts" class="alert alert-info">
```
sevein: Should we use our standard `<div class="description">...</div>` instead? See: https://imgur.com/a/5hH4r.

apps/qubit/modules/settings/actions/securityAction.class.php
```diff
@@ -62,11 +62,13 @@ protected function populateSecurityForm()
     $limitAdminIp = QubitSetting::getByName('limit_admin_ip');
     $requireSslAdmin = QubitSetting::getByName('require_ssl_admin');
     $requireStrongPasswords = QubitSetting::getByName('require_strong_passwords');
+    $googleMapsApiKey = QubitSetting::getByName('google_maps_api_key');
```
sevein: I'm not sure if the security page is the best place for the Google Maps JavaScript API key. It's not even a secret, is it? I'd personally prefer to see it (by order of preference): 1) in the same place where you enable the map element (under default page elements, see: https://imgur.com/a/M0tqq), or 2) in a new settings section ("Third-party integrations"?), or 3) under global, which is kind of a sink of all kind of settings.

plugins/sfIsdiahPlugin/modules/sfIsdiahPlugin/templates/indexSuccess.php
```diff
@@ -73,8 +73,8 @@
 
 <?php end_slot() ?>
 
-<?php if (isset($latitude) && isset($longitude) && null !== $key = sfConfig::get('app_google_maps_api_key')): ?>
-  <div id="front-map" class="simple-map" data-key="<?php echo $key ?>" data-latitude="<?php echo $latitude ?>" data-longitude="<?php echo $longitude ?>"></div>
+<?php if (isset($latitude) && isset($longitude) && null !== $keyProperty = QubitSetting::getByName('google_maps_api_key')): ?>
+  <div id="front-map" class="simple-map" data-key="<?php echo $keyProperty->value ?>" data-latitude="<?php echo $latitude ?>" data-longitude="<?php echo $longitude ?>"></div>
```
sevein: This is correct but is it necessary?  `lib/filter/QubitSettingsFilter.class.php` loads sfConfig with the settings in `QubitSetting`. In theory `sfConfig::get('app_google_maps_api_key')` should still return the value of `google_maps_api_key` even after the migration to `QubitSetting`. Not 100% sure but give it a shot? Take a look at the filter too if you can!

apps/qubit/modules/digitalobject/actions/metadataComponent.class.php
```diff
@@ -40,5 +40,17 @@ public function execute($request)
     {
       $this->denyFileNameByPremis = true;
     }
+
+    // Provide Google Maps API key to template
+    $keyProperty = QubitSetting::getByName('google_maps_api_key');
```
sevein: If you use the sfConfig store instead you'll save a hit to the database,  settings are populated in `sfConfig` from our cache backend.

plugins/sfIsadPlugin/modules/sfIsadPlugin/templates/indexSuccess.php
```diff
@@ -326,7 +326,7 @@
 <?php if (0 < count($resource->digitalObjects)): ?>
 
   <div class="digitalObjectMetadata">
-    <?php echo get_partial('digitalobject/metadata', array('resource' => $resource->digitalObjects[0])) ?>
+    <?php echo get_component('digitalobject', 'metadata', array('resource' => $resource->digitalObjects[0], 'infoObj' => $resource)) ?>
```
sevein, line 329: Good catch!

sevein (changes requested):
Pretty good, only a few changes suggested!

sevein:
What's the public ticket? #10232 is private, we should name branches after the public ticket id right?

sevein:
Remember to name the branch and the pull request after the public Redmine ticket next time pleaseeee... 🙏

apps/qubit/modules/settings/actions/securityAction.class.php
mcantelon: Hrm... maybe we can put it into global until there are other API keys to group with it.

plugins/sfIsdiahPlugin/modules/sfIsdiahPlugin/templates/indexSuccess.php
mcantelon: Sounds good.

apps/qubit/modules/settings/actions/securityAction.class.php
sevein: Makes sense!

mcantelon:
Oops... PR and commits reference public ticket, but I forgot to rename the branch. Sorry. :(

mcantelon:
OK, ready for another review.

sevein:
Can we merge the translations? Their classes also have the same names?

mcantelon:
I've never dealt with translations so not sure what you mean.

sevein:
Sorry, I meant the database migrations.

sevein:
This is the reason of my comment:
`arMigration0146.class.php` => `class arMigration0146`
`arMigration0147.class.php` => `class arMigration0146`

In any case, I understand that a migration is necessary for `digital_object_geolocation` because we want it to be enabled by default. However, do we need a migration for `google_maps_api_key` knowing that the default is to **not** have a key?

mcantelon:
Oops... sorry I'll fix that.

mcantelon:
OK, fixed!

lib/task/migrate/migrations/arMigration0146.class.php
```diff
@@ -0,0 +1,48 @@
...
+    if (null === QubitSetting::getByName('google_maps_api_key'))
+    {
+      $setting = new QubitSetting;
+      $setting->setName('google_maps_api_key');
+      $setting->save();
+    }
+
+    return true;
+  }
+}
```
sevein, line 50: Do we need this migration?

data/fixtures/settings.yml
```diff
@@ -779,6 +779,9 @@ QubitSetting:
   QubitSetting_swordDepositDir:
     name: sword_deposit_dir
     value: /tmp
+  QubitSetting_googleMapsApiKey:
+    name: google_maps_api_key
+    value:
```
sevein: Do we need this?

apps/qubit/modules/settings/actions/globalAction.class.php
```diff
@@ -364,6 +366,16 @@ protected function updateGlobalSettings()
       $setting->save();
     }
 
+    // Google Maps Javascript API key
+    if (null !== $googleMapsApiKey = $thisForm->getValue('google_maps_api_key'))
+    {
+      $setting = QubitSetting::getByName('google_maps_api_key');
```
sevein: We can save the migration for `google_maps_api_key` if you create the object on demand, right here, if `$setting` is `NULL`. You did something very similar in the following commit a while ago: f3041d0d291437d5a0816f8c3488b36661e5f1f2.

sevein (changes requested):
I'm still hoping that we can save that extra migration, can we? Me 15 mins ago:

> In any case, I understand that a migration is necessary for digital_object_geolocation because we want it to be enabled by default. However, do we need a migration for google_maps_api_key knowing that the default is to not have a key?

apps/qubit/modules/settings/actions/globalAction.class.php
mcantelon: Cool... okay I'll implement that.

mcantelon: Okay, I've gotten rid of the Google Maps API key migration.

sevein (approved):

sevein:
Squash, please!

mcantelon:
Merged.

PR #440, opened 2016-09-21
Delete user CLI tool, refs #10334

mcantelon:
Added delete user CLI tool (needed to delete "demo" user after using purge CLI tool at the beginning of an import bash script).

### DIFF
--- a/lib/task/tools/deleteUserTask.class.php
+++ b/lib/task/tools/deleteUserTask.class.php
@@ -1,0 +1,103 @@
+<?php
+
+/*
+ * This file is part of the Access to Memory (AtoM) software.
+ *
+ * Access to Memory (AtoM) is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * Access to Memory (AtoM) is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with Access to Memory (AtoM).  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+/**
+ * Delete AtoM user
+ *
+ * @package    symfony
+ * @subpackage task
+ * @author     Mike Cantelon <mike@artefactual.com>
+ */
+class deleteUserTask extends arBaseTask
+{
+  protected function configure()
+  {
+    $this->addArguments(array(
+      new sfCommandArgument('username', sfCommandArgument::REQUIRED, 'The username to delete.')
+    ));
+
+    $this->addOptions(array(
+      new sfCommandOption('application', null, sfCommandOption::PARAMETER_OPTIONAL, 'The application name', true),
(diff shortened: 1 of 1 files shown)


sevein:
Deleting a user from the database will disassociate his notes before deletion. We [warn the administrator](https://github.com/artefactual/atom/blob/qa/2.4.x/lib/model/QubitUser.php#L54-L59) in the user interface, I wonder if we should add a note in the task description (help message) or if you want to go further, require a `--confirm` argument or similar when related notes have been found?

mcantelon:
Sounds good... I'll incorporate a check for that and flag to force dissociation.

mcantelon:
OK, I've updated the branch with the tweak.

lib/task/tools/deleteUserTask.class.php
```diff
@@ -0,0 +1,106 @@
...
+Delete a user.
+EOF;
+  }
+
+  public function execute($arguments = array(), $options = array())
+  {
+    parent::execute($arguments, $options);
+
+    // Abort if deletion isn't forced or confirmed
+    if (!$options['force'])
```
MikeFE: can combine these two ifs with && I think, since short circuiting

lib/task/tools/deleteUserTask.class.php
```diff
@@ -0,0 +1,106 @@
...
+    // If user is an administrator, abort if the user is the only administrator
+    if ($user->hasGroup(QubitAclGroup::ADMINISTRATOR_ID))
+    {
+      $criteria = new Criteria;
+      $criteria->add(QubitAclUserGroup::GROUP_ID, QubitAclGroup::ADMINISTRATOR_ID);
+      $adminCount = count(QubitAclUserGroup::get($criteria));
+
+      if ($adminCount == 1)
+      {
+        throw new Exception('This is the only administrator: deletion aborted.');
```
MikeFE, line 81: good call ;)

lib/task/tools/deleteUserTask.class.php
mcantelon: Ah good catch.

lib/task/tools/deleteUserTask.class.php
```diff
@@ -0,0 +1,106 @@
...
+    $this->addArguments(array(
+      new sfCommandArgument('username', sfCommandArgument::REQUIRED, 'The username to delete.')
+    ));
+
+    $this->addOptions(array(
+      new sfCommandOption('application', null, sfCommandOption::PARAMETER_OPTIONAL, 'The application name', true),
+      new sfCommandOption('env', null, sfCommandOption::PARAMETER_REQUIRED, 'The environment', 'cli'),
+      new sfCommandOption('connection', null, sfCommandOption::PARAMETER_REQUIRED, 'The connection name', 'propel'),
+      new sfCommandOption('force', 'f', sfCommandOption::PARAMETER_NONE, 'Delete without confirmation', null),
+      new sfCommandOption('update-notes', 'n', sfCommandOption::PARAMETER_NONE, 'Dissassociate any notes the user has created', null)
```
sevein, line 40: Wouldn't be better to use mode `PARAMETER_OPTIONAL` with `false` default in `--force` and `--update-notes`? Not really important if it doesn't make a difference, feel free to ignore.

sevein:
:+1:

lib/task/tools/deleteUserTask.class.php
mcantelon: I think PARAMETER_OPTIONAL needs a value, whereas PARAMETER_NONE can just be a flag (and I just checked... if flag not supplied $option[<flag name>] will === false).

mcantelon:
Thanks guys! Closed and merged.

PR #443, opened 2016-10-11
Background bulk export from clipboard, Refs #9998

sbreker:


apps/qubit/modules/object/actions/exportAction.class.php
```diff
@@ -0,0 +1,159 @@
...
+
+  protected function doBackgroundExport($request)
+  {
+    // Export type, CSV or XML?
+    $exportType = $request->getParameter('format', 'xml');
+
+    // We will use this later to redirect users back to the importSelect page
+    if (isset($this->getRoute()->resource))
+    {
+      $exportRoute = array($this->getRoute()->resource, 'module' => 'object', 'action' => 'importSelect', 'type' => $exportType);
```
jraddaoui: Shouldn't it be to the export action? If so, I think you don't need this check nor the redirect in the following exception catch.

apps/qubit/modules/object/actions/exportAction.class.php
```diff
@@ -0,0 +1,159 @@
...
+      // Let user know import has started
+      sfContext::getInstance()->getConfiguration()->loadHelpers(array('Url'));
+      $jobManageUrl = url_for(array('module' => 'jobs', 'action' => 'browse'));
+      $message = '<strong>Export of descriptions initiated.</strong> Check <a href="'. $jobManageUrl . '">job management</a> page to download the results when it has completed.';
+      $this->context->user->setFlash('notice', $this->context->i18n->__($message));
+    }
+    catch (sfException $e)
+    {
+      $this->context->user->setFlash('error', $e->getMessage());
+      $this->redirect($exportRoute);
```
jraddaoui: If I was right before, leaving the user in the same route with the flash alert should be enough. You could use `return sfView::SUCCESS;`

apps/qubit/modules/object/actions/exportAction.class.php
```diff
@@ -0,0 +1,159 @@
...
+    {
+      $this->form->bind($request->getPostParameters());
+
+      if ($this->form->isValid())
+      {
+        $this->processForm();
+
+        $this->doBackgroundExport($request);
+
+        $this->setTemplate('exportResults');
```
jraddaoui: I'd redirect to the previous page, with the flash alert should be enough here too. That way we don't need the exportResults template, which redirects you to the export page again. I'm not completely sure that the following will work, but it should be something like `$this->redirect($request->getReferer());`

apps/qubit/modules/object/templates/exportResultsSuccess.php
```diff
@@ -0,0 +1,7 @@
+<h1><?php echo __('Export initiated') ?></h1>
+
+<section class="actions">
+  <ul>
+    <li><?php echo link_to(__('Back'), array('module' => 'object', 'action' => 'export'), array('class' => 'c-btn')) ?></li>
+  </ul>
+</section>
```
jraddaoui: I'd remove this template if you could redirect to the previous page where the export was initiated

apps/qubit/modules/object/templates/exportSuccess.php
```diff
@@ -0,0 +1,103 @@
...
+              <option value="xml"><?php echo __('XML') ?></option>
+              <option value="csv" selected="selected"><?php echo __('CSV') ?></option>
+            </select>
+          <?php endif; ?>
+
+          <?php if ('csv' != $type): ?>
+            <select name="format">
+              <option value="xml" selected="selected"><?php echo __('XML') ?></option>
+              <option value="csv"><?php echo __('CSV') ?></option>
+            </select>
```
jraddaoui: An inline if in the option elements may be better in here to avoid repeating the whole select block.

apps/qubit/modules/object/templates/exportSuccess.php
```diff
@@ -0,0 +1,103 @@
...
+              </label>
+              <label>
+                <input name="includeAllLevels" type="checkbox"/>
+                <?php echo __('Include all levels of description') ?>
+              </label>
+              <div class="hidden" id="exportLevels">
+                <?php echo $form->levels
+                  ->label(__('Levels of description'))
+                  ->help(__('Select the levels of description to be included in the export. If no levels are selected, the export will fail. You can use the control (Mac ⌘) and/or shift keys to multi-select values from the Levels of description menu. Descriptions that are descendants of levels not included in the export will also be excluded.'))
+                  ->renderRow(array('class' => 'form-autocomplete')) ?>
```
jraddaoui: Not sure if the "form-autocomplete" class is needed here

apps/qubit/modules/object/templates/exportSuccess.php
```diff
@@ -0,0 +1,103 @@
...
+              </div>
+              <br />
+              <?php $taxonomy = QubitTaxonomy::getById(QubitTaxonomy::LEVEL_OF_DESCRIPTION_ID) ?>
```
jraddaoui: Is this intended? I think it's not being used

apps/qubit/modules/object/templates/exportSuccess.php
```diff
@@ -0,0 +1,103 @@
...
+
+
+      </fieldset>
+      </div>
+    </section>
+
+    <section class="actions">
+      <ul>
+        <li><input class="c-btn c-btn-submit" type="submit" id="exportSubmit" value="<?php echo __('Export') ?>"/></li>
+        <li><?php echo link_to(__('Cancel'), array('module' => 'user', 'action' => 'clipboard'), array('class' => 'c-btn')) ?></li>
```
jraddaoui: I'd use the "referer" in here too, I think it will keep the params in the clipboard page if there are any. And if will be ready for the future if we link to here from other places. The request can be accessed from the template with `$sf_request`

lib/job/arInformationObjectCsvExportJob.php
```diff
@@ -67,7 +68,7 @@ public function runJob($parameters)
 
     // Export CSV to temp directory
     $this->info($this->i18n->__('Starting export to %1.', array('%1' => $tempPath)));
-    $itemsExported = $this->exportResults($tempPath);
+    $itemsExported = $this->exportResults($tempPath, $this->params);
```
jraddaoui: `$this->params` should be available in there without sending it

lib/job/arInformationObjectCsvExportJob.php
```diff
@@ -121,6 +122,17 @@ protected function exportResults($path)
       {
         $writer->exportResource($resource);
 
+        // export descendants if configured
+        if (true != $this->params['current-level-only'])
```
jraddaoui: `if (!$this->params['current-level-only'])` looks better to me

lib/job/arXmlExportJob.php
```diff
@@ -0,0 +1,190 @@
...
+        {
+          $zip->addFile($path . DIRECTORY_SEPARATOR . $file, $file);
+        }
+      }
+
+      $zip->close();
+    }
+
+    return $success;
+  }
```
jraddaoui: The createZipForDownload function could be moved to arBaseJob and removed from both CSV and XML export jobs

lib/job/arXmlExportJob.php
```diff
@@ -0,0 +1,190 @@
...
+  /**
+   * Export search results as CSV
+   *
+   * @param string  Path of file to write CSV data to
+   *
+   * @return null
+   */
+  protected function exportResults($path, $params)
+  {
+    $exitCode= 0;
```
jraddaoui: Missing space

jraddaoui:
Hi @sbreker, nice work!

I've added a few comments along the changes and, in general, I wonder if this should be working with actors and repositories. I could not see any way to add those to the clipboard and, in both jobs, it looks like all kind of slugs can be sent but only the information object index is queried.

Also, I'd appreciate if you could update the pull request title and description with the commit title and description you will use in the merge. Thanks!

apps/qubit/modules/object/actions/exportAction.class.php
```diff
@@ -0,0 +1,159 @@
...
+      if ('CSV' == strtoupper($exportType))
+      {
+        QubitJob::runJob('arInformationObjectCsvExportJob', $options);
+      }
+      else
+      {
+        QubitJob::runJob('arXmlExportJob', $options);
+      }
+
+      // Let user know import has started
```
jraddaoui: Export ;)

apps/qubit/modules/object/actions/exportAction.class.php
jraddaoui: You may need to check that the "referer" exists and if it doesn't I'd let it reach the default template

apps/qubit/modules/object/templates/exportSuccess.php
jraddaoui: You may need to check that the "referer" exists and if it doesn't I'd redirect to the clipboard page (until this can be reached from other places)

apps/qubit/modules/object/actions/exportAction.class.php
sbreker: Good catch - cut and paste error on my part.  Will adjust this.

apps/qubit/modules/object/actions/exportAction.class.php
sbreker: Changing this to `return sfView::SUCCESS` does not seem to work. There are two undesirable effects when I use that:
- I want to have the 'type' specified on the URL to pre-set one of the dropdowns.
- If execution is in this catch block we have an error.  On error, I want to do a http GET on this form and re-load with the 'type' set.  Returning sfView::SUCCESS appears to do a POST and makes it look like the export completed successfully. I have tried 'sfView::ERROR' but still see the form being POSTed.

apps/qubit/modules/object/actions/exportAction.class.php
sbreker: Sounds good - I've made this change.  The flash alert does still show up with a link to the job management page.  Not sure what you mean by 'default template' - if (null == getReferer()) then I redirect to the export page: `$this->redirect(array('module' => 'object', 'action' => 'export'));`

apps/qubit/modules/object/templates/exportResultsSuccess.php
sbreker: Removed.

apps/qubit/modules/object/templates/exportSuccess.php
sbreker: Done.

apps/qubit/modules/object/templates/exportSuccess.php
sbreker: Tested and you are correct. I have removed the array from renderRow().

apps/qubit/modules/object/templates/exportSuccess.php
sbreker: Removed.

apps/qubit/modules/object/templates/exportSuccess.php
sbreker: Yes, I can see the value in this in case we add additional links to this form. Change made.

lib/job/arInformationObjectCsvExportJob.php
sbreker: Done.

lib/job/arInformationObjectCsvExportJob.php
sbreker: Completed.

lib/job/arXmlExportJob.php
sbreker: Change made.

lib/job/arXmlExportJob.php
sbreker: Completed.

apps/qubit/modules/object/actions/exportAction.class.php
sbreker: Fixed.  :)

sbreker:
The work to add export functionality for the additional entity types (repos & actors) has been broken out into other tickets.  This task completes the WebUI changes and creates the XML export background job.

apps/qubit/modules/object/actions/exportAction.class.php
jraddaoui: It could be stored in `$this->type` making it available in the template instead of sending it in the request. I think we could remove the checks to build the URL and avoid the redirect that way.

apps/qubit/modules/object/actions/exportAction.class.php
```diff
@@ -0,0 +1,166 @@
...
+      $this->response->addJavaScript('exportOptions', 'last');
+
+      if (isset($request->type))
+      {
+        $this->type = $request->type;
+      }
+
+      $this->title = $this->context->i18n->__('Export');
+    }
+  }
```
jraddaoui, line 152: I'd do all this else block at the beginning of the function, outside the method condition, that way it will be all ready even if the form is not valid in a POST.

apps/qubit/modules/object/actions/exportAction.class.php
```diff
@@ -0,0 +1,166 @@
...
+        $this->doBackgroundExport($request);
+
+        if (null !== $redirectUrl = $request->getReferer())
+        {
+          $this->redirect($redirectUrl);
+        }
+        else
+        {
+          $this->redirect(array('module' => 'object', 'action' => 'export'));
+        }
```
jraddaoui: If you place the next else work at the beginning of the function like I say in the previous comment, this else and redirect to the same page shouldn't be needed.

apps/qubit/modules/object/actions/exportAction.class.php
jraddaoui: If it reaches the end of the execute method without a redirect or a return it will load the default template, the same as doing `return sfView::SUCCESS` at the end

apps/qubit/modules/object/actions/exportAction.class.php
```diff
@@ -0,0 +1,166 @@
...
+
+    // We will use this later to redirect users back to the export page
+    if (isset($this->getRoute()->resource))
+    {
+      $exportRoute = array($this->getRoute()->resource, 'module' => 'object', 'action' => 'export', 'type' => $exportType);
+    }
+    else
+    {
+      $exportRoute = array('module' => 'object', 'action' => 'export', 'type' => $exportType);
+    }
```
jraddaoui: Actually, this looks like it comes from the import page, where you can import directly to a resource, but it should not be needed in the export page (AFAIK it can't be reached from a resource). And, as I said in other comments, if you move the `$this->limit` population outside the method check in the execute function, you should be able to access `$type` in the template in all cases.

jraddaoui:
Thanks @sbreker, just a few more comments about the redirect and the request method check that I think it will make it a little cleaner.

A detailed description in the merged commit (and the PR) will help to understand the feature and, for following PRs, if you keep the commit history intact while it's being reviewed, it makes it easier for the reviewer, but that may be my personal opinion ;)

Thanks!!

apps/qubit/modules/menu/templates/_clipboardMenu.php
```diff
@@ -15,13 +15,10 @@
...
-
-      <span data-plural-label="<?php echo __('descriptions selected') ?>" data-single-label="<?php echo __('description selected') ?>">
-        <?php if ($count != 1): ?>
-          <?php echo $count . ' ' . __('descriptions selected') ?>
-        <?php else: ?>
-          <?php echo $count . ' ' . __('description selected') ?>
-        <?php endif; ?>
+      <span id="count-block" data-information-object-label="<?php echo sfConfig::get('app_ui_label_informationobject') ?>" data-actor-object-label="<?php echo sfConfig::get('app_ui_label_actor') ?>" data-repository-object-label="<?php echo sfConfig::get('app_ui_label_repository') ?>">
+        <?php foreach ($countByType as $objectType => $countType): ?>
+          <?php echo $objectTypes[$objectType] . ' count: ' . $countType ?><br />
```
MikeFE: don't forget __() for the 'count:' string

apps/qubit/modules/object/actions/exportAction.class.php
```diff
@@ -0,0 +1,166 @@
...
+      }
+      else
+      {
+        QubitJob::runJob('arXmlExportJob', $options);
+      }
+
+      // Let user know export has started
+      sfContext::getInstance()->getConfiguration()->loadHelpers(array('Url'));
+      $jobManageUrl = url_for(array('module' => 'jobs', 'action' => 'browse'));
+      $message = '<strong>Export of descriptions initiated.</strong> Check <a href="'. $jobManageUrl . '">job management</a> page to download the results when it has completed.';
```
MikeFE: do we need i18n here too?

apps/qubit/modules/object/actions/exportAction.class.php
MikeFE: can you use $this->var = blah when using return sfView::SUCCESS? I can't seem to get it to work!

jraddaoui: It works for me. Could it be that it was not set in the URL? It's checking it to populate the var.

And that's other thing I don't like very much, we're calling the same thing three different names: format, type and exportType in the same action, which creates a lot of confusion and unneeded code.

apps/qubit/modules/menu/templates/_clipboardMenu.php
jraddaoui: Thanks Mike ;)

apps/qubit/modules/object/actions/exportAction.class.php
jraddaoui: That's wrapped after, which I have no idea if it works ...

sevein: That's not going to work, the i18n extractor finds the strings statically. You need to put the string literal inside `__()` AFAIK.

If you need replacements, use:

``` php
__('My name is %1%', array('%1%' => $name));
```

We try not to include HTML tags in our i18n strings, e.g.:

``` php
__('Export of descriptions initiated. %1%Check job management%2% page to download the results when it has completed.', array(
    '%1%' => sprintf('<a href="%s">', url_for(...)),
    '%2' => '</a>'));
```

I think the idea is to avoid having the translator writing HTML, but I'm not sure if AtoM is consistently doing that, probably not.

apps/qubit/modules/menu/templates/_clipboardMenu.php
sevein: You have to use replacements, remember!

`<?php echo __('%1% count: %2%', array('%1%' => $objectTypes[$objectType], '%2%' => $countType)); ?>`

The translator will see `%1% count: %2%` which is going to be really hard to translate without contextual markers but I don't think they're supported in Symfony. Django does: https://docs.djangoproject.com/en/1.10/topics/i18n/translation/#contextual-markers.

apps/qubit/modules/object/actions/exportAction.class.php
sbreker: I think we may have been misunderstanding each other.  I have changed this and have code working using the `return sfView::SUCCESS`.

apps/qubit/modules/object/actions/exportAction.class.php
sbreker: Change made.

apps/qubit/modules/object/actions/exportAction.class.php
sbreker: Change made.

apps/qubit/modules/object/actions/exportAction.class.php
sbreker: I've removed this code.

apps/qubit/modules/object/actions/exportAction.class.php
```diff
@@ -0,0 +1,158 @@
...
+  public function execute($request)
+  {
+    parent::execute($request);
+
+    $this->response->addJavaScript('exportOptions', 'last');
+
+    if (isset($request->type))
+    {
+      $this->type = $request->type;
+    }
```
jraddaoui: I think a default value is needed in `$this->type` for GET requests, as it's being used in the template in the type field. I'd replace this if with the way it's being done in the doBackgroundExport function:

`$this->type = $request->getParameter('type', 'csv');`

apps/qubit/modules/object/actions/exportAction.class.php
```diff
@@ -0,0 +1,158 @@
...
+        }
+
+        break;
+    }
+  }
+
+  protected function doBackgroundExport($request)
+  {
+    // Export type, CSV or XML?
+    $this->type = $request->getParameter('type', 'csv');
```
jraddaoui: It could be removed from here after it's added in the execute function.

jraddaoui:
👍 Thanks @sbreker !!

PR #445, opened 2016-10-14
Fix pagination in term index page, refs #10426

jraddaoui:
In terms where the related information objects are shown (places, subjects
and genres) a default limit is needed to make the pager work properly


sevein:
Sorry, can you remind me why is this necessary? I was thinking on [this snippet](https://github.com/artefactual/atom/blob/qa/2.4.x/apps/qubit/modules/default/actions/browseAction.class.php#L160-L164), which populates that value in `$this->limit`.

jraddaoui:
Right, thanks @sevein!

PR #447, opened 2016-10-18
Background file import error, refs #10137

sbreker:
An issue was discovered in testing where the import file uploaded via
the WebUI was no longer available for the background import job once
it was launched.

This change moves the uploaded import file to the uploads/tmp
directory before launching the background job. Also added code to the
background job (arFileImportJob) to remove the temp file from
uploads/tmp after the file has been processed.

### DIFF
--- a/apps/qubit/modules/object/actions/importSelectAction.class.php
+++ b/apps/qubit/modules/object/actions/importSelectAction.class.php
@@ -95,6 +95,13 @@ class ObjectImportSelectAction extends DefaultEditAction
     }
   }
 
+  /**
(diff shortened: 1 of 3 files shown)


sevein:
Nice! I really like `moveUploadFile`. Sounds like you don't really have many hours left right? If you do, it would be nice to move this function somewhere else like the `Qubit` class (a bit of a sink but it will do it) and reuse it in `object/import` but also `digitalobject/upload`, etc... if you don't have time, maybe file a issue so we don't forget?

sevein:
:+1:

PR #448, opened 2016-10-24
Fix treeview_type setting lookup, refs #10256

sevein:
Redmine: https://projects.artefactual.com/issues/10256

Make sure that we look up the source value of `treeview_type` and not a translation.

It also works in `stable/2.3.x`.


jraddaoui:
👍

PR #449, opened 2016-10-26
Prevent duplicating fields when using match-and-update, #10150

MikeFE:
Previously, if you run a csv:import task with --update="match-and-update", if the same event dates and access points are in the updated CSV, it will duplicate those fields in AtoM. This fix:
- Prevents access points from duplicating when in match-and-update mode
- Prevents event start/end dates from duplicating when in match-and-update mode

### DIFF
--- a/lib/QubitFlatfileImport.class.php
+++ b/lib/QubitFlatfileImport.class.php
@@ -666,10 +666,19 @@ class QubitFlatfileImport
 
       foreach ($fields as $field)
       {
-        // Found a field that doesn't match, so this isn't a duplicate. Move onto the next event and check...
-        if ($existingEvent->$field != $event->$field)
+        // Use special logic when comparing dates, see dateStringsEqual for details.
+        if (false !== strpos(strtolower($field), 'date'))
         {
-          $match = false;
+          $match = $match && $this->dateStringsEqual($existingEvent->$field, $event->$field);
+        }
+        else
+        {
+          $match = $match && $existingEvent->$field === $event->$field;
+        }
+
+        // Event fields differ, don't bother checking other fields since these aren't equal
+        if (!$match)
+        {
           break;
         }
       }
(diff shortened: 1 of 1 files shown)


lib/QubitFlatfileImport.class.php
```diff
@@ -1766,6 +1812,40 @@ public function createObjectTermRelation($objectId, $termId)
...
+   *
+   * @return bool  True if this relation already exists, false otherwise.
+   */
+  private function objectTermRelationExists($objectId, $termId)
+  {
+    $c = new Criteria;
+    $c->add(QubitObjectTermRelation::OBJECT_ID, $objectId);
+    $c->add(QubitObjectTermRelation::TERM_ID, $termId);
+
+    return null !== QubitObjectTermRelation::getOne($c);
```
mcantelon, line 1845: We could avoid using the ORM for performance.

lib/QubitFlatfileImport.class.php
```diff
@@ -1766,6 +1812,40 @@ public function createObjectTermRelation($objectId, $termId)
   }
 
   /**
+   * Check whether or not a term relation already exists for this info object.
+   *
+   * @param integer $subjectId  The term or actor we're relating to.
```
mcantelon, line 1817: Physical objects too IIRC.

lib/QubitFlatfileImport.class.php
```diff
@@ -1766,6 +1812,40 @@ public function createObjectTermRelation($objectId, $termId)
   }
 
   /**
+   * Check whether or not a term relation already exists for this info object.
```
mcantelon, line 1815: (Might not be a term.)

lib/QubitFlatfileImport.class.php
```diff
@@ -1766,6 +1812,40 @@ public function createObjectTermRelation($objectId, $termId)
...
+   *
+   * @return bool  True if this relation already exists, false otherwise.
+   */
+  private function relationExists($subjectId, $objectId)
+  {
+    $c = new Criteria;
+    $c->add(QubitRelation::OBJECT_ID, $objectId);
+    $c->add(QubitRelation::SUBJECT_ID, $subjectId);
+
+    return null !== QubitRelation::getOne($c);
```
mcantelon, line 1828: We could avoid using the ORM for performance.

mcantelon (commented):
Looks good overall. A few possible minor tweaks.

PR #450, opened 2016-10-28
Import changes, refs #10472, #10016 and #10017

jraddaoui:
- Add repository CSV import via UI, refs #10472:
  - Add repository option to CSV import form
  - Fix CSV import form submission
  - Remove merge-existing option and allow update and other options
  - Disable/enable search index based on option
- Add update options to repo CSV import, refs #10017:
  - Display update fields based on import object type
  - Set update options in QubitFlatfileImport
  - Use update options on repository CSV import
- Update options for actor CSV import, refs #10016:
  - Add new update, skip and limit options to the actor CSV import task
  - Merge actor and repo update logic in the same function
  - Fix delete method for actors and repositories

### DIFF
--- a/lib/QubitFlatfileImport.class.php
+++ b/lib/QubitFlatfileImport.class.php
@@ -124,6 +124,39 @@ class QubitFlatfileImport
     }
   }
 
+  public function setUpdateOptions($options)
+  {
+    if ($options['limit'])
+    {
+      $this->limitToId = $this->getIdCorrespondingToSlug($options['limit']);
+    }
+
+    // Are there params set on --update flag?
+    if ($options['update'])
+    {
+      // Parameters for --update are validated in csvImportBaseTask.class.php.
+      switch ($options['update'])
+      {
+        case 'delete-and-replace':
+          // Delete any matching records, and re-import them (attach to existing entities if possible).
+          $this->deleteAndReplace = true;
+          break;
+
+        case 'match-and-update':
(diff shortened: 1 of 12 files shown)


lib/QubitFlatfileImport.class.php
```diff
@@ -962,6 +992,104 @@ private function setInformationObjectByKeymap($legacyId)
...
+    }
+
+    // Check existing repo by auth. form of name
+    $query = "SELECT object.id
+      FROM object JOIN actor_i18n i18n
+      ON object.id = i18n.id
+      WHERE i18n.authorized_form_of_name = ?
+      AND object.class_name = ?;";
+
+    $statement = QubitFlatfileImport::sqlQuery($query, array($this->columnValue('authorizedFormOfName'), $this->className));
```
MikeFE, line 1016: I think we should use QubitPdo::\* when possible for queries of this nature, sqlQuery is going to be removed eventually since it just copies the functionality and is only used in 2-3 places as opposed to QubitPdo which is everywhere.

jraddaoui: Hi @MikeFE, I like the idea, but right now it's being used 11 times in QubitFlatfileImport.class.php, I'd rather change it when we replace it everywhere. Please, let me know if you're okay with it. 😉

And thanks for the tip, I'll not use it in the XML import changes I'm doing.

PR #454, opened 2016-11-02
Description updates/index functions, refs #10458

mcantelon:
* Added ElasticSearch indexing of functions
* Changed description updates page to use ElasticSearch
* Redesigned description updates page

### DIFF
--- a/apps/qubit/modules/search/actions/descriptionUpdatesAction.class.php
+++ b/apps/qubit/modules/search/actions/descriptionUpdatesAction.class.php
@@ -29,12 +29,10 @@ class SearchDescriptionUpdatesAction extends sfAction
   public static
     $NAMES = array(
       'className',
-      'dateStart',
-      'dateEnd',
+      'startDate',
+      'endDate',
       'dateOf',
-      'publicationStatus',
-      'limit',
-      'sort'
+      'publicationStatus'
     );
 
   protected function addField($name)
@@ -54,15 +52,15 @@ class SearchDescriptionUpdatesAction extends sfAction
 
         break;
 
-      case 'dateStart':
+      case 'startDate':
         $this->form->setValidator($name, new sfValidatorDate(array(), array('invalid' => $this->context->i18n->__('Invalid start date'))));
-        $this->form->setWidget($name, new sfWidgetFormInput(array(), array('style' => 'width: auto')));
+        $this->form->setWidget($name, new sfWidgetFormInput);
 
         break;
 
-      case 'dateEnd':
+      case 'endDate':
         $this->form->setValidator($name, new sfValidatorDate(array(), array('invalid' => $this->context->i18n->__('Invalid end date'))));
-        $this->form->setWidget($name, new sfWidgetFormInput(array(), array('style' => 'width: auto')));
+        $this->form->setWidget($name, new sfWidgetFormInput);
 
         break;
 
@@ -89,18 +87,6 @@ class SearchDescriptionUpdatesAction extends sfAction
         $this->form->setWidget($name, new arWidgetFormSelectRadio(array('choices' => $choices, 'class' => 'radio inline')));
 
         break;
-
-      case 'limit':
-        $this->form->setValidator($name, new sfValidatorString);
-        $this->form->setWidget($name, new sfWidgetFormInputHidden);
-
-        break;
-
-      case 'sort':
-        $this->form->setValidator($name, new sfValidatorString);
-        $this->form->setWidget($name, new sfWidgetFormInputHidden);
-
-        break;
     }
   }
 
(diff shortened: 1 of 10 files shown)


apps/qubit/modules/search/actions/descriptionUpdatesAction.class.php
```diff
@@ -118,150 +104,95 @@ public function execute($request)
...
+      'publicationStatus' => 'all'
     );
 
     $this->form->bind($request->getRequestParameters() + $request->getGetParameters() + $defaults);
 
     if ($this->form->isValid())
     {
       $this->className = $this->form->getValue('className');
-
+      $this->nameColumnDisplay = ($this->className == 'QubitInformationObject') ? 'Title' : 'Name';
```
sevein: Not using i18n.

apps/qubit/modules/search/templates/_updatesSearch.php
```diff
@@ -0,0 +1,76 @@
...
+          <div class="filter-left">
+            <?php echo $form->className
+              ->label(__('Type'))
+              ->renderRow() ?>
+          </div>
+
+          <div class="filter-center">
+            <label class="date-of-label"><?php echo __('Date of') ?></label>
+            <div class="date-of">
+              <?php foreach($form->getWidgetSchema()->dateOf->getChoices() as $value => $translatedText): ?>
```
sevein: `foreach` is a statement not a func

mcantelon:
Note: the `_updatesSearch.php` template generates the radio button HTML manually to output it in a non-standard way so it looks like the analyst's mockup for it.

apps/qubit/modules/search/templates/descriptionUpdatesSuccess.php
```diff
@@ -1,134 +1,124 @@
...
-        <?php echo $form->publicationStatus
-          ->label(__('Publication status (%1% only)', array('%1%' => sfConfig::get('app_ui_label_informationobject'))))
-          ->renderRow() ?>
+<?php end_slot() ?>
 
-        <button type="submit" class="btn"><?php echo __('Search') ?></button>
+<?php slot('content') ?>
 
-      </form>
+  <?php echo get_partial('search/updatesSearch', array(
```
sevein, line 25: Can you change the style of this component so it doesn't float on top of the `.table`? To make it visually more attractive I think we need to create some room between the search widget and the table.

This is what I am seeing now:

screenshot

I would expect something like:

screenshot

apps/qubit/modules/search/actions/descriptionUpdatesAction.class.php
```diff
@@ -118,150 +104,95 @@ public function execute($request)
...
+      'startDate' => date('Y-m-d', strtotime('-1 month')),
+      'endDate' => date('Y-m-d'),
       'dateOf' => 'CREATED_AT',
-      'publicationStatus' => 'all',
-      'limit' => '10',
-      'sort' => 'updatedDown'
+      'publicationStatus' => 'all'
     );
 
     $this->form->bind($request->getRequestParameters() + $request->getGetParameters() + $defaults);
```
sevein: It's unclear to me that we need to use getRequestParameters in this context bound to the form, do we?

apps/qubit/modules/search/actions/descriptionUpdatesAction.class.php
```diff
@@ -118,150 +104,95 @@ public function execute($request)
...
+    $this->pager = new QubitSearchPager($resultSet);
+    $this->pager->setMaxPerPage($limit);
+    $this->pager->setPage($this->request->getParameter('page', 1));
 
-    // End date at midnight
-    if (null != $this->form->getValue('dateEnd'))
-    {
-      $dateEnd = $this->form->getValue('dateEnd').' 24:00:00';
-    }
+    return $resultSet;
```
sevein: Do you need to return the `$resultSet`? It doesn't seem that you're making use of it and it's carred to the template via the pager, right?

apps/qubit/modules/search/actions/descriptionUpdatesAction.class.php
```diff
@@ -118,150 +104,95 @@ public function execute($request)
...
+  }
 
-    // Add date criteria
-    switch ($dateOf = $this->form->getValue('dateOf'))
+  private function addDateRangeQuery(&$queryBool)
+  {
+    switch ($this->form->getValue('dateOf'))
```
sevein: Regarding to separation of roles, should this function receive the value of `dateOf` as a parameter instead of looking up the form?

apps/qubit/modules/search/actions/descriptionUpdatesAction.class.php
```diff
@@ -118,150 +104,95 @@ public function execute($request)
...
-    if (null != $this->form->getValue('dateEnd'))
-    {
-      $dateEnd = $this->form->getValue('dateEnd').' 24:00:00';
-    }
+    return $resultSet;
+  }
 
-    // Add date criteria
-    switch ($dateOf = $this->form->getValue('dateOf'))
+  private function addDateRangeQuery(&$queryBool)
```
sevein: I believe that objects are always passed by reference in PHP5 and you don't need to use the `&` operator.

> In PHP 5 there is a new Object Model. PHP's handling of objects has been completely rewritten, allowing for better performance and more features. In previous versions of PHP, objects were handled like primitive types (for instance integers and strings). The drawback of this method was that semantically the whole object was copied when a variable was assigned, or passed as a parameter to a method. In the new approach, objects are referenced by handle, and not by value (one can think of a handle as an object's identifier).

apps/qubit/modules/search/actions/descriptionUpdatesAction.class.php
```diff
@@ -118,150 +104,95 @@ public function execute($request)
...
-      case 'updatedDown':
-      default:
-       $criteria->addDescendingOrderByColumn(QubitObject::UPDATED_AT);
+        $queryBool->addMust($bothDatesQueryBool);
     }
+  }
 
-    // Add fallback criteria for name
-    if ('nameDown' == $this->sort || 'nameUp' == $this->sort)
+  private function addDateRangeQueryClause(&$queryBool, $field)
```
sevein: Same here, objects are always passed by reference.

apps/qubit/modules/search/actions/descriptionUpdatesAction.class.php
```diff
@@ -118,150 +104,95 @@ public function execute($request)
...
+  {
+    if (null !== $this->form->getValue('startDate'))
```
sevein: Same here, receive parameter instead of form lookup?

apps/qubit/modules/search/actions/descriptionUpdatesAction.class.php
```diff
@@ -118,150 +104,95 @@ public function execute($request)
...
-      $criteria = QubitCultureFallback::addFallbackCriteria($criteria, $this->form->getValue('className'));
+      $queryBool->addMust(new \Elastica\Query\Range($field, array('gte' => $this->form->getValue('startDate'))));
     }
 
-    // Page results
-    $this->pager = new QubitPager($this->className);
-    $this->pager->setCriteria($criteria);
-    $this->pager->setMaxPerPage($this->form->getValue('limit'));
-    $this->pager->setPage($this->request->getParameter('page', 1));
+    if (null !== $this->form->getValue('endDate'))
```
sevein: Same here, receive parameter instead of form lookup?

apps/qubit/modules/search/templates/_updatesSearch.php
```diff
@@ -0,0 +1,76 @@
+<section class="advanced-search-section">
```
sevein: This is just a general thought about _updatesSearch.php. Why do you implent it as its own partial? Is it for future composability of reusability? Do you think there will be a case for it?

apps/qubit/modules/search/templates/_updatesSearch.php
```diff
@@ -0,0 +1,76 @@
...
+              ->label(__('Type'))
+              ->renderRow() ?>
+          </div>
+
+          <div class="filter-center">
+            <label class="date-of-label"><?php echo __('Date of') ?></label>
+            <div class="date-of">
+              <?php foreach($form->getWidgetSchema()->dateOf->getChoices() as $value => $translatedText): ?>
+                <label>
+                  <input type="radio" name="dateOf" value="<?php echo $value ?>" <?php echo $form->getValue('dateOf') == $value ? 'checked' : '' ?>>
```
sevein, line 28: I've read your note about why you're using raw HTML here but wouldn't be more correct to implement our own widget form like `arWidgetFormSelectRadio` to avoid having too much logic in our templates? I wouldn't do it if you're running out of time though.

apps/qubit/modules/search/templates/_updatesSearch.php
```diff
@@ -0,0 +1,76 @@
...
+                  <?php echo $translatedText ?>
+                </label>
+              <?php endforeach; ?>
+            </div>
+          </div>
+
+          <div class="filter-right">
+            <label class="publication-status-label"><?php echo __('Publication status (%1% only)', array('%1%' => sfConfig::get('app_ui_label_informationobject'))) ?></label>
+            <div class="publication-status">
+              <?php foreach($form->getWidgetSchema()->publicationStatus->getChoices() as $value => $translatedText): ?>
```
sevein: `foreach` is a statement.

apps/qubit/modules/search/templates/descriptionUpdatesSuccess.php
```diff
@@ -1,134 +1,124 @@
...
-              <?php echo $result->createdAt ?>
-            <?php endif; ?>
-          </td>
 
-          <td>
-            <?php if ('QubitInformationObject' == $className || 'QubitActor' == $className || 'QubitRepository' == $className): ?>
-              <?php echo get_component('informationobject', 'clipboardButton', array('slug' => $result->slug, 'wide' => true)) ?>
-            <?php endif; ?>
-          </td>
+            <td>
```
sevein: Shouldn't this `<td></td>` be contained by the `if` clause? Otherwise we end up with an empty column / weird visual glitch when the user is search for `QubitFunction` or `QubitTerm`.

apps/qubit/modules/search/templates/descriptionUpdatesSuccess.php
```diff
@@ -1,134 +1,124 @@
...
+          <?php endif; ?>
+        </tr>
+      </thead><tbody>
+        <?php foreach ($pager->getResults() as $result): ?>
 
-<?php end_slot() ?>
+          <?php $doc = $result->getSource() ?>
 
-<?php slot('content') ?>
+          <tr class="<?php echo 0 == @++$row % 2 ? 'even' : 'odd' ?>">
```
sevein: I would recomend to use `table-striped` instead. We used it in `inventorySuccess.php` and I think it worked nicely IIRC.

apps/qubit/modules/search/templates/descriptionUpdatesSuccess.php
```diff
@@ -1,134 +1,124 @@
...
+              <div class="separator" style="display: inline;">/</div>
+              <a href="#" class="none">None</a>
+            </th>
+          <?php endif; ?>
+        </tr>
+      </thead><tbody>
+        <?php foreach ($pager->getResults() as $result): ?>
 
-<?php end_slot() ?>
+          <?php $doc = $result->getSource() ?>
```
sevein: We tend to use `getData()` instead of `getSource()` (see `\Elastic\Result`) just in case we want to access outter fields of the document and not just the stuff in `_source`. Some of the helpers you're using later may be expecting them to be available down the road.

js/dominion.js
```diff
@@ -833,7 +833,8 @@
 
   $(function ()
     {
-      var $advancedSearch = $('body.informationobject.browse');
+      // Find search for if on an appropriate page
+      var $advancedSearch = $('body.informationobject.browse').add('body.search.descriptionUpdates');
```
sevein: I think that a comma-separated list of selectors will be faster because only one jQuery object is built. The docs of add() give more details.

plugins/arElasticSearchPlugin/lib/model/arElasticSearchFunction.class.php
```diff
@@ -0,0 +1,65 @@
...
+  {
+    $errors = array();
+
+    $sql  = 'SELECT function.id';
+    $sql .= ' FROM '.QubitFunction::TABLE_NAME.' function';
+    $sql .= ' JOIN '.QubitObject::TABLE_NAME.' object ON function.id = object.id';
+    $sql .= ' WHERE object.class_name = ?';
+    $sql .= ' ORDER BY function.lft';
+
+    $actors = QubitPdo::fetchAll($sql, array('QubitFunction'));
```
sevein, line 32: I've never realized we use fetchAll here (in similar classes for other entities). The only reason why we do it seems to be that `count()`. It sounds like we should use SQL instead for that count and then make another query using just `fetch` so we can use a scrollable cursor in the loop to save memory right? It doesn't have to happen now though unless you have time :)

sevein (changes requested):
It works. I'm including many comments in my review but some of them suggest changes that are probably not realistic considering how much time you have left and that's ok, we can address later but it's good to keep them in mind or discuss if you want.

apps/qubit/modules/search/actions/descriptionUpdatesAction.class.php
mcantelon: Good question... I'll see if it works without it.

apps/qubit/modules/search/templates/_updatesSearch.php
mcantelon: Just to make the parent template less cluttered. I could put it in there too.

js/dominion.js
mcantelon: I tried using the command-separated initially and for some reason couldn't get it to work. But I'll do a quick re-try.

mcantelon:
OK, I've updated, addressing all feedback I think (except for removing fetchAll and creating a arWidgetFormSelectRadio variant).

sevein:
Thank you, looks great! :+1:

sevein (approved):

mcantelon:
Thanks! Merged.

PR #455, opened 2016-11-03
Improve actor matching on IO import, refs #10506

jraddaoui:
Fetch or create a QubitActor record based on the actor name, the
imported IO repository and the update options. Update the actor
history in matches from the same repository when using the match
and update option. For a full description, check:

https://projects.artefactual.com/issues/10506


sevein:
Test

PR #456, opened 2016-11-04
Ensure slug separator value is dash, Refs #10276

sbreker:
This change hard codes the slug separator value to dash ('-') instead of
pulling this value from the 'Reference code separator' setting.

A migration file is also included that will replace any period or forward-
slash characters found in the slug table with a dash ('-').

### DIFF
--- a/lib/model/QubitSlug.php
+++ b/lib/model/QubitSlug.php
@@ -27,22 +27,25 @@ class QubitSlug extends BaseSlug
 
   public static function random($length = 12)
   {
-    $separator = sfConfig::get('app_separator_character', '-');
+    $separator = '-';
 
     // Adapted from http://stackoverflow.com/questions/5615490/random-code-generator/5615957#5615957
     $alphabet = '23456789abcdefghkmnpqrstwxyz';
-    $alphabet_size = strlen($alphabet);
+    $alphabetSize = strlen($alphabet);
 
-    $block_length = 4;
-    $num_blocks = $length / $block_length;
+    $blockLength = 4;
+    $numBlocks = ceil($length / $blockLength);
 
     $slug = '';
-    for ($i = 0; $i < $num_blocks; $i++) {
-      for ($j = 0; $j < $block_length; $j++) {
-        $slug .= $alphabet[mt_rand(0, $alphabet_size - 1)];
+    for ($i = 0; $i < $numBlocks; $i++)
+    {
+      for ($j = 0; $j < $blockLength; $j++)
+      {
+        $slug .= $alphabet[mt_rand(0, $alphabetSize - 1)];
       }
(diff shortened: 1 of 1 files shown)


lib/model/QubitSlug.php
```diff
@@ -27,22 +27,25 @@ class QubitSlug extends BaseSlug
...
+    for ($i = 0; $i < $num_blocks; $i++)
+    {
+      for ($j = 0; $j < $block_length; $j++)
+      {
         $slug .= $alphabet[mt_rand(0, $alphabet_size - 1)];
       }
 
-      if ($i != $num_blocks - 1) {
+      if ($i != $num_blocks - 1)
+      {
```
sevein, line 48: Now that you're here... can we use `$camelCase` style for the vars?

lib/task/migrate/migrations/arMigration0147.class.php
```diff
@@ -0,0 +1,45 @@
...
+  /**
+   * Upgrade
+   *
+   * @return bool True if the upgrade succeeded, False otherwise
+   */
+  public function up($configuration)
+  {
+    // Correct slug records created with a separator of . or / instead of -.
+    QubitPdo::prepareAndExecute('UPDATE slug SET slug = REPLACE(slug, ".", "-") WHERE slug LIKE "%.%"');
+    QubitPdo::prepareAndExecute('UPDATE slug SET slug = REPLACE(slug, "/", "-") WHERE slug LIKE "%/%"');
```
sevein: That's cool. Should we chain the two REPLACE functions in one single query to avoid to passes to the entire table? Also, I've noticed that this issue has been tagged for v2.3.1 however I'm not sure if we're supposed to deliver new migrations in minor releases. Can you confirm that with @djjuhasz, please? If we did it, I think we'd need to reuse v139 and bump the others - it's too complicated.

sevein (changes requested):

lib/model/QubitSlug.php
sbreker: Absolutely.

lib/task/migrate/migrations/arMigration0147.class.php
sbreker: David had indeed suggested going to route of using a migration. For 2.3.1, I do have a tools:run script that could be used in place of a migration.

sevein: Cool!

sbreker: David confirmed that using a migration for this is not ideal. Dropping migration 147 from this commit in favour of tools:run script.

sevein:
:+1:

PR #457, opened 2016-11-05
Add lazy load to full width treeview, refs #10522

jraddaoui:
- Add data URL and params instead of load and render the full tree
  on the first load. All the ancestors of the resource and their children
  will be loaded on first load, after that it will load the children of the
  opened node if they are not fetched yet.
- Allow others fetch methods in QubitPdo::fetchAll()
- Upgrade jsTree to v3.3.3


lib/QubitPdo.class.php
```diff
@@ -30,13 +30,14 @@ class QubitPdo
...
+  public static function fetchAll($query, $parameters = array(), $options = array())
   {
     $readStmt = self::prepareAndExecute($query, $parameters);
 
-    $fetchedRows = $readStmt->fetchAll(\PDO::FETCH_CLASS);
-    $readStmt->closeCursor();
+    $fetchMode = isset($options['fetchMode']) ? $options['fetchMode'] : PDO::FETCH_CLASS;
+    $fetchedRows = $readStmt->fetchAll($fetchMode);
 
+    $readStmt->closeCursor();
```
sevein, line 40: I would say that unset() does already take care of that indirectly?

js/fullWidthTreeView.js
```diff
@@ -60,12 +60,13 @@
...
     // Scroll to active node
-    var active_node = null;
-    if ( active_node = $('li [selected_on_load]')[0])
-    {
-      active_node.scrollIntoView(true);
-      $('body')[0].scrollIntoView(true);
-    }
+    $('#fullwidth-treeview').bind('ready.jstree', function() {
+      var active_node = $('li[selected_on_load="true"]')[0];
+      if (active_node) {
```
sevein: Should we use `active_node === undefined` and `camelCase` style? i.e. `activeNode`.

sevein: Hey, totally unrelated but can you take this chance to clean up that script to instantiate `$("#fullwidth-treeview")` only once?

sevein: And nitpicking... we don't really type our brackets that way in JavaScript...

sevein (approved):
This is fancy! Minor observations but 👍 .

jraddaoui:
Merged with squash in https://github.com/artefactual/atom/commit/d1a2f45127f06847a7178e5f153914f856c98099

PR #458, opened 2016-11-05
Fix sfConfig settings source value, refs #10256

jraddaoui:
The setting values stored in sfConfig using the '__source' suffix
are now populated even if the setting source culture matches
the current culture.


sevein:
This looks good to me but I'm not sure if I can understand what was the purpose of `AND source.CULTURE <> ?`, can you? Maybe just to speed up the query to avoid the actual join to happen when it wasn't going to be used anyways?

jraddaoui:
It looks like that, when the source value was not used that row was not needed because it's the same as the current row. I'll merge tomorrow and cherry-pick the commit to stable/2.3.x

Thanks!

sevein:
Thank you!

PR #459, opened 2016-11-08
Level of desc filter on export, Refs #10000

sbreker:
This change adds the ability for users to restrict a CSV or XML export to
specific Levels of Description when exporting from the WebUI.

The levels of description are passed to the CSV and XML export jobs in an
array. If the array is empty, it is assumed that all levels of description
will be exported.

I have refactored arXmlExportJob to remove code that was calling the CLI
export task directly. This background job is now modelled off
arInformationObjectCsvExportJob and queries Elasticsearch directly. It
leverages the static export functions from exportBulkBaseTask.

In addition, I have corrected a validation bug in exportOptions.js.


jraddaoui:
👍  Great job!

PR #460, opened 2016-11-08
Add keymap matching EAD XML import, Refs #9992

sbreker:
Adding keymap matching logic to EAD XML import code as a fallback in case
info obj matching on id, title and repo does not match.

Keymap matching for EAD XML records will be based on
- EAD url extracted from XML file
- sourceName - populated by original uploaded file name
- target name of "information_object"


MikeFE:
:+1:

PR #461, opened 2016-11-09
Add export form help text, Refs #10009

sbreker:
Added the help icon and expandable help text box to the new CSV/XML export
form.

### DIFF
--- a/js/exportOptions.js
+++ b/js/exportOptions.js
@@ -17,6 +17,7 @@
       this.$exportSubmit = this.$element.find('input[id="exportSubmit"]');
       this.$exportDiv = this.$element.find('div[id="export-options"]');
       this.$formatSelect = this.$element.find('select[name="format"]');
+      this.$genericHelpIcon = this.$element.find('a.generic-help-icon');
 
       this.init();
       this.listen();
@@ -38,6 +39,7 @@
       this.$includeAllLevels.on('change', $.proxy(this.onIncludeAllLevelsChange, this));
       this.$objectType.on('change', $.proxy(this.onObjectTypeChange, this));
       this.$exportSubmit.on('click', $.proxy(this.onExportSubmit, this));
+      this.$genericHelpIcon.on('click', $.proxy(this.toggleGenericHelp, this));
     },
 
     setDefaults: function()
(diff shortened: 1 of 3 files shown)


sevein:
:+1:

js/exportOptions.js
```diff
@@ -144,6 +146,22 @@
         .append(this.$exportDiv.data('export-alert-message'))
         .prependTo($('#wrapper.container'));
       }
+    },
+
+    toggleGenericHelp: function (e)
+    {
+      e.preventDefault();
+
+      if(this.$genericHelpIcon.toggleClass('open').hasClass('open'))
```
sevein: `if (`

sbreker: Will fix!  :)

js/exportOptions.js
```diff
@@ -144,6 +146,22 @@
...
+      {
+        this.$genericHelpIcon.attr('aria-expanded', true);
+      }
+      else
+      {
+        this.$genericHelpIcon.attr('aria-expanded', false);
+      }
```
sevein: Small suggestion, what do you think?

``` js
var expanded = this.$genericHelpIcon.toggleClass('open').hasClass('open');
this.$genericHelpIcon.attr('aria-expanded', expanded);
```

sevein: (not really important)

sevein:
Sorry I :+1:'ed but on a second thought, should we use the default tooltip description style that we use in our other forms?

sevein:
See `7a3f83bb68a6d29d25f9a56b5a94f4c26a6c2866` (change made in `apps/qubit/modules/settings/templates/pageElementsSuccess.php`). Would that work? It'd make the code much simpler.

sbreker:
Hello! I implemented this to match the help icon on the Adv Search page as requested by Dan.  The implementation of the CSS for the one on the Adv Search page was implemented in a pretty specific way for the dates on that page, so I created a more generic version of the css in the buttons.less file that could potentially be reused.

That commit you reference above just displays the help text as static text in the table cell for me? Is that intended?

sevein:
Yes, that's intended, but I think what you've done is :+1: - please go ahead! I like that you use CSS classes so it's reusable.

js/exportOptions.js
sbreker: I like!  Incorporated...

sevein:
Thank you! 👍

PR #462, opened 2016-11-09
Prevent export of excluded descendant, Refs #10000

sbreker:
CSV exports were incorrectly including levels of description that were
below parent levels of description that had been excluded from the export.

This change corrects this - the CSV export job now leverages the
getDescendantsForExport() method of QubitInformationObject which recursively
checks each level in the nested set hierarchy.


sevein:
:+1:

PR #463, opened 2016-11-14
Enable ACL checks on export, Refs #10000

sbreker:
Historically, ACL checks are disabled for CLI tasks. This applied also to
asynchronous background jobs. CSV and XML export background jobs launched
from the Clipboard required ACL checks to be run. This change allows a
background job to conditionally enable ACL checks.

arBaseJob has two new methods that activate ACL checks when called. This is
optional for each background job so as not to impact existing functionality.
Currently, the only two background jobs that call these methods to activate
ACL checks are arInformationObjectExportJob and arXmlExportJob.

### DIFF
--- a/lib/job/arBaseJob.class.php
+++ b/lib/job/arBaseJob.class.php
@@ -60,9 +60,11 @@ class arBaseJob extends Net_Gearman_Job_Common
     // Net_Gearman_Job_Exception to avoid breaking the worker
     try
     {
+      $this->signIn();
       $this->createJobsDownloadsDirectory();
       $this->runJob($parameters);
       QubitSearch::getInstance()->flushBatch();
+      $this->signOut();
 
       $this->info($this->i18n->__('Job finished.'));
     }
@@ -269,5 +271,33 @@ class arBaseJob extends Net_Gearman_Job_Common
     }
 
     return $success;
+  }
+
+  /**
+   * Set job owner in user Context. ACL checks require this to be set.
+   * Job owner's user is grabbed from the QubitJob instance.
+   * @param int none
+   *
+   * @return null
+   */
+  protected function signIn()
+  {
+    $user = QubitUser::getById($this->job->userId);
+    sfContext::getInstance()->user->signIn($user);
+  }
+
+  /**
+   * Clean up job owner & user Context.
+   *
+   * @param none
+   *
+   * @return null
+   */
+  protected function signOut()
+  {
+    // Need to delete the ACL instance because we are in a gearman worker loop.
+    // Calling destruct() forces a new QubitAcl instance for each job.
+    QubitAcl::destruct();
+    sfContext::getInstance()->user->signOut();
   }
 }
--- a/lib/model/QubitInformationObject.php
+++ b/lib/model/QubitInformationObject.php
@@ -672,7 +672,6 @@ class QubitInformationObject extends BaseInformationObject
   public function getDescendantsForExport($options = array())
   {
     $descendants = array();
-    $fromClipboard = isset($options['params']['fromClipboard']) ? true : false;
     $levels = isset($options['levels']) ? $options['levels'] : array();
     $numLevels = count($levels);
 
@@ -680,8 +679,8 @@ class QubitInformationObject extends BaseInformationObject
     {
       $addCondition = true;
 
-      // If we're not in a CLI or BG job enviroment, check ACL
-      if (!is_using_cli() || $fromClipboard)
+      // If we're not in a CLI enviroment, check ACL
+      if ('cli' != sfContext::getInstance()->getConfiguration()->getEnvironment())
       {
         $addCondition = QubitAcl::check($child, 'read');
       }
--- a/plugins/qbAclPlugin/lib/QubitAcl.class.php
+++ b/plugins/qbAclPlugin/lib/QubitAcl.class.php
@@ -74,6 +74,18 @@ class QubitAcl
   }
 
   /**
+   * Each request is completely unique in PHP however our workers run in a loop
+   * where state is held across jobs. This class was built as a singleton so we
+   * need this extra method to void the instance when needed when running ACL
+   * checks from a worker process.
+   *
+   */
+  public static function destruct()
+  {
+    self::$_instance = null;
+  }
+
+  /**
    * Test user access to the given resource
    *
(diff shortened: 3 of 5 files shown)


sevein:
If the `cli` environment is going to be different from the context of workers, I wonder if we should create a custom `worker` environment. Making the distinction between these two at that level of abstraction may provide us with better way to control behavior in our application, without having to pass extra arguments around, change signatures, etc...

Here is where we decide which environment a particular task is going to use: https://github.com/artefactual/atom/blob/qa/2.4.x/lib/task/jobs/clearJobsTask.class.php#L44. We could add custom settings for a new `worker` environment here: https://github.com/artefactual/atom/blob/qa/2.4.x/apps/qubit/config/settings.yml.tmpl even though I think we could just start with the same settings that `cli` has. In [factories.yml](https://github.com/artefactual/atom/blob/qa/2.4.x/config/factories.yml) we could also pass a new parameter to one of the factories (e.g. `user`) if that was necessary, for example, to bootstrap `myUser` with a new parameter specific to the `worker` environment.

I'm not sure if it's the right approach though. I'm curious to hear thoughts from @mcantelon @MikeFE @jraddaoui.

sbreker:
Ultimately, I was trying to come up with a solution that could be triggered on a case by case basis to avoid triggering ACL checks for all background jobs as I didn't know what the fallout of that would be.

jraddaoui:
@sevein's solution sounds great, if I'm understanding well, we only need to duplicate the current env. for 'cli' calling it 'worker' and use it in the jobWorkerTask, and with that you won't need to pass the backgroundJob option all around.

But we are still going to need to do the log in to share the user in the jobs, right? Maybe that's something that can be done without passing the user id as a param and using the related QubitJob userId instead.

lib/job/arBaseJob.class.php
```diff
@@ -270,4 +270,31 @@ protected function createZipForDownload($path)
...
+  /**
+   * Identify this job as a background job to differentiate it from a standard
+   * CLI task. This is used to trigger ACL checks which normally short-circuit
+   * an allow full access for CLI tasks.
+   *
+   * @param array   Refence to job parameters
+   *
+   * @return null
+   */
+  protected function isBackgroundJob(&$parameters)
```
MikeFE: can we change the function name here since it's a setter? The function name makes me think it's some sort of checker function that returns a bool

sevein: I agree but I think that method is going to be gone very soon. He's giving a try to that `worker` environment instead.

MikeFE: yeah I spoke with Steve and it sounds like this is changing anyway. Didn't know, sorry!

lib/model/QubitInformationObject.php
```diff
@@ -672,18 +672,17 @@ protected function updateNestedSet($connection = null)
...
     {
       $addCondition = true;
 
-      // If we're not in a CLI or BG job enviroment, check ACL
-      if (!is_using_cli() || $fromClipboard)
+      // If we're not in a CLI enviroment, check ACL
+      if ('cli' != sfContext::getInstance()->getConfiguration()->getEnvironment())
       {
-        $addCondition = QubitAcl::check($child, 'read');
+        $addCondition = QubitAcl::check($child, 'read', $options);
```
sevein: What exactly is in `$options` that you need to pass to `QubitAcl`? `levels`? `public`? If any, perhaps you could add a small comment to clarify.

plugins/qbAclPlugin/lib/QubitAcl.class.php
```diff
@@ -73,6 +73,11 @@ public static function getInstance()
     return self::$_instance;
   }
 
+  public static function destruct()
```
sevein, line 83: This method could use some docs to describe our very particular use case, perhaps something like:

```
Each request is completely unique in PHP however our workers run in a loop
where state is held across jobs. This class was built as a singleton so we
need this extra method to void the instance when needed.
```

sevein:
I've added a couple of comments but 👍 !
I'm glad that this is working out but the issue proves that running long-running workers off our codebase is definitely not what we built this for originally and there are big chances that we'll need to keep patching things to fix new bugs unless we implement new workers differently. Nothing stop us to implement the next worker ability in a different platform, language, framework, etc...

sevein (changes requested):
I've added a couple of comments but 👍 !
I'm glad that this is working out but the issue proves that running long-running workers off our codebase is definitely not what we built this for originally and there are big chances that we'll need to keep patching things to fix new bugs. Nothing stop us to implement the next worker ability in a different platform, language, framework, etc...

lib/model/QubitInformationObject.php
sbreker: No longer necessary.  Missed removing it when I was revising the work.

plugins/qbAclPlugin/lib/QubitAcl.class.php
sbreker: I added a comment at the call but missed adding one here.  Good idea.

lib/model/QubitInformationObject.php
```diff
@@ -672,16 +672,15 @@ protected function updateNestedSet($connection = null)
...
     foreach ($this->getChildren()->orderBy('lft') as $child)
     {
       $addCondition = true;
 
-      // If we're not in a CLI or BG job enviroment, check ACL
-      if (!is_using_cli() || $fromClipboard)
+      // If we're not in a CLI enviroment, check ACL
+      if ('cli' != sfContext::getInstance()->getConfiguration()->getEnvironment())
       {
         $addCondition = QubitAcl::check($child, 'read');
```
sbreker, line 685: Removed $options array.

sevein (approved):

lib/job/arBaseJob.class.php
```diff
@@ -270,4 +272,32 @@ protected function createZipForDownload($path)
 
     return $success;
   }
+
+  /**
+   * Set job owner in user Context. ACL checks require this to be set.
+   * Job owner's user is grabbed from the QubitJob instance.
+   * @param int none
```
MikeFE, line 279: int none? :P

sbreker: I've removed this.  😄

PR #464, opened 2016-11-14
Display import source archival rec, Refs #10193

sbreker:
If the user is logged in and viewing an archival record, display any source-
names in the keymap table associated with that record in the administrative
section of that record.


sevein:
:+1:

PR #466, opened 2016-11-17
CSV import keep-digital-objects, Refs #10548

sbreker:
Add new CLI option '--keep-digital-objects'. Using this option with the
--update option will stop all updates to existing digital objects. New digital
objects will still be added however.


MikeFE:
:+1:

PR #467, opened 2016-11-17
Dev/issue 10579 import and transform tweaks

mcantelon:
A couple of small changes to make it easier to write CSV transformations:
* Made keymap entry fetching method static
* Added '"copy" method (to copy a value from one column to another)

### DIFF
--- a/lib/QubitFlatfileImport.class.php
+++ b/lib/QubitFlatfileImport.class.php
@@ -2156,7 +2156,7 @@ class QubitFlatfileImport
    *
    * @return stdClass  result object
    */
-  public function fetchKeymapEntryBySourceAndTargetName($sourceId, $sourceName, $targetName)
+  public static function fetchKeymapEntryBySourceAndTargetName($sourceId, $sourceName, $targetName)
   {
     $query = "SELECT target_id, id FROM keymap
       WHERE source_id=? AND source_name=? AND target_name=?";
--- a/lib/QubitFlatfileImport.class.php
+++ b/lib/QubitFlatfileImport.class.php
@@ -298,6 +298,14 @@ class QubitFlatfileImport
   }
 
   /**
+   * Copy one column value to another column in internal representation of current row
+   */
+  function copy($sourceColumn, $destinationColumn)
+  {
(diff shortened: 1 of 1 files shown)


lib/QubitFlatfileImport.class.php
```diff
@@ -2148,7 +2156,7 @@ public function createRightAndRelation($options)
    *
    * @return stdClass  result object
    */
-  public function fetchKeymapEntryBySourceAndTargetName($sourceId, $sourceName, $targetName)
+  public static function fetchKeymapEntryBySourceAndTargetName($sourceId, $sourceName, $targetName)
```
MikeFE, line 2159: Don't forget to switch the $self-> and $this-> to QubitFlatfileImport:: instead when we actually call it. There's a few places:

```
$ git grep -n fetchKeymapEntryBySourceAndTargetName
lib/QubitFlatfileImport.class.php:944:    $mapEntry = $this->fetchKeymapEntryBySourceAndTargetName(
lib/QubitFlatfileImport.class.php:1882:  public function fetchKeymapEntryBySourceAndTargetName($sourceId, $sourceName, $targetName)
lib/task/import/csvImportTask.class.php:201:      $keyMapEntry = QubitFlatfileImport::fetchKeymapEntryBySourceAndTargetName(
lib/task/import/csvImportTask.class.php:611:            if ($mapEntry = $self->fetchKeymapEntryBySourceAndTargetName(
lib/task/import/csvImportTask.class.php:645:        $keymap = $self->fetchKeymapEntryBySourceAndTargetName(
lib/task/import/csvImportTask.class.php:850:            $accessionMapEntry = $self->fetchKeymapEntryBySourceAndTargetName(
```

(apparently we were calling it once as a static already mistakenly, heh)

MikeFE: Apparently according to http://stackoverflow.com/a/7983863 you _can_ actually call static methods like that. My bad. :)

MikeFE: :+1:

mcantelon: Danke!

mcantelon:
Merged.

PR #468, opened 2016-11-18
Exclude types when indexing, refs #10581

mcantelon:
Added an option, to the CLI task for search index populations, to exclude one or more AtoM document types.

One question: should we have an option to display the types that will be index or just do it by default without the ability to toggle it?

### DIFF
--- a/lib/task/search/arPopulateTask.class.php
+++ b/lib/task/search/arPopulateTask.class.php
@@ -30,6 +30,7 @@ class arSearchPopulateTask extends sfBaseTask
     $this->addOptions(array(
       new sfCommandOption('application', null, sfCommandOption::PARAMETER_OPTIONAL, 'The application name', 'qubit'),
       new sfCommandOption('env', null, sfCommandOption::PARAMETER_REQUIRED, 'The environment', 'cli'),
+      new sfCommandOption('exclude-types', null, sfCommandOption::PARAMETER_OPTIONAL, 'Exclude document type(s) (command-separated) from indexing'),
       new sfCommandOption('verbose', 'v', sfCommandOption::PARAMETER_NONE, 'If passed, progress is displayed for each object indexed')));
 
     $this->namespace = 'search';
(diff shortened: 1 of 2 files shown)


MikeFE:
I'd just display it by default if that's easier, simpler is better. :)

lib/task/search/arPopulateTask.class.php
```diff
@@ -48,6 +50,6 @@ public function execute($arguments = array(), $options = array())
 
     new sfDatabaseManager($this->configuration);
 
-    QubitSearch::getInstance()->populate();
+    QubitSearch::getInstance()->populate($options);
```
sevein: Can we explicitly pass here the options we want here to avoid undefined behaviour? This reminds me to the use of `locals()` in Django views, that we try to avoid.

mcantelon: Yeah, probably a good idea.

plugins/arElasticSearchPlugin/lib/arElasticSearchPlugin.class.php
```diff
@@ -280,20 +315,29 @@ public function populate()
 
     foreach ($this->mappings as $typeName => $typeProperties)
     {
-      $camelizedTypeName = sfInflector::camelize($typeName);
-      $className = 'arElasticSearch'.$camelizedTypeName;
+      if (!count($this->excludeTypes) || !in_array(strtolower($typeName), $this->excludeTypes))
```
MikeFE: could you remove the !count($this->excludeTypes) part to this if statement? Since excludeTypes will be an empty array and in_array will return false in that case anyway? 

it just took my head for a spin a bit when I saw if (!count($this->excludeTypes) || ...) then right after that if (count($this->excludeTypes)) :smiley:

mcantelon: Sounds good.

mcantelon: Actually it doesn't really need to be an object property too... will just make it a standard variable.

mcantelon:
Updated... made display of document types that'll be indexed happen all the time.

MikeFE:
:+1:

sevein:
Mike, could you please add an example to `$this->detailedDescription`? It's hard to figure what kind of format you expect unless you look at the code. Other than that, 👍  me too!

mcantelon:
Awesome... thanks folks.

mcantelon:
Merged.

PR #469, opened 2016-11-18
Generate archival description reports asynchronously

MikeFE:
Redmine: https://projects.artefactual.com/issues/10309

This PR takes the old AtoM archival description report generation code and makes it asynchronous. Reports can be generated into stand alone HTML or CSV documents. Furthermore, it allows the user to access previously generated reports if they have permission.

Notes:

- The HTML reports use inline CSS since the HTML documents are meant to be stand alone files the user can download / view in their browser later if they want. Referring to an external site's CSS didn't seem very reliable in this case, so some minimal inline CSS was provided at the top of these documents.


sevein:
I haven't looked at your PR in detail but the area where you render the reports caught my attention. If you don't make use of just plain `__()` I think that the messages won't be reachable by `i18n:extract`. You may have to alias the function and verify that `i18n:extract` is happy with it, I've never tried such thing myself. But we need these messages to be translatable.

Something that you could try is to repurpose Symfony's view-controller machinery to get the rendering working in the worker so it works similarly to how we do it generally. You probably have no time left for such change but I'd recomment to take a look at [how the execution filter](https://github.com/artefactual/atom/blob/c705d3804698361c3b77f26d2e37fcf6b450a757/vendor/symfony/lib/filter/sfExecutionFilter.class.php#L119-L164) does the rendering. We always init the context in our CLI tasks so I wonder if that stuff is just as easy as it looks without much extra bootstrapping. If that stuff is going to take you a couple of hours I think it's totally worth the extra effort.

apps/qubit/modules/informationobject/actions/boxLabelAction.class.php
```diff
@@ -0,0 +1,80 @@
...
+        'objectId' => $this->resource->id,
+        'reportType' => 'boxLabel',
+        'reportTypeLabel' => $this->context->i18n->__('Box label'),
+        'reportFormat' => $this->form->format->getValue(),
+    );
+
+    QubitJob::runJob('arGenerateReportJob', $params);
+
+    $reportsUrl = url_for(array($this->resource, 'module' => 'informationobject', 'action' => 'reports'));
+    $message = $this->context->i18n->__('Report generation has started, please check the <a href="'.
```
mcantelon: Nitpick: You could use a %1 placeholder rather than string concatenation.

apps/qubit/modules/informationobject/actions/itemOrFileListAction.class.php
```diff
@@ -0,0 +1,148 @@
...
+        'sortBy' => $this->form->sortBy->getValue(),
+        'reportFormat' => $this->form->format->getValue(),
+        'includeThumbnails' => $includeThumbnails
+    );
+
+    QubitJob::runJob('arGenerateReportJob', $params);
+
+    $reportsUrl = url_for(array($this->resource, 'module' => 'informationobject', 'action' => 'reports'));
+    $message = $this->context->i18n->__('Report generation has started, please check the <a href="'.
+                                        $reportsUrl.'">reports</a> page again soon.');
```
mcantelon: Nitpick: could use placeholder here too instead of string concat.

sevein: Agree, that message cannot be extracted, they need to be string literals.

apps/qubit/modules/informationobject/actions/reportsAction.class.php
```diff
@@ -25,6 +25,70 @@ class InformationObjectReportsAction extends sfAction
...
+        if (file_exists($path))
+        {
+          if (!sfContext::getInstance()->user->isAuthenticated() &&
+              in_array($type, array('storageLocations', 'boxLabel')))
+          {
+            continue;
+          }
+
+          $this->existingReports[] = array(
+            'path' => sfConfig::get('siteBaseUrl').DIRECTORY_SEPARATOR.$path,
```
sevein: I think you shouldn't use `DIRECTORY_SEPARATOR` here but just the plain `/` literal.

apps/qubit/modules/informationobject/templates/boxLabelCriteria.php
```diff
@@ -5,23 +5,19 @@
 <?php end_slot() ?>
 
 <?php slot('title') ?>
-  <h1><?php echo __('File list - report criteria') ?></h1>
+  <h1><?php echo __($type.' - report criteria') ?></h1>
```
sevein: This is not going to work with i18n, we need to use gettext replacements.

apps/qubit/modules/informationobject/templates/itemOrFileListCriteria.php
```diff
@@ -0,0 +1,37 @@
+<?php decorate_with('layout_2col') ?>
+
+<?php slot('sidebar') ?>
+  <?php include_component('informationobject', 'contextMenu') ?>
+<?php end_slot() ?>
+
+<?php slot('title') ?>
+  <h1><?php echo __($type.' list - report criteria') ?></h1>
```
sevein: Same here.

apps/qubit/modules/informationobject/templates/itemOrFileListError.php
```diff
@@ -1,9 +1,9 @@
 <div class="content">
-    <h1><?php echo __('Item list') ?></h1>
+    <h1><?php echo __($type.' list') ?></h1>
```
sevein: Same here.

apps/qubit/modules/informationobject/templates/itemOrFileListError.php
```diff
@@ -1,9 +1,9 @@
...
 
     <h1 class="label"><?php echo __('No results') ?></h1>
 
-    <p><?php echo __('Oops, we couldn\'t find any item level descriptions.') ?></p>
+    <p><?php echo __("Oops, we couldn't find any ".strtolower($type).' level descriptions.') ?></p>
```
sevein: Same here.

apps/qubit/modules/informationobject/templates/storageLocationsCriteria.php
```diff
@@ -5,27 +5,19 @@
 <?php end_slot() ?>
 
 <?php slot('title') ?>
-  <h1><?php echo __('Item list - report criteria') ?></h1>
+  <h1><?php echo __($type.' - report criteria') ?></h1>
```
sevein: Same here.

apps/qubit/modules/informationobject/templates/storageLocationsSuccess.php
```diff
@@ -1,44 +1,57 @@
...
+        </tr>
+      <?php endforeach; ?>
+    </tbody>
+  </table>
 
-<div id="result-count">
-  <?php echo __('Showing %1% results', array('%1%' => $foundcount)) ?>
-</div>
+  <div id="result-count">
+    <?php echo $this->i18n->__('Showing %1% results', array('%1%' => count($results))) ?>
```
sevein, line 54: :+1:

apps/qubit/modules/informationobject/templates/storageLocationsCriteria.php
MikeFE: OK, got it.

sevein: hahaha sorry

MikeFE: no worries! It'll save time when I come fix it soon if I know where all I did the mistake hehe

apps/qubit/modules/informationobject/actions/reportsAction.class.php
MikeFE: I agree, I read an article after saying it's pointless unless you're using it to explode a path a system call passes back to you. Missed this one!

MikeFE:
Had Mike C take one final look after I fixed the i18n strings to use placeholders.

PR #470, opened 2016-11-21
Repository CSV import fixes, refs #10017

jraddaoui:
- Do not create contact info or maintenance note if there is no data
- Rename phone and notes columns to telephone and maintenanceNote
- Avoid creating duplicated contact info and maintenance note while updating
- Avoid adding/updating blank fields in contact info and maintenance note


sevein:
:+1:

PR #472, opened 2016-11-24
Restore finding aid settings form, Refs #10259

sbreker:
This commit restores lib/form/SettingsFindingAidForm.class.php.

The Admin->Settings->Finding Aid settings page will now work as expected.


sevein:
Interesting that we missed that!

:+1:

PR #474, opened 2016-11-29
Use EasyRDF in sfSkosPlugin, refs #10621

sevein:
Redmine: https://projects.artefactual.com/issues/10621


lib/model/QubitJob.php
```diff
@@ -117,21 +117,24 @@ public function getCompletionDateString()
...
   public function getStatusString()
   {
+    $i18n = sfContext::getInstance()->i18n;
+    $unknown = $i18n->__('Unknown');
+
     if (!isset($this->statusId))
     {
-      return 'unknown';
+      return $unknown;
     }
```
jraddaoui: Not a big deal and I'm not completely sure, but I think that if the statusId is not set it will fall in the switch default case.

lib/job/arJobLogger.class.php
```diff
@@ -0,0 +1,57 @@
...
+    $this->options = $options;
+
+    if (isset($this->options['level']))
+    {
+      $this->setLogLevel($this->options['level']);
+    }
+
+    if (!isset($this->options['job']))
+    {
+      throw new sfException('Missing parameter job');
```
jraddaoui: I'm not a native but I think "Missing job parameter" sounds better.

plugins/sfSkosPlugin/lib/sfSkosPlugin.class.php
```diff
@@ -17,359 +17,315 @@
  * along with Access to Memory (AtoM).  If not, see <http://www.gnu.org/licenses/>.
  */
 
+class sfSkosPluginException extends Exception {}
+
```
jraddaoui: I'd expect to find this in a different file, like QubitApiException.class.php

plugins/sfSkosPlugin/lib/sfSkosPlugin.class.php
```diff
@@ -17,359 +17,315 @@
...
-    // XPath selector for expanded RDF syntax
-    $rdfsel = "rdf:Description[rdf:type[@rdf:resource='http://www.w3.org/2004/02/skos/core#Concept']]";
+  public function __construct($taxonomyId, $options = array())
+  {
+    $this->logger = isset($options['logger']) ? $options['logger'] : new sfNoLogger(new sfEventDispatcher);
 
-    // Get all concepts
-    $concepts = $skos->xpath->query("skos:Concept | $rdfsel");
+    $this->i18n = sfContext::getInstance()->i18n;
+    $this->graph = new EasyRdf_Graph;
```
jraddaoui: We could initiate ```$this->graph``` after the checks bellow

plugins/sfSkosPlugin/lib/sfSkosPlugin.class.php
```diff
@@ -17,359 +17,315 @@
...
+    $term = new QubitTerm;
+    $term->parentId = $parentId;
+    $term->taxonomyId = $this->taxonomy->id;
+
+    $this->setPrefLabel($term, $concept);
+    $this->setAltLabels($term, $concept);
+    $this->setScopeNote($term, $concept);
+    $this->setUriSourceNote($term, $concept);
+
+    // Map dc.coverage to term.code for place terms. Hacky Hackerton was here.
```
jraddaoui, line 195: :smile:

plugins/sfSkosPlugin/lib/sfSkosPlugin.class.php
```diff
@@ -17,359 +17,315 @@
...
 
-      if ($parentUri instanceof DOMAttr)
-      {
-        $this->terms[$uri->nodeValue]->parent = $this->terms[$parentUri->nodeValue];
-        $this->terms[$uri->nodeValue]->save();
-      }
+      $otherName = new QubitOtherName;
+      $otherName->typeId = QubitTerm::ALTERNATIVE_LABEL_ID;
+      $otherName->sourceCulture = $item->getLang();
+      $otherName->__set('name', $item->getValue(), array('culture' => $lang));
```
jraddaoui: No need to change it, but I think you can use ```$otherName->setName($item->getValue(), array('culture' => $lang));``` for a little more magic in cases like this

plugins/sfSkosPlugin/lib/sfSkosPlugin.class.php
```diff
@@ -17,359 +17,315 @@
...
+    {
+      foreach ($c->allResources('skos:related') as $r)
       {
-        if (
-          $r['subject'] == $objectUri->nodeValue && $r['object'] == $subjectUri->nodeValue ||
-          $r['subject'] == $subjectUri->nodeValue && $r['object'] == $objectUri->nodeValue)
-        {
-          continue 2;
-        }
+        $relations->insert($c->get('atom:id')->getValue(), $r->get('atom:id')->getValue());
```
jraddaoui, line 156: Love it!

plugins/sfSkosPlugin/lib/sfSkosPlugin.class.php
```diff
@@ -17,359 +17,315 @@
...
+    if (0 < count($this->unsupportedLanguages))
+    {
+      $this->errors[] = $this->i18n->__('The following languages are used in the dataset imported but not supported by AtoM: %1%', array('%1%' => implode(",", array_keys($this->unsupportedLanguages))));
     }
 
-    return $this;
+    // Re-index parent term so numberOfDescendants reflects the changes
+    if (QubitTerm::ROOT_ID != $this->parent->id)
+    {
+      QubitSearch::getInstance()->update($this->parent);
```
jraddaoui, line 180: Not a big deal, as the terms are not so big in the ES index, but we could use partial updates in here.

plugins/sfSkosPlugin/modules/sfSkosPlugin/actions/importAction.class.php
```diff
@@ -57,39 +88,78 @@ public function execute($request)
...
 
-          $this->skos = sfSkosPlugin::parse($doc, array('taxonomy' => $this->taxonomy->id, 'parent' => $this->parent));
+          return;
         }
+
+        $this->getUser()->setFlash('notice', $this->context->i18n->__('Import file initiated. Check %1%job %2%%3% to view the status of the import.', array(
+          '%1%' => sprintf('<a href="%s">', $this->context->routing->generate(null, array('module' => 'jobs', 'action' => 'report', 'id' => $job->id))),
+          '%2%' => $job->id,
+          '%3%' => '</a>'
+        )), array('persist' => false));
```
jraddaoui, line 136: Could you check the persist option in the other flash alerts we set? e.g.: https://github.com/artefactual/atom/blob/qa/2.4.x/apps/qubit/modules/informationobject/actions/updatePublicationStatusAction.class.php#L106

plugins/sfSkosPlugin/test/unit/uniqueTest.php
```diff
@@ -0,0 +1,127 @@
+<?php
```
jraddaoui: I'd name it 'uniqueRelationsTest.php'

jraddaoui (commented):
Impressive! I've added a few minor comments but it looks great overall.

lib/job/arJobLogger.class.php
fiver-watson: Agree! :D

lib/model/QubitJob.php
sevein: You're right! I guess this is the result of having switched to Python for a while now, where `AttributeError` would be raised if you try to access to a member of an object that is not defined. I'll fix it!

lib/job/arJobLogger.class.php
sevein: Hehehe thank you!

plugins/sfSkosPlugin/lib/sfSkosPlugin.class.php
sevein: That'd be cool but where would you take the new value from? It'd be weird to instantiate a `arElasticSearchTermPdo` only for that reason. If I make the calculation then I'm duplicating code. If I use the ORM I'd need to hydrate the object again I believe?
It'd be very nice if the `partialUpdate` method was able to generate the new value for me in case I don't give him once...

plugins/sfSkosPlugin/lib/sfSkosPlugin.class.php
sevein: I actually didn't write that! I think it was Jack!

plugins/sfSkosPlugin/modules/sfSkosPlugin/actions/importAction.class.php
sevein: On that particular example I think it's good that it's persisted so it survives to the redirection, right? In any case, there is a total of 16 uses of `setFlash`... it sounds like we should do that on a separate pull request?

jraddaoui: Sounds good!

jraddaoui:
:+1:

PR #475, opened 2016-11-30
Option: use DB for CSV import match, refs #10624

mcantelon:
Redmine: https://projects.artefactual.com/issues/10624

Added option, to CSV import CLI tool, to use the database, rather than the
ElasticSearch index, when using import modes that check if the row matches
existing data.

I added this as I'm working on an import where I am importing on top of an
existing database and it takes a long time to index the existing data so I
don't want to have to do that before indexing just so the matching will work.


sevein:
If this is not happening, should we close it?

PR #477, opened 2016-11-30
CSV import using existing parent IDs, refs #10626

mcantelon:
Added logic to CSV information object import to, if a parent ID in a row
doesn't correspond to a keymap entry, check to see if the parent is present in
existing information object data.

This adds the ability to import children of information objects that have been
created by entry through the web UI (rather than only being able to import
children of information objects that have been created by importing data).

### DIFF
--- a/lib/task/import/csvImportTask.class.php
+++ b/lib/task/import/csvImportTask.class.php
@@ -624,9 +624,13 @@ EOF;
             {
               $parentId = $mapEntry->target_id;
             }
+            else if (null !== QubitInformationObject::getById($self->rowStatusVars['parentId']))
+            {
+              $parentId = $self->rowStatusVars['parentId'];
+            }
             else
             {
-              $error = sprintf('legacyId %s: could not find parentId %s in key_map table. Setting parent to root...',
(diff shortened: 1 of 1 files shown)


lib/task/import/csvImportTask.class.php
```diff
@@ -624,9 +624,12 @@ public function execute($arguments = array(), $options = array())
             {
               $parentId = $mapEntry->target_id;
             }
-            else
+            elseif (null !== QubitInformationObject::getById($self->rowStatusVars['parentId']))
```
MikeFE: both work but we usually put a space between else if

lib/task/import/csvImportTask.class.php
```diff
@@ -624,9 +624,12 @@ public function execute($arguments = array(), $options = array())
...
             {
-              $error = sprintf('legacyId %s: could not find parentId %s in key_map table. Setting parent to root...',
+              $parentId = $self->rowStatusVars['parentId'];
+            }
+            else {
```
MikeFE: can you pop that { onto its own line?

MikeFE:
:+1:

mcantelon:
Merged.

PR #478, opened 2016-12-08
CSV import missing fields in php7, Refs #10633

sbreker:
This change corrects an issue discovered when running AtoM under PHP 7.x.

Under PHP 7.x, the CSV import from the command line would not import all
fields from the CSV. This fix has been tested with PHP 5.5.9 and does
not have any side effects.

### DIFF
--- a/lib/task/import/csvImportTask.class.php
+++ b/lib/task/import/csvImportTask.class.php
@@ -1121,13 +1121,13 @@ EOF;
       return (substr_count($text, '|')) ? '* '. str_replace("|", "\n* ", $text) : $text;
     };
 
-    $import->addColumnHandler('levelOfDescription', function(&$self, $data)
+    $import->addColumnHandler('levelOfDescription', function($self, $data)
     {
       $self->object->setLevelOfDescriptionByName($data);
     });
 
     // Map value to taxonomy term name and take note of taxonomy term's ID
-    $import->addColumnHandler('radGeneralMaterialDesignation', function(&$self, $data)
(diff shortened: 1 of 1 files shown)


lib/QubitFlatfileImport.class.php
```diff
@@ -1206,7 +1206,7 @@ protected function rowProcessingBeforeSave($row)
         // otherwise, if column is data and a handler for it is set, use it
         call_user_func_array(
           $self->handlers[$columnName],
-          array($self, $value)
+          array(&$self, $value)
```
sevein: I don't understand why in first place the function where this is happening prefers the object to be passed by reference if we already get an object identifier by default. I think I would try instead to update the func declaration to `function($self, $index, $columnName, $value)` and leave the array declaration as `array($self, $value)`. That would make the code much easier to understand IMO. More here: http://php.net/manual/en/language.oop5.references.php,

> A PHP reference is an alias, which allows two different variables to write to the same value. As of PHP 5, an object variable doesn't contain the object itself as value anymore. It only contains an object identifier which allows object accessors to find the actual object. When an object is sent by argument, returned or assigned to another variable, the different variables are not aliases: they hold a copy of the identifier, which points to the same object.

I think we should embrace the fact that in PHP function arguments are always going to be passed by value, where in objects these are going to be copies of the identifier values, not copies of the actual objects values.

sbreker: Makes sense.  I have completed the change to remove the references.  Tested on both PHP7 and 5.5.

sevein:
👍

PR #479, opened 2016-12-12
CSV import accession mapping, Refs #10602

sbreker:
This change addresses an error that can show up during CSV import when the
column 'accessionNumber' is included in the CSV import file.

Previously, the import program would attempt to find the accession identifier
in the keymap table, and if that failed it would create a new accession
record with that identifier. In the case where the keymap record could not be
found, but the accession record already existed, an error would display and
the import would halt.

I have changed the logic so that the import will try to find a corresponding
keymap record to use and if that fails, it will try to directly find an
accession record in the database. If keymap search is successful, that
accession record will be used. If the secondary lookup against the database
is successful, then that accession record will be used. If both keymap and
direct lookup fail, a new accession record will be created.

### DIFF
--- a/lib/task/import/csvImportTask.class.php
+++ b/lib/task/import/csvImportTask.class.php
@@ -869,20 +869,26 @@ EOF;
             // If no entry found, create accession and entry
             if (!$accessionMapEntry)
             {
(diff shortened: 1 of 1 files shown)


mcantelon:
Hi Steve... looks good to me!

PR #480, opened 2016-12-13
PDF derivatives issue 16.04, Refs #10675

sbreker:
Redmine: https://projects.artefactual.com/issues/10675

This change corrects an issue discovered under Ubuntu 16.04 where the
ImageMagick 'identify' utility incorrectly indicates the mime type of a PDF
file is PBM. Detecting mime types of any file type is now delegated to
the PHP fileinfo library instead of using the 'identify' command.

### DIFF
--- a/plugins/sfThumbnailPlugin/lib/sfImageMagickAdapter.class.php
+++ b/plugins/sfThumbnailPlugin/lib/sfImageMagickAdapter.class.php
@@ -195,7 +195,7 @@ class sfImageMagickAdapter
 
         $this->sourceWidth = $width;
         $this->sourceHeight = $height;
-        $this->sourceMime = $this->mimeMap[strtolower($type)];
+        $this->sourceMime = $this->getMimeType($image);
       }
     }
     else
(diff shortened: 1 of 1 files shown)


plugins/sfThumbnailPlugin/lib/sfImageMagickAdapter.class.php
```diff
@@ -217,6 +217,25 @@ public function loadFile($thumbnail, $image)
...
+   *
+   * @param string  file we want the mime type for
+   *
+   * @return string  Mime type
+   */
+  public static function getMimeType($file)
+  {
+    // Use fileinfo to figure out file mimetype.
+    $file_info = new finfo(FILEINFO_MIME);
+    $mimeInfo = $file_info->buffer(file_get_contents($file));
```
sevein: Wouldn't be more efficient to use `finfo::file` or `finfo_file` so we don't have to load the whole file into a string?

sbreker: Good catch.  I have updated and tested with big files and there is a noticeable performance increase.

sevein:
👍

PR #481, opened 2016-12-14
Prevent note duplication on update, refs #10150

MikeFE:
When using match-and-update mode for csv:import, prevent duplicate notes from
being created.


mcantelon:
Looks good to me!

PR #482, opened 2016-12-16
Create always keymap on IO CSV import, refs #10144

jraddaoui:
When importing files with the same name and legacy ids if a new entry is not
created in the keymap table it will always get the resource from the previous
file. Make sure to get the last entry while checking.


mcantelon:
Looks good to me!

PR #484, opened 2016-12-20
Correct keymap matching EAD XML import, Refs #9992

sbreker:
Dropping clause '$importSchema == 'ead'' as it does not work ($importSchema
is not available in this function) and is redundant since eadUrl will only be
set when loading an EAD XML file.


sevein:
👍

sbreker:
Thanks!

PR #486, opened 2016-12-21
Fix event place relation edit, refs #10708

jraddaoui:
- Fix place population in first load of edit event dialog
- Change how the object term relation is created/updated/deleted


sevein:
👍

PR #487, opened 2016-12-22
renderDate leading zeroes on day, Refs #10641

sbreker:
This change addresses an issue with date formatting where if the day had a
leading zero it would be formatted incorrectly when displayed on the screen.
e.g. 2016-12-03 would display as 2016-123

Note that this issue would only occur when the month was a two digit month as
in the example above.

This corrects the issue and forces single digit month and date to display
with a leading zero. Including the '0*' within the parentheses
will force any leading zero to be part of the replacement string sub-
pattern.
e.g. 2016-6-9 will display as 2016-06-09
     2016-12-6 will display as 2016-12-06

### DIFF
--- a/lib/Qubit.class.php
+++ b/lib/Qubit.class.php
@@ -39,8 +39,11 @@ class Qubit
   // support zero day or zero month
   public static function renderDate($value)
   {
-    // Natural number
-    $natural = '([1-9]\d*)';
+    // Natural number - This will not strip leading zeroes.
+    // Note that this will NOT add leading zeroes if they are not present.
+    // e.g. 2016-12-08 will be displayed as 2016-12-08
+    //      2016-12-8 will be displayed as 2016-12-8
+    $natural = '(0*[1-9]\d*)';
 
     // Zero and separator, trim leading zeros
     $zero = '0*(0[-\/])';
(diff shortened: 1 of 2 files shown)


sevein:
That function is used in more than 30 files across the application. Would you be willing to add a simple test to verify that we're not introducing regressions? I know it's late but probably not a bad idea to define the expectations that we have on this function. This is a quick example, feel free to add more tests cases to the array: https://gist.github.com/sevein/56a4c06c733b5fc2c2f61cb0f7c866b4. You only need to run `php test/unit/qubitTest.php` from the command line to see the results.

lib/Qubit.class.php
```diff
@@ -51,9 +51,9 @@ public static function renderDate($value)
     // Trim leading zeros from year, month, and day, and trailing zero day and
     // month
     $pattern = "/0*(\d+)(?:([-\/])$natural(?:([-\/])$natural
-          |$trim)
-        |([-\/])$zero$natural
-        |$trim$trim)/x";
+      |$trim)
+      |([-\/])$zero$natural
+      |$trim$trim)/x";
```
sevein, line 58: Not necessary but if you want to improve the indentation maybe you should consider a more radical approach like:

```
/
  0*
  (\d+)              # Group 1
  (?:
      ([-\/])        # Group 2
      $natural       # Group 3
      (?:
        ([-\/])      # Group 4
        $natural     # Group 5
      |
        $trim
      )
    |
      ([-\/])        # Group 6
      $zero          # Group 7
      $natural       # Group 8
    |
      $trim
      $trim
  )
/x
```

Based on https://goo.gl/3QgAHH.

sbreker:
Good idea re: the unit tests.  Added.

test/unit/qubitTest.php
```diff
@@ -45,3 +45,38 @@ public function getPathInfoPrefix()
 
 $t->is(Qubit::pathInfo('/aaa/bbb/ccc/ddd'), '/aaa/bbb/ccc/ddd',
   '"::pathInfo()" without prefix');
+
+
+/*
+ * Qubit::renderDate
++ */
```
sevein: That `+` is not supposed to be there :)

sevein:
I've fixed it and merged it for you.

sevein:
Please remember next time to file a public ticket first!

PR #488, opened 2016-12-22
SKOS: fix concept matcher, refs #10621

sevein:
Redmine: https://projects.artefactual.com/issues/10621

getRootConcepts() makes an effort to return only top concepts so we can harvest
the hierarchy from top to bottom by looking at skos:ConceptScheme. This commit
fixes a case where this element had no skos:hasToConcept properties, causing
the importer to believe that there were no concepts available.


sbreker:
👍

PR #489, opened 2016-12-27
Ensure Mime property set early, Refs #10675

sbreker:
This change corrects an issue where the mime class property was not yet
set when getExtract() was called in sfThumbnailPLugin.

The check to determine if the plugin should use pdfinfo to extract pages
and get a page count was made dependent on mime type instead of file
extension in 10675. An issue was uncovered where the mime type was not
yet available when the decision to use pdfinfo needed to be made. This
issue has been corrected by setting mime type earlier in the code.


sevein:
👍

PR #492, opened 2017-01-05
Institutional Scoping migration conf, Refs #10733

sbreker:
Institutional Scoping enhancement for AtoM. When activated, the feature
will allow users to be scoped to a specific repository such that browsing
and searching will restict a user to a specific repository. Repository
theming will apply to pages when scoped to give the user visual cues
indicating they are browsing with a repository. A new search block will
be present on the left hand side from which browse and search will be done.

This is the first of several commits to pull this into 2.4.x

This commit includes:
- migration 149 to set up new settings and menu items
- settings and menu item config in menus.yml and settings.yml
- Global Settings page change to add new option to "Enable insitutional
scoping"

### DIFF
--- a/apps/qubit/modules/actor/actions/browseAction.class.php
+++ b/apps/qubit/modules/actor/actions/browseAction.class.php
@@ -81,6 +81,19 @@ class ActorBrowseAction extends DefaultBrowseAction
   {
     parent::execute($request);
 
+    if (isset($request->repos) && ctype_digit($request->repos))
+    {
+      $this->repos = QubitRepository::getById($request->repos);
+
+      // Add repo to the user session as realm
+      $this->context->user->setAttribute('search-realm', $request->repos);
+    }
+    elseif (sfConfig::get('app_enable_institutional_scoping'))
+    {
+      // Remove search realm
+      $this->context->user->removeAttribute('search-realm');
+    }
+
     if (1 === preg_match('/^[\s\t\r\n]*$/', $request->subquery))
     {
       $this->search->queryBool->addMust(new \Elastica\Query\MatchAll());
--- a/apps/qubit/modules/actor/actions/browseAction.class.php
+++ b/apps/qubit/modules/actor/actions/browseAction.class.php
@@ -88,7 +88,7 @@ class ActorBrowseAction extends DefaultBrowseAction
       // Add repo to the user session as realm
       $this->context->user->setAttribute('search-realm', $request->repos);
     }
-    elseif (sfConfig::get('app_enable_institutional_scoping'))
+    else if (sfConfig::get('app_enable_institutional_scoping'))
     {
       // Remove search realm
       $this->context->user->removeAttribute('search-realm');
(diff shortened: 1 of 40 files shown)


lib/model/QubitMenu.php
```diff
@@ -52,7 +52,15 @@ public function getPath($options = array())
     $aliases = array(
       '%profile%' => sfContext::getInstance()->routing->generate(null, array('module' => 'user', 'slug' => sfContext::getInstance()->user->getUserSlug())),
       '%currentId%' => sfContext::getInstance()->request->id,
-      '%currentSlug%' => @sfContext::getInstance()->request->getAttribute('sf_route')->resource->slug
+      '%currentSlug%' => @sfContext::getInstance()->request->getAttribute('sf_route')->resource->slug,
+
+      // 'currentRealm' is used by menu items on the Institutional Block as part of the enable_institutional_scoping feature.
+      // Try using search-realm if available, else try resource->id from sf_route, else try repos param on request.
+      '%currentRealm%' => (@sfContext::getInstance()->user->getAttribute('search-realm')
```
MikeFE: I'd avoid both using @ and chaining ternary operators and do this before $aliases is assigned if possible. The PHP docs warn against chaining ternary ops since there are subtleties in PHP that are non-obvious when you do it, and the @ operator... I think it's better to explicitly check things with isset() and things like that.

apps/qubit/modules/actor/actions/browseAction.class.php
```diff
@@ -81,6 +81,19 @@ public function execute($request)
...
     parent::execute($request);
 
+    if (isset($request->repos) && ctype_digit($request->repos))
+    {
+      $this->repos = QubitRepository::getById($request->repos);
+
+      // Add repo to the user session as realm
+      $this->context->user->setAttribute('search-realm', $request->repos);
+    }
+    elseif (sfConfig::get('app_enable_institutional_scoping'))
```
MikeFE: It's a nitpick but we use 'else if' when outside of templates & using the weird alternative syntax. 

elseif as all one word is only ever required in the `<?php elseif (...): ?>` style syntax.

apps/qubit/modules/search/actions/boxComponent.class.php
```diff
@@ -21,6 +21,14 @@ class SearchBoxComponent extends sfComponent
 {
   public function execute($request)
   {
+    // if the institutional scoping setting is on, search is always global.
+    if (sfConfig::get('app_enable_institutional_scoping'))
+    {
+      $this->repository = NULL;
```
MikeFE: lowercase nulls please

plugins/sfIsadPlugin/modules/sfIsadPlugin/actions/stylesheetComponent.class.php
```diff
@@ -0,0 +1,73 @@
...
+    else if (!sfConfig::get('app_enable_institutional_scoping'))
+    {
+      if (!isset($request->getAttribute('sf_route')->resource))
+      {
+        return sfView::NONE;
+      }
+
+      $resource = $request->getAttribute('sf_route')->resource;
+
+      switch (true)
```
MikeFE, line 39: heheh, I hadn't seen this in the wild before but I remember someone arguing for its case. It works I guess :)

MikeFE:
:+1:

PR #493, opened 2017-01-10
EAD XML delete replace/keymap issues, Refs #9992

sbreker:
This commit corrects 2 issues found in the EAD XML import when trying to
match records in the import file to records in the DB.

Both issues were related to the fact that with XML import matching, only the
top level record in the import file needs to match a record in the database.
On import, once the top level record was matched, the entirerest hierarchy
for that record in the database was being deleted so attempting to match
child records to records in the database was breaking.

Changes made:
- When 'delete-and-replace' and 'skip-unmatched' is specified, only match
the top level record in the import file to the DB. If it matches, delete
the DB hierarchy and import all records in the file.
- Only attempt keymap matching for the top level record in the XML import
file.
- Only create keymap matching records for the top level record in the XML
import file.

Additional change: move saving of i18n instance to the top of the import()
method otherwise the call to validateOptions() will fail if it outputs a
message.

### DIFF
--- a/lib/QubitXmlImport.class.php
+++ b/lib/QubitXmlImport.class.php
@@ -39,11 +39,12 @@ class QubitXmlImport
 
   public function import($xmlFile, $options = array(), $xmlOrigFileName = null)
   {
+    // Needs to be created before validateOptions() is called.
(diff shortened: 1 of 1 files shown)


sevein:
👍

PR #495, opened 2017-01-12
Fix thumbnailPath in ES, refs #10758

jraddaoui:
The getFullPath method from QubitDigitalObject takes care of
the differences between external and internal resources


djjuhasz:
I found the same bug yesterday and had the same fix @jraddaoui :) :+1:

PR #496, opened 2017-01-12
Fix digital object regeneration task, refs #10221

djjuhasz:
* Fix bug that prevented creation of derivatives when "type" not set
* Add validation of "type" option
* QubitDigitalObject: Don't generate thumbnail in createRepresentation when
  $usageId = REFERENCE_ID
* QubitDigitalObject: Use standard functions for generating reference and
  thubmnail images


mcantelon:
LGTM!

PR #497, opened 2017-01-13
Fix bug with video and audio derivs, refs #10221

djjuhasz:
Correctly handle reference and thumbnail $usageId in createRepresentations()
for video and audio files.


mcantelon:
Looks good to me.

PR #498, opened 2017-01-13
Inst Scoping repo holdings list, Refs #10733

sbreker:
The change breaks the repository holdings component list and pager out into
it's own component so that it can be re-used in the holdingsInstitution
component.

Altered the _contextMenu template so the new list component is included with
the call to the original holdings component, and with the holdingsInstitution
component which is activated when Institutional Scoping feature is turned ON.


MikeFE:
:+1: LGTM

PR #499, opened 2017-01-16
Update swiftmailer, refs #10772

sevein:
Redmine: https://projects.artefactual.com/issues/10772

Updated swiftmailer to v5.4.5, which addresses a security issue recently found. `vendor/symfony/test/unit/mailer/sfMailerTest.php` looks ok after the update.

Also, deleted unused mail function in `QubitOai`. With this, I actually think that neither the plain mail() or swiftmailer are used at all in AtoM.


sbreker:
👍

PR #501, opened 2017-01-18
Inst Scoping ensure 2 col layout, Refs# 10733

sbreker:
This change creates an exception in the information object browse template
to ensure that the 2 column layout is kept when the Institutional Scoping
feature is turned ON.

When Insitutional Scoping is OFF and a user performs a search that returns
zero results, the information object browse template changes to a 1 col
layout removing the sidebar. The change creates an exception for that rule
so that when Institutional Scoping is ON, the 2 column layout will be used
and the sidebar including the Institutional Block will be displayed, which
will provide context to the user and allow the user to perform another
scoped search.


MikeFE:
:+1:

PR #502, opened 2017-01-23
Add EAD support in OAI responses and generate XML metadata asynchronously, refs #10797

mcantelon:
Redmine: https://projects.artefactual.com/issues/10797

EAD support for OAI-PMH API:
* EAD support added for ListMetadataFormats, GetRecord, and ListRecords verb responses
* ListRecords verb response limited to one record per page if EAD metadata format is requested
* ListIdentifiers verb response changed to show only top-level records if EAD metadata format is requested

Cache EAD and DC XML representations of information objects:
* Job type for caching a single information object's XML representations as files
* CLI task for caching all information object's XML representations as files (cache:xml-representations)
* Migration for caching all information object's XML representations as files
* Trigger caching from web UI in appropriate places
* Link directly, in web UI, to cached EAD and DC XML files

OAI-PMH performance enhancements:
* Optimize QubitOai class's getOaiSets and getMatchingOaiSet methods
* Updated GetRecord and ListRecords verb responses to use cached EAD/DC metadata

Fixes:
* Add resumptionToken support to ListSets verb response
* Make getCollections pageable

Cleanup:
* Unused templates and redundant code removed


data/fixtures/settings.yml
```diff
@@ -3,7 +3,7 @@ QubitSetting:
     name: version
     editable: 0
     deleteable: 0
-    value: 148
+    value: 150
```
sevein: Steve missed that (148 » 149), he is fixing it soon. You can rebase later!

lib/oai/QubitOai.class.php
```diff
@@ -64,6 +64,25 @@ public static function checkBadArgument($keys, $requestValidKeys, $mandatoryKeys
...
+   * Check that metadata format is valid
+   *
+   * @param string    metadata format
+   * @return bool  is the metadata format valid or not
+   */
+  public static function checkValidMetadataFormat($metadataFormat)
+  {
+    foreach (self::getMetadataFormats() as $formatData)
+    {
+      if ($formatData['prefix'] == $metadataFormat)
```
sevein: I tend to prefer the `===` operator unless you have a reason not to.

lib/model/QubitInformationObject.php
```diff
@@ -440,6 +440,11 @@ public static function getUpdatedRecords($options = array())
       $criteria = QubitAcl::addFilterDraftsCriteria($criteria);
     }
 
+    if (isset($options['topLevel']) && $options['topLevel'])
```
sevein: Nitpick... wouldn't be more correct to write the following?

```php
if (isset($options['topLevel']) && true === $options['topLevel']) { }
```

Or do you really want to support any truthy value?

plugins/arOaiPlugin/modules/arOaiPlugin/templates/_listSets.xml.php
```diff
@@ -5,4 +5,7 @@
            <setName><?php echo esc_specialchars(strval($set->getName())) ?></setName>
         </set>
     <?php endforeach ?>
+  <?php if ($remaining > 0): ?>
+    <resumptionToken><?php echo $resumptionToken?></resumptionToken>
```
sevein: Missing whitespace.

plugins/arOaiPlugin/modules/arOaiPlugin/templates/_listSets.xml.php
```diff
@@ -5,4 +5,7 @@
...
+  <?php endif?>
```
sevein: Missing whitespace.

lib/oai/QubitOai.class.php
```diff
@@ -294,16 +294,29 @@ public static function addOaiSet($oaiSet)
...
    */
-  public static function getMatchingOaiSet($setSpec, $oaiSets)
+  public static function getMatchingOaiSet($setSpec)
   {
-    foreach ($oaiSets as $oaiSet)
+    // Check additional sets, if enabled
+    $setting = QubitSetting::getByName('oai_additional_sets_enabled');
+    $setsEnabled = (null === $setting) ? false : $setting->value;
+
+    if ($setsEnabled)
```
sevein: Could the following snippet replace L300-L303?

```php
if (sfConfig::get('app_oai_additional_sets_enabled__source', false))
{
  // ...
}
```

plugins/arOaiPlugin/modules/arOaiPlugin/actions/getRecordComponent.class.php
```diff
@@ -30,14 +30,9 @@ class arOaiPluginGetRecordComponent extends arOaiPluginComponent
...
   {
-    $request->setRequestFormat('xml');
-    $this->date = gmdate('Y-m-d\TH:i:s\Z');
-
     $oai_local_identifier_id = QubitOai::getOaiIdNumber($request->identifier);
-    $this->informationObject = QubitInformationObject::getRecordByOaiID($oai_local_identifier_id);
-    $request->setAttribute('informationObject', $this->informationObject);
-
-    $this->path = $request->getUriPrefix().$request->getPathInfo();
+    $this->record = QubitInformationObject::getRecordByOaiID($oai_local_identifier_id);
```
sevein: Can you take this chance to use `$snakeCase`?

data/fixtures/settings.yml
sevein: It's been fixed! https://github.com/artefactual/atom/pull/503

sevein:
I've found that the code used by `GetRecord` fails when the object cannot be found with the identifier given:

```
PHP message: PHP Fatal error:  Call to a member function getOaiIdentifier() on a non-object in /usr/share/nginx/atom/plugins/arOaiPlugin/modules/arOaiPlugin/templates/_getRecord.xml.php on line 5
```

lib/task/arCacheDescriptionXmlTask.class.php
```diff
@@ -0,0 +1,93 @@
...
+        printf("Cached DC XML for information object %s.\n", $io->id);
+      }
+    }
+
+    print "Done.\n";
+  }
+
+  public static function cacheXmlRepresentation($resource, $format)
+  {
+    $tempFilePath = sys_get_temp_dir() . DIRECTORY_SEPARATOR . 'export_'. $format .'_'. $io->id;
```
sevein: We should use `tmpfile()`, right?

lib/task/arCacheDescriptionXmlTask.class.php
```diff
@@ -0,0 +1,93 @@
...
+  }
+
+  public static function exportAll()
+  {
+    arXmlExportSingleFileJob::createExportDestinationDirs();
+    exportBulkBaseTask::includeXmlExportClassesAndHelpers();
+
+    print "Caching XML representations of information objects...\n";
+
+    foreach(QubitInformationObject::getAll() as $io)
```
sevein: `foreach` is a statement, we normally insert a whitespace after it.

lib/model/QubitInformationObject.php
MikeFE: How do you feel about
`if (!empty($options['topLevel']))`

It's still a fairly loose check but we do it all over the application already and it's more condensed.

sevein:
When I see static functions like `storeExport` or `getPath` used from multiple places I feel that the functionality is not properly encapsulated and the responsibilities are scattered. I also think that you would have a hard time if you wanted to test that code. I think that tasks like `exportBulkBaseTask` and `arCacheDescriptionXmlTask` or jobs like `arXmlExportSingleFileJob` should just be glue code. They're similar to web controllers in that there is some input and you send commands to your models or external classes, collect the output and return.

One example, if you `git grep arXmlExportSingleFileJob` you'll see at least two places making use of its static method `getPath`. That feels wrong, doesn't it? It's like if you defined a static method in a controller (subclass of sfAction) so it's shared with other controllers. We don't do that because it's just glue code and if there is something in common between two controllers we try to create an intermediate class (e.g. DefaultEditAction).

Let me know what you think. You've written this code very recently so there are chances that you can think of other ways to arrange it without much effort. I would start with `arXmlExportSingleFileJob`, how would that class look like if all the logic of actual exporting the file was in a new class?

lib/oai/QubitOai.class.php
mcantelon: Sounds good... what does the "__source" do?

sevein: It gets the source value without fallback, which is useful when you don't really care about translated content.

mcantelon: Ah, nice trick.

mcantelon:
I've refactored things, creating a QubitInformationObjectXmlCache class to get rid of the overuse of static methods.

sevein:
Where is `QubitInformationObjectXmlCache` defined?

mcantelon:
Oops... have added QubitInformationObjectXmlCache.class.php now.

lib/QubitInformationObjectXmlCache.class.php
```diff
@@ -0,0 +1,217 @@
...
+   */
+  public function export($resource)
+  {
+    // Only cache top-level information object's EAD XML
+    if ($resource->parentId == QubitInformationObject::ROOT_ID)
+    {
+      $this->cacheXmlRepresentation($resource, 'ead');
+    }
+
+    $this->cacheXmlRepresentation($resource, 'dc');
```
MikeFE: this would seem to get called even if we're dealing with EAD and the if evaluates true above here, right? is that desirable?

lib/QubitInformationObjectXmlCache.class.php
```diff
@@ -0,0 +1,217 @@
...
+   * @param string  path to temporary file containing XML
+   * @param integer  ID of information object to cache
+   * @param string  format of XML ("dc" or "ead")
+   *
+   * @return null
+   */
+  protected function storeXmlExport($filePath, $objectId, $format)
+  {
+    copy($filePath, $this->getFilePath($objectId, $format));
+    $skipLines = ($format == 'ead') ? 2 : 1;
```
MikeFE: please comment why this is needed

lib/QubitInformationObjectXmlCache.class.php
```diff
@@ -0,0 +1,217 @@
...
+  /**
+   * Rewrite file skipping lines.
+   *
+   * @param string  source file
+   * @param string  destination file
+   * @param integer  number of lines to skip
+   *
+   * @return string  URL of XML representation
+   */
+  protected function rewriteFileSkippingLines($source, $destination, $skipLines = 0)
```
MikeFE: please elaborate on why we're skipping lines + rewriting the file, this function seems to be doin' some tricky stuff.

lib/QubitInformationObjectXmlCache.class.php
mcantelon: This function's meant to export both formats, yes (but EAD only if it's a top-level element).

lib/QubitInformationObjectXmlCache.class.php
mcantelon: I've added more commenting in about the rationale for making the two versions.

lib/job/arXmlExportSingleFileJob.class.php
```diff
@@ -0,0 +1,101 @@
...
+  /**
+   * @see arBaseJob::$requiredParameters
+   */
+  protected $params = array();
+
+  public function runJob($parameters)
+  {
+    $this->params = $parameters;
+
+    if (!isset($this->params['objectId']))
```
sevein: I think that you can assign an array of parameters you require in `$extraRequiredParameters` and `arBaseJob` will do this for you, e.g. see `arGenerateJobReport`.

lib/job/arXmlExportSingleFileJob.class.php
```diff
@@ -0,0 +1,101 @@
...
+    {
+      throw new sfException($this->i18n->__('Information object %1% does not eist', array('%1%' => $this->params['objectId'])));
+    }
+
+    try
+    {
+      // Print warnings/notices here too, as they are often important.
+      $errLevel = error_reporting(E_ALL);
+
+      $cache = new QubitInformationObjectXmlCache;
```
sevein, line 80: Should we pass the job logger to `QubitInformationObjectXmlCache`? I'm worried that these exports may take really long. It would be nice to log some events from `QubitInformationObjectXmlCache`? I did that [here](https://github.com/artefactual/atom/blob/qa/2.4.x/lib/job/arFileImportJob.class.php#L72) and it worked great. The log is being stored in the database (unfortunately) so I don't recommend to be too chatty, but some details would be awesome. Remember that the Job UI lets the user see the output of a job as it is generated.

lib/job/arXmlExportSingleFileJob.class.php
```diff
@@ -0,0 +1,101 @@
...
+
+    if (null === $resource)
+    {
+      throw new sfException($this->i18n->__('Information object %1% does not eist', array('%1%' => $this->params['objectId'])));
+    }
+
+    try
+    {
+      // Print warnings/notices here too, as they are often important.
+      $errLevel = error_reporting(E_ALL);
```
sevein, line 78: This is very interesting. @MikeFE mentioned something about error reporting.

With this, the errors will be sent to the standard stream, is that right? Is that what we want? We could setup an error handler so the errors are also sent to the logger, but I'm not sure if that would be too chatty or undesirable. What do you think? I'm also curious why warnings are important to be shown but not important enough to be fixed right away, is it too spread? Should we file an issue?

lib/model/QubitInformationObject.php
```diff
@@ -567,6 +572,114 @@ public function canInheritRepository($repositoryId)
...
+   * Initate asynchronous jobs to export EAD and DC XML.
+   */
+  public function createXmlExports()
+  {
+    // Export top-level parent as EAD
+    $params = array(
+      'objectId' => $this->getCollectionRoot()->id,
+      'format'   => 'ead'
+    );
+
```
sevein: Nitpick, is this blank line necessary? The next block `// Export as DC` does not use it.

lib/model/QubitInformationObject.php
```diff
@@ -567,6 +572,114 @@ public function canInheritRepository($repositoryId)
...
+  public function urlForEadExport()
+  {
+    if ($this->getPublicationStatus()->statusId == QubitTerm::PUBLICATION_STATUS_PUBLISHED_ID && file_exists($this->pathToEadExport()))
+    {
+      return sfConfig::get('siteBaseUrl') .'/'. $this->pathToEadExport();
+    }
+    else
+    {
+      sfContext::getInstance()->getConfiguration()->loadHelpers('Url');
+      return url_for(array($this, 'module' => 'sfEadPlugin', 'sf_format' => 'xml'));
```
sevein: Perhaps we can replace `url_for` with `generate` from `routing` (in `sfContext`) as `url_for` is a template helper.

lib/QubitInformationObjectXmlCache.class.php
```diff
@@ -0,0 +1,229 @@
...
+   * @param string  path to temporary file containing XML
+   * @param integer  ID of information object to cache
+   * @param string  format of XML ("dc" or "ead")
+   *
+   * @return null
+   */
+  protected function storeXmlExport($filePath, $objectId, $format)
+  {
+    // Copy unmodified XML to downloads subdirectory
+    copy($filePath, $this->getFilePath($objectId, $format));
```
sevein: Is this ok? Can you invoke `getFilePath` like that when it is static? Sorry my PHP is rusty.

lib/model/QubitInformationObject.php
```diff
@@ -567,6 +572,114 @@ public function canInheritRepository($repositoryId)
...
+
+  /**
+   * Return file path to EAD XML, regardless of whether it's available.
+   *
+   * @param
+   * @return string  file path of EAD XML
+   */
+  public function pathToEadExport($contentsOnly = false)
+  {
+    return QubitInformationObjectXmlCache::getFilePath($this->getCollectionRoot()->id, 'ead', $contentsOnly);
```
sevein: I think this is a good story to rethink `QubitInformationObjectXmlCache` a bit further. If `QubitInformationObjectXmlCache` took a `$resource` in the constructor you could do something like:

```php
$foo = new QubitInformationObjectXmlCache($this->getCollectionRoot());
$foo->getFilePath();
```

Would that make sense? That would require some extra refactoring, mainly I think `QubitInformationObjectXmlCache` would need to be repurposed to have a single purpose: one `$resource`. Now it's a combo of one or many (`exportAll()`). Don't follow this to the letter, it's just a thought.

lib/model/QubitInformationObject.php
```diff
@@ -567,6 +572,114 @@ public function canInheritRepository($repositoryId)
...
+  public function urlForDcExport()
+  {
+    if ($this->getPublicationStatus()->statusId == QubitTerm::PUBLICATION_STATUS_PUBLISHED_ID && file_exists($this->pathToDcExport()))
+    {
+      return sfConfig::get('siteBaseUrl') .'/'. $this->pathToDcExport();
+    }
+    else
+    {
+      sfContext::getInstance()->getConfiguration()->loadHelpers('Url');
+      return url_for(array($this, 'module' => 'sfDcPlugin', 'sf_format' => 'xml'));
```
sevein: Same here, I think we should avoid template helpers here.

lib/QubitInformationObjectXmlCache.class.php
MikeFE: I think this is [legitimate PHP](https://gist.github.com/MikeFE/349c70de0932a19d7f33f31264a4cfb3). But most languages don't let you do it like this, so if there are still some hours to spare on the ticket I'd recommend using classname::method() instead.

lib/job/arXmlExportSingleFileJob.class.php
MikeFE: Hilariously it seems the PHP docs don't mention if error_reporting() errors go to stderr vs. stdout, but it's one of those two (I guess stderr?). I don't have time to test it at the moment. 

The original rationale for me turning that on was I was touching our EAD export code at various times and there were like 15+ PHP warnings and notice issues that had been sitting there in the code for years. So I enabled that to hopefully draw attention to those undefined variable and related issues to coders working on that code subsequently if any more came up. Maybe it's not the best idea.

lib/QubitInformationObjectXmlCache.class.php
mcantelon: Cool... I'll use the "self::" notation.

MikeFE: :+1:

lib/job/arXmlExportSingleFileJob.class.php
mcantelon: I've added logger support...

When in the database do log entries get stored? I didn't see any obvious table names for storing log entries...

mcantelon:
I added a QubitInformationObjectXmlCacheResource class to, hopefully, further clean things up.

lib/job/arXmlExportSingleFileJob.class.php
sevein: Would the following line in the [worker front-end settings](https://github.com/artefactual/atom/blob/qa/2.4.x/apps/qubit/config/settings.yml.tmpl#L14) be equivalent?

```yml
---
worker:
  .settings:
    logging_enabled: true
    error_reporting: <?php echo E_ALL."\n" ?>
```

sevein:
:+1:

sevein:
Can we have @fiver-watson testing the functionality before you merge?

mcantelon:
I've added the ability to toggle the XML updating.

sevein:
Looking good!

mcantelon:
Merged.

PR #503, opened 2017-01-25
Inst Scoping bump DB version, Refs #10733

sbreker:
Bump DB version in settings.yml to match the new migration for the Inst
Scoping work.


sevein:
:+1:

PR #504, opened 2017-01-27
Inst Scoping search text, Refs #10733

sbreker:
This change controls which search box will contain the searched-for text
after a page refresh when the Institutional Scoping feature is turned on.

- When scoped to an insitution and an institutional search is performed
using the search field in the institutional block, retain the searched-for
text in this search field. Do not display this searched-for text in the
global search box in the header.

- if a institutional search has been performed, and the repo facet is removed
from the results page removing scoping, copy the searched-for text to the
global search box in the header.

- if a user performs a global search using the header search box and on the
results page the institutional facet is ADDED to the results causing the
session to scope to a repository, move the search term from the global
search box in the header to the Institutional Block search field.

Note that this feature only takes effect if Institutional Scoping feature is
ON.


sevein:
👍

PR #505, opened 2017-01-30
Improve sort options in clipboard, refs #10834

jraddaoui:
Redmine: https://projects.artefactual.com/issues/10834

Use the same sort options as the information object browse page


sevein:
:+1:

PR #507, opened 2017-01-30
Always display date range in search results, refs #10838

sevein:
See [#10838](https://projects.artefactual.com/issues/10838) for more details.


sbreker:
👍

PR #509, opened 2017-01-31
Install cached config files issue, Refs #10648

sbreker:
Redmine: https://projects.artefactual.com/issues/10648

Clear PHP opcode cache. This was added to correct issue where occasionally
during installation, the cache will contain the vendor skeleton .yml
files, which override the AtoM config files that should be written to the cache
folder during install. This prevents the i18n and Qubit helpers being
loaded (from apps/qubit/config/settings.yml) triggering the i18n errors
during the installation process.


sevein:
👍

PR #510, opened 2017-01-31
Multiple fixes, refs #10838 #10839

sevein:
This pull request addresses some issues found in the following tickets once they were merged:
- https://projects.artefactual.com/issues/10838 (#507)
- https://projects.artefactual.com/issues/10839 (#508)

### DIFF
--- a/apps/qubit/modules/informationobject/templates/inventorySuccess.php
+++ b/apps/qubit/modules/informationobject/templates/inventorySuccess.php
@@ -49,18 +49,7 @@
               <?php endif; ?>
             </td>
             <td>
-              <?php if (isset($doc['dates'])): ?>
-                <?php foreach ($doc['dates']->getRawValue() as $date): ?>
-                  <?php if (isset($date['startDateString'])
-                    || isset($date['endDateString'])
-                    || null != get_search_i18n($date, 'date', array('culture' => $culture))): ?>
-                    <?php echo Qubit::renderDateStartEnd(get_search_i18n($date, 'date', array('culture' => $culture)),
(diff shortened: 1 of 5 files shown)


apps/qubit/modules/informationobject/templates/inventorySuccess.php
```diff
@@ -49,18 +49,7 @@
               <?php endif; ?>
             </td>
             <td>
-              <?php if (isset($doc['dates'])): ?>
-                <?php foreach ($doc['dates']->getRawValue() as $date): ?>
-                  <?php if (isset($date['startDateString'])
-                    || isset($date['endDateString'])
-                    || null != get_search_i18n($date, 'date', array('culture' => $culture))): ?>
```
sbreker, line 56: Should you add  to the array `'allowEmpty => 'false'` to prevent return of empty string?

sevein: I'm not sure. I'm trying to replicate the existing behaviour. Should I?

sevein: Oh, it's a date not a title, that's the reason why you thought it'd be good to use `allowEmpty`, right?

sbreker: Ok - yes - if it's a date I expect it should return null making your comparison work.  I was just debugging an issue on a call to get_search_i18n() so have strings on my mind!  😄

sbreker:
Looks good.  👍

PR #511, opened 2017-02-01
Fix Descr Update repository display, Refs# 10730

sbreker:
The 'Repository' column on the Description Updates page was always displaying
'Untitled'. This change corrects this and the repository will now display for
each record.


sevein:
👍

PR #512, opened 2017-02-01
Import-bulk output 'error' text, Refs #10729

sbreker:
The import:bulk tast output messages were always prefixed with the static
text 'Error'. This text has been removed since this output will contain status
messages as well.


MikeFE:
:+1:

PR #514, opened 2017-02-06
Fix realm in search box links, refs #10295

jraddaoui:
When the multi-repository setting is set to 'no', no realm check-boxes are
added to the search-box drop-down, adding the 'repos=all' parameter to
the 'all matching ...' links, which leads to a page without results

Redmine: https://projects.artefactual.com/issues/10295#note-9


sbreker:
👍   Look good.

PR #515, opened 2017-02-10
Add support for other entities on the clipboard

MikeFE:
Add clipboard/export support to repos and actors

This new feature adds 'add to clipboard' buttons for both repositories and
actors. The clipboard now has an additional drop down box to allow the user to
change the clipboard view between the three entity types (archival descs,
repositories and actors), to clear the clipboard for each entity type, and to
export each entity type.

Asynchronous jobs to export actors (both CSV and EAC XML), and repositories
(CSV) were also added. Additionally, the export repository task that was
in its own branch was integrated into this work.

sortPicker has been generalized into genericPicker where you can specify the
drop down label and options. You can still call sortPicker as it just invokes
genericPicker with sort-specific labels hard coded.

To summarize:

- Archival descriptions, actors and repositories can now all be added to
  clipboard

- Asynchronous jobs to export all these three entity types are in place

- You can switch between entity types in the clipboard via a dropdown menu

- You can clear the clipboard of each entity type depending on the currently
  selected type.

- Made a more generic sortPicker, "genericPicker."


jraddaoui:
Wow, this is quite a pull request @MikeFE!

I guess it will be too much work to split it in various PR at this point, but I think we should. Or, if we don't have time to do that, at least we should improve a lot the description (and the commit message that will be used in the squashed merge) as it doesn't even talk about the asynchronous jobs.

For the next time we work in big features, I'd love to advice that we start the code review process earlier, that way we could have spotted this sooner, reviewed the clipboard part and merged it, to avoid conflicts in the subsequent PRs, etc. Also, I'd appreciate clearer individual commit messages and better descriptions.

About the code, it looks great overall! But I'll take a deeper look tomorrow or on Monday, just in case we can split the PR or improve the description.

MikeFE:
Sounds good Radda. Yeah, this PR got way larger than originally anticipated since this feature went way over hours and we vastly under estimated how much work we were going to do on it unfortunately, but I'm totally open to discussion on how we can avoid this sort of thing in the future.

regards

jraddaoui:
Great @MikeFE! Please, could you improve the description then? Thanks!

MikeFE:
OK I rewrote the PR description and gave it my best shot. Hopefully that's what you were looking for :)

apps/qubit/modules/default/templates/_genericPicker.php
```diff
@@ -0,0 +1,39 @@
+
```
jraddaoui: We don't use to start a template with an empty line

apps/qubit/modules/default/templates/_sortPicker.php
```diff
@@ -1,43 +1,2 @@
...
-
-      </ul>
-
-    </div>
-
-  </div>
-
-</div>
+<?php echo get_partial('default/genericPicker', array('options' => $sf_data->getRaw('options'),
+                       'param' => 'sort', 'label' => __('Sort by'))) ?>
```
jraddaoui: But we use to add a new line at the end of all files. Check: http://stackoverflow.com/questions/729692/why-should-text-files-end-with-a-newline

If you use Sublime Text, that can be set in a config setting, so it automatically adds a new line at the end if it's not there when saving.

apps/qubit/modules/object/actions/exportAction.class.php
```diff
@@ -82,23 +115,24 @@ protected function doBackgroundExport($request)
...
-                                       'slugs' => $this->context->user->getClipboard()->getAll()),
-                     'current-level-only' => ('on' == $request->getParameter('includeDescendants')) ? false : true,
-                     'public' => ($request->getParameter('includeDrafts') == 'on') ? false : true,
-                     'objectType' => $request->getParameter('objectType'),
-                     'levels' => $levelsOfDescription);
+    $options = array(
+      'params' => array('fromClipboard' => true, 'slugs' => $this->context->user->getClipboard()->getAll()),
+      'current-level-only' => 'on' !== $request->getParameter('includeDescendants'),
+      'public' => 'on' !== $request->getParameter('includeDrafts'),
+      'objectType' => $request->getParameter('objectType'),
```
jraddaoui: You could use `$this->objectType` in here as it was obtained before with a default value

apps/qubit/modules/repository/templates/_browseCardView.php
```diff
@@ -21,7 +21,7 @@
         <?php endif; ?>
       </a>
       <div class="bottom">
-        <p><?php echo $authorizedFormOfName ?></p>
+        <p><?php echo get_component('object', 'clipboardButton', array('slug' => $doc['slug'], 'wide' => true)) ?><?php echo $authorizedFormOfName ?></p>
```
jraddaoui: I think we should use the small clipboard button and place it in the right side as it takes a lot of place in those cards footer. You may need to fix the text ellipsis or place the button in another container.

apps/qubit/modules/repository/templates/_browseTableView.php
```diff
@@ -53,6 +53,8 @@
 
         <?php echo link_to(render_title(get_search_i18n($doc, 'authorizedFormOfName', array('allowEmpty' => false,
                            'culture' => $selectedCulture))), array('module' => 'repository', 'slug' => $doc['slug'])) ?>
+
+        <?php echo get_component('object', 'clipboardButton', array('slug' => $doc['slug'], 'wide' => true)) ?>
```
jraddaoui: Same here, it doesn't look right alongside the auth. form of name as it changes the button position. I'd use the small button in a new column at the end of the table, without column header title, so it takes only the button width. We probably don't have time to add the column toggle (All/None) in the column header, like in the description updates page, but we should check it, just in case it's an easy change.

apps/qubit/modules/user/actions/clipboardAction.class.php
```diff
@@ -25,25 +25,29 @@
...
   public function execute($request)
   {
     if ('print' == $request->getGetParameter('media'))
     {
       $this->getResponse()->addStylesheet('print-preview', 'last');
     }
 
-    $slugs = $this->context->user->getClipboard()->getAll();
+    // Get entity type class name
+    $this->entityType = 'Qubit'.ucfirst($request->getGetParameter('type', 'informationObject'));
```
jraddaoui: This variable could be created after `$this->type` is set, to avoid getting the parameter from the request twice.

apps/qubit/modules/user/actions/clipboardClearAction.class.php
```diff
@@ -36,6 +38,7 @@ public function execute($request)
     }
 
     $slugs = $this->context->user->getClipboard()->getAll();
+    $this->typeLabel = $this->getTypeLabel();
 
     // Redirect to clipboard page if the clipboard is empty
     if (count($slugs) == 0)
```
jraddaoui: We probably need to check `$slugs[$this->type]` in here, like the changes in `clipboardAction.class.php`

lib/flatfile/config/QubitActor-atom.yml
```diff
@@ -0,0 +1,21 @@
...
+  - internalStructures
+  - generalContext
+  - descriptionIdentifier
+  - institutionIdentifier
+  - rules
+  - status
+  - levelOfDetail
+  - revisionHistory
+  - sources
+  - maintenanceNotes
```
jraddaoui: New line at the end.

lib/flatfile/config/QubitActor.yml
```diff
@@ -0,0 +1,28 @@
...
+  institutionResponsibleIdentifier:  institutionIdentifier
+  descriptionStatus:                 status
+  entityType:                        typeOfEntity
+  descriptionDetail:                 levelOfDetail
+
+property:
+
+note:
+
+cacheTaxonomies:
```
jraddaoui: ;) the same, new line

lib/task/migrate/migrations/arMigration0137.class.php
```diff
@@ -47,7 +47,7 @@ public function up($configuration)
         $menu->name = 'clearClipboard';
         $menu->path = 'user/clipboardClear';
         $menu->sourceCulture = 'en';
-        $menu->label = 'Clear selections';
+        $menu->label = 'Clear all selections';
```
jraddaoui, line 50: I think this is a 2.3 migration, so instances upgraded from 2.3 to 2.4 won't get the label change. I think we must create a new migration script.

plugins/sfIsaarPlugin/modules/sfIsaarPlugin/templates/indexSuccess.php
```diff
@@ -35,6 +35,11 @@
 
     <section id="action-icons">
       <ul>
+        <li class="separator"><h4><?php echo __('Clipboard') ?></h4></li>
+
+        <li class="clipboard">
+          <?php echo get_component('object', 'clipboardButton', array('slug' => $resource->slug, 'wide' => true)) ?>
+        </li>
```
jraddaoui, line 42: I think this button is missing in the repository index page.

jraddaoui (changes requested):
Hi @MikeFE,

Thank you for the description update, we should use it as the commit body when we merge this work. I'd add a small note about the generic picker changes in there too.

I've added a few comments, some that I think must be addressed before merging and others not so important. I also have the feeling that we could have covered all export cases with only one export job, as there is a lot of duplicated code in those classes, but it may be too late to change it now.

Great work!

apps/qubit/modules/default/templates/_sortPicker.php
MikeFE: hmm OK I use VSCode and didn't notice it wasn't adding the newlines! I'll look into it :)

plugins/arDominionPlugin/css/less/buttons.less
```diff
@@ -170,6 +170,10 @@ button.clipboard-wide {
   }
 }
 
+button.clipboard.clipboard-repository-browse {
```
jraddaoui: You could have place this inside the `button.clipboard` section bellow using `&.clipboard-repository-browse`. The transformation to CSS will be the same but it makes a nicer LESS with the nested rules.

lib/task/migrate/migrations/arMigration0150.class.php
```diff
@@ -0,0 +1,48 @@
...
+  /**
+   * Upgrade
+   *
+   * @return bool True if the upgrade succeeded, False otherwise
+   */
+  public function up($configuration)
+  {
+    if (null !== $menuItem = QubitMenu::getByName('clearClipboard'))
+    {
+      $menuItem->label = 'Clear all selections';
```
jraddaoui: Could you remove the changes in lib/task/migrate/migrations/arMigration0137.class.php, please.

sevein: I think we shouldn't change previous migrations, right? We should only update state from newer migrations.

sevein: @MikeFE I think you should replace `$menuItem->label = 'Clear all selections';` with:

```php
$menuItem->setLabel('Clear all selections', array('culture' => 'en'));
```

lib/task/migrate/migrations/arMigration0150.class.php
```diff
@@ -0,0 +1,48 @@
...
+      $menuItem->sourceCulture = 'en';
```
sevein: This one you don't need it because it should already have that value in it.

jraddaoui:
Hi @MikeFE, looking great!

Now that we have the button in both repository view options I think we should add it to the card (digital object) view in the IO browse page, it should be an easy addition. Also, for those card views, since we are placing the clipboard button to the right in all cases, I'd place it in the right bottom corner of the card instead of at the left-bottom corner.

And please, add a little note about the generic picker in the description. Thanks!

jraddaoui:
For following PRs, could you avoid to force push until just before the merge. It's not a big deal but it keeps the commit history and review messages in place ;)

jraddaoui:
Hi @MikeFE,

After all, for the clipboard button inside the cards (DO browse and repository browse in card view) we could have avoided passing the parameter and adding the class by adding the CSS rules for `.masonry-brick button.clipboard`, it will remove some code but feel free to leave it be. But please, remove the unneeded changes in lib/task/migrate/migrations/arMigration0137.class.php.

Thanks!

jraddaoui (approved):

MikeFE:
merged into qa/2.4.x 4403962532fb87df6623c335a7fc66f5fbd8b214

jraddaoui:
Hi @MikeFE, 

Please, delete the development branches after merge. We're "starting" to have a big list of branches :)

Thanks!!

jraddaoui:
Also, the commit message used in the merge doesn't have a reference to the Redmine ticket ...

PR #516, opened 2017-02-15
Correct event date behaviour on save, Refs #10278

sbreker:
An issue was reported where if modifications to existing dates were made,
and a new date was added, and then the description was saved that the new
date would be saved but the modified dates would not. This issue affects
all description templates.

On informationObject save, without any modifications to the dates,
eventsRelatedByobjectId[] would contain all events. In addition, with
modifications to one of many dates, eventsRelatedByobjectId[] would contain
all events associated with the informationObject. When there was a combination
of new records and updated ones, only the new records were contained in
eventsRelatedByobjectId[], causing only those updates to be saved.

Ensuring that eventsRelatedByobjectId[] was always populated with the date
events corrects this issue.

When using the MODS and RAD templates where an actor is also saved on the
same dialog, the same issue was occurring when there was a combination of
new actors/dates and updated actors or dates. This has also been addressed
in this fix.

### DIFF
--- a/apps/qubit/modules/event/actions/editComponent.class.php
+++ b/apps/qubit/modules/event/actions/editComponent.class.php
@@ -146,7 +146,15 @@ class EventEditComponent extends sfComponent
         if (isset($item['id']))
         {
           $params = $this->context->routing->parse(Qubit::pathInfo($item['id']));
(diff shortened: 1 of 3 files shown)


mcantelon:
Looks good to me!

PR #519, opened 2017-02-18
Fix blurry thumbnails, refs #9803

mcantelon:
Fixed issue with replacement of thumbnails where dimension restrictions during
replacement differed from dimension restrictions during creation.


sevein:
LGTM

mcantelon:
Thanks! Merged.

PR #520, opened 2017-02-18
Fixed minor REST API issues, refs #10047

mcantelon:
* Fixed class naming inconsistencies
* Made all REST API exception instances return an error message
* Added logic so errors are returned as JSON


mcantelon:
I ended up fixing it in the QubitApiAction class (didn't see your update in time, Jesús), but I could re-implement if it'd be more elegant.

mcantelon:
Oops, found an issue with it. Closing PR until I fix.

PR #523, opened 2017-02-22
Fix duplication issues for alternative identifiers and notes, refs #8400

MikeFE:
This ticket is actually 3 separate but related bugs. This PR fixes all 3 (each with their own commit). The bugs were:

- When duplicating an archival description, if the user modifies an existing _note_ that was copied from the source archival description before hitting save, the change to the note also changes in the source archival description.

- When duplicating an archival description, if the user modifies an existing _alternative identifier_ that was copied from the source archival description before hitting save, the change to the alt identifier also changes in the source archival description.

- When duplicating an archival description, any alternative identifiers from the source archival description don't actually get copied to the new duplicate record (despite showing in the edit screen). So that means when the user hits save, said alternative identifier won't be saved to the new record. If the user modified that alt identifier in the edit page, it'll be modified in the source record (2nd bug in this list).

I feel like these 3 commits should just be merged as is without squashing, since they're 3 different, self-contained fixes. What do you guys think?


sevein (approved):
Nicely done! 👍

PR #524, opened 2017-02-22
Dev/issue 10581 exclude type fixes/cleanup

mcantelon:
* Removed un-implemented task option
* Added task option to display available document types
* Added mention of new task option to task's help text
* Fixed inaccurate comment in search plugin code

### DIFF
--- a/lib/task/search/arPopulateTask.class.php
+++ b/lib/task/search/arPopulateTask.class.php
@@ -30,8 +30,7 @@ class arSearchPopulateTask extends sfBaseTask
     $this->addOptions(array(
       new sfCommandOption('application', null, sfCommandOption::PARAMETER_OPTIONAL, 'The application name', 'qubit'),
       new sfCommandOption('env', null, sfCommandOption::PARAMETER_REQUIRED, 'The environment', 'cli'),
-      new sfCommandOption('exclude-types', null, sfCommandOption::PARAMETER_OPTIONAL, 'Exclude document type(s) (command-separated) from indexing'),
-      new sfCommandOption('verbose', 'v', sfCommandOption::PARAMETER_NONE, 'If passed, progress is displayed for each object indexed')));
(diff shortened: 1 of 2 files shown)


lib/task/search/arPopulateTask.class.php
```diff
@@ -44,17 +44,33 @@ protected function configure()
...
     $excludeTypes = (!empty($options['exclude-types'])) ? explode(',', strtolower($options['exclude-types'])) : null;
 
     QubitSearch::getInstance()->populate($excludeTypes);
   }
+
+  protected function availableDocumentTypes()
+  {
+    $types = array_keys(QubitSearch::getInstance()->loadMappings()->asArray());
+    sort($types);
+    return $types;
```
MikeFE: return sort($types); ?

lib/task/search/arPopulateTask.class.php
```diff
@@ -44,17 +44,33 @@ protected function configure()
...
+    }
+
     new sfDatabaseManager($this->configuration);
 
     $excludeTypes = (!empty($options['exclude-types'])) ? explode(',', strtolower($options['exclude-types'])) : null;
 
     QubitSearch::getInstance()->populate($excludeTypes);
   }
+
+  protected function availableDocumentTypes()
```
MikeFE: can be private, we're unlikely to inherit from this class and need protected

MikeFE:
:+1: other than 2 nitpicks

lib/task/search/arPopulateTask.class.php
mcantelon: Unfortunately PHP's sort is weird and sorts in place, returning a boolean (success or failure).

lib/task/search/arPopulateTask.class.php
mcantelon: Ah true, will change that.

mcantelon:
Merged.

PR #525, opened 2017-02-23
i18n extract translatable settings, refs #10469

jraddaoui:
Extract only translatable settings from /data/fixtures/settings.yml
to the XLIFF files.

- Add array of translatable scopes in QubitSetting
- Add scope check in QubitI18nConsolidatedExtract
- Remove unneeded translations in /data/fixtures/settings.yml


sevein (approved):
👏

PR #526, opened 2017-02-23
Use JSON in API errors, refs #10047

mcantelon:
Redmine: https://projects.artefactual.com/issues/10047

* Added catch-all /api route to generate endpoint not found exception
* Added filter to handle API exceptions and return all errors (except sfStopException) as JSON
* Fixed class naming inconsistencies
* Made all REST API exception instances specify an error message

### DIFF
--- a/plugins/arRestApiPlugin/lib/arRestApiPluginFilter.class.php
+++ b/plugins/arRestApiPlugin/lib/arRestApiPluginFilter.class.php
@@ -1,0 +1,73 @@
+<?php
+
+/*
+ * This file is part of the Access to Memory (AtoM) software.
+ *
+ * Access to Memory (AtoM) is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * Access to Memory (AtoM) is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with Access to Memory (AtoM).  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+class arRestApiPluginFilter extends sfFilter
+{
+  public function execute($filterChain)
+  {
+    sfConfig::set('sf_web_debug', false);
+
+    try
+    {
+      $filterChain->execute();
+    }
+    catch (QubitApi404Exception $e)
+    {
+      $this->setErrorResponse(404, 'not-found', $e->getMessage());
+    }
+    catch (QubitApiNotAuthorizedException $e)
+    {
+      header('HTTP/1.0 401 Unauthorized');
+      $this->setErrorResponse(401, 'not-authorized', $e->getMessage());
+    }
+    catch (QubitApiForbiddenException $e)
+    {
+      $this->setErrorResponse(403, 'forbidden', $e->getMessage());
+    }
+    catch (QubitApiBadRequestException $e)
+    {
+      $this->setErrorResponse(400, 'bad-request', $e->getMessage());
+    }
+    catch (sfStopException $e)
+    {
+      // Ignore stop exceptions
+      return;
+    }
+    catch (Exception $e)
+    {
+      $this->setErrorResponse(500, 'internal-error', $e->getMessage());
+    }
+  }
+
+  private function setErrorResponse($httpStatusCode, $errorId, $errorMessage)
+  {
+    $response = sfContext::getInstance()->response;
+
+    $response->setHttpHeader('Content-Type', 'application/json; charset=utf-8');
+    $response->setStatusCode($httpStatusCode);
+
+    // Translate exception into response data
+    $responseData = array(
+      'id' => $errorId,
+      'message' => $errorMessage
+    );
+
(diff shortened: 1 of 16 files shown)


plugins/arRestApiPlugin/lib/arRestApiPluginFilter.class.php
```diff
@@ -0,0 +1,73 @@
...
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with Access to Memory (AtoM).  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+class arRestApiPluginFilter extends sfFilter
+{
+  public function execute($filterChain)
+  {
+    sfConfig::set('sf_web_debug', false);
```
sevein, line 24: Can you remember why we're doing this?

I'm sure it's still relevant and this is probably a good place, but it'd be nice to add a comment if you remember.

plugins/arRestApiPlugin/modules/api/actions/informationobjectsUpdateAction.class.php
```diff
@@ -30,7 +30,7 @@ protected function put($request, $payload)
 
     if (QubitInformationObject::ROOT_ID === (int)$this->io->id)
     {
-      throw new QubitApiForbiddenException;
+      throw new QubitApiForbiddenException('Forbidden');
```
sevein: It'd be nice to have a default exception message so we don't have to give it one. If you changed `QubitApiForbiddenException` as the following:

```php
class QubitApiForbiddenException extends QubitApiException
{
  const MESSAGE = 'Forbidden';

  const STATUS_CODE = 403;

  public function __construct($message = null, $code = 0, Exception $previous = null)
  {
    $message = is_null($message) ? self::MESSAGE : $message;

    return parent::__construct($message, $code, $previous);
  }

  public function getStatusCode()
  {
    return self::STATUS_CODE;
  }
}
```

Then `throw new QubitApiForbiddenException;` would be enough but we could still add a custom message when necessary. Does that sound like a good idea?

plugins/arRestApiPlugin/lib/exception/QubitApi404Exception.class.php
```diff
@@ -17,6 +17,6 @@
  * along with Access to Memory (AtoM).  If not, see <http://www.gnu.org/licenses/>.
  */
 
-class QubitAPI404Exception extends Exception
+class QubitApi404Exception extends Exception
```
sevein: Should we rename to `QubitApiNotFoundException` to be consistent with the other classes where the status code is never included?

sevein (changes requested):
Looking good. I have one suggestion but it's not critical. If you define a
new intermediate class `QubitApiException` and establish the following
hierarchy:

```
QubitApiException
 ├── QubitNotFoundException
 ├── QubitApiNotAuthorizedException
 ├── QubitApiForbiddenException
 └── QubitApiBadRequestException
```

You could add some more context to the exception (status code, etc...) and use
that to simplify the implementation of the `arRestApiPluginFilter` filter,
e.g.: (quick draft, have not tested)

```php
class arRestApiPluginFilter extends sfFilter
{
  public function execute($filterChain)
  {
    sfConfig::set('sf_web_debug', false);
    try
    {
      $filterChain->execute();
    }
    catch (sfStopException $e)
    {
      // Ignore stop exceptions
      return;
    }
    catch (Exception $e)
    {
      $this->setErrorResponse($e);
    }
  }

  private function setErrorResponse(Exception $error)
  {
    $response = sfContext::getInstance()->response;
    $response->setHttpHeader('Content-Type', 'application/json; charset=utf-8'

    // Defaults for unknown API exception.
    $statusCode = 500;
    $errorMsg = 'Internal error';
    $errorId = 'internal-error';

    if (is_subclass_of($error, 'QubitApiException'))
    {
      $statusCode = $error->getStatusCode();
      $errorMsg = $error->getMessage();
      $errorId = $error->getId();
    }

    // Translate exception into response data
    $payload = arRestApiPluginUtils::arrayToJson(array('id' => $errorId, 'message' => $errorMsg));

    $response->setContent($payload);
  }
}
```

You could even define a new `QubitApiUnknownException` that wraps the unknown exception and have the signature of `setErrorResponse` only receive QubitApiException instances (type hinting). That way you wouldn't even need to call `is_subclass_of` and have the defaults for the unknown error in `QubitApiUnknownException` like the others (`500`, `Internal error`, etc...).

plugins/arRestApiPlugin/lib/arRestApiPluginFilter.class.php
mcantelon: Hrm... good question. I'll see what happens when I take it out (maybe in dev mode filters inject some debugging-related content?).

plugins/arRestApiPlugin/modules/api/actions/informationobjectsUpdateAction.class.php
mcantelon: Good idea!

sevein: I think this is a better approach :smile: 
https://github.com/symfony/http-kernel/tree/master/Exception

plugins/arRestApiPlugin/lib/arRestApiPluginFilter.class.php
sevein: Right, I think it's safer if we leave it! Thanks for taking a look.

sevein (approved):
👍

mcantelon:
Thanks!!! Merged.

PR #528, opened 2017-02-23
Upgrade ES to 5.0, refs #10847

sbreker:
Redmine: https://projects.artefactual.com/issues/10847


jraddaoui:
It's looking great @sbreker!

mcantelon:
Good on you for taking this on, Steve... tons of work!

sevein:
@sbreker, there are two commits in your branch changing permission bits and a third one reverting the changes. We'll squash into one commit but I think it'd be better if it gets fixed before.

sbreker:
Hey @sevein! Yes, I noticed that after the fact and corrected the perms in the later commit.  I figured it would disappear during the final squash/merge/commit.  When you say 'fixed before', what/how do you mean?

sevein:
> When you say 'fixed before', what/how do you mean?

I meant that you could try to rewrite your commits so we don't confuse the reviewer with unnecessary changes.

PR #529, opened 2017-02-23
Fix exportCsv referer issue, refs #10174

mcantelon:
The exportCsv page (informationobject/exportCsv) was generating a 500 if
visited directly from the browser because it always tries to redirect to
a referer. Fixed by making sure the referer is a valid URL before
redirecting (otherwise redirecting to the information objects browse
page.)


MikeFE:
:+1:

mcantelon:
Thanks! Merged.

PR #530, opened 2017-02-24
Warning translation and CSS fixes, refs #10473

mcantelon:
* Wrapped warning, when base URL not set, with translation helper
* Changed CSS specifier used for warning to a class instead of an ID


MikeFE:
:+1:

mcantelon:
Danke! Merged.

PR #532, opened 2017-03-01
Fix OAI virtual set enabling bug, refs #10929

mcantelon:
Fixed a bug that prevented OAI virtual sets from working.


MikeFE:
hahaha, prefixing oai_ to all the setting names in the database seems kind of silly/redundant if the scope is oai as well already, right? But changing it all is probably beyond the scope of fixing this bug for now.

:+1:

mcantelon:
Yeah, it is kind of weird haha. Thanks for the CR!

PR #533, opened 2017-03-09
Improve sorting options to term page, refs #10958

MikeFE:
Added sort by date and reference code when on a term page that has linked
archival description results.

### DIFF
--- a/apps/qubit/modules/term/templates/indexSuccess.php
+++ b/apps/qubit/modules/term/templates/indexSuccess.php
@@ -254,11 +254,12 @@
           <?php endif; ?>
 
         <div id="sort-header">
-          <?php echo get_partial('default/sortPicker',
-            array(
-              'options' => array(
-                'lastUpdated' => __('Most recent'),
-                'alphabetic' => __('Alphabetic')))) ?>
+          <?php echo get_partial('default/sortPicker', array(
+            'options' => array(
+              'lastUpdated' => __('Most recent'),
+              'alphabetic'  => __('Alphabetic'),
+              'identifier'  => __('Reference code'),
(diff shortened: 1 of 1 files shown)


apps/qubit/modules/term/templates/indexSuccess.php
```diff
@@ -254,11 +254,12 @@
...
-            array(
-              'options' => array(
-                'lastUpdated' => __('Most recent'),
-                'alphabetic' => __('Alphabetic')))) ?>
+          <?php echo get_partial('default/sortPicker', array(
+            'options' => array(
+              'lastUpdated' => __('Most recent'),
+              'alphabetic'  => __('Alphabetic'),
+              'identifier'  => __('Reference code'),
+              'date'        => __('Date')))) ?>
```
sevein, line 262: You need to update [this](https://github.com/artefactual/atom/blob/a93ba709ba3b65e781c02bd5af1fff1c529e5095/apps/qubit/modules/term/actions/indexAction.class.php#L276-L288) too, right?

MikeFE: I thought so at first, but this change applies to linked descriptions underneath the terms indexSuccess page, not to the terms themselves. See http://imgur.com/a/1dlJa for clarification.

MikeFE: the ref code and date case is handled here in the linked descriptions, just like the regular descriptions browse page: https://github.com/artefactual/atom/blob/qa/2.4.x/apps/qubit/modules/informationobject/actions/browseAction.class.php#L527

sevein (approved):

PR #535, opened 2017-03-09
Term delete skip nested set option, refs #10962

mcantelon:
Added the ability to, when programatically deleting a term node, set the
disableNestedSetUpdating to true so the nested set isn't updated during
the delete. This allows much faster batch deletion of terms and the nested
set can be updated manually afterwards.

### DIFF
--- a/lib/model/QubitTerm.php
+++ b/lib/model/QubitTerm.php
@@ -303,6 +303,7 @@ class QubitTerm extends BaseTerm
     {
       foreach ($children as $child)
       {
+        $child->disableNestedSetUpdating = $this->disableNestedSetUpdating;
         $child->delete($connection);
       }
     }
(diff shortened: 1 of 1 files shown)


lib/model/QubitTerm.php
```diff
@@ -336,7 +337,24 @@ public function delete($connection = null)
 
     QubitSearch::getInstance()->delete($this);
 
-    parent::delete($connection);
+    // Replicate parent class delete functionality, skipping nested set updating
```
MikeFE, line 340: Is this a very common feature we'll need? I'm pretty uneasy about this change since it seems we're explicitly skipping over BaseTerm::delete() and going straight to QubitObject::delete(). BaseTerm::delete might not do much right now, but in the future if functionality needs to be added to it it won't be expected at all that our QubitTerm is bypassing it.

If this isn't a common feature we'll need I suggest just stashing a .patch somewhere for future reference if it comes up again.

mcantelon: That's a valid take. It's not something we'll commonly need... only thing I can think of is for migrations where you're migrating onto an estisting AtoM instance where data has to be pruned off. Managing patches is a bit of a pain, but yes, this PR could present a pitfall in the future. Maybe I should take another look at modifying the parent class (which would be trivial, except it gets auto-generated by a builder class).

jraddaoui: I agree with @MikeFE.

@mcantelon have you investigated how hard would it be to do this for all models with nested set in their base class, by extending and modifying sfPropelBuildModelTask and related classes. I'd guess it's not an easy change, but I think it will be safer than this.

mcantelon: Yeah, agreed... I'll take a look at this approach.

PR #537, opened 2017-03-15
Add FW treeview display settings, refs #10909

jraddaoui:
Add new settings to choose what is displayed in the full width treeview nodes,
available options are: identifier/referenceCode/none, LoD yes/no, dates yes/no.

All treeview settings have been relocated to a new setting page.

Redmine: https://projects.artefactual.com/issues/10909


sbreker:
Looks great Radda.  I was wondering about the params fields that start with ':' like ':id' and ':culture' but that appears to be related to what QubitPdo is expecting.
👍

jraddaoui:
Merged in https://github.com/artefactual/atom/commit/b4e633774d18325cd1cfa1b4a74b54c9d0029955

PR #539, opened 2017-03-20
Fix term-related io results sorting, refs #10958

MikeFE:
The previous work to sort info objects related to terms by reference code or
date was missing some logic in the controller--this commit adds the proper
logic to do it.


mcantelon:
LGTM!

PR #540, opened 2017-03-21
Add D&D support to FW treeview, refs #10973

jraddaoui:
Add ability to change order of siblings via drag-and-drop in the full-width treeview

Main changes:
- Add and configure dnd plugin for jsTree
- Add move request on node drop and show alerts with result, reload tree if fail
- Add FW treeview move action to validate request, launch arObjectMoveJob
  and respond with the success or error message in JSON
- Allow moving between siblings in arObjectMoveJob
- Add system to run sensitive jobs in serie, avoiding them to run in parallel
  if multiple worker instances are available

Other changes:
- Remove app-alert class, only used to add margin
- Improvements in arObjectMoveJob logging
- Make parent id param optional in arObjectMoveJob

### DIFF
--- a/apps/qubit/modules/informationobject/actions/fullWidthTreeViewMoveAction.class.php
+++ b/apps/qubit/modules/informationobject/actions/fullWidthTreeViewMoveAction.class.php
@@ -96,7 +96,7 @@ class InformationObjectFullWidthTreeViewMoveAction extends sfAction
       'targetPosition' => $newPosition > $oldPosition ? 'after' : 'before'
     );
 
-    // QubitJob::runJob('arObjectMoveJob', $params);
+    QubitJob::runJob('arObjectMoveJob', $params);
 
     sfContext::getInstance()->getConfiguration()->loadHelpers(array('Url'));
 
--- a/apps/qubit/modules/informationobject/actions/fullWidthTreeViewMoveAction.class.php
+++ b/apps/qubit/modules/informationobject/actions/fullWidthTreeViewMoveAction.class.php
@@ -52,11 +52,39 @@ class InformationObjectFullWidthTreeViewMoveAction extends sfAction
     $oldPosition = $request->getParameter('oldPosition');
     $newPosition = $request->getParameter('newPosition');
 
-    if (empty($oldPosition) || empty($newPosition))
+    // Empty or non numeric positions are not allowed
+    if (!is_numeric($oldPosition) || !is_numeric($newPosition))
     {
       $this->response->setStatusCode(400);
 
-      return $this->renderText(json_encode(array('error' => $i18n->__('Move failed: new and old positions required'))));
+      return $this->renderText(json_encode(array('error' => $i18n->__('Move failed: new and old positions required as numbers'))));
+    }
+
+    // Moving to the same position
+    if ($oldPosition == $newPosition)
+    {
+      $this->response->setStatusCode(400);
+
+      return $this->renderText(json_encode(array('error' => $i18n->__('Move not needed: new and old positions are the same'))));
+    }
+
+    // Check current positions to avoid mismatch
+    $sql = "SELECT id FROM information_object WHERE parent_id = :parentId ORDER BY lft;";
+    $params = array(':parentId' => $this->resource->parentId);
+    $children = QubitPdo::fetchAll($sql, $params, array('fetchMode' => PDO::FETCH_ASSOC));
+
+    if (array_search(array('id' => $this->resource->id), $children) != $oldPosition)
+    {
+      $this->response->setStatusCode(400);
+
+      return $this->renderText(json_encode(array('error' => $i18n->__('Move failed: mismatch in current position'))));
+    }
+
+    if ($newPosition >= count($children))
+    {
+      $this->response->setStatusCode(400);
+
+      return $this->renderText(json_encode(array('error' => $i18n->__('Move failed: new position outside the range'))));
     }
 
     $this->response->setStatusCode(201);
--- a/apps/qubit/modules/informationobject/actions/fullWidthTreeViewMoveAction.class.php
+++ b/apps/qubit/modules/informationobject/actions/fullWidthTreeViewMoveAction.class.php
@@ -93,7 +93,7 @@ class InformationObjectFullWidthTreeViewMoveAction extends sfAction
     $jobManageLink = '<a href="'. $jobManageUrl . '">'. $i18n->__('job management') .'</a>';
 
     $message = '<strong>'. $i18n->__('Move initiated.') .'</strong> ';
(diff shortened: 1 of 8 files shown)


apps/qubit/modules/informationobject/actions/fullWidthTreeViewMoveAction.class.php
```diff
@@ -0,0 +1,102 @@
...
+
+      return $this->renderText(json_encode(array('error' => $i18n->__('Move failed: new and old positions required as numbers'))));
+    }
+
+    // Moving to the same position
+    if ($oldPosition == $newPosition)
+    {
+      $this->response->setStatusCode(400);
+
+      return $this->renderText(json_encode(array('error' => $i18n->__('Move not needed: new and old positions are the same'))));
```
sbreker, line 68: Instead of throwing an error here - is it possible to just throw a status message or something stating "Item not moved - nothing done", short-circuit and leave them where they are at?  This isn't really an error, and to throw them to an error page seems overkill.

apps/qubit/modules/informationobject/actions/fullWidthTreeViewMoveAction.class.php
```diff
@@ -0,0 +1,102 @@
...
+    {
+      $this->response->setStatusCode(401);
+
+      return $this->renderText(json_encode(array('error' => $i18n->__('Move not allowed: log in required'))));
+    }
+
+    // Check user authorization
+    if (!QubitAcl::check($this->resource, 'update'))
+    {
+      $this->response->setStatusCode(403);
```
sbreker, line 47: Use QubitAcl::forwardUnauthorized();  ?

apps/qubit/modules/informationobject/actions/fullWidthTreeViewMoveAction.class.php
sbreker: Ajax flash message?

lib/job/arBaseJob.class.php
```diff
@@ -263,4 +292,31 @@ protected function signOut()
...
+     AND job.name IN $jobNames
+     ORDER BY object.created_at;";
+
+    $params = array(':statusId' => QubitTerm::JOB_STATUS_IN_PROGRESS_ID);
+
+    // It will be at least one sensitive job running, this one
+    $runningJobs = QubitPdo::fetchAll($sql, $params, array('fetchMode' => PDO::FETCH_ASSOC));
+
+    // If this job is the first one, it can be fully executed
+    return $this->job->id === $runningJobs[0]['id'];
```
sbreker: If you haven't, I'd check to make sure that there is no strange interaction between a waiting mutex job, and an admin running 'jobs:clear' from the CLI.  It should never occur under expected circumstances, but is it worth adding a check for the condition where zero job records are returned?

sbreker (commented):
Just a few comments regarding user experience and one regarding a rather unlikely edge-case. Overall looks really great!

apps/qubit/modules/informationobject/actions/fullWidthTreeViewMoveAction.class.php
jraddaoui: Hi Steve, this action returns "always" a JSON response, that's why I'm setting the status this way.

apps/qubit/modules/informationobject/actions/fullWidthTreeViewMoveAction.class.php
jraddaoui: Yes, this error messages are added like that so they can be displayed as alerts in the GUI after the Ajax request. But this is a special one, as we have the same check before the call in the JS code, so you should never see it in normal circumstances. However, this could be reached in the future differently and I want to let the requester know this is a bad request (400) instead of a successful (200) or an accepted (201) one.

lib/job/arBaseJob.class.php
jraddaoui: Good catch Steve, the JobsDeleteAction doesn't delete running jobs, but I didn't know about this task and it looks like it's only showing a warning, so running jobs could be actually deleted. I'll add check, but in that case something else bad is going to happen ...

lib/job/arBaseJob.class.php
```diff
@@ -312,10 +312,14 @@ protected function canBeFullyExecuted()
      ORDER BY object.created_at;";
 
     $params = array(':statusId' => QubitTerm::JOB_STATUS_IN_PROGRESS_ID);
-
-    // It will be at least one sensitive job running, this one
     $runningJobs = QubitPdo::fetchAll($sql, $params, array('fetchMode' => PDO::FETCH_ASSOC));
 
+    // Edge case where the QubitJobs are cleared while this one is waiting
+    if (count($runningJobs) == 0)
```
sbreker: nitpick: Yoda order should be used.

lib/job/arBaseJob.class.php
```diff
@@ -44,12 +44,13 @@ class arBaseJob extends Net_Gearman_Job_Common
...
-   * the amount of seconds indicated in $waitForRetryTime, until the previous ones are finished.
-   * Due to the limitations of the Gearman job server, the waiting jobs will block the workers
-   * executing them until they are finished.
+   * the amount of seconds indicated in $waitForRetryTime, until the previous ones are finished
+   * or the maximun amount of tries ($maxTries) is reached. Due to the limitations of the Gearman
+   * job server, the waiting jobs will block the workers executing them until they are ended.
    */
   protected $avoidParallelExecutionJobs = array('arObjectMoveJob', 'arFileImportJob');
   protected $waitForRetryTime = 10;
+  Protected $maxRetries = 10;
```
sbreker: It'd be nice if these vars were individually configurable per job type - but that's likely out of scope now with only a few hours left. Perhaps can be addressed if/when the critical section guarding is changed to use db locks at some point.  We may discover a need to tune the timeouts in production if run time varies from the testing system.

lib/job/arBaseJob.class.php
jraddaoui: Any reason?

lib/job/arBaseJob.class.php
jraddaoui: I thought about that, they could be overwritten in the job sub-classes to have values per job, but in the end the wait time depends more on the blocking job than on the waiting job, and that was a little more work.

This wait/retry system will only be used in sites with multiples workers available and when various sensitive jobs are running at the same time. I think we only setup a worker in our production sites at the moment, so that shouldn't be a problem.

lib/job/arBaseJob.class.php
sbreker: You're right - it would blow up with a typo '=' regardless.  Please disregard.

jraddaoui:
Merged in https://github.com/artefactual/atom/commit/fe5db3029f4ae292ca2437f9b517d05cdac73c1c

PR #542, opened 2017-03-23
Skip objects during XML cache export. refs #11002

mcantelon:
Added an option to the cache::xml-representations CLI task to optionally skip
information objects during export. This allows interrupted or aborted exports
to be resumed.

### DIFF
--- a/lib/QubitInformationObjectXmlCache.class.php
+++ b/lib/QubitInformationObjectXmlCache.class.php
@@ -81,12 +81,17 @@ class QubitInformationObjectXmlCache
    *
    * @return null
    */
-  public function exportAll()
+  public function exportAll($options = array())
   {
+    $skip = isset($options['skip']) ? $options['skip'] : 0;
+
     // Get not-root and published information objects
     $criteria = new Criteria;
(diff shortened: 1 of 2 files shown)


lib/QubitInformationObjectXmlCache.class.php
```diff
@@ -81,12 +81,17 @@ public function export($resource)
    *
    * @return null
    */
-  public function exportAll()
+  public function exportAll($options = array())
   {
+    $skip = isset($options['skip']) ? $options['skip'] : 0;
```
sevein, line 86: Do you have to do this? I'd assume `0` is the default if the user didn't pass the argument.

sevein (approved):
LGTM! Minor issue mentioned, not a big deal.

lib/QubitInformationObjectXmlCache.class.php
mcantelon: Just doing that to keep the class self contained (avoiding an "Undefined index" notice in case anything else ends up calling the method without providing that in the $options array).

mcantelon:
Awesome... thanks!

PR #543, opened 2017-03-28
Batch mode support for ES deletes, refs #10029

mcantelon:
Add the ability to delete ElasticSearch documents using batch mode and don't
process the deletion batch until add batched document creation has occurred.

### DIFF
--- a/plugins/arElasticSearchPlugin/lib/arElasticSearchPlugin.class.php
+++ b/plugins/arElasticSearchPlugin/lib/arElasticSearchPlugin.class.php
@@ -42,13 +42,21 @@ class arElasticSearchPlugin extends QubitSearchEngine
   public $index = null;
 
   /**
-   * elasticsearch bulk API makes it possible to perform many index/delete
-   * operations in a single call. This can greatly increase the indexing speed.
-   * This array will be used to store documents in batches.
+   * Elasticsearch bulk API makes it possible to perform many operations in a
+   * single call. This can greatly increase the indexing speed.
+   *
+   * This array will be used to store documents to add in a batch.
    *
    * @var array
    */
-  private $batchDocs = array();
+  private $batchAddDocs = array();
+
+  /**
+   * This array will be used to store documents to delete in a batch.
+   *
+   * @var array
+   */
+  private $batchDeleteDocs = array();
 
   /**
    * Mappings configuration, mapping.yml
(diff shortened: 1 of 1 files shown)


mcantelon:
I could add some logic to the flushBatch method to not create a document if it's slated to be deleted, but that'd take a little more time and add some code complexity (and adding then deleting a bit of an edge case) so I've kept it simple for now.

plugins/arElasticSearchPlugin/lib/arElasticSearchPlugin.class.php
```diff
@@ -484,13 +515,27 @@ public function delete($object)
...
     {
-      $this->index->getType(get_class($object))->deleteById($object->id);
+      // The document being deleted may not have been added to the index yet (if it's
+      // still queued up in $this->batchAddDocs) so create a document object representing
+      // the document to be deleted and add this document object to the batch delete
+      // queue.
+      $document = new \Elastica\Document($object->id);
+      $document->setType(get_class($object));
+
+      $this->batchDeleteDocs[] = $document;
```
sbreker, line 527: At the end of addDocument() a call to flushBatch() is made if the number of batchAddDocs is greater than batchSize. In delete(), there is no similar concept - will the queue of batchDeleteDocs only get flushed when addDocument finally triggers flushBatch()?  Should a similar condition and call to flushBatch() be included in delete() here?

plugins/arElasticSearchPlugin/lib/arElasticSearchPlugin.class.php
```diff
@@ -405,14 +437,13 @@ public function addDocument($data, $type)
...
+      $this->batchAddDocs[] = $document;
 
-      // If we have a full batch, send in bulk
-      if (count($this->batchDocs) >= $this->batchSize)
+      // If we have a full batch, send additions and deletions in bulk
+      if (count($this->batchAddDocs) >= $this->batchSize)
       {
-        $this->index->addDocuments($this->batchDocs);
-        $this->batchDocs = array();
+        $this->flushBatch();
```
sbreker, line 446: flushBatch() is called here but not in delete() - see other comment.

plugins/arElasticSearchPlugin/lib/arElasticSearchPlugin.class.php
mcantelon: You're right... queued deletes could sit there infinitely. I'll fix!

sbreker:
👍   Looks good!

mcantelon:
Closing (this is merged, but looked like I didn't merge it in the way that auto-closes the issue - oops).

PR #544, opened 2017-03-29
Sidebar treeview quick search fixes, refs #10173

jraddaoui:
- Add culture fallback in results title
- Avoid warning when treeview is set to full-width mode
- Remove treeview margin

### DIFF
--- a/apps/qubit/modules/informationobject/templates/_treeView.php
+++ b/apps/qubit/modules/informationobject/templates/_treeView.php
@@ -13,7 +13,7 @@
   </li>
 </ul>
 
-<div id="treeview" data-current-id="<?php echo $resource->id ?>" data-sortable="<?php echo $sortable ? 'true' : 'false' ?>">
+<div id="treeview" data-current-id="<?php echo $resource->id ?>" data-sortable="<?php echo empty($sortable) ? 'false' : 'true' ?>">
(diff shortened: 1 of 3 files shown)


apps/qubit/modules/informationobject/templates/_treeView.php
```diff
@@ -13,7 +13,7 @@
   </li>
 </ul>
 
-<div id="treeview" data-current-id="<?php echo $resource->id ?>" data-sortable="<?php echo $sortable ? 'true' : 'false' ?>">
+<div id="treeview" data-current-id="<?php echo $resource->id ?>" data-sortable="<?php echo isset($sortable) && $sortable ? 'true' : 'false' ?>">
```
MikeFE: I wonder if empty could work here to make this shorter

MikeFE:
:+1:

apps/qubit/modules/informationobject/templates/_treeView.php
jraddaoui: Great point Mike. From [PHP: empty](http://php.net/manual/en/function.empty.php) ...

> No warning is generated if the variable does not exist. That means empty() is essentially the concise equivalent to !isset($var) || $var == false.

PR #545, opened 2017-03-29
Update IO LODs when normalizing LODs, refs #10698

jraddaoui:
When the normalice taxonomy task is executed over the levels of description
taxonomy, the information object table needs to be updated accordingly.

- Extend arBaseTask in taxonomyNormalizeTask
- Use QubitPdo for raw SQL queries in taxonomyNormalizeTask
- Small logging improvements with search index note
- Cosmetic changes

### DIFF
--- a/lib/task/taxonomy/taxonomyNormalizeTask.class.php
+++ b/lib/task/taxonomy/taxonomyNormalizeTask.class.php
@@ -102,21 +102,23 @@ EOF;
 
   protected function populateTaxonomyNameUsage(&$names, $taxonomyId, $culture)
   {
-    $sql = "SELECT t.id, i.name FROM term t \r
-      INNER JOIN term_i18n i ON t.id=i.id \r
-      WHERE t.taxonomy_id=? AND i.culture=? \r
+    $sql = "SELECT t.id, i.name FROM term t
+      INNER JOIN term_i18n i ON t.id=i.id
+      WHERE t.taxonomy_id=:id AND i.culture=:culture
       ORDER BY t.id";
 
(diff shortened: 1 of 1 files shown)


lib/task/taxonomy/taxonomyNormalizeTask.class.php
```diff
@@ -67,19 +67,21 @@ protected function execute($arguments = array(), $options = array())
...
-    $this->log("Normalizing for '". $options['culture'] ."' culture...");
-
     // Look up taxonomy ID using name
     $taxonomyId = $this->getTaxonomyIdByName($arguments['taxonomy-name'], $options['culture']);
     if (!$taxonomyId)
     {
       throw new sfException("A taxonomy named '". $arguments['taxonomy-name'] ."' not found for culture '". $options['culture'] ."'.");
     }
 
+    $this->log("Normalizing for '". $options['culture'] ."' culture...");
```
mcantelon: Minor nitpick: using sprintf rather than string concatenation might be cleaner.

mcantelon (approved):
Nicely done! I made a comment on possible further cleanup by using sprintf rather than string concatenation, but on second thoughts there's string concatenation used throughout the script so might be better to just leave it as-is for now.

lib/task/taxonomy/taxonomyNormalizeTask.class.php
mcantelon: On second thought, concatenation's used throughout the script and might be more time than it's worth to change. :/

jraddaoui:
Thanks Mike. Merged in https://github.com/artefactual/atom/commit/dd6b087573d2ad012fd89ea4e9c8c1af10c9abdd

PR #546, opened 2017-03-29
Fix clipboard entity type issues, refs #10011

MikeFE:
- Make CSS for add clipboard button during repository card view consistent with archival description card view
- Fix bug where adding repositories to clipboard when on the repository view page (as opposed to browsing/searching) was broken.


jraddaoui (approved):
👍

PR #547, opened 2017-03-29
Fix "Use For" label for browsing terms ref #10949

MikeFE:
The code was incorrect when rendering the "Use For" alternative labels when
browsing terms. It'd always show "Untitled" before if you ever displayed a
term's Use For label(s).


jraddaoui (approved):
Looks good, but I think it may have been better to do it in a helper function in QubitHelper. Feel free to leave it as it is if you don't have the time to change it.

PR #541, opened 2017-03-22
New inst. browse menu item refs #10733

timhutch:
Adds menu item for browse digital objects, child of institutional browse.

Since the browse digital objects page has been reconfigured as a filtered view of the information objects browse, this allows the addition of a scoped digital objects browse as part of the "virtual institutional database" functionality.

PR #550, opened 2017-04-11
Dev/issue 9970 job schema fix

mcantelon:
Added a new taxonomy, and corresponding terms, for the three job statuses (completed, in progress, and halted due to an error). Make the statusId property of the QubitJob model a foreign key to QubitTerm. Added migration to make these changes on old versions of the database.

### DIFF
--- a/data/fixtures/settings.yml
+++ b/data/fixtures/settings.yml
@@ -3,7 +3,7 @@ QubitSetting:
     name: version
     editable: 0
     deleteable: 0
-    value: 149
+    value: 152
   milestone:
     name: milestone
     editable: 0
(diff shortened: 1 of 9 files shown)


data/fixtures/settings.yml
```diff
@@ -3,7 +3,7 @@ QubitSetting:
     name: version
     editable: 0
     deleteable: 0
-    value: 149
+    value: 150
```
sevein: This should be moved to 152.

sevein: Related: #515 and #541.

sevein (changes requested):
@mcantelon, this is looking great. Glad we're fixing this.

You may have noticed that previous PRs did not bump the db version in the fixture. You have a chance to solve that here if you want. Related: #515 and #541.

data/fixtures/settings.yml
mcantelon: Ah, I didn't notice that the bumped number was less than my migration. Will fix. Thanks!

sbreker:
Thanks Mike.  Good catch Sevein!  I completely forgot about the version
setting.

PR #552, opened 2017-04-13
Added term usage CSV export task, refs #11058

mcantelon:
Added a task to export, as CSV, terms, in a given taxonomy, that have been
assocated with information objects.

### DIFF
--- a/lib/flatfile/config/QubitTerm.yml
+++ b/lib/flatfile/config/QubitTerm.yml
@@ -1,0 +1,8 @@
+# Columns that map directly to term properties
+direct:
+  - id
+  - parentId
+  - taxonomyId
+  - name
+  - sourceCulture
+  - culture
--- a/lib/task/export/csvExportTermUsageTask.class.php
+++ b/lib/task/export/csvExportTermUsageTask.class.php
@@ -1,0 +1,138 @@
+<?php
+
+/*
+ * This file is part of the Access to Memory (AtoM) software.
+ *
+ * Access to Memory (AtoM) is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * Access to Memory (AtoM) is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with Access to Memory (AtoM).  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+/**
+ * Bulk export data about usage of term as CSV
+ *
+ * @package    symfony
+ * @subpackage task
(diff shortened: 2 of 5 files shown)


lib/flatfile/config/QubitTerm.yml
```diff
@@ -0,0 +1,9 @@
+# Columns that map directly to term properties
+direct:
+  - id
+  - parentId
+  - taxonomyId
+  - code
+  - sourceCulture
+  - name
+  - culture
```
jraddaoui: Do we really need this file?

lib/task/export/csvExportTermUsageTask.class.php
```diff
@@ -0,0 +1,139 @@
...
+      $criteria->add(QubitTaxonomy::ID, $options['taxonomy-id']);
+
+      if (null === QubitTaxonomy::getOne($criteria))
+      {
+        throw new sfException('Invalid taxonomy-id.');
+      }
+
+      return $options['taxonomy-id'];
+    }
+    else if(isset($options['taxonomy-name']))
```
jraddaoui: We normally use elseif and it's missing the space after the if.

lib/task/export/csvExportTermUsageTask.class.php
```diff
@@ -0,0 +1,139 @@
...
+    /*
+
+    Idea for future:
+
+    1. Count uses with SELECT DISTINCT t.id, COUNT(object_id) AS uses, r.object_id FROM term t INNER JOIN object_term_relation r ON r.term_id=t.id WHERE t.taxonomy_id=35 GROUP BY(id) ORDER BY t.id;
+    2. Add uses to "usage" export config and change QubitEport so setColumn will work before exportResource
+    3. LEFT JOIN could include unused terms so we could count those too
+    4. Add --mode=used|both|unused-only ("used" as default) to flip between LEFT and INNER join (and maybe add criteria for specific modes)
+
+    */
```
jraddaoui: Could you remove the empty lines at the beginning and end and add one after the comment block? And I prefer the full block ...

/*
 *
 *
 */

lib/task/export/csvExportTermUsageTask.class.php
```diff
@@ -0,0 +1,139 @@
...
+    $format = 'SELECT DISTINCT t.id FROM %s t INNER JOIN %s r ON r.term_id=t.id WHERE t.taxonomy_id=? ORDER BY t.id';
```
jraddaoui: I think we should add another JOIN in this query with the object table, and check the class_name column to find only relations with QubitInformationObject.

jraddaoui (changes requested):
Looks great Mike! Just a few notes.

jraddaoui:
It looks like the commit message refers to a client ticket, could you create a public one explaining the task, etc. And I'm not sure if @fiver-watson would like to add some docs about it.

mcantelon:
Updated and ready for another review... thanks!

mcantelon:
(Oops, didn't mean to close PR.)

MikeFE:
LGTM :+1: -- don't forget to make the merge commit msg point to a public ticket when merging into mainline AtoM (can use Redmine's duplicate ticket functionality if one doesn't exist).

mcantelon:
Thanks Mike!

PR #554, opened 2017-04-19
CSV import language/script fixes, refs #10075

mcantelon:
During CSV import of information objects, language and script-related columns
weren't being checked or having their capitalization normalized. Added error
dispay when invalid values are encountered and capitalization normalization.


MikeFE:
:+1:

PR #555, opened 2017-04-27
Change FW treeview node format, refs #10909

jraddaoui:
Separate the dates using a comma instead of parentheses


MikeFE:
:+1: lgtm

PR #557, opened 2017-05-04
Add clipboard icon to inventory, refs #11119

mcantelon:


apps/qubit/modules/informationobject/templates/inventorySuccess.php
```diff
@@ -34,6 +34,7 @@
           <?php echo get_partial('default/sortableTableHeader',
             array('label' => __('Date'), 'name' => 'date', 'size' => '24%')) ?>
           <th width="8%"><?php echo sfConfig::get('app_ui_label_digitalobject') ?></th>
+          <th />
```
sevein: I don't think that's valid HTML because `th` is a non-void element?

sevein (changes requested):
One small observation.
Also, the author of your commit reads "vagrant".

mcantelon:
Thanks! I've fixed those two issues.

sevein (approved):

PR #558, opened 2017-05-14
Dev/issue 11119 inventory clipboard

mcantelon:
* Add modifications to enable job creation by unauthenticated users
* Support clipboard export by unauthenticated users
* Persist alerts for each unauthenticated job
* Add ability for unauthenticated users to delete individual jobs
* Add class for managing an unauthenticated user's jobs

### DIFF
--- a/apps/qubit/modules/default/actions/alertsComponent.class.php
+++ b/apps/qubit/modules/default/actions/alertsComponent.class.php
@@ -1,0 +1,64 @@
+<?php
+
+/*
+ * This file is part of the Access to Memory (AtoM) software.
+ *
+ * Access to Memory (AtoM) is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * Access to Memory (AtoM) is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with Access to Memory (AtoM).  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+/**
+ * Alerts component
+ *
+ * @package AccesstoMemory
+ * @subpackage default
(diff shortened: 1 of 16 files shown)


mcantelon:
Seems adding this to the delete method of QubitJob might make sense:

```
if (isset($job->downloadPath))
{
  unlink($job->downloadPath);
}
```

EDIT: I ended up adding this to the PR.

sbreker:
I did not pick these commits and test locally.  All looks good in terms of formatting I think.  Looking at the code I have a couple questions:

- The levels of description selector on the export page should be tested to ensure it still works if a user is unauthenticated.  I am unsure if it would be affected by this - the LOD filter does depend on ACL checks in the bg job.

- If I am understanding how this works correctly, the job id's are held in the user context in the unauthenticated user's browser I think?  If so, is there a danger that a malicious request could be created that could contain ANY job id for deletion?

mcantelon:
Thanks Steve. Good call on the levels of description... will test. The job IDs, AFAIK, are only referenced server-side (except for the download file... I'll look at if I can obfuscate that).

mcantelon:
I've added a commit message that's more detailed and added logic to obfuscate job IDs in job download filenames.

sbreker:
Look good Mike!  👍

MikeFE:
Hi guys, I wrote the original jobs system and if I recall (and could have outdated knowledge here), but QubitJob->delete() is primarily just called when you click 'clear jobs' in the job management page (will delete just job rows in the database for that particular user), or when you call the CLI task jobs:clear (will clear all the rows in the jobs table).

The original intention was just to let people remove cluttered job status lists where there can be hanging jobs stuck in 'running' due to technical issues or things like that. The intention wasn't to unlink the files generated from the jobs as well in the process.

I think this would be worth discussing esp. with Dan as well if this is the behavior you want. If it is, I think we should change the jobs:clear task and clear button in the jobs management page to inform the user it's going to delete the generated file too. Otherwise, I think we should create a separate method such as deleteJobFile() and use that instead when required, and not pack that behavior into the generic delete().

MikeFE:
Actually, on second thought, we're using the db entries to let AtoM know where to point to the files anyway, so if those rows are gone the files are only seeable by someone with CLI access at that point. I'm less against this change now. 👍 

However, I feel like we should still warn users if files are about to be deleted.

mcantelon:
Merged! I'll look into adding a confirmation before deletion warning of file deletion.

PR #561, opened 2017-05-17
Add identifier mask for descriptions, refs #11162

MikeFE:
This PR implements a similar feature to our existing accession mask settings,
except for archival descriptions' identifiers instead.

Feature details:

- New admin->settings fields for identifier mask, identifier counter, and
  identifier mask enable / disable.

- If identifier mask is enabled, when creating new descriptions the identifier
  will automatically fill in an auto-generated identifier in the form for the
  user, based on the mask & counter.

- There will be a button underneath the identifier text box when editing that
  will use an XHR to manually generate an identifier from an end point and fill
  the box in.

- In the identifier mask (like accession mask), %<letter> indicates the
  identifier generation should interpret this with strftime. #i should be
  interpreted as the place to insert the current counter value.

- If the user edits the identifier either during the creation process if
  identifier mask is enabled, or after the user clicks generate identifier
  manually, the counter will not be incremented.

### DIFF
--- a/apps/qubit/modules/informationobject/actions/editAction.class.php
+++ b/apps/qubit/modules/informationobject/actions/editAction.class.php
@@ -138,7 +138,7 @@ class InformationObjectEditAction extends DefaultEditAction
         $this->parent = QubitInformationObject::getById(QubitInformationObject::ROOT_ID);
         $this->form->setDefault('parent', $this->context->routing->generate(null, array($this->parent, 'module' => 'informationobject')));
       }
-      
+
       if (isset($getParams['repository']))
       {
         $this->resource->repository = QubitRepository::getById($this->request->repository);
@@ -156,6 +156,9 @@ class InformationObjectEditAction extends DefaultEditAction
       {
         $this->publicationStatusId = sfConfig::get('app_defaultPubStatus', QubitTerm::PUBLICATION_STATUS_DRAFT_ID);
       }
+
+      // If creating new description and identifier mask is set to on, auto-generate next identifier.
+      $this->handleIdentifierFromMask();
     }
   }
 
@@ -781,6 +784,26 @@ class InformationObjectEditAction extends DefaultEditAction
       {
         $this->resource->informationObjectsRelatedByparentId[] = $childLevel;
       }
+    }
+  }
+
+  /**
+   * If identifier mask is enabled, set our new info obj to an identifier generated
+   * from the mask. Also increment the mask counter.
+   */
+  private function handleIdentifierFromMask()
+  {
+    if (null === $maskEnabled = QubitSetting::getByName('identifier_mask_enabled'))
+    {
+      throw new sfException('identifier_mask_enabled setting not found--is your database upgraded?');
+    }
+
+    if ($maskEnabled->value)
+    {
+      $this->resource->identifier = QubitInformationObject::generateIdentiferFromMask();
+      $counter = QubitInformationObject::getIdentifierCounter();
+      $counter->value++;
+      $counter->save();
     }
   }
 }
--- a/apps/qubit/modules/informationobject/actions/editAction.class.php
+++ b/apps/qubit/modules/informationobject/actions/editAction.class.php
@@ -600,6 +600,7 @@ class InformationObjectEditAction extends DefaultEditAction
     $this->deleteNotes();
     $this->updateChildLevels();
     $this->removeDuplicateRepositoryAssociations();
+    $this->incrementMaskCounter();
   }
 
   public function execute($request)
@@ -789,7 +790,7 @@ class InformationObjectEditAction extends DefaultEditAction
 
   /**
    * If identifier mask is enabled, set our new info obj to an identifier generated
-   * from the mask. Also increment the mask counter.
+   * from the mask.
    */
   private function handleIdentifierFromMask()
   {
@@ -801,9 +802,23 @@ class InformationObjectEditAction extends DefaultEditAction
     if ($maskEnabled->value)
     {
       $this->resource->identifier = QubitInformationObject::generateIdentiferFromMask();
-      $counter = QubitInformationObject::getIdentifierCounter();
-      $counter->value++;
-      $counter->save();
     }
+
+    $this->mask = $maskEnabled->value; // Pass if we're using an identifier generated from mask to template
+  }
+
+  /**
+   * If the user is using an identifier generated from the mask, increment the mask counter.
+   */
+  private function incrementMaskCounter()
+  {
+    if (!filter_var($this->request->getPostParameter('usingMask'), FILTER_VALIDATE_BOOLEAN))
+    {
+      return;
+    }
+
+    $counter = QubitInformationObject::getIdentifierCounter();
+    $counter->value++;
+    $counter->save();
   }
 }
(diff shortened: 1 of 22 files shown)


apps/qubit/modules/informationobject/actions/editAction.class.php
```diff
@@ -783,4 +787,38 @@ protected function updateChildLevels()
...
     }
   }
+
+  /**
+   * If identifier mask is enabled, set our new info obj to an identifier generated
+   * from the mask.
+   */
+  private function handleIdentifierFromMask()
+  {
+    if (null === $maskEnabled = QubitSetting::getByName('identifier_mask_enabled'))
```
jraddaoui: We normally use sfConfig to obtain the settings value, that way you can set a default one ... `sfConfig::get('app_identifier_mask_enabled', false)`. Like this you could remove the exception and make only one if check.

apps/qubit/modules/settings/actions/globalAction.class.php
```diff
@@ -213,6 +219,43 @@ protected function updateGlobalSettings()
       }
     }
 
+    // Identifier mask enabled
+    if (null !== $identifierMaskEnabled = $thisForm->getValue('identifier_mask_enabled'))
+    {
+      if (null === $setting = QubitSetting::getByName('identifier_mask_enabled'))
+      {
+        $setting = new QubitSetting;
```
jraddaoui: Why do you create this one in here an not the others? We use to do this when the settings are not added via migration script, but you're adding them.

apps/qubit/modules/settings/actions/globalAction.class.php
```diff
@@ -213,6 +219,43 @@ protected function updateGlobalSettings()
...
+
+      // Force sourceCulture update to prevent discrepency in settings between cultures
+      $setting->setValue($identifierMask, array('sourceCulture' => true));
+      $setting->save();
+    }
+
+    // Identifier counter
+    if (null !== $identifierCounter = $thisForm->getValue('identifier_counter'))
+    {
+      if (ctype_digit($identifierCounter) && $identifierCounter > -1)
```
jraddaoui: ctype_digit will return false if the value is negative

lib/model/QubitInformationObject.php
```diff
@@ -3317,4 +3317,51 @@ public function deleteFullHierarchy()
...
   }
+
+  /**
+   * Get current identifier counter for identifier mask from database.
+   *
+   * @return QubitSetting  The identifier counter setting (use ->value to get value).
+   */
+  public static function getIdentifierCounter()
+  {
+    if (null === $counter = QubitSetting::getByName('identifier_counter'))
```
jraddaoui, line 3328: The same about sfConfig in here

lib/model/QubitInformationObject.php
```diff
@@ -3317,4 +3317,51 @@ public function deleteFullHierarchy()
...
+  /**
+   * Generate identifier based on identifier mask and current counter.
+   *
+   * @return string  The generated identifier.
+   */
+  public static function generateIdentiferFromMask()
+  {
+    $counter = self::getIdentifierCounter();
+
+    if (null === $mask = QubitSetting::getByName('identifier_mask'))
```
jraddaoui: ;)

apps/qubit/modules/informationobject/templates/_identifierOptions.php
```diff
@@ -0,0 +1,11 @@
+<div class="section">
+  <a href="#" class="identifier-action" id="alternative-identifiers">
+    <?php echo __('Add alternative identifier(s)') ?>
```
jraddaoui: Good missing i18n catch!

apps/qubit/modules/informationobject/templates/_identifierOptions.php
```diff
@@ -0,0 +1,11 @@
+<div class="section">
```
jraddaoui, line 1: One thing about this partial is that the alternative identifiers partial is not used in the DC and MODS templates, so the "Add alternative identifier(s)" link shouldn't appear in there, and placed in here it will (without working). If you need the two links together in the GUI you could send an option to this template with a boolean to show the alternative identifiers link or not based on the IO template.

plugins/sfDcPlugin/modules/sfDcPlugin/templates/editSuccess.php
```diff
@@ -42,6 +42,8 @@
           ->label(__('Identifier').' <span class="form-required" title="'.__('This is a mandatory element.').'">*</span>')
           ->renderRow() ?>
 
+        <?php echo get_partial('informationobject/identifierOptions', array('mask' => $mask)) ?>
```
jraddaoui: We shouldn't need the alternative identifiers link in here

plugins/sfModsPlugin/modules/sfModsPlugin/templates/editSuccess.php
```diff
@@ -41,6 +41,8 @@
           ->help(__('Contains a unique standard number or code that distinctively identifies a resource.'))
           ->renderRow() ?>
 
+        <?php echo get_partial('informationobject/identifierOptions', array('mask' => $mask)) ?>
```
jraddaoui: We shouldn't need the alternative identifiers link in here

jraddaoui (changes requested):
Hi @MikeFE ,

Great attention to the user input, this is looking really good!

I've added some comments with a few things that I think should be addressed.

Nice job!

apps/qubit/modules/informationobject/templates/_identifierOptions.php
MikeFE: Yeah, I actually noticed this and was struggling on what to do. I figured maybe if we add the alt identifier label to the other standards, we could then add in alternative identifier functionality to them too in another ticket soon. But I can pass a boolean in as well to hide it for now as you suggest. Maybe there's a reason mods and dc don't have alternative identifiers? I'm not sure!

apps/qubit/modules/settings/actions/globalAction.class.php
MikeFE: oh weird, I'm not sure why I did that hahaha. I was mostly just copying what other things were doing in that file. I'll remove it.

apps/qubit/modules/settings/actions/globalAction.class.php
MikeFE: hahaha I just copied again what accession counter was doing. Good catch, I'll fix them both.

jraddaoui:
Only the setting access in lib/model/QubitInformationObject.php getIdentifierCounter() not using sfConfig yet ;)

MikeFE:
Aaaah you're so fast! I'm writing you an explanation in Redmine, one second :)

jraddaoui:
Hi Mike, answered in there. Please, for the next time lets try to keep the code review discussion in one place. 😄

MikeFE:
Okay Radda, I looked into why sfConfig::get on identifier_counter wasn't working and our suspicions were correct. If I added QubitCache::getInstance()->removePattern('settings:i18n:*') like what the global settings form does when you save it, sfConfig::get will return the correct (updated) counter value. I'm going to keep it QubitSetting since I think that's preferable to clearing all our settings cache every time we want to increment the counter.

MikeFE:
I guess we could target the removePattern to something more specific to the counter than *, but I still think that's a bit hacky and probably you're losing performance due to pattern matching/cache clearing.

jraddaoui:
Awesome! Thanks for looking into it Mike!

I looks like there are some conflicts in the PR, but feel free to merge when they're fixed.

jraddaoui (approved):
👍

PR #562, opened 2017-05-18
Add default for QubitPdo::modify params, #10971

MikeFE:
We were calling QubitPdo::modify with only one parameter throughout the
application. This fixes the issue by making the second parameter optional
(now defaults to empty array).


sbreker:
Looks good to me.  👍

PR #564, opened 2017-05-18
Avoid empty zip on csv export, refs #10907

MikeFE:
Previously, if you went to export a csv of archival descriptions but no
descriptions matched the criteria (e.g. exporting with 'export drafts'
unchecked but only draft records are in the clipboard), AtoM would create a
blank .zip file instead of indicating to the user that no matching records
were found to be exported.

This fix rectifies that situation.


mcantelon:
:+1:

PR #565, opened 2017-05-20
Optimize and fix EAD import, refs #11110

jraddaoui:
- Improve physical objects import:

Get the collection root only once and use it to make the existing physical
objects check with RAW SQL instead of the ORM.

- Improve events association:

  - IO events are saved when the IO is saved so there is no need to do it twice.
  - Avoid the use of magic methods
  - Associate the events once per description instead of once per method

- Fix origination export:

Avoid node duplication per creator instead of wrapping them all and add
event id to children nodes to allow event association on import.

- Fix origination import:

Return and track imported actors from the origination element to fix
creation events association.

- Fix geogname export:

Use the event id instead of the object term relation id to fix the event
association on import.

PS 1. There is a lot more info on the [Redmine ticket.
](https://projects.artefactual.com/issues/11110)

PS 2. I'll create and use a public ticket for the merge.


lib/model/QubitInformationObject.php
```diff
@@ -2599,25 +2604,25 @@ public function setPhysicalObjectByName($physicalObjectName, $options)
     }
   }
 
-  public function importPhysicalObject($location, $name = false, $type = false, $label = false)
+  public function importPhysicalObject($location, $name, $options = array())
   {
-    if ($label && $type)
+    if ($options['label'] && $options['type'])
```
MikeFE: Is there a case where this ever gets called when these option elements don't exist? If so, we should check isset first right?

lib/model/QubitInformationObject.php
```diff
@@ -2626,28 +2631,18 @@ public function importPhysicalObject($location, $name = false, $type = false, $l
       : false;
 
     // Check for an existing physical object within this collection with the same name
-    if ($this->parentId)
+    if ($options['collectionId'])
```
MikeFE: Same here, is this always going to be a valid element in $options?

lib/model/QubitInformationObject.php
```diff
@@ -2626,28 +2631,18 @@ public function importPhysicalObject($location, $name = false, $type = false, $l
...
         $name,
         $location,
         ($term) ? $term->id : null,
-        $topLevelParent->id
+        $options['collectionId']
       );
     }
 
-    // There was no existing physical object to attach, create a new one.
-    if (!isset($object) || $object === null)
```
MikeFE, line 2650: Good catch :)

lib/model/QubitPhysicalObject.php
```diff
@@ -152,25 +152,41 @@ public static function getPhysicalObjectsByNameAndLocation($name, $location, $ty
...
   public static function checkPhysicalObjectExistsInCollection($name, $location, $typeId, $collectionId)
   {
+    if (!isset($collectionId))
+    {
+      return;
+    }
+
+    // Get collection lft and rgt values from the database as they are not
+    // always being updated in the class cache in multi-level imports
+    $sql = "SELECT lft, rgt FROM information_object WHERE id = :id;";
```
MikeFE: Very minor nitpick... probably just need single quotes here and below in L172

lib/model/QubitPhysicalObject.php
```diff
@@ -152,25 +152,41 @@ public static function getPhysicalObjectsByNameAndLocation($name, $location, $ty
...
-        if ($topLevelParent->id === $collectionId)
-        {
-          return $physObj;
-        }
+        return $physObj;
       }
     }
 
-    return null;
+    return;
```
MikeFE: we shouldn't need this right?

lib/model/QubitInformationObject.php
jraddaoui: I think the warning occurs when options is not initiated as a default array, otherwise in php `if (null)` is evaluated as false.

lib/model/QubitInformationObject.php
jraddaoui: I think the warning occurs when options is not initiated as a default array, otherwise in php `if (null)` is evaluated as false.

lib/model/QubitPhysicalObject.php
jraddaoui: I'm not completely sure if this is needed with double quotes because of the params, I'll double check ;)

lib/model/QubitPhysicalObject.php
jraddaoui: Right! :)

lib/model/QubitInformationObject.php
MikeFE: I think it will throw a PHP Notice if 'collectionId' doesn't exist, so maybe wrapping it with !empty(...) is best? If 'collectionId' should be there always though, feel free to disregard.

MikeFE:
Nothing major I can see besides a couple of minor things. :+1:

lib/model/QubitInformationObject.php
jraddaoui: You're right, it shows a PHP Notice:  Undefined index, thanks!

lib/QubitXmlImport.class.php
```diff
@@ -1023,25 +1047,26 @@ private function associateEvents()
     {
       $event = $values['event'];
 
-      if (!empty($event))
+      if (empty($event))
       {
-        $place = array_key_exists('place', $values) ? $values['place'] : NULL;
-        $actor = array_key_exists('actor', $values) ? $values['actor'] : NULL;
+        continue;
```
MikeFE, line 1052: great change!

lib/QubitXmlImport.class.php
```diff
@@ -681,23 +695,33 @@ private function processMethods(&$domNode, &$importDOM, $methods, &$currentObjec
...
 
-              // invoke the object and method defined in the schema map
-              $obj = call_user_func_array(array( & $currentObject, $methodMap['Method']), $parameters);
+              // Invoke the object and method defined in the schema map
+              $result = call_user_func_array(array( & $currentObject, $methodMap['Method']), $parameters);
 
               // If an actor/event object was returned, track that
               // in the events cache for later cleanup
-              if(!empty($obj))
+              if($currentObject instanceof QubitInformationObject && !empty($result))
```
MikeFE: Minor nitpick, space here after if (I realize it was like this before too heheh)

MikeFE:
:+1:

jraddaoui:
Merged with --squash in https://github.com/artefactual/atom/commit/ba89f62d392e3eda1f77fa1f79e1ec1ca2801c9e

PR #567, opened 2017-05-25
Fix EAD XML cache deletion issue, refs #11191

mcantelon:
Fixed an issue with deletion of cached EAD XML caused by checking the
publication status of the child node rather than the top-level
parent.


MikeFE:
:+1:

MikeFE:
Oops, please ensure that's 2 space indentation in the code blocks. Looks good otherwise.

PR #568, opened 2017-05-29
Extend the functionality of the i18n:remove-html-tags command

g011um:
The i18n:remove-html-tags command has been extended to include more
tables, thanks to a submission to the AtoM list by Clara Rosales.
https://groups.google.com/d/msg/ica-atom-users/_xdBK0ucegg/RQnNM5DKBAAJ

HTML character entities (ccedil, ecirc etc) are also translated back to
their non-HTML entity representations.

### DIFF
--- a/lib/task/i18n/i18nRemoveHtmlTagsTask.class.php
+++ b/lib/task/i18n/i18nRemoveHtmlTagsTask.class.php
@@ -172,14 +172,13 @@ EOF;
   }
 
   /**
-   * Determine what information object i18n columns are populated
-   * and update them.
+   * Determine which i18n columns are populated and update them.
    *
    * @param stdClass $io  row of information object i18n data
    *
    * @return integer  number of columns changed
    */
-  private function processInformationObjectI18nHtml(&$io, $tableName, $columns)
+  private function processI18nHtml(&$io, $tableName, $columns)
   {
     // Determine what column values contain HTML
     $columnValues = array();
--- a/lib/task/i18n/i18nRemoveHtmlTagsTask.class.php
+++ b/lib/task/i18n/i18nRemoveHtmlTagsTask.class.php
@@ -219,8 +219,7 @@ EOF;
 
     foreach($columnValues as $column => $value)
     {
-      // Only update if tags are found
-      //if ($value != strip_tags($value)) {
+      // Only update if tags or HTML entities are found
       if (($value != strip_tags($value)) || ($value != html_entity_decode($value))) {
         $transformedValue = $this->transformHtmlToText($value);
 
--- a/lib/task/i18n/i18nRemoveHtmlTagsTask.class.php
+++ b/lib/task/i18n/i18nRemoveHtmlTagsTask.class.php
@@ -27,28 +27,76 @@
  */
 class i18nRemoveHtmlTagsTask extends sfBaseTask
 {
-  public static $columns = array(
-    'title',
-    'alternate_title',
-    'edition',
-    'extent_and_medium',
-    'archival_history',
-    'acquisition',
-    'scope_and_content',
-    'appraisal',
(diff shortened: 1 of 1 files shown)


g011um:
This is a fairly minor change to the code really.  It changes the "public static $columns" from an array of column names to a nested array of table and columns names, then wraps the existing Artefactual code in a foreach loop.

An additional check is made to determine if the field contains HTML entities (in addition to the existing check that looked for just HTML tags).  That will cause HTML entities like **& ccedil ;** to be turned back into proper characters (like **ç**)

fiver-watson:
Thanks for submitting this, @g011um ! I have created the following issue ticket: 

* https://projects.artefactual.com/issues/11207

I will be asking one of our developers to code review this soon. Cheers!

lib/task/i18n/i18nRemoveHtmlTagsTask.class.php
```diff
@@ -128,23 +179,23 @@ public function execute($arguments = array(), $options = array())
    *
    * @return integer  number of columns changed
    */
-  private function processInformationObjectI18nHtml(&$io)
+  private function processInformationObjectI18nHtml(&$io, $tableName, $columns)
```
MikeFE: Can we rename this function since it no longer is specific to information objects?

lib/task/i18n/i18nRemoveHtmlTagsTask.class.php
```diff
@@ -161,15 +212,16 @@ private function processInformationObjectI18nHtml(&$io)
...
+    // Assemble query and note parsed column values
     $values = array();
 
     $query = 'UPDATE '. $table .' SET ';
 
     foreach($columnValues as $column => $value)
     {
       // Only update if tags are found
-      if ($value != strip_tags($value)) {
+      //if ($value != strip_tags($value)) {
```
MikeFE: Can probably just remote this line now.

MikeFE:
Looks pretty good g011um! I had 2 suggestions, other than that, there are a couple stylistic nitpicks but I noticed this task is already violating a bunch of our style rules. I'll take care of those if you implement my 2 suggestions.

Thanks!

lib/task/i18n/i18nRemoveHtmlTagsTask.class.php
g011um: Sorry Mike; that's sloppy.  Didn't mean to commit that.  Easy fix.

lib/task/i18n/i18nRemoveHtmlTagsTask.class.php
g011um: Do you have a suggestion?  Is processI18nHtml too generic?

MikeFE: I think processI18nHtml is fine. :+1:

lib/task/i18n/i18nRemoveHtmlTagsTask.class.php
```diff
@@ -80,32 +129,34 @@ public function execute($arguments = array(), $options = array())
...
-      $columnsChanged = $this->processInformationObjectI18nHtml($io);
-
-      // Update total column values changed
-      if ($columnsChanged)
+      while ($io = $statement->fetch(PDO::FETCH_OBJ))
       {
-        $changedCount++;
-        $columnsChangedCount += $columnsChanged;
+        // Process HTML in row's columns
+        $columnsChanged = $this->processInformationObjectI18nHtml($io, $tableName, $columns);
```
MikeFE: Don't forget to rename here too :P

MikeFE:
Thanks for the submission g011um.

committed 9c71043cf2e8a1fdd300c5bbb9c7526efb721b86

PR #569, opened 2017-06-01
Description updates repository filter, refs #11221

jraddaoui:
Add a select field with all the repositories to the description updates
page to allow filtering by institution when information object is selected

- Use (and rename) search repositories cache key for the field options
- Add new field in a new row alongside the pub. status field to allow
  showing/hidding both at the same time on type change
- Add clipboardToggleAll.js via view.yml instead of in the action
- Add toggleIoOptions.js in the same way to show/hide information
  object actions based on the type field value

### DIFF
--- a/apps/qubit/modules/informationobject/actions/browseAction.class.php
+++ b/apps/qubit/modules/informationobject/actions/browseAction.class.php
@@ -183,7 +183,7 @@ class InformationObjectBrowseAction extends DefaultBrowseAction
         $criteria->addAscendingOrderByColumn('authorized_form_of_name');
 
         $cache = QubitCache::getInstance();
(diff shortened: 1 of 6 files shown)


js/toggleIoOptions.js
```diff
@@ -0,0 +1,45 @@
+(function ($) {
```
MikeFE: This feature is for showing IO updates filtered by associated repository, right? I think this file name is a bit too generic for such a specific feature it's handling. I would rename it to something a bit more informative if I were you, but I don't think changing the class name is as necessary if you change filename and make a brief comment at the top about what this file does.

Maybe toggleIoUpdateLogOptions.js or something like that? Not sure. :)

PR #572, opened 2017-06-20
Prevent article removal slugs, Refs #11260

sbreker:
English articles "a," "an," and the "the" will no longer be stripped
from slugs.

Post deployment be sure to run CLI tasks:
- php symfony propel:generate-slugs --delete
- php symfony search:populate


jraddaoui (approved):
👍

PR #573, opened 2017-06-26
CSV export include drafts issue, Refs# 10006

sbreker:
This commit corrects an issue where the ACL checks on background export
jobs was not running correctly.

arBaseJob method signIn() was not being called because it was blocked
by a condition dependent on isAuthenticated() returning true. In the
background job context, isAuthenticated() will always return false before
signIn() is called.


MikeFE:
:+1:

PR #577, opened 2017-07-01
Fix event deletion, refs #10278

jraddaoui:
Existing events can be deleted after the form fields are processed, so we
don't want them in the related events array of the current resource where,
when the resource is saved, they are tried to be saved causing an error.
They need to be saved independently after the fields are processed.


sbreker:
Great job!  I picked it to a test branch to test the date saving bug - all working correctly!

👍

PR #579, opened 2017-07-04
sf_culture validation issue, Refs #10673

sbreker:
I have added code to filter bad sf_culture values. I have changed the
check in sfUser.class.php to instead verify that it is a valid
sf_culture, and if not, fall back to:

- use the culture defined in the user session
- use the default culture set in settings.yml

This prevents 500 errors and prevents the ability to inject paths and
files into the cache directory.


sbreker:
Hi Mike

I kept that since the vendor/symfony code was previously using that technique.  The only change was to ensure that the URL parameter sf_culture is validated correctly, instead of blindly using it if it's populated.

MikeFE:
:+1:

PR #580, opened 2017-07-05
Add setting to escape string queries, refs #11299

jraddaoui:
To avoid Elasticsearch errors while searching using usual special chars,
like '/' in an instance where that's the reference code separator, a new
setting has been added to determine which chars should be escaped
before performing the string query.

- Add escape_queries setting
- Add escapeTerm function to arElasticSearchPluginUtil
- Use new escapeTerm function in ES string queries
- Add unit test

### DIFF
--- a/apps/qubit/modules/settings/actions/globalAction.class.php
+++ b/apps/qubit/modules/settings/actions/globalAction.class.php
@@ -114,7 +114,7 @@ class SettingsGlobalAction extends sfAction
       'identifier_counter' => (isset($identifierCounter)) ? intval($identifierCounter->getValue(array('sourceCulture'=>true))) : 1,
       'separator_character' => (isset($separatorCharacter)) ? $separatorCharacter->getValue(array('sourceCulture'=>true)) : null,
       'inherit_code_informationobject' => (isset($inheritCodeInformationObject)) ? intval($inheritCodeInformationObject->getValue(array('sourceCulture'=>true))) : 1,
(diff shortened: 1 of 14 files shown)


apps/qubit/modules/settings/actions/globalAction.class.php
```diff
@@ -273,6 +275,16 @@ protected function updateGlobalSettings()
       $setting->save();
     }
 
+    // Escape queries, add setting if it's not already created (to avoid adding it in a migration)
+    if (null === $setting = QubitSetting::getByName('escape_queries'))
+    {
+      $setting = QubitSetting::createNewSetting('escape_queries', null);
```
MikeFE, line 281: Do you think it'd make sense to make the default here the set of special characters that make ES produce an error instead of null?

MikeFE:
Great work, I've been wanting a fix like this for a long time! :+1:

apps/qubit/modules/settings/actions/globalAction.class.php
jraddaoui: Hi @MikeFE, the default value is set in the form and added to the setting after this condition. We have also decided to leave it empty by default to avoid unexpected changes in upgrades.

MikeFE:
👍

jraddaoui:
Merged in https://github.com/artefactual/atom/commit/813c5e06b269f9adb10750f00fd8b0566a5db726

PR #581, opened 2017-07-13
Genre access point fixes, refs #11281

jraddaoui:
- Allow EAD import of genreform with source attribute (except with RAD value)
- Fix genre directly related IOs facet


MikeFE:
Looks great. :+1:

PR #583, opened 2017-07-13
Fix i18n repository region filter, refs #11121

MikeFE:
This fix will make it so when browsing repositories and setting the region
filter, only show regions in the dropdown for the current culture. This also
fixes repo browse so the region filter works for non-English cultures.

NOTE: This filter does not support culture fallback! You can only set the
region filter on terms for the current culture.

### DIFF
--- a/apps/qubit/modules/repository/actions/browseAction.class.php
+++ b/apps/qubit/modules/repository/actions/browseAction.class.php
@@ -40,7 +40,7 @@ class RepositoryBrowseAction extends DefaultBrowseAction
               'size' => 10),
       'regions' =>
         array('type' => 'term',
-              'field' => 'contactInformations.i18n.en.region.untouched',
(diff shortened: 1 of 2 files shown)


apps/qubit/modules/repository/actions/browseAction.class.php
```diff
@@ -88,6 +88,7 @@ protected function populateFacet($name, $ids)
 
   public function execute($request)
   {
+    $this->setI18nFieldCultures(); // Must call this first as parent::execute() calls addFacets().
```
jraddaoui: I'd prefer this comment on top but not a big deal

jraddaoui (approved):

jraddaoui:
Merged in https://github.com/artefactual/atom/commit/b4477a2e10ab58e734ef06dde123bcaaeba6834c

PR #585, opened 2017-07-13
Change places/subjects search operator ref #11349

MikeFE:
Change the default search operator for subjects and places to AND (previously
was OR). Note, this change will affect all taxonomy searches as well, not
just subjects and places. :)


jraddaoui (approved):

MikeFE:
953664ee59a2053611068fced1d849de6356310e

PR #586, opened 2017-07-19
Dev/issue 11097 clipboard export improve

mcantelon:
* Allow unauthenticated export jobs to be deleted only after job is complete
* Fixed Internet Explorer-specific clipboard issue

### DIFF
--- a/apps/qubit/modules/default/actions/alertsComponent.class.php
+++ b/apps/qubit/modules/default/actions/alertsComponent.class.php
@@ -65,8 +65,11 @@ class DefaultAlertsComponent extends sfComponent
           );
           $alertType = $alertTypes[$job->getStatusId()];
 
-          // Add as conditional alert
+          // If job is complete, allow it to be deleted by the user
           $deleteUrl = $this->context->controller->genUrl('jobs/delete?id='. $job->id);
+          $deleteUrl = ($job->getStatusId() == QubitTerm::JOB_STATUS_COMPLETED_ID) ? $deleteUrl : null;
+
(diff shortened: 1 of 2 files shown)


apps/qubit/modules/default/actions/alertsComponent.class.php
```diff
@@ -65,8 +65,11 @@ public function execute($request)
           );
           $alertType = $alertTypes[$job->getStatusId()];
 
-          // Add as conditional alert
+          // If job is complete, allow it to be deleted by the user
           $deleteUrl = $this->context->controller->genUrl('jobs/delete?id='. $job->id);
+          $deleteUrl = ($job->getStatusId() == QubitTerm::JOB_STATUS_COMPLETED_ID) ? $deleteUrl : null;
+
+          // Add as conditional alert
           array_push($this->conditionalAlerts, array('type' => $alertType, 'message' => $message, 'deleteUrl' => $deleteUrl));
```
sbreker, line 73: Is it OK that we could push an alert where $deleteUrl is null?

mcantelon: This is for controlling what happens when the X (to close a notification) is clicked. If the $deleteUrl is null then clicking X will have the standard behaviour of dismissing the notification.

sbreker: Sounds good!

sbreker: Looks good!

PR #589, opened 2017-07-27
Avoid exception in migration, refs #11386

jraddaoui:


sevein (approved):
This makes sense since the user may have modified the menu node.

PR #592, opened 2017-07-28
Fix export form, refs #11384

jraddaoui:
Use proper form fields for object type and export format


MikeFE:
Do you remember where the actual regression in #11384 was (i.e. the cause of the bug)? Great catch with moving the form to use sfWidgets. :+1:

jraddaoui:
Hi Mike,  I don't really know why `$request->getParameter()` for the "format" param (before "type") doesn't work but it works for the "objectType". Not even setting the value manually in the URL I have managed to get a default value, it's really a mystery :/

I couldn't find any place where we call this route with that param and we change the default via JS when needed so I think it's okay to leave it like that for now ;)

Thanks!

PR #593, opened 2017-08-01
Add --repository to io delete task, refs #11344

MikeFE:
Add a new --repository option to tools:delete-description. This will make the
slug specified in the CLI argument point to a repository slug instead, and all
descriptions associated under the specified repository will be deleted.

### DIFF
--- a/lib/task/tools/deleteDescriptionTask.class.php
+++ b/lib/task/tools/deleteDescriptionTask.class.php
@@ -86,6 +86,11 @@ EOF;
    */
   private function confirmDeletion($noConfirmation)
   {
+    if ($noConfirmation)
+    {
+      return true;
+    }
+
     switch ($this->resourceType)
     {
       case 'QubitRepository':
@@ -95,12 +100,12 @@ EOF;
       case 'QubitInformationObject':
         $confirmWarning = sprintf('WARNING: You are about to delete the record "%s" and %d descendant records.',
                                   $this->resource->getTitle(array('cultureFallback' => true)),
-                                  max(0, count($this->resource->descendants->andSelf()) - 1));
+                                  count($this->resource->descendants));
         break;
     }
 
-    if ($noConfirmation || $this->askConfirmation(array($confirmWarning,
-      'Are you sure you want to proceed? (y/N)'), 'QUESTION_LARGE', false))
+    if ($this->askConfirmation(array($confirmWarning, 'Are you sure you want to proceed? (y/N)'),
+                               'QUESTION_LARGE', false))
     {
       return true;
     }
@@ -133,10 +138,10 @@ EOF;
    */
   private function deleteDescriptions($root)
   {
-    $descriptions = $root->descendants->andSelf('rgt');
+    $descriptions = $root->descendants->andSelf()->orderBy('rgt');
     $this->logSection(sprintf('[%s] Deleting description "%s" (slug: %s, +%d descendants)', strftime('%r'),
                               $root->getTitle(array('cultureFallback' => true)),
-                              $root->slug, max(0, count($descriptions) - 1)));
+                              $root->slug, count($root->descendants)));
 
     foreach ($descriptions as $desc)
     {
--- a/lib/task/tools/deleteDescriptionTask.class.php
+++ b/lib/task/tools/deleteDescriptionTask.class.php
@@ -17,15 +17,18 @@
  * along with Access to Memory (AtoM).  If not, see <http://www.gnu.org/licenses/>.
  */
 
-class deleteDescriptionTask extends sfBaseTask
+class deleteDescriptionTask extends arBaseTask
 {
+  private $nDeleted = 0;
+
   /**
    * @see sfTask
    */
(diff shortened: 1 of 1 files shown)


lib/task/tools/deleteDescriptionTask.class.php
```diff
@@ -49,44 +53,115 @@ protected function configure()
    */
   public function execute($arguments = array(), $options = array())
   {
-    $databaseManager = new sfDatabaseManager($this->configuration);
-    $conn = $databaseManager->getDatabase('propel')->getConnection();
-    sfContext::createInstance($this->configuration);
+    parent::execute($arguments, $options);
```
jraddaoui, line 56: Nice!

lib/task/tools/deleteDescriptionTask.class.php
```diff
@@ -49,44 +53,115 @@ protected function configure()
...
+    switch ($this->resourceType)
+    {
+      case 'QubitRepository':
+        $confirmWarning = sprintf('WARNING: You are about to delete all the records under the repository "%s".',
+                                  $this->resource->getAuthorizedFormOfName(array('cultureFallback' => true)));
+        break;
+      case 'QubitInformationObject':
+        $confirmWarning = sprintf('WARNING: You are about to delete the record "%s" and %d descendant records.',
+                                  $this->resource->getTitle(array('cultureFallback' => true)),
+                                  max(0, count($this->resource->descendants->andSelf()) - 1));
```
jraddaoui: `count($this->resource->descendants)` is not enough in here? I think `count(null)` equals to 0 ...

lib/task/tools/deleteDescriptionTask.class.php
```diff
@@ -49,44 +53,115 @@ protected function configure()
...
+        break;
+    }
 
-    if (null === $informationObject = QubitInformationObject::getBySlug($arguments['slug']))
+    if ($noConfirmation || $this->askConfirmation(array($confirmWarning,
```
jraddaoui: Not a big deal, but checking `$noConfirmation` first could avoid counting descendants in the previous switch

lib/task/tools/deleteDescriptionTask.class.php
```diff
@@ -49,44 +53,115 @@ protected function configure()
...
+    }
+
+    // User wishes to proceed, delete targeted information objects.
+    switch ($this->resourceType)
+    {
+      case 'QubitRepository':
+        $this->deleteDescriptionsFromRepository();
+        break;
+      case 'QubitInformationObject':
+        $this->deleteDescriptions($this->resource);
```
jraddaoui, line 74: You don't need to send `$this->resource` as it's a global var.

lib/task/tools/deleteDescriptionTask.class.php
```diff
@@ -49,44 +53,115 @@ protected function configure()
...
 
-    $n = 0;
+  /**
+   * Delete specified description & its descendants from AtoM.
+   *
+   * @param $root  A top level QubitInformationObject which will be deleted along with its descendants.
+   */
+  private function deleteDescriptions($root)
+  {
+    $descriptions = $root->descendants->andSelf('rgt');
```
jraddaoui: Wow! Can you really set the order in there? I have always used `->orderBy()`

lib/task/tools/deleteDescriptionTask.class.php
```diff
@@ -49,44 +53,115 @@ protected function configure()
...
+    $this->logSection(sprintf('[%s] Deleting description "%s" (slug: %s, +%d descendants)', strftime('%r'),
+                              $root->getTitle(array('cultureFallback' => true)),
+                              $root->slug, max(0, count($descriptions) - 1)));
```
jraddaoui: Same here

jraddaoui (changes requested):

lib/task/tools/deleteDescriptionTask.class.php
MikeFE: Well apparently you _can_ do that. But it's incorrect :D

lib/task/tools/deleteDescriptionTask.class.php
MikeFE: Good catch, missed that in a refactor.

lib/task/tools/deleteDescriptionTask.class.php
MikeFE: I didn't know that, neat

lib/task/tools/deleteDescriptionTask.class.php
MikeFE: Actually, I remember why I did it now. I also call deleteDescriptions but from inside deleteDescriptionsFromRepository, when it needs to delete something other than $this->resource.

jraddaoui: Right!

jraddaoui:
:+1:

MikeFE:
91584d8

PR #594, opened 2017-08-02
Fixed IE issue with export form, refs #11118

mcantelon:
In Internet Explorer 11, select options can't be hidden using CSS.

Fixed issue, caused by this, by changing code to rewrite select options.


sbreker:
Lgtm.  Perhaps add some comments why this is this way, and that this is a workaround for IE11 so this does not get modified my someone following along after.

sbreker:
Thanks!  Lgtm!  👍

PR #595, opened 2017-08-03
Fix issue with relative image paths, refs #10692

mcantelon:
Fix issue, with migration-related task for comparing files in a directory to
values in a CSV file, where it wasn't working correctly for files in
subdirectories.


jraddaoui (approved):
👍

PR #596, opened 2017-08-04
Deletion nested set update disabling, refs #10962

mcantelon:
Added the ability to, when programatically deleting a term node, set the disableNestedSetUpdating to true so the nested set isn't updated during the delete. This allows much faster batch deletion of terms and the nested set can be updated manually afterwards.

Implemented by modifying the QubitObjectBuilder class to generate model code that suppports this and updated existng model code.

### DIFF
--- a/lib/model/QubitActor.php
+++ b/lib/model/QubitActor.php
@@ -25,6 +25,8 @@
  */
 class QubitActor extends BaseActor
 {
+  public $disableNestedSetUpdating = false;
+
   const
     ROOT_ID = 3;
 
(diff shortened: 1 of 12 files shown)


jraddaoui (commented):
Looks great @mcantelon!

The only think I would add is a public flag for the property in the related Qubit... classes, I think we already have it in the QubitInformationObject and QubitTerm classes, but it will make clear that this option is available in the other classes.

jraddaoui (approved):

mcantelon:
Waiting until 2.4 is released to merge into dev branch.

PR #597, opened 2017-08-09
CSV import issue qubitParentSlug, Refs# 11409

sbreker:
This commit fixes an issue where CSVs with the qubitParentSlug column
populated would not import without errors.

Root cause of this issue was that the QubitFlatfileImport method
getIdCorrespondingToSlug() was not being referenced correctly from
functions in csvImportTask.class.php.

### DIFF
--- a/lib/QubitFlatfileImport.class.php
+++ b/lib/QubitFlatfileImport.class.php
@@ -2236,7 +2236,7 @@ class QubitFlatfileImport
     }
   }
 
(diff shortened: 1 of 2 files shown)


MikeFE:
Wow, apparently those function calls date all the way back to the initial Github commit from when we migrated from Subversion (if I'm glancing at Git blame right). Crazy how the bug didn't pop up until now?

Looks good to me though!

PR #598, opened 2017-08-10
Offer to replace export if it exists, refs #11082

mcantelon:
The CSV term usage export was appending rows if exporting to an existng
CSV file. Added log to, if the CSV export file already exists, ask the user
if they want to replace it (aborting if they decide not to).

### DIFF
--- a/lib/task/export/csvExportTermUsageTask.class.php
+++ b/lib/task/export/csvExportTermUsageTask.class.php
@@ -62,8 +62,17 @@ class csvExportTermUsageTask extends exportBulkBaseTask
     $sf_context = sfContext::createInstance($configuration);
     $conn = $this->getDatabaseConnection();
 
+    $this->exportFileReplacePrompt($arguments['path']);
     $itemsExported = $this->exportToCsv($this->determineTaxonomyId($options), $arguments['path'], $options['items-until-update']);
-    $this->log(sprintf("\nExport complete (%d terms exported).", $itemsExported));
+
+    if ($itemsExported)
+    {
+      $this->log(sprintf("\nExport complete (%d terms exported).", $itemsExported));
+    }
+    else
+    {
+      $this->log("No term usages found to export.");
+    }
   }
 
   private function determineTaxonomyId($options)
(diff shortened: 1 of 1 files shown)


lib/task/export/csvExportTermUsageTask.class.php
```diff
@@ -102,6 +102,17 @@ private function determineTaxonomyId($options)
 
   private function exportToCsv($taxonomyId, $exportPath, $rowsUntilUpdate)
   {
+    // Offer to delete export file, if it already exists
```
MikeFE: To avoid exportToCsv doing too many things, I'd consider moving this into its own function (private function checkReplaceFile()?), and maybe moving the check up into execute() before exportToCsv() is called. What do you think?

mcantelon: Makes sense. I'll implement that.

mcantelon: (I've made that change and tweaked a couple other things.)

lib/task/export/csvExportTermUsageTask.class.php
```diff
@@ -117,20 +139,23 @@ private function exportToCsv($taxonomyId, $exportPath, $rowsUntilUpdate)
...
     {
-      $resource = QubitTerm::getById($row->id);
-      $writer->setColumn('name', $resource->getName(array('cultureFallback' => true)));
-      $writer->setColumn('use_count', $row->use_count);
-      $writer->exportResource($resource);
+      // Instantiate CSV writer using "usage" column ordering
+      $writer = new QubitFlatfileExport($exportPath, 'usage');
+      $writer->loadResourceSpecificConfiguration('QubitTerm');
+
+      while($row = $result->fetch(PDO::FETCH_OBJ))
```
MikeFE: space after while here

lib/task/export/csvExportTermUsageTask.class.php
```diff
@@ -100,6 +109,19 @@ private function determineTaxonomyId($options)
     throw new sfException('Either the taxonomy-id or taxonomy-name must be used to specifiy a taxonomy.');
   }
 
+  private function exxportFileReplacePrompt($exportPath)
```
MikeFE: typo in func name

MikeFE:
Had a couple minor cosmetic suggestions. Looks good otherwise. :+1:

mcantelon:
Thanks! I've made the tweaks so ready to merge (will have to see if we want to merge it into 2.4 or not).

PR #599, opened 2017-08-10
Regen derivatives no-overwrite flag, Refs# 10055

sbreker:
Added the digitalobject:regen-derivatives --no-overwrite flag.

When set, the task won't overwrite existing derivatives.


lib/task/digitalobject/digitalObjectRegenDerivativesTask.class.php
```diff
@@ -40,13 +40,14 @@ protected function configure()
...
       new sfCommandOption('skip-to', null, sfCommandOption::PARAMETER_OPTIONAL, 'Skip regenerating derivatives until a certain filename is encountered', null),
+      new sfCommandOption('no-overwrite', 'n', sfCommandOption::PARAMETER_NONE, 'Don\'t overwrite existing derivatives (and no confirmation message)', null),
     ));
 
     $this->namespace = 'digitalobject';
     $this->name = 'regen-derivatives';
     $this->briefDescription = 'Regenerates digital object derivative from master copy';
     $this->detailedDescription = <<<EOF
-FIXME
+Regenerate digital object derivatives from master copy.
```
MikeFE, line 50: Hehe, I think this was like this for years. Good catch!

MikeFE:
:+1:

PR #602, opened 2017-08-15
Arch Descr move and more buttons, Refs# 11410

sbreker:
Removing change to add an ACL check so that 'move' & 'more' buttons
could be accessed by users in the 'translate' group. This did not
achieve the desired behaviour.


MikeFE:
:+1:

PR #603, opened 2017-08-17
Fix description deletion approach, refs #11404

MikeFE:
There was a bug related to the lft/rgt values not staying up to date during a
single ORM query while we're deleting things and the nested set is updating,
which sometimes resulted in unrelated records being accessed in the loop.

Splitting up the ORM queries on a per-top-leve-description basis with getById
rectifies this.


sbreker:
Code review complete - looks good!   👍

MikeFE:
d23b2c2ca8c7927ff47f1f4eebf82159ef69b6e9

PR #604, opened 2017-08-18
Fixed MODS creator import issue, refs #11052

mcantelon:
Fixed issue with MODS import where what's exported can't be reimported.


sbreker:
Looks good.  👍

PR #607, opened 2017-08-21
Dev/issue 9765 ldap config

mcantelon:
* Makes attribute used by LDAP to bind configurable
* Fixes check for user cass to make it work with LDAP user class


sbreker:
Looks good!  👍

PR #608, opened 2017-08-22
Dev/issue xxx clipboard sets

mcantelon:
Add ability to save/load clipboards, refs #11360
    
    * Changed UI so AJAX is used to relay clipboard information so, with changes
      to Varnish configuration, it won't be cached for anonymous users
    * Added clipboard save and save item models
    * Added ability to load/save clipboards in UI (when loading a clipboard the
      user can select if they want to merge it with their current clipboard or
      replace it)
    * Added a maximum age (in days) setting for saved clipboards
    * Added data expiry task (supporting deletion of expired clipboards)
    
    Also made minor changes to clipboard page:
    
    * Removed relevance from clipboard sort options.
    * Removed limit on clipboard items printed.

### DIFF
--- a/apps/qubit/modules/default/actions/editAction.class.php
+++ b/apps/qubit/modules/default/actions/editAction.class.php
@@ -201,7 +201,11 @@ class DefaultEditAction extends sfAction
 
     $this->form = new sfForm;
 
-    $this->earlyExecute();
+    // Call early execute logic, if defined by a child class
+    if (method_exists($this, 'earlyExecute'))
+    {
+      call_user_func(array($this, 'earlyExecute'));
+    }
 
     // Mainly used in autocomplete.js, this tells us that the user wants to
     // reuse existing objects instead of adding new ones.
--- a/apps/qubit/modules/user/actions/clipboardLoadAction.class.php
+++ b/apps/qubit/modules/user/actions/clipboardLoadAction.class.php
@@ -1,0 +1,150 @@
+<?php
+
+/*
+ * This file is part of the Access to Memory (AtoM) software.
+ *
+ * Access to Memory (AtoM) is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * Access to Memory (AtoM) is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with Access to Memory (AtoM).  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+class UserClipboardLoadAction extends DefaultEditAction
+{
+  // Arrays not allowed in class constants
+  public static
+    $NAMES = array(
+      'password',
+      'mode');
+
+  protected function addField($name)
+  {
+    switch ($name)
+    {
+      case 'password':
+        $this->form->setValidator('password', new sfValidatorString(array('required' => true)));
+        $this->form->setWidget('password', new sfWidgetFormInput);
+
+        break;
+
+      case 'mode':
+        $this->form->setDefault('mode', 'merge');
+        $this->form->setValidator('mode', new sfValidatorString);
+        $choices = array(
+          'merge' => $this->context->i18n->__('Merge saved clipboard with existing clipboard results'),
+          'replace' => $this->context->i18n->__('Replace existing clipboard results with saved clipboard')
+        );
+        $this->form->setWidget('mode', new sfWidgetFormSelect(array('choices' => $choices)));
+
+        break;
+    }
+  }
+
+  protected function processField($field)
+  {
+    switch ($field->getName())
+    {
+      case 'password':
+        $this->password = $this->form->getValue($field->getName());
+
+        break;
+
+      case 'mode':
+        $this->mode = $this->form->getValue($field->getName());
+
+        break;
+    }
+  }
+
+  public function execute($request)
+  {
+    parent::execute($request);
+
+    if ($request->isMethod('post'))
+    {
+      $this->form->bind($request->getPostParameters());
+
+      if ($this->form->isValid())
+      {
+        $this->processForm();
+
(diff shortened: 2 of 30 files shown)


apps/qubit/modules/user/actions/clipboardLoadAction.class.php
```diff
@@ -0,0 +1,155 @@
...
+  // Arrays not allowed in class constants
+  public static
+    $NAMES = array(
+      'password',
+      'mode');
+
+  protected function earlyExecute()
+  {
+
+  }
```
jraddaoui: What is this for? There is no earlyExecute in the parent class to overwrite, right?

apps/qubit/modules/user/actions/clipboardLoadAction.class.php
```diff
@@ -0,0 +1,155 @@
...
+  private function addSavedClipboardItems($password, $mode)
+  {
+    // Get saved clipboard corresponding to password
+    $criteria = new Criteria;
+    $criteria->add(QubitClipboardSave::PASSWORD, $password);
+    $save = QubitClipboardSave::getOne($criteria);
+
+    if ($save === null)
+    {
+      return null;
```
jraddaoui: `return;` is the same

apps/qubit/modules/user/actions/clipboardSaveAction.class.php
```diff
@@ -0,0 +1,113 @@
...
+      $message = $this->context->i18n->__('No items in clipboard to save.');
+      $this->context->user->setFlash('error', $message);
+    }
+    else
+    {
+      // Generate unique password
+      if (false === $password = $this->getUniquePassword())
+      {
+        $message = $this->context->i18n->__('Clipboard ID generation failure. Please try again.');
+        $this->context->user->setFlash('error', $message);
```
jraddaoui, line 37: You may want to stop execution after this (return or redirect)

apps/qubit/modules/user/actions/clipboardSaveAction.class.php
```diff
@@ -0,0 +1,113 @@
...
+      }
+
+      $this->saveClipboard($allSlugs, $password);
+
+      // Inform user of progress
+      $messageText = 'Clipboard saved. Clipoard ID is <b>%1%</b>. Please write this number down. ';
```
jraddaoui: It's a little confusing for me calling it password in one end and ID in the other.

config/schema.yml
```diff
@@ -54,6 +54,18 @@ propel:
...
 
+  clipboard_save:
+    id: { type: integer, primaryKey: true, autoIncrement: true, inheritanceKey: false }
+    user_id: { type: integer, required: false, foreignTable: user, foreignReference: id, onDelete: setnull  }
+    password: varchar(255)
+    created_at: timestamp
+
+  clipboard_save_item:
+    id: { type: integer, primaryKey: true, autoIncrement: true, inheritanceKey: false }
+    save_id: { type: integer, required: false, foreignTable: clipboard_save, foreignReference: id, onDelete: setnull  }
```
jraddaoui: I'd use `cascade` in here.

js/clipboard.js
```diff
@@ -186,6 +188,43 @@
...
+            {
+              slugsDisplayed.push($(this).attr('data-clipboard-slug'));
+            }
+        });
+
+      var slugsDisplayedString = slugsDisplayed.join(',');
+
+      // Query for status of displayed slugs and current clipboard counts
+      $.ajax({
+        url: this.$element.data('clipboard-status-url') + '?slugs=' + slugsDisplayedString,
```
jraddaoui: This URL may become quite long. I'd send the slugs as part of the request body

lib/task/tools/expireDataTask.class.php
```diff
@@ -0,0 +1,165 @@
...
+
+    // Delete clipbooard saves and save items
+    $deletedCount = 0;
+
+    foreach(QubitClipboardSave::get($criteria) as $save)
+    {
+      $itemCriteria = new Criteria;
+      $itemCriteria->add(QubitClipboardSaveItem::SAVE_ID, $save->id);
+
+      foreach(QubitClipboardSaveItem::get($itemCriteria) as $item)
```
jraddaoui: With `onDelete=cascade` this should not be needed

jraddaoui (changes requested):
Great job @mcantelon!

Nicely done to allow other types in the expire task in the future. I've added a few small comments.

apps/qubit/modules/user/actions/clipboardLoadAction.class.php
mcantelon: Yeah, I'm doing a bit of "cargo cult"-ing here. IIRC without this added to the class a fatal error gets generated and I noticed that a bunch of other classes do the same thing with this method. I'll see if I can track down what's calling this method and add some logic to see if the method exists before calling it. EDIT: found where it's caused and will fix.

apps/qubit/modules/user/actions/clipboardSaveAction.class.php
mcantelon: Good point.

apps/qubit/modules/user/actions/clipboardSaveAction.class.php
mcantelon: Yeah, this is a bit confusing. I called it "password" internally as that's more an accurate description of its purpose (given it's meant to be randomized rather than be a sequential ID).

lib/task/tools/expireDataTask.class.php
mcantelon: Good call.

mcantelon:
Thanks for the review, Radda... I've updated the PR.

apps/qubit/modules/default/actions/editAction.class.php
```diff
@@ -201,7 +201,11 @@ public function execute($request)
 
     $this->form = new sfForm;
 
-    $this->earlyExecute();
+    // Call early execute logic, if defined by a child class
+    if (method_exists($this, 'earlyExecute'))
+    {
+      call_user_func(array($this, 'earlyExecute'));
+    }
```
jraddaoui, line 208: Awesome!

jraddaoui (approved):
Thanks @mcantelon!

I'm not sure if the models need to be regenerated after the "on delete" changes in clipboard_save_item. Feel free to merge if it's not needed.

PR #609, opened 2017-08-23
Hide physical locations for --public ref #8993

MikeFE:
Hide physical location data from resulting csv in csv:export while using the
--public option and setting physical locations to not visible in the visible
elements module.


MikeFE:
https://github.com/artefactual/atom/blob/qa/2.4.x/lib/flatfile/csvInformationObjectExport.class.php#L95 wasn't being activated since $this->options wasn't being set in the csv:export task.

mcantelon:
LGTM! :+1:

PR #611, opened 2017-08-26
Check digitalObjectPath during csv import #11448

MikeFE:
Previously it was blindly trying to copy from a file located at
digitalObjectPath even if it didn't exist or we didn't have read perms. Now we
will display a warning to the user and gracefully skip trying to create the
digital object.

### DIFF
--- a/lib/task/import/csvImportTask.class.php
+++ b/lib/task/import/csvImportTask.class.php
@@ -162,41 +162,6 @@ EOF;
       QubitTaxonomy::PHYSICAL_OBJECT_TYPE_ID     => 'physicalObjectTypes'
     ));
 
-    // Allow default parent ID to be overridden by CLI options
-    if ($options['default-parent-slug'])
-    {
-      $defaultParentId = QubitFlatfileImport::getIdCorrespondingToSlug($options['default-parent-slug']);
-
-      if (!$options['quiet'])
-      {
-        print 'Parent ID of slug "'. $options['default-parent-slug'] .'" is '. $defaultParentId;
-      }
-    }
-    else if ($options['default-legacy-parent-id'])
-    {
-      // Attempt to fetch keymap entry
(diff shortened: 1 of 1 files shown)


lib/task/import/csvImportTask.class.php
```diff
@@ -139,43 +139,14 @@ public function execute($arguments = array(), $options = array())
...
-      $defaultStatusId = QubitTerm::PUBLICATION_STATUS_PUBLISHED_ID;
-    }
-
-    // TODO: this may be unnecessary as it may now be part of Qubit trunk
-    // create note term if it doesn't yet exist
-    QubitFlatfileImport::createOrFetchTerm(
-      QubitTaxonomy::NOTE_TYPE_ID,
-      'Language note'
-    );
+    $skipRows = $options['skip-rows'] ?: 0;
```
mcantelon, line 147: Didn't know about "?:"... slick.

lib/task/import/csvImportTask.class.php
```diff
@@ -1198,6 +1110,82 @@ public function execute($arguments = array(), $options = array())
...
+    {
+      if (false === $keyMapEntry = QubitFlatfileImport::fetchKeymapEntryBySourceAndTargetName(
+          $options['default-legacy-parent-id'], $sourceName, 'information_object'))
+      {
+        throw new sfException('Could not find keymap entry for default legacy parent ID '.
+                              $options['default-legacy-parent-id']);
+      }
+
+      $parentId = $keyMapEntry->target_id;
+      $this->log("Using default parent ID $parentId (legacy parent ID {$options['default-legacy-parent-id']})");
```
mcantelon, line 1180: Using sprintf instead of the curly braces to embed the option value might be cleaner, but that's subjective I guess so feel free to disregard.

mcantelon (approved):
One super minor style comment, but is good as-is! :+1:

MikeFE:
d059730

PR #612, opened 2017-08-29
Fix 'uploads/tmp/' dir creation, refs #11461

jraddaoui:
Make sure 'uploads/' is created first

### DIFF
--- a/lib/Qubit.class.php
+++ b/lib/Qubit.class.php
@@ -324,13 +324,8 @@ class Qubit
    */
   public static function moveUploadFile($file)
   {
-    // Create tmp dir if it doesn't exist already.
+    Qubit::createUploadDirsIfNeeded();
     $tmpDir = sfConfig::get('sf_upload_dir').'/tmp';
-    if (!file_exists($tmpDir))
-    {
-      mkdir($tmpDir);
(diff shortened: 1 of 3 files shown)


sbreker:
Looks good.  👍

lib/Qubit.class.php
```diff
@@ -351,4 +346,24 @@ public static function moveUploadFile($file)
...
+   */
+  public static function createUploadDirsIfNeeded()
+  {
+    $uploadsPath = sfConfig::get('sf_upload_dir');
+    $directories = array(
+      $uploadsPath,
+      $uploadsPath . DIRECTORY_SEPARATOR . 'tmp'
+    );
+
+    foreach ($directories as $dirPath)
```
MikeFE, line 361: Could use mkdir(..., ..., true) instead of a loop here to create both folders I think.

jraddaoui: Yes, but someone was complaining about some issues with the permissions on Windows, so I decided to leave it like that, at least we're not calling chmod(), heheh

sbreker: With this implementation if the need arose, we could add additional folders to the array.  🤔

PR #613, opened 2017-09-05
Protect user routes, Refs# 11427

sbreker:
These routes were able to be accessed with malformed urls (slug missing
from URL path). Access should only be allowed if the request contains
the user resource:
e.g.
<baseurl>/<userslug>/user or <baseurl>/<userslug>/user/delete


mcantelon:
This looks good to me. You could make parent action classes (UserExistingAction and UserRelateResourceAction or something) to do the two kinds of checks and subclass, but that might be overkill.

jraddaoui:
This was merged in https://github.com/artefactual/atom/commit/447031605da771a9e963fe4f68195a101d6dca50

PR #614, opened 2017-09-08
Remove APCu back. compatibility dep, refs #11387

mcantelon:
Remove reliance on APCu's API backwards compatibility to eliminate need to install APCu-bc. Fixed issue with creation time and modification time not working with older versions of APCu.


sbreker:
I did a quick grep for 'apc_' - there are a couple of apc_* calls in sfInstall.class.php inserted to fix an install caching issue which I think will need to be updated.  

I also see apc_* calls in vendor's symfony and symfony2 dirs - I am not sure if these will need updating too.

It'll be great to have this done and not worry about this dependency anymore!

sbreker:
See note on ticket. I am unsure about omitting sfInstall from this change.  It clears opcache and apc cache in succession - it does not short circuit if opcache_reset is present.

mcantelon:
I've clarified the sfInstall code.

sbreker:
👍  Looks good Mike.  Thanks for updating that sfInstall code!

mcantelon:
Merged.

PR #620, opened 2017-09-19
Dev/issue 11311 csv format check

mcantelon:
Added way to safeguard against users attempting to import non-CSV files as CSV.

Given any text file can technically be a CSV file, created logic to check the CSV header row and check what
column(s) are found to determine whether something seems like a legitimate CSV import.

* Added check of CSV header columns when attempting CSV importing, looking for columns
  specific to type of data being imported
* Got rid of necessity of having more than one CSV import/export config file for each type of data
* Updated CSV import/export config files so there's a place to find out what CSV columns can be used:
  * Added CSV import/export config files for some data types that didn't have one
  * Added specification of what CSV columns exist to base CSV import/export config files


jraddaoui (approved):
Looks great @mcantelon!

Please, make sure to merge with squash and reference the ticket in the commit message.

PR #621, opened 2017-09-20
Move max image width setting, Refs# 9308

sbreker:
This change moves the setting "Maximum image width (pixels)" from the
Global Settings page to Admin->Settings->Digital Object Derivatives
page.


mcantelon:
Looks good... didn't see any issues! :+1:

PR #622, opened 2017-09-21
Fix to APCu class loader, ref #11387

mcantelon:
Fixed a small but critical error.


sbreker:
Looks great - thanks!

PR #623, opened 2017-09-22
ES5 XML import query error, Refs# 11527

sbreker:
Change ES query in getByTitleIdentifierAndRepo() to use a match query
instead of term query. Match query should be used for exact match as it
compares to exact strings instead of analyzed text.

Prevent null values from getting passed in to query. Querying on null
values triggers errors under ES5.

Fix issue in QubitXmlImport where repository name was not being passed
in to getByTitleIdentifierAndRepo() for EAD XML child records.


mcantelon:
Good catch... code looks good! :+1:

MikeFE:
:+1:

PR #625, opened 2017-09-30
Use scroll API for CSV export job, refs #11534

mcantelon:
Fixes issue with CSV information object export job where large result sets
can't be retrieved from Elasticsearch without pagination.


mcantelon:
Closing as I'm going to fix some other jobs as part of a PR with wider scope.

PR #626, opened 2017-10-01
Dev/issue 11534 csv export job

mcantelon:
* Use scroll API for export jobs

Fixes issue with export jobs where large result sets can't be retrieved from Elasticsearch without pagination.

### DIFF
--- a/lib/job/arRepositoryCsvExportJob.class.php
+++ b/lib/job/arRepositoryCsvExportJob.class.php
@@ -37,7 +37,7 @@ class arRepositoryCsvExportJob extends arBaseJob
     $this->params = $parameters;
 
     // Create query increasing limit from default
-    $this->search = new arElasticSearchPluginQuery(1000000000);
+    $this->search = new arElasticSearchPluginQuery(arElasticSearchPluginUtil::SCROLL_SIZE);
     $this->search->queryBool->addMust(new \Elastica\Query\Terms('slug', $this->params['params']['slugs']));
 
     $this->downloadFileExtension = 'csv';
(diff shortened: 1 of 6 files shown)


lib/job/arRepositoryCsvExportJob.class.php
```diff
@@ -73,16 +73,17 @@ protected function exportResults($path)
...
 
-    foreach ($resultSet as $hit)
+    // Scroll through results, caching IDs in file, and iterate through IDs in file (given scroll result will expire)
+    foreach ($identifierIteratorFile = arElasticSearchPluginUtil::cacheScrolledSearchResultIdentifiersAsIterableFileObject($search) as $id)
     {
-      if (null === $resource = QubitRepository::getById($hit->getId()))
+      if (null === $resource = QubitRepository::getById($id))
       {
-        $this->error($this->i18n->__('Cannot fetch repository, id: '.$hit->getId()));
+        $this->error($this->i18n->__('Cannot fetch repository, id: '. $id));
```
jraddaoui: I don't know if we go with spaces around the `.` operator or not, but try not to use both options at the same time ;)

jraddaoui (commented):
It looks great @mcantelon!

I just wonder why did you decide to use a temporary file to cache the scroll results instead of our usual cache system. It shouldn't be a huge deal if we're only storing the hit ids.

Thanks!

lib/job/arRepositoryCsvExportJob.class.php
mcantelon: Haha I think I picked that up with Drupal standards or something (no space next to string, but space between var and concatenator). Maybe I'll go with no space as I'm not sure we have a standard.

mcantelon:
I was thinking of avoiding putting all IDs into memory, but I'm probably "prematurely optimizing" so I'll look into using our normal cache system.

lib/job/arRepositoryCsvExportJob.class.php
mcantelon: (Actually I'll use the __ method's sprintf-ish substitution as that's cleaner anyways and as-is it'll mess up translation I think.)

jraddaoui (approved):
Looks great, thanks Mike!

PR #627, opened 2017-10-02
Correct parameters to logSection, Refs# 11542

sbreker:
Under PHP 7.1, ArgumentCountError as now being elevated to an error
instead of a warning. This commit corrects the number of parameters being
passed to logSection() in the following CLI tasks:

- lib/task/digitalobject/digitalObjectDeleteTask.class.php
- lib/task/digitalobject/digitalObjectLoadTask.class.php
- lib/task/import/csvRepositoryImportTask.class.php
- lib/task/import/importDipObjectsTask.class.php
- lib/task/tools/deleteDescriptionTask.class.php

These jobs were failing to run under PHP 7.1 giving the error:

"PHP Fatal error: Uncaught ArgumentCountError: Too few arguments to
function sfCommandApplicationTask::logSection()"

### DIFF
--- a/lib/task/digitalobject/digitalObjectDeleteTask.class.php
+++ b/lib/task/digitalobject/digitalObjectDeleteTask.class.php
@@ -70,7 +70,7 @@ EOF;
 
     foreach ($ios as $io)
     {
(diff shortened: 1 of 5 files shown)


mcantelon:
Looks good! :+1:

PR #628, opened 2017-10-11
Fix CSV/XML import from IO page, refs #11597

jraddaoui:
- Do not process all request params in defaultEditAction, like parent
- Send only parent id to the import job as PDO objects can't be
  serialized to be added as job params


sbreker:
Looks great Radda!  👍

jraddaoui:
Cherry-picked to stable in https://github.com/artefactual/atom/commit/fcdee43df3f736bc589891e6339a9ae866e40e19

PR #629, opened 2017-10-13
Add MARC XML parser and import task

jraddaoui:
- Add arMarcXmlParser, extending QubitSaxParser. It parses and creates
  terms from a MARC XML file into a determinated taxonomy. Some
  assumptions have been made on how the labels and related/broader
  terms are parsed based on the example XML file used for development.
  This assumptions may not follow completely the MARC21 standart but
  this parser should be a good starting point for future development.

- Add arMarcXmlTermsImportTask, example call:

  `php symfony import:marc-xml /path/to/marc.xml taxonomy-slug`

  It parses the given XML file to the taxonomy (indicated by its
  slug) using the new arMarcXmlParser.

- Small tweak in QubitSaxParser to allow function calls on element tags
  with namespaces, replacing ':' by '_' in the tag names.

### DIFF
--- a/lib/QubitSaxParser.class.php
+++ b/lib/QubitSaxParser.class.php
@@ -111,15 +111,16 @@ class QubitSaxParser
     $this->storeAncestorIfNesting();
     $this->lastTagStartLevel = $this->level;
 
-    // Store current tag data
-    $this->tag = $tag;
+    // Store current tag data, concatenate namespace
+    // with '_' instead of ':' to allow function calls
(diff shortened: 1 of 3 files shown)


mcantelon:
Hi @jraddaoui ... I'm continuing work on this (Marc's pretty intense... yikes). I'll keep adding to this branch, if that doesn't create any issues, rather than making a new one.

jraddaoui:
Glad you can use part of it @mcantelon, I hope it's in a good shape ;)

PR #633, opened 2017-11-23
Fix FA dates of creation, rev and del, refs #11727

jraddaoui:
Fix path to include a paragraph element like it's being exported in EAD


sbreker (approved):
Looks good Radda!  👍

PR #634, opened 2017-11-23
Map LODs to EAD levels on export, refs #11723

jraddaoui:
Allow some LOD name variations to be exported as proper EAD levels so,
for example, a 'Sub-series' LOD is exported as:

`<c level="subseries">`

Instead of:

`<c level="otherlevel" otherlevel="sub-series">`

This affects how they are interpreted when Finding Aids are generated.


sbreker (approved):
Looks good!

PR #635, opened 2017-11-23
ES 5.x: fix fields included in all, refs #10847

jraddaoui:
String fields are now two different types in the ES mapping: 'keyword'
and 'text'. We now need to check those types instead of 'string' to
determine if a field should be included in the 'all' query.


sbreker (approved):
Looks good.  Good catch!

sbreker:
I think this should be Refs #10847.

jraddaoui:
Right, thanks @sbreker. For this one and the others ;)

PR #640, opened 2018-01-05
Stable/2.4.x Active Directory Integration

AnthonyCSheehy:
This addition based on the LDAP module adds basic Active Directory login integration support for logging in to AtoM. 

Our Archivist wants users to use their existing credentials and auto-creation of accounts in AtoM for our ~740 pupils and ~300 staff. Importing is not possible as AD stores credentials in a non-reversible format as per policy. Only a handful of staff will be given additional permissions. 

The Active Directory module differs from the LDAP module in only a handful of ways: 

1. Allows for LDAPS using a URI (e.g. LDAPS://<server>[:<port>]) rather than a host and port. For this to work the CA Certificate of the issuing authority that issues the LDAP server's certificate must be entered in to the file specified in /etc/ldap/ldap.conf (or equivalent.)

2. Uses the UserPrincipalName (UPN) to identify the user to the AD server. This is the Windows 2003+ user login of <username>@<domain> that replaced the <domain>\<username> a long time ago. This allows for identification of users throughout the directory, not just on a specified Organisational Unit within the directory. 

3. Reduces the number of settings to 2 - the URI and the Search Base - which should be the root of the directory.

4. Uses the email address to store the UPN and identify the returning user. It pulls the name attribute from the directory to store as the user name. The additional benefit of this is that a user that is already created in AtoM will not have the account overridden or re-created (should the email address match the UPN.)

I created a separate class so that the LDAP module can continue to be developed and this class will not interfere with that Module.

### DIFF
--- a/apps/qubit/modules/settings/actions/adAction.class.php
+++ b/apps/qubit/modules/settings/actions/adAction.class.php
@@ -1,0 +1,100 @@
+<?php
+
+/*
+ * This file is part of the Access to Memory (AtoM) software.
+ *
+ * Access to Memory (AtoM) is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * Access to Memory (AtoM) is distributed in the hope that it will be useful,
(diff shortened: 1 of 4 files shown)


AnthonyCSheehy:
Closed for pulling against the wrong version???

PR #643, opened 2018-01-06
Changes in IO ES documents

jraddaoui:
This could be merged with `--ff-only` as each commit references a different issue, if no changes are needed or they can be amended in the last commit.

Please, check the commit messages for more information.


mcantelon:
👍

PR #645, opened 2018-01-08
Permissive slugs, Refs# 11761

sbreker:
Add setting to Admin page to trigger new slug generation.

### DIFF
--- a/lib/task/migrate/migrations/arMigration0159.class.php
+++ b/lib/task/migrate/migrations/arMigration0159.class.php
@@ -1,0 +1,51 @@
+<?php
+
+/*
+ * This file is part of the Access to Memory (AtoM) software.
+ *
+ * Access to Memory (AtoM) is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * Access to Memory (AtoM) is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with Access to Memory (AtoM).  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+/*
+ * Add new setting for enabling institutional scoping.
+ *
+ * @package    AccesstoMemory
+ * @subpackage migration
+ */
+class arMigration0159
+{
+  const
+    VERSION = 159, // The new database version
+    MIN_MILESTONE = 2; // The minimum milestone required
+
+  /**
+   * Upgrade
+   *
+   * @return bool True if the upgrade succeeded, False otherwise
+   */
+  public function up($configuration)
+  {
+    if (null === QubitSetting::getByName('permissive_slug_creation'))
+    {
+      $setting = new QubitSetting;
+      $setting->name = 'permissive_slug_creation';
+      $setting->value = 0;
+      $setting->editable = 1;
+      $setting->culture = 'en';
+      $setting->save();
+    }
+
+    return true;
+  }
+}
--- a/lib/task/migrate/migrations/arMigration0159.class.php
+++ b/lib/task/migrate/migrations/arMigration0159.class.php
@@ -40,11 +40,14 @@ class arMigration0159
     {
       $setting = new QubitSetting;
       $setting->name = 'permissive_slug_creation';
-      $setting->value = 0;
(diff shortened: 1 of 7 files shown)


lib/task/migrate/migrations/arMigration0159.class.php
```diff
@@ -0,0 +1,54 @@
...
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with Access to Memory (AtoM).  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+/*
+ * Add new setting for enabling institutional scoping.
```
jraddaoui: Comment from other migration

lib/task/migrate/migrations/arMigration0159.class.php
```diff
@@ -0,0 +1,54 @@
...
+  public function up($configuration)
+  {
+    if (null === QubitSetting::getByName('permissive_slug_creation'))
+    {
+      $setting = new QubitSetting;
+      $setting->name = 'permissive_slug_creation';
+      $setting->value = QubitSlug::SLUG_RESTRICTIVE;
+      $setting->editable = 1;
+      $setting->culture = 'en';
+      $setting->save();
```
jraddaoui: Just to be sure the value is added to the source culture, I prefer:

```
$setting->source_culture = 'en';
$setting->setValue(QubitSlug::SLUG_RESTRICTIVE, array('culture' => 'en'));
```

lib/task/migrate/migrations/arMigration0159.class.php
```diff
@@ -0,0 +1,54 @@
...
+    }
+
+    $sql = "ALTER TABLE `slug` MODIFY `slug` VARCHAR(255) CHARACTER SET utf8 COLLATE utf8_bin";
```
jraddaoui: Do we need to modify the SQL schema to create the database with this change in new installs? Also, does it require to rebuild the models?

jraddaoui: Oh, I see it now. Nicely done ;)

plugins/sfInstallPlugin/lib/sfInstall.class.php
```diff
@@ -531,4 +533,17 @@ public static function addSymlinks()
 
     return $symlinks;
   }
+
+  /**
+   * Make any additional sql changes that cannot be made via schema.yml.
+   * Needs to be run for new installs and the tools:purge task.
```
jraddaoui, line 539: Should it be called from that task then?

test/unit/slugifyTest.php
```diff
@@ -0,0 +1,72 @@
...
+  $t->is(QubitSlug::slugify($given, QubitSlug::SLUG_RESTRICTIVE), $expected, "slugify(SLUG_RESTRICTIVE) creates slug from text: $given as $expected");
+}
+
+// Test Permissive mode slug creation.
+foreach ($testsPermissive as $item)
+{
+  list($given, $expected) = $item;
+
+  $t->is(QubitSlug::slugify($given, QubitSlug::SLUG_PERMISSIVE), $expected, "slugify(SLUG_PERMISSIVE) creates slug from text: $given as $expected");
+}
```
jraddaoui, line 72: Awesome!

jraddaoui (approved):
Looks great! Just a couple of minor comments.

lib/task/migrate/migrations/arMigration0159.class.php
sbreker: You are right!  Will take this out...

lib/task/migrate/migrations/arMigration0159.class.php
sbreker: Our version of Propel has a bug where I cannot set individual column collation.  Was forced to do that in sfInstall plugin for new installs and the CLI purge task instead.

plugins/sfInstallPlugin/lib/sfInstall.class.php
sbreker: sfInstall::loadData() is called from the tools:purge task here:

https://github.com/artefactual/atom/blob/qa/2.5.x/lib/task/tools/purgeTask.class.php#L115

test/unit/slugifyTest.php
sbreker: Thanks!  :)

jraddaoui:
Thanks! Please, merge in one single commit. And I think it looks better with refs in lowercase, as it's after a comma an with the # in the issue number.

jraddaoui:
Merged in 648d15a

PR #647, opened 2018-01-12
Fix search results CSV export, refs #11860

jraddaoui:
Do not include descendants in the exported records, add only
the records returned from the search query.


mcantelon:
Looks good to me! 👍

jraddaoui:
4e040a2 in stable/2.4.x

PR #648, opened 2018-01-16
Fix PHP 7 issue with build SQL task, refs #11639

mcantelon:
The propel:build-sql task was generating an error in PHP 7 due to the
DOMDocument's load method being called statically (although it's supposed
to work in PHP 7). Removed static call.


sbreker (approved):
Looks good!

PR #652, opened 2018-01-19
Avoid job temp folder collision, refs #11882

jraddaoui:
To create the export job temporary directory where the files will be added
before they are compressed and added to the downloads folder, use a MD5
hash created from instance info, job id and the current Epoch time to avoid
collisions when multiple AtoM instances are available on the same machine
and in instances where the database is regenerated from another dump (like
it's done in sites with public and private instances), where the job id
could be repeated, adding the export results to an existing export folder.


mcantelon (approved):
Looks good to me!

jraddaoui:
94ccc9b in stable/2.4.x

PR #653, opened 2018-01-20
Fix show all link in DO imageflow, refs #11669

jraddaoui:
Show only the DOs from the descendants of the current resource instead
of all the DOs in the collection:

- Send ancestor id instead of collection id from show all link.
- Allow filtering by ancestor in arElasticSearchPluginQuery.
- Add ancestor filter tag.

We could have used the collection parameter and change the related facet
to use the ancestors field, but the collection filter shows all descendants
and itself and in this case we only want the descendants.

Minor related changes:

- Fix add/remove realm in institutional scoping enabled instances.
- Fix Finding Aids search filter tag icon.


mcantelon:
👍

PR #654, opened 2018-01-25
Dev/issue 11075 perms

mcantelon:
* Added group-based security restrictions to a number of actions that didn’t have any
* Replaced some security-related action logic with security.yml rules
* Added method to user class for checking, using security.yml files, if a user has permission to use a module’s action
* Added logic to the menu model to check if a user has permission to access an action before displaying a menu item for it
* Added logic to limit display of the add menu to groups that can add content

### DIFF
--- a/lib/myUser.class.php
+++ b/lib/myUser.class.php
@@ -25,6 +25,10 @@ class myUser extends sfBasicSecurityUser implements Zend_Acl_Role_Interface
   // storage to save a list of selected information objects
   private $clipboard = null;
 
+  // Module-specific permissions get temporarily stored here for access checks
+  protected
+    $security = array();
+
   /**
    * Required for Zend_Acl_Role_Interface
    */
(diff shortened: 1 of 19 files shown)


lib/myUser.class.php
```diff
@@ -207,6 +211,66 @@ public function listGroups()
...
+    if ($file = sfContext::getInstance()->getConfigCache()->checkConfig($securityFilePath, true))
+    {
+      require($file);
+    }
+
+    // Get credentials, using security.yml parsing convention
+    $credentials = $this->getModuleSecurityValue($action, 'credentials');
+
+    // Allow access if action isn't secured or user has appropriate credentials
+    if (!$this->getModuleSecurityValue($action, 'is_secure', false) || $this->hasCredential($credentials))
```
jraddaoui: We could return the condition directly.

lib/myUser.class.php
```diff
@@ -207,6 +211,66 @@ public function listGroups()
...
+   *
+   * @return boolean
+   */
+  public function checkModuleActionAccess($module, $action)
+  {
+    // Set security property to module's security configuration
+    $securityFilePath = 'modules/'. $module .'/config/security.yml';
+    if ($file = sfContext::getInstance()->getConfigCache()->checkConfig($securityFilePath, true))
+    {
+      require($file);
```
jraddaoui, line 235: Is this loading the config. file inside `$this->security`? It'd be a lot clearer with `sfYaml::load()`.

jraddaoui (changes requested):
Looks great! Just a couple of minor suggestions.

lib/myUser.class.php
mcantelon: It's using a mechanism, used in other permission processing in Symfony, that returns the filename of a PHP file if it has already been parsed from YAML, so it should be slightly more performant.

I could always add logic to it does something similar with another layer of caching though.

jraddaoui: I didn't know, thanks @mcantelon!

Could you add a comment in the code explaining that? I think it will be helpful.

mcantelon: Will do... thanks!

PR #655, opened 2018-01-25
Use PDO data binding gen slugs task, Refs# 11899

sbreker:
Use PDO param/value binding. This ensures special chars are escaped
correctly in the insert statement.


jraddaoui (approved):
Nice!

PR #657, opened 2018-01-26
Skip empty rows during CSV import, refs #11646

mcantelon:
During CSV import, check each row column to make sure that there is at least
one non-blank value before importing the row.

### DIFF
--- a/lib/QubitFlatfileImport.class.php
+++ b/lib/QubitFlatfileImport.class.php
@@ -611,6 +611,13 @@ class QubitFlatfileImport
     {
       if ($this->status['rows'] >= $skipRows)
       {
+        // Skip blank rows, but keep track of rows parsed
+        if (!$this->rowContainsData($item))
+        {
+          $this->status['rows']++;
+          continue;
+        }
+
         if (!$timerStarted)
         {
           $this->startTimer();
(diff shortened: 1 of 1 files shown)


lib/QubitFlatfileImport.class.php
```diff
@@ -611,6 +611,13 @@ public function csv($fh, $skipRows = 0)
     {
       if ($this->status['rows'] >= $skipRows)
       {
+        // skip blank rows, but keep track of rows parsed
```
jraddaoui: Uppercase

lib/QubitFlatfileImport.class.php
```diff
@@ -653,6 +660,24 @@ public function csv($fh, $skipRows = 0)
     $this->executeClosurePropertyIfSet('completeLogic');
   }
 
+  /**
+   * Filter out elements containing blank strings and check if any elements
+   * remain
+   *
+   * @param array $row  array of column values
```
jraddaoui: Uppercase in the second array

jraddaoui (approved):
Looks good! Just a couple of cosmetic changes.

sbreker:
Looks good.  See Radda's comments above.

PR #658, opened 2018-01-31
Translate link to add new row via JS, refs #11605

mcantelon:
Updated multiJs.js, and templates that make use of it, so the "Add new" link
can be translated into different languages.


sbreker (approved):
Looks good Mike!

PR #659, opened 2018-02-05
Avoid unknown hidden fields in forms, refs #11950

jraddaoui:
Control what fields are added as hidden inputs to the IO adv. search
and repository adv. filters forms.


MikeFE:
If someone were to use garbage data in 'allowed' hidden fields would this potentially break our code too Radda? Or do you think it's safe? Other than that concern, :+1:

jraddaoui:
Hi @MikeFE, the garbage data is escaped if it's part of the value, so that shouldn't be a problem.

jraddaoui:
ddb0d9e4c3230e1d5257a8b35ce019054818b2da in stable/2.4.x

PR #660, opened 2018-02-06
Fix name access point lists #11828

MikeFE:
Previously, non-subject authority records were being added to the name access
point lists of archival descriptions. This was by design, however we have
decided to change this so only authority records that are actually subjects of
the archival description appear under the name access points. This PR also
removes the duplicate link between creators (or other roles) and the
"Subject of" lists when viewing an authority record page.


jraddaoui (approved):
Looks good!

PR #662, opened 2018-02-06
Improve description autocomplete fields #9255

MikeFE:
Change autocomplete ES search query when editing archival descriptions to use 
AND operator for terms instead of a wildcard query.


jraddaoui (commented):
Without the wildcards, will it hit 'Example record' when only  'Exam' is entered? If tit does, it looks good.

MikeFE:
Indeed it does :)

Thanks

PR #667, opened 2018-02-20
Update descendant inherited fields refs #11855

MikeFE:
For fields such as creator, repository and 'part of', descendant records were
not being updated on-the-fly to reflect said inherited fields from their parent
records when they were updated. This fix adds an async job to update descendant
records when the parent changes the mentioned fields.

### DIFF
--- a/apps/qubit/modules/informationobject/actions/editAction.class.php
+++ b/apps/qubit/modules/informationobject/actions/editAction.class.php
@@ -597,6 +597,8 @@ class InformationObjectEditAction extends DefaultEditAction
       $this->getUser()->setFlash('notice', $message);
     }
 
+    QubitJob::runJob('arUpdateDescendantsJob', array('tldId' => $this->resource->id));
+
     $this->deleteNotes();
     $this->updateChildLevels();
     $this->removeDuplicateRepositoryAssociations();
--- a/apps/qubit/modules/informationobject/actions/editAction.class.php
+++ b/apps/qubit/modules/informationobject/actions/editAction.class.php
@@ -597,10 +597,9 @@ class InformationObjectEditAction extends DefaultEditAction
       $this->getUser()->setFlash('notice', $message);
     }
 
-    QubitJob::runJob('arUpdateDescendantsJob', array('tldId' => $this->resource->id));
-
     $this->deleteNotes();
     $this->updateChildLevels();
+    $this->updateDescendantInheritedFields();
     $this->removeDuplicateRepositoryAssociations();
     $this->incrementMaskCounter();
   }
@@ -817,5 +816,22 @@ class InformationObjectEditAction extends DefaultEditAction
     $counter = QubitInformationObject::getIdentifierCounter();
     $counter->value++;
     $counter->save();
+  }
+
+  /**
+   * Trigger a job to update inherited fields in descendant records. Flash job started message to user.
+   */
+  private function updateDescendantInheritedFields()
+  {
+    $job = QubitJob::runJob('arUpdateDescendantsJob', array('tldId' => $this->resource->id));
+
+    $params = array(
+      '%1%' => sprintf('<a href="%s">', $this->context->routing->generate(null, array('module' => 'jobs', 'action' => 'report', 'id' => $job->id))),
+      '%2%' => $job->id,
+      '%3%' => '</a>'
+    );
+
+    $message = $this->context->i18n->__('Updating inherited fields for descendants. Check %1%job %2%%3% to view the status of the job.', $params);
+    $this->getUser()->setFlash('notice', $message);
   }
 }
--- a/apps/qubit/modules/informationobject/actions/editAction.class.php
+++ b/apps/qubit/modules/informationobject/actions/editAction.class.php
@@ -823,6 +823,11 @@ class InformationObjectEditAction extends DefaultEditAction
    */
   private function updateDescendantInheritedFields()
   {
+    if ($this->resource->rgt - $this->resource->lft <= 1)
+    {
+      return;
+    }
+
     $job = QubitJob::runJob('arUpdateDescendantsJob', array('tldId' => $this->resource->id));
 
     $params = array(
--- a/lib/job/arUpdateDescendantsJob.class.php
+++ b/lib/job/arUpdateDescendantsJob.class.php
@@ -19,7 +19,6 @@
 
 /**
  *
- *
  * @package    symfony
  * @subpackage jobs
  */
@@ -43,8 +42,6 @@ class arUpdateDescendantsJob extends arBaseJob
     $searchIo = new arElasticSearchInformationObject;
     $searchIo->recursivelyUpdateInformationObjects($parameters['tldId'], count($this->topLevelDesc->descendants),
       'updateInheritedFields');
-
-
     $this->info($this->i18n->__('Update(s) completed.'));
     $this->job->setStatusCompleted();
     $this->job->save();
--- a/lib/job/arUpdateDescendantsJob.class.php
+++ b/lib/job/arUpdateDescendantsJob.class.php
@@ -1,0 +1,54 @@
+<?php
+
+/*
+ * This file is part of the Access to Memory (AtoM) software.
+ *
+ * Access to Memory (AtoM) is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * Access to Memory (AtoM) is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with Access to Memory (AtoM).  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+/**
+ *
+ *
+ * @package    symfony
+ * @subpackage jobs
+ */
+
+class arUpdateDescendantsJob extends arBaseJob
+{
+  /**
+   * @see arBaseJob::$requiredParameters
+   */
+  protected $extraRequiredParameters = array('tldId');
+
+  public function runJob($parameters)
+  {
+    if (null === $this->topLevelDesc = QubitInformationObject::getById($parameters['tldId']))
+    {
+      $this->error($this->i18n->__('Invalid description id: %1', array('%1' => $parameters['tldId'])));
+      return false;
+    }
+
+    $this->info($this->i18n->__("Updating child descriptions' inherited fields from ancestors (id: %1)", array('%1' => $this->topLevelDesc->id)));
+    $searchIo = new arElasticSearchInformationObject;
+    $searchIo->recursivelyUpdateInformationObjects($parameters['tldId'], count($this->topLevelDesc->descendants),
+      'updateInheritedFields');
+
+
+    $this->info($this->i18n->__('Update(s) completed.'));
+    $this->job->setStatusCompleted();
+    $this->job->save();
+
+    return true;
+  }
+}
--- a/lib/job/arUpdateDescendantsJob.class.php
+++ b/lib/job/arUpdateDescendantsJob.class.php
@@ -41,7 +41,9 @@ class arUpdateDescendantsJob extends arBaseJob
     $this->info($this->i18n->__("Updating child descriptions' inherited fields from ancestors (id: %1)", array('%1' => $this->topLevelDesc->id)));
     $searchIo = new arElasticSearchInformationObject;
     $searchIo->recursivelyUpdateInformationObjects($parameters['tldId'], count($this->topLevelDesc->descendants),
-      'updateInheritedFields');
+      'updateInheritedFields', array('repository' => $this->topLevelDesc->getRepository(),
+      'inheritedCreators' => $this->topLevelDesc->getCreators()));
+
     $this->info($this->i18n->__('Update(s) completed.'));
     $this->job->setStatusCompleted();
     $this->job->save();
--- a/plugins/arElasticSearchPlugin/lib/model/arElasticSearchInformationObject.class.php
+++ b/plugins/arElasticSearchPlugin/lib/model/arElasticSearchInformationObject.class.php
@@ -41,13 +41,31 @@ class arElasticSearchInformationObject extends arElasticSearchModelBase
     $this->count = QubitPdo::fetchColumn($sql, array(QubitInformationObject::ROOT_ID));
 
     // Recursively descend down hierarchy
-    $this->recursivelyAddInformationObjects(QubitInformationObject::ROOT_ID, $this->count);
+    $this->recursivelyUpdateInformationObjects(QubitInformationObject::ROOT_ID, $this->count, 'updateAllFields');
 
     return $this->errors;
   }
 
-  public function recursivelyAddInformationObjects($parentId, $totalRows, $options = array())
+  /**
+   * Descends an archival description hierarchy and calls a specified update function on each description to update it
+   * in the ES index.
+   *
+   * @param int $parentId  The parent archival description of the current one.
+   * @param int $totalRows The total archival description count in the hierarchy
+   *
+   * @param string $updateFunc The string indicating the name of the update function within $this class instance.
+   *                           The update function takes the returned row object from the getChildren query and
+   *                           returns an instance of arElasticSearchInformationObjectPdo for the current description.
+   *
+   * @param array $options  An array which we use to pass down 'inherited' fields like repository and inheritedCreators.
+   */
+  public function recursivelyUpdateInformationObjects($parentId, $totalRows, $updateFunc, $options = array())
   {
+    if (!isset(self::$conn))
+    {
+      self::$conn = Propel::getConnection();
+    }
+
     // Get information objects
     if (!isset(self::$statements['getChildren']))
     {
@@ -69,20 +87,13 @@ class arElasticSearchInformationObject extends arElasticSearchModelBase
     {
       try
       {
-        $node = new arElasticSearchInformationObjectPdo($item->id, $options);
-        $data = $node->serialize();
-
-        QubitSearch::getInstance()->addDocument($data, 'QubitInformationObject');
-
-        self::$counter++;
-
-        $this->logEntry($data['i18n'][$data['sourceCulture']]['title'], self::$counter);
+        $node = $this->$updateFunc($item, $options);
 
         // Descend hierarchy
         if (1 < ($item->rgt - $item->lft))
         {
           // Pass ancestors, repository and creators down to descendants
-          $this->recursivelyAddInformationObjects($item->id, $totalRows, array(
+          $this->recursivelyUpdateInformationObjects($item->id, $totalRows, $updateFunc, array(
             'ancestors'  => array_merge($node->getAncestors(), array($node)),
             'repository' => $node->getRepository(),
             'inheritedCreators' => array_merge($node->inheritedCreators, $node->creators)));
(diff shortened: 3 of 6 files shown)


lib/job/arUpdateDescendantsJob.class.php
```diff
@@ -0,0 +1,54 @@
...
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with Access to Memory (AtoM).  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+/**
+ *
```
jraddaoui, line 21: Extra line.

lib/job/arUpdateDescendantsJob.class.php
```diff
@@ -0,0 +1,54 @@
...
+    {
+      $this->error($this->i18n->__('Invalid description id: %1', array('%1' => $parameters['tldId'])));
+      return false;
+    }
+
+    $this->info($this->i18n->__("Updating child descriptions' inherited fields from ancestors (id: %1)", array('%1' => $this->topLevelDesc->id)));
+    $searchIo = new arElasticSearchInformationObject;
+    $searchIo->recursivelyUpdateInformationObjects($parameters['tldId'], count($this->topLevelDesc->descendants),
+      'updateInheritedFields');
+
```
jraddaoui: Extra line.

plugins/arElasticSearchPlugin/lib/model/arElasticSearchInformationObject.class.php
```diff
@@ -95,6 +111,56 @@ public function recursivelyAddInformationObjects($parentId, $totalRows, $options
...
+    // Add "Part of" title if this isn't a top level description
+    if (null !== $partOf = $node->getPartOf())
+    {
+      $data['partOf'] = $partOf;
+    }
+
+    if ($data)
+    {
+      $data['_id'] = $item->id;
+      QubitSearch::getInstance()->addDocument($data, 'QubitInformationObject');
```
jraddaoui: This will replace the existing document with a new one that will only have the inherited data, everything else will be removed. Please, take a look at `partialUpdate`:

https://github.com/artefactual/atom/blob/qa/2.5.x/plugins/arElasticSearchPlugin/lib/arElasticSearchPlugin.class.php#L456-L486

As noted in that function comment, the way the merge works in `updateDocument` means that, if a field is not present in the new data, it won't be deleted from the document, so it should be set to `null` to do so.

Right now, only the transcript fields are expecting partial updates. See the following comment:

https://github.com/artefactual/atom/blob/qa/2.5.x/plugins/arElasticSearchPlugin/lib/model/arElasticSearchInformationObjectPdo.class.php#L1070-L1075

And how the partial update is done from the finding aid job:

https://github.com/artefactual/atom/blob/qa/2.5.x/lib/job/arFindingAidJob.class.php#L186
https://github.com/artefactual/atom/blob/qa/2.5.x/lib/job/arFindingAidJob.class.php#L275
https://github.com/artefactual/atom/blob/qa/2.5.x/lib/job/arFindingAidJob.class.php#L319-L326

If we handle `null` values this way, most likely we won't need changes in the queries and aggregations made over those fields, but we should double check. We may need to use empty arrays instead as this are "multifields".

plugins/arElasticSearchPlugin/lib/model/arElasticSearchInformationObjectPdo.class.php
```diff
@@ -1334,10 +1377,10 @@ public function serialize()
...
       );
 
-      $serialized['partOf']['id'] = $collectionRootId;
-      $serialized['partOf']['sourceCulture'] = $rootSourceCulture;
-      $serialized['partOf']['slug'] = $rootSlug;
-      $serialized['partOf']['i18n'] = $i18nFields;
+      if (null !== $partOf = $this->getPartOf($collectionRootId))
+      {
+        $serialized['partOf'] = $partOf;
+      }
```
jraddaoui: We are doing things twice in here. You could replace the entire `if` wrapping this one with this one and remove the `$collectionRootId = $this->getCollectionRootId();` call made before the `if`, as you're getting it in the new `getPartOf` function, which also makes unnecessary to pass it to the function.

apps/qubit/modules/informationobject/actions/editAction.class.php
```diff
@@ -597,6 +597,8 @@ protected function processForm()
       $this->getUser()->setFlash('notice', $message);
     }
```
jraddaoui: Like with the publication status change made on top and like we do with other jobs, could you add a flash notice telling that job has been started to update the descendants?

apps/qubit/modules/informationobject/actions/editAction.class.php
```diff
@@ -597,6 +597,8 @@ protected function processForm()
...
 
+    QubitJob::runJob('arUpdateDescendantsJob', array('tldId' => $this->resource->id));
```
jraddaoui: What about other places where a description could be edited? Shouldn't we call the job from the IO save method?

We should also avoid the job execution if the description has no descendants, checking that `rgt - lft > 1`, like in:

https://github.com/artefactual/atom/blob/qa/2.5.x/apps/qubit/modules/informationobject/templates/updatePublicationStatusSuccess.php#L29

lib/job/arUpdateDescendantsJob.class.php
```diff
@@ -0,0 +1,54 @@
...
+ */
+
+/**
+ *
+ *
+ * @package    symfony
+ * @subpackage jobs
+ */
+
+class arUpdateDescendantsJob extends arBaseJob
```
jraddaoui, line 26: I think we could improve a lot this job to accept other big ES documents updates without a big amount of changes.

I'd do the following:
- Rename the job to arUpdateIoEsDocumentsJob (or similar).
- Require the following parameters:
  - An array of IO ids.
  - A boolean to update the IOs from the array or not.
  - A boolean to update their descendants or not.
- Change the process logic to:
  - Update the IOs from the array based on the first boolean.
  - Update their descendants based on the second boolean.

That way we could use this job in this case and in other big IO updates, like when a term or an actor is updated:

https://github.com/artefactual/atom/blob/qa/2.5.x/apps/qubit/modules/term/actions/editAction.class.php#L578-L589
https://github.com/artefactual/atom/blob/qa/2.5.x/lib/model/QubitActor.php#L146-L153

plugins/arElasticSearchPlugin/lib/model/arElasticSearchInformationObjectPdo.class.php
```diff
@@ -1063,6 +1063,49 @@ private function getActRights()
...
+  public function getPartOf($collectionRootId)
+  {
+    $collectionRootId = $this->getCollectionRootId();
+
+    if ($collectionRootId && $collectionRootId != $this->id)
+    {
+      $rootSlug = QubitPdo::fetchColumn('SELECT slug FROM slug WHERE object_id=?', array($collectionRootId));
+      if (!$rootSlug)
+      {
+        throw new sfException("No slug found for information object $collectionRootId");
```
jraddaoui, line 1081: Throwing an exception in here will avoid the entire document to be added to the ES index. I think it's better to have it without part of the `parOf` data than not having the document in the index.

plugins/arElasticSearchPlugin/lib/model/arElasticSearchInformationObjectPdo.class.php
```diff
@@ -1063,6 +1063,49 @@ private function getActRights()
...
+      }
+
+      $rootSourceCulture = QubitPdo::fetchColumn('SELECT source_culture FROM information_object WHERE id=?',
+                                                array($collectionRootId));
+      if (!$rootSourceCulture)
+      {
+        throw new sfException("No source culture found for information object $collectionRootId");
```
jraddaoui: Same here.

jraddaoui (changes requested):
Hi @MikeFE,

We could do a big improvement to the job that should not require too much time and it will allow us to improve other parts of the application where a lot of ES documents are being updated.

See also my other comments bellow.

lib/job/arUpdateDescendantsJob.class.php
jraddaoui: Ideally, this job could accept a partial update option, where only some fields are updated instead of the entire IOs, but that could be done in the future. This option should only be considered in the IOs from the array, descendants should always use partial updates, like it's being done.

Later, we could rename and add other resources to this job, like QubitActor and QubitTerm.

plugins/arElasticSearchPlugin/lib/model/arElasticSearchInformationObjectPdo.class.php
MikeFE: ah good call, yeah I missed that when refactoring

apps/qubit/modules/informationobject/actions/editAction.class.php
MikeFE: agreed, I will implement it

plugins/arElasticSearchPlugin/lib/model/arElasticSearchInformationObjectPdo.class.php
MikeFE: The first condition in the if is from a result from a db query indicating an info object exists. If one exists without a slug and isn't ROOT_ID, I think that's data corruption? I don't think it's wise to continue the code path here.

apps/qubit/modules/informationobject/actions/editAction.class.php
```diff
@@ -816,4 +817,26 @@ private function incrementMaskCounter()
...
+
+    $params = array(
+      '%1%' => sprintf('<a href="%s">', $this->context->routing->generate(null, array('module' => 'jobs', 'action' => 'report', 'id' => $job->id))),
+      '%2%' => $job->id,
+      '%3%' => '</a>'
+    );
+
+    $message = $this->context->i18n->__('Updating inherited fields for descendants. Check %1%job %2%%3% to view the status of the job.', $params);
+    $this->getUser()->setFlash('notice', $message);
+  }
```
jraddaoui, line 841: I still think the job should be triggered from the IO save, as we now have an 'update on match' type of import and other places that may be making changes to the IOs.

plugins/arElasticSearchPlugin/lib/model/arElasticSearchInformationObject.class.php
```diff
@@ -95,6 +111,59 @@ public function recursivelyAddInformationObjects($parentId, $totalRows, $options
...
+    QubitSearch::getInstance()->addDocument($data, 'QubitInformationObject');
+
+    self::$counter++;
+    $this->logEntry($data['i18n'][$data['sourceCulture']]['title'], self::$counter);
+
+    return $node;
+  }
+
+  /**
+   * Update ES fields descendants inherit from their ancestors. This assumes QubitInformationObject atm.
```
jraddaoui, line 128: We are in arElasticSearchInformationObject ;)

lib/job/arUpdateDescendantsJob.class.php
```diff
@@ -0,0 +1,53 @@
...
+    {
+      $this->error($this->i18n->__('Invalid description id: %1', array('%1' => $parameters['tldId'])));
+      return false;
+    }
+
+    $this->info($this->i18n->__("Updating child descriptions' inherited fields from ancestors (id: %1)", array('%1' => $this->topLevelDesc->id)));
+    $searchIo = new arElasticSearchInformationObject;
+    $searchIo->recursivelyUpdateInformationObjects($parameters['tldId'], count($this->topLevelDesc->descendants),
+      'updateInheritedFields', array('repository' => $this->topLevelDesc->getRepository(),
+      'inheritedCreators' => $this->topLevelDesc->getCreators()));
```
jraddaoui, line 45: Cosmetic change: at first glace it makes me think `updateInheritedFields` and `inheritedCreators` belongs to the same array, I'd put the array in a new line.

plugins/arElasticSearchPlugin/lib/model/arElasticSearchInformationObject.class.php
```diff
@@ -95,6 +111,59 @@ public function recursivelyAddInformationObjects($parentId, $totalRows, $options
...
+   */
+  private function updateInheritedFields($item, $options = array())
+  {
+    $node = new arElasticSearchInformationObjectPdo($item->id, $options);
+    $data = array();
+
+    if (isset($options['repository']) && null !== $repository = $node->getRepository())
```
jraddaoui, line 135: Do you really need to check it's in the options array?

plugins/arElasticSearchPlugin/lib/model/arElasticSearchInformationObject.class.php
```diff
@@ -95,6 +111,59 @@ public function recursivelyAddInformationObjects($parentId, $totalRows, $options
...
+    {
+      $data['repository'] = arElasticSearchRepository::serialize($repository);
+    }
```
jraddaoui, line 138: Otherwise set `$data['repository']` to `null` or empty array (you'll need to check what's better in this 'multifields')

plugins/arElasticSearchPlugin/lib/model/arElasticSearchInformationObject.class.php
```diff
@@ -95,6 +111,59 @@ public function recursivelyAddInformationObjects($parentId, $totalRows, $options
...
+
+    if (isset($options['inheritedCreators']))
```
jraddaoui, line 140: Again, not sure why you need to check this.

plugins/arElasticSearchPlugin/lib/model/arElasticSearchInformationObject.class.php
```diff
@@ -95,6 +111,59 @@ public function recursivelyAddInformationObjects($parentId, $totalRows, $options
...
+    {
+      foreach ($options['inheritedCreators'] as $creator)
+      {
+        if (!isset($data['inheritedCreators']))
+        {
+          $data['inheritedCreators'] = array();
```
jraddaoui, line 146: You'll need an empty array or null in the end, so I'd create it before the foreach to avoid this if.

jraddaoui: Also it's called `creators` in the index, not `inheritedCreators`. This should have failed while testing the changes.

plugins/arElasticSearchPlugin/lib/model/arElasticSearchInformationObject.class.php
```diff
@@ -95,6 +111,59 @@ public function recursivelyAddInformationObjects($parentId, $totalRows, $options
...
+    $data = array();
+
+    if (isset($options['repository']) && null !== $repository = $node->getRepository())
+    {
+      $data['repository'] = arElasticSearchRepository::serialize($repository);
+    }
+
+    if (isset($options['inheritedCreators']))
+    {
+      foreach ($options['inheritedCreators'] as $creator)
```
jraddaoui, line 142: You're going only over the parent creators in this foreach. In the index and in the GUI we only inherit the parent creators if the IO doesn't have creators by itself. Check:

https://github.com/artefactual/atom/blob/qa/2.5.x/plugins/arElasticSearchPlugin/lib/model/arElasticSearchInformationObjectPdo.class.php#L80-L87
https://github.com/artefactual/atom/blob/qa/2.5.x/plugins/arElasticSearchPlugin/lib/model/arElasticSearchInformationObjectPdo.class.php#L1228-L1240

You should do the foreach over `$node->creators` and `$node->inheritedCreators`. Only one of then will have data after the constructor. Actually, to avoid having two foreach, you could do it over an array merge of both arrays (and change it in arElasticSearchInformationObjectPdo too, please).

plugins/arElasticSearchPlugin/lib/model/arElasticSearchInformationObject.class.php
```diff
@@ -95,6 +111,59 @@ public function recursivelyAddInformationObjects($parentId, $totalRows, $options
...
+        $creatorNode = new arElasticSearchActorPdo($creator->id);
+        $data['inheritedCreators'][] = $creatorNode->serialize();
+      }
+    }
+
+    // Add "Part of" title if this isn't a top level description
+    if (null !== $partOf = $node->getPartOf())
+    {
+      $data['partOf'] = $partOf;
+    }
```
jraddaoui, line 158: Otherwise it should be set to null or empty array, like the other data.

plugins/arElasticSearchPlugin/lib/model/arElasticSearchInformationObject.class.php
```diff
@@ -95,6 +111,59 @@ public function recursivelyAddInformationObjects($parentId, $totalRows, $options
...
+
+    // If any of the inherited fields have been set to other than null, do partial update
+    if (array_filter($data))
+    {
+      QubitSearch::getInstance()->partialUpdate(QubitInformationObject::getById($item->id), $data);
+    }
```
jraddaoui, line 164: We should do the partial update in any case, to remove the fields when needed.

plugins/arElasticSearchPlugin/lib/model/arElasticSearchInformationObjectPdo.class.php
```diff
@@ -1063,6 +1063,50 @@ private function getActRights()
...
+   * Get an array outlining which top level description the current description is part of.
+   *
+   * @return array  An array containing the id and various i18n titles of the top level description.
+   *                If this description is already a TLD or another issue occurs, return null.
+   */
+  public function getPartOf()
+  {
+    $collectionRootId = $this->getCollectionRootId();
+
+    if ($collectionRootId && $collectionRootId != $this->id)
```
jraddaoui, line 1076: You could reverse the condition and return null directly in here, to avoid having all the code bellow indented to just return null after the if.

plugins/arElasticSearchPlugin/lib/model/arElasticSearchInformationObject.class.php
```diff
@@ -95,6 +111,59 @@ public function recursivelyAddInformationObjects($parentId, $totalRows, $options
...
+    // Add "Part of" title if this isn't a top level description
+    if (null !== $partOf = $node->getPartOf())
+    {
+      $data['partOf'] = $partOf;
+    }
+
+    // If any of the inherited fields have been set to other than null, do partial update
+    if (array_filter($data))
+    {
+      QubitSearch::getInstance()->partialUpdate(QubitInformationObject::getById($item->id), $data);
```
jraddaoui, line 163: This looks a little overkill in this case, after all the work to improve the population using raw queries, I'd try to avoid having to hydrate the entire IO to do the partial update call. The partial update function only requires the entire resource in the case the document is not found in the index:

https://github.com/artefactual/atom/blob/qa/2.5.x/plugins/arElasticSearchPlugin/lib/arElasticSearchPlugin.class.php#L484

I'd change the partialUpdate function, to only require the class of the resource, the id and the partial data. And, if the update fails because the document is not found in the index, I'd do the hydration in there, using the class name and the id before calling the full update (and change the calls from the finding aid job).

jraddaoui (changes requested):
Looking better @MikeFE , I've added a few more comments ;)

lib/job/arUpdateDescendantsJob.class.php
jraddaoui: Feel free to disregard this if running out of time. We can create a new feature ticket to not forgot about the enhancement in the future.

plugins/arElasticSearchPlugin/lib/model/arElasticSearchInformationObjectPdo.class.php
jraddaoui: I agree this is a data corruption issue on the ancestor, but I don't think it's enough to remove all the descendants from the index. The search populate task will show an error at the end about the ancestor issue but, if this is called from the GUI, and somehow the ancestor lost its slug, you may end up loosing a lot of records without noticing.

plugins/arElasticSearchPlugin/lib/model/arElasticSearchInformationObject.class.php
jraddaoui: Feel free to disregard this too if running out of time. We can create a new feature ticket to not forgot about the enhancement in the future.

jraddaoui:
Thank you very much for this work @MikeFE!

I'll move to it to a new PR to avoid the new reviewer the need of reading all this comments.

jraddaoui:
Moved to #684

PR #669, opened 2018-02-22
Accession Import col fix, refs #11993

sbreker:
Remove reference (&) from $self param to addColumnHandler. Passing $self
as a reference was preventing the column handler functions from running.

This change will allow the correct import of the accession columns:
- acquisitionDate
- resourceType
- acquisitionType
- processingStatus
- processingPriority


mcantelon:
👍

MikeFE:
:+1:

PR #670, opened 2018-02-26
Avoid pagination over 10K records, refs #11999

jraddaoui:
In Elasticsearch versions after 1.x they introduced a new setting to avoid
memory issues in deep pagination, which limits pagination by default to
10,000 records. This value could be increased but the requirements will
be too high in instances with a really high amount of records.

Other solutions, like the 'Search after' and 'Scroll' APIs added to ES doesn't
suit our needs in most of the cases. Therefore, this commit avoids pagination
over 10,000 records when hitting the ES index.

The following pages will redirect the user to the first page with an alert when
they go over 10,000 records:

- IO browse
- Actor browse
- Repository browse
- Accessions browse
- Clipboard
- Move action
- Description updates
- IO inventory report
- Taxonomy index page
- Term index page (IO results)

The following sidebar lists will stop paginating after 10,000 records. They
won't show any warning and will allow paginating back but not after:

- Actor index page (related descriptions lists)
- Repository index page (holdings list)
- Repository index page (maintained actors list)
- Term index page (taxonomy terms list)

The following API endpoint will return an error message about the issue
when paginating over 10,000 records:

- IO browse endpoint

This lists and the API endpoint are good candidates to implement the 'Search
after' API in the future.

The pager template has not been modified, so it will still show the pages over
10,000 records, this is also something that could be improved later.

This commit also fixes a regression in arInformationObjectCsvExportJob.

### DIFF
--- a/apps/qubit/modules/actor/actions/relatedInformationObjectsAction.class.php
+++ b/apps/qubit/modules/actor/actions/relatedInformationObjectsAction.class.php
@@ -33,6 +33,13 @@ class ActorRelatedInformationObjectsAction extends sfAction
     $limit = sfConfig::get('app_hits_per_page', 10);
     $culture = $this->context->user->getCulture();
 
+    // Avoid pagination over 10000 records
+    if ((int)$limit * (int)$request->page > 1000)
+    {
+      // Return nothing to not break the list
+      return;
+    }
+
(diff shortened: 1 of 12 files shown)


sbreker:
Looks great Radda!   👍

sbreker (approved):
CR looks great!

jraddaoui:
Squashed in d8011c8f60a33351a8cd047f0a47b6779ba7c101

PR #671, opened 2018-02-27
Allow alternative sort directions, refs #12000

jraddaoui:
- Include another generic picker with sort direction options to the
  default sort picker and rename it to 'sortPickers'.
- Layout and style improvements.
- Handle sortDir param in actions.
- Rename sort option labels.

Pages that will have sort direction options after this changes:

- Actor browse
- Function browse
- IO browse
- Repository browse
- Rights holders browse
- Taxonomy index
- Term index
- Clipboard
- Accession browse
- Donor browse


sbreker (approved):
Looks good @jraddaoui!

PR #672, opened 2018-03-03
Prevent description self-relation, refs #11960

mcantelon:
Don't allow a description to be related to itself and show a warning if user
attempts.

### DIFF
--- a/apps/qubit/modules/informationobject/actions/editAction.class.php
+++ b/apps/qubit/modules/informationobject/actions/editAction.class.php
@@ -451,12 +451,25 @@ class InformationObjectEditAction extends DefaultEditAction
               break;
 
             case 'relatedMaterialDescriptions':
-              $relation->typeId = QubitTerm::RELATED_MATERIAL_DESCRIPTIONS_ID;
+              // Don't allow self-relations
+              if (!isset($this->resource->id) || ($this->resource->id != $item->id))
+              {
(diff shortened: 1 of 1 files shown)


apps/qubit/modules/informationobject/actions/editAction.class.php
```diff
@@ -451,7 +451,16 @@ protected function processField($field)
...
+              // Don't allow self-relations
+              if (!isset($this->resource->id) || ($this->resource->id != $item->id))
+              {
+                $relation->typeId = QubitTerm::RELATED_MATERIAL_DESCRIPTIONS_ID;
+              }
+              else
+              {
+                $message = $this->context->i18n->__('Self-relation ignored.');
+                $this->getUser()->setFlash('notice', $message);
+              }
```
jraddaoui, line 463: Hi @mcantelon, is the relation not being created or it's being created without a type id after this change? It looks like the type id is not a required column, so I guess it's being created, which is not ideal.

mcantelon: Yikes... yeah, that's definitely not optimal. I'll fix.

jraddaoui: Nice! thanks!

jraddaoui (approved):

PR #673, opened 2018-03-05
Permission-related fix to menu items, refs #11075

mcantelon:
This fixes a permissions-related issue preventing the "Subjects" and
"Places" links from being displayed on the home page and in the "Browse"
menu.

The permission checking logic in the menu display relies on permissions from
security.yml files. The menu items meant to direct users to the index pages
for the place and subject taxonomies had paths that linked to the taxonomy
browse action instead which would then redirect to the taxonomy index action.
Because of restrictions on the taxonomy browse action this was causing the
menu items for places and subjects not to appear in menus for
non-authenticated users.

This is fixed by:

* Fixing an issue with the taxonomy module’s index action that prevented it
  from working correctly when called with an ID as a GET parameter
* Moving access control for the taxonomy module’s index action from
  security.yml to the action logic so place and subject can be accessible to all
* Changing the paths in the menu items for the place and subject taxonomies,
  using a data migration, to so the menu paths aren’t reliant on the taxonomy
  module browse action to redirect to the taxonomy module index action

This commit also removes some unnecessary logic from the taxonomy browse action.


apps/qubit/modules/taxonomy/actions/browseAction.class.php
```diff
@@ -21,71 +21,6 @@ class TaxonomyBrowseAction extends sfAction
...
-        break;
-    }
-
-    $this->pager = new QubitPager('QubitTerm');
-    $this->pager->setCriteria($criteria);
-    $this->pager->setMaxPerPage($request->limit);
-    $this->pager->setPage($request->page);
-
-    $this->terms = $this->pager->getResults();
+    $this->redirect(array('module' => 'taxonomy', 'action' => 'list'));
```
jraddaoui, line 24: Nice. We could also remove the "browseSuccess" template after this.

apps/qubit/modules/taxonomy/actions/indexAction.class.php
```diff
@@ -49,6 +49,17 @@ public function execute($request)
       $this->forward404();
     }
 
+    // Restrict access (except to places and subject taxonomies)
+    $unrestrictedTaxonomies = array(QubitTaxonomy::PLACE_ID, QubitTaxonomy::SUBJECT_ID);
```
jraddaoui: I think @fiver-watson was asking for the genres taxonomy to be included in here too. We could also add a browse menu item for it now that we're adding a migration to modify them, if Dan likes the idea.

lib/task/migrate/migrations/arMigration0160.class.php
```diff
@@ -0,0 +1,65 @@
...
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with Access to Memory (AtoM).  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+/*
+ * Add new setting for permissive slug generation. DB column collation change.
```
jraddaoui: Hehe, it always happens.

lib/task/migrate/migrations/arMigration0160.class.php
```diff
@@ -0,0 +1,65 @@
...
+    $criteria->add(QubitMenu::NAME, 'browseSubjects');
+
+    $menu = QubitMenu::getOne($criteria);
+
+    if ($menu !== null)
+    {
+      $menu->path = 'taxonomy/index?id=35';
+      $menu->save();
+    }
+
```
jraddaoui, line 62: We're caching the menus, which I believe is the reason of @fiver-watson not seeing the changes in the homepage. I'm not sure if this will be an issue in all the upgrades as it's using the "ViewCacheManager" and probably removing the ".cache" folder is enough. But, could you try to add something like what we use in the menu edit action to this migration? See:

https://github.com/artefactual/atom/blob/qa/2.5.x/apps/qubit/modules/menu/actions/editAction.class.php#L151-L156

jraddaoui (changes requested):
Hi @mcantelon,

Looking great! Just a couple of suggestions and a comment about the genres taxonomy that may need feedback from @fiver-watson.

Thanks!

lib/task/migrate/migrations/arMigration0160.class.php
mcantelon: LOL... good catch. ;D

lib/task/migrate/migrations/arMigration0160.class.php
mcantelon: Sounds good.

mcantelon: Running `sfContext::getInstance()->getViewCacheManager()` from a CLI script returns null so seems like there's no view cache manager available when running from CLI.

One thing we might want to do though anyway, after running the migrate task, is clear *all* the caches which I think can be done with this code (which I think does the same thing as `php symfony cc`):

    $cacheClear = new sfCacheClearTask(sfContext::getInstance()->getEventDispatcher(), new sfFormatter());
    $cacheClear->run();

What do you think's the best way to go? If we want to add clearing of all caches to the migrate task then I'll make sure it'll work and will add that, but will make it a separate issue, just to keep the commits granular.

jraddaoui: I have vague recollection about `php symfony cc` not clearing the menus cache, that's why I asked. But you're right, if it's not as easy as using that piece of code, we can revisit later if needed.

Thanks for looking into this @mcantelon!

jraddaoui (approved):

lib/task/migrate/migrations/arMigration0160.class.php
mcantelon: Thanks!

PR #674, opened 2018-03-08
Deaccession CLI import task, refs #12020

sbreker:
Add new CLI task to import Deaccession records.

Default behaviour:
- deaccession row much match an existing accession number to be
imported. If match not found, an error will be displayed and the row
will not be loaded.

- deaccession row will be checked to make sure that it does not exactly
match an existing row in the database. If it does, an error will be
displayed and the row will not be loaded.

- dupe checking can be overridden so that the entire csv will be loaded
without checking for dupes using the --ignore-duplicates option.

### DIFF
--- a/lib/task/import/csvDeaccessionImportTask.class.php
+++ b/lib/task/import/csvDeaccessionImportTask.class.php
@@ -1,0 +1,201 @@
+<?php
+
+/*
+ * This file is part of the Access to Memory (AtoM) software.
+ *
+ * Access to Memory (AtoM) is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * Access to Memory (AtoM) is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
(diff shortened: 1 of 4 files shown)


lib/task/import/csvDeaccessionImportTask.class.php
```diff
@@ -0,0 +1,203 @@
...
+        'scope',
+        'accessionNumber',
+      ),
+
+      // Import logic to load deaccession
+      'rowInitLogic' => function(&$self)
+      {
+        $accessionIdentifier =  $self->rowStatusVars['accessionNumber'];
+
+        $ignoreDuplicates = ($self->status['options']['ignore-duplicates']) ? true : false;
```
mcantelon: Is this row necessary?

lib/task/import/csvDeaccessionImportTask.class.php
```diff
@@ -0,0 +1,203 @@
...
+ *
+ * @package    symfony
+ * @subpackage task
+ * @author     Steve Breker <sbreker@artefactual.com>
+ */
+class csvDeaccessionImportTask extends csvImportBaseTask
+{
+  protected $namespace           = 'csv';
+  protected $name                = 'deaccession-import';
+  protected $briefDescription    = 'Import csv deaccession data';
```
mcantelon: Ultra-nitpick with CSV capitalization. ;)

mcantelon:
Other than a couple of super minor nitpicks, looks good!

sbreker:
Suggested changes made.

mcantelon:
👍

PR #675, opened 2018-03-08
Contact info form translation fix, refs #11602

mcantelon:
Fixed inability to translate labels for a number of fields and text for
several buttons.

### DIFF
--- a/apps/qubit/modules/contactinformation/templates/_edit.php
+++ b/apps/qubit/modules/contactinformation/templates/_edit.php
@@ -56,6 +56,9 @@ $rowTemplate = json_encode(<<<value
 value
 );
 
+$submitText = __('Submit');
+$cancelText = __('Cancel');
+$addNewText = __('Add new');
 echo javascript_tag(<<<content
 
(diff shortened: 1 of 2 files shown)


jraddaoui (approved):
Looks great @mcantelon!! I wouldn't add the 'message.xml' files changes, though. They will be overwritten when we pull from Transifex (or whatever system we may end up using), so no big deal.

jraddaoui:
Actually, don't we need to add labels to the fields without them? E.g.:
https://github.com/artefactual/atom/blob/ebca194d7cc3064f65133b2ef82b03386484beee/apps/qubit/modules/contactinformation/templates/_edit.php#L107

Like it's being done in:
https://github.com/artefactual/atom/blob/ebca194d7cc3064f65133b2ef82b03386484beee/apps/qubit/modules/contactinformation/templates/_edit.php#L127-L129

mcantelon:
Ah, I think you're right. Fixed.

jraddaoui:
Nice, thanks! Looks good!

PR #677, opened 2018-03-12
Fix accession counter, refs #11700

mcantelon:
The accession counter, used to create accession identifiers, wasn't being
incremented.

Fixed this by:
* Changing when accession counter is incremented
* Adding an availability check, to avoid identifiers already in use, when
  determining the next accession identifier to use
* Minor refactor to share mask generation code between accessions and
  information objects

### DIFF
--- a/lib/Qubit.class.php
+++ b/lib/Qubit.class.php
@@ -366,4 +366,32 @@ class Qubit
       }
     }
   }
+
+  /**
+   * Generate an identifier using a counter value and a mask
+   *
+   * @param int $counter  current counter value
+   * @param string $mask  mask
+   *
+   * @return string
+   */
+  public static function generateIdentifierFromCounterAndMask($counter, $mask)
+  {
+    return preg_replace_callback('/([#%])([A-z]+)/', function($match) use ($counter)
+    {
+      if ('%' == $match[1])
+      {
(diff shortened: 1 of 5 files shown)


jraddaoui (commented):
Hi @mcantelon,

This is looking great but we should put some thinking on this transaction(s).

After this changes, we're getting the value in  `nextAccessionNumber` but incrementing it in `incrementAccessionCounter` (where the transaction happens). Which may lead to race conditions in simultaneous saves.

Moreover, we have a transaction filter that acts on each request and I'm not completely sure how MySQL and Propel work with nested transactions. However we still need to have this transaction for saves that occur outside the request cycle, like in the accession CSV import, or move it to those places.

jraddaoui (changes requested):
It looks like propel ignores nested transactions:

http://propelorm.org/Propel/documentation/06-transactions.html#nested-transactions

So we should be good by moving this transaction to wrap all the operations in `nextAvailableIdentifier`.

mcantelon:
Thanks @jraddaoui ... I've gotten rid of the transaction in the increment function and wrapped the contents of nextAvailableIdentifier in a transaction.

jraddaoui (approved):
Awesome, thanks @mcantelon!

PR #678, opened 2018-03-14
Permission-related fix to menu items, refs #11075

mcantelon:
Fixed issue with unnecessarily re-assigning action's resource after it has been properly assigned.


jraddaoui (approved):
I'd change the commit message to better reflect this changes (keeping the reference).

PR #679, opened 2018-03-14
Fix accession counter, refs #11700

mcantelon:
Added missing commit of transaction.


jraddaoui (approved):
Like in #678, I know it's related to the previous fix, but I'd use a different commit message to better reflect what this change is doing (maintaining the issue reference).

PR #680, opened 2018-03-16
Deaccessions CLI import enhancement, Refs #12020

sbreker:
This commit modified the deaccession import CLI task to separate out
the identifier field. In the import CSV, there are now two columns:
- accessionNumber: used to match to accession
- deaccessionNumber: used to populate the identifier field in the DB.
  Shows up in the "Deaccession Number" field in the WebUI.


mcantelon:
👍

PR #681, opened 2018-03-19
Accession counter incrementing fix, refs #11700

mcantelon:
Added helper function to check whether mask's enabled and fixed inconsistent
accession counter incrementing.

Changes to logic relating to accession identifiers during database insert:
* Removed automatic population of identifier, via the mask, if no accession
  identifier is specified
* Added logic to, if an identifier is specified and mask is enabled, increment
  the accession counter

### DIFF
--- a/plugins/qtAccessionPlugin/lib/model/QubitAccession.php
+++ b/plugins/qtAccessionPlugin/lib/model/QubitAccession.php
@@ -26,10 +26,21 @@ class QubitAccession extends BaseAccession
 
   protected function insert($connection = null)
   {
-    // If no identifier has been specified, use next available one
-    if (!$this->identifier)
+    // If identifier has been specified and the mask is enabled, increment the counter
+    if (!empty($this->identifier) && self::maskEnabled())
     {
-      $this->identifier = self::nextAvailableIdentifier();
+      $con = Propel::getConnection();
+      try
+      {
+        $con->beginTransaction();
+        self::incrementAccessionCounter();
+        $con->commit();
+      }
+      catch (PropelException $e)
+      {
+        $con->rollback();
+        throw $e;
+      }
     }
 
     if (!isset($this->slug))
@@ -79,6 +90,12 @@ class QubitAccession extends BaseAccession
     $criteria->add(QubitRelation::SUBJECT_ID, $this->id);
 
     return 0 < count(QubitRelation::get($criteria));
+  }
+
+  public static function maskEnabled()
+  {
+    $setting = QubitSetting::getByName('accession_mask_enabled');
+    return (null === $setting || boolval($setting->getValue(array('sourceCulture' => true))));
   }
 
   public static function nextAccessionNumber()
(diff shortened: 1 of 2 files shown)


plugins/qtAccessionPlugin/lib/model/QubitAccession.php
```diff
@@ -26,10 +26,10 @@ public function __toString()
 
   protected function insert($connection = null)
   {
-    // If no identifier has been specified, use next available one
-    if (!$this->identifier)
+    // If identifier has been specified and the mask is enabled, increment the counter
+    if (!empty($this->identifier) && QubitAccession::maskEnabled())
```
jraddaoui: We could use `self` instead of `QubitAccession` like inside the `if`.

plugins/qtAccessionPlugin/lib/model/QubitAccession.php
```diff
@@ -26,10 +26,10 @@ public function __toString()
...
     {
-      $this->identifier = self::nextAvailableIdentifier();
+      self::incrementAccessionCounter();
```
jraddaoui: We may need to wrap the `incrementAccessionCounter` function content in a transaction too as it may be called from CLI in here after this changes, and the transaction in `nextAvailableIdentifier` won't be working in that case.

plugins/qtAccessionPlugin/lib/model/QubitAccession.php
```diff
@@ -81,6 +81,12 @@ public function isAccrual()
     return 0 < count(QubitRelation::get($criteria));
   }
 
+  public static function maskEnabled()
+  {
+    $setting = QubitSetting::getByName('accession_mask_enabled');
+    return (null === $setting || boolval($setting->getValue(array('sourceCulture' => true))));
```
jraddaoui, line 98: This feels wrong to me, are we sure that if the setting is missing (a rare case) we should consider that the mask is enabled? Otherwise, we could use sfConfig to get its value.

jraddaoui (changes requested):
Thanks @mcantelon! Some comments bellow.

plugins/qtAccessionPlugin/lib/model/QubitAccession.php
mcantelon: Ah, true! I'll fix.

plugins/qtAccessionPlugin/lib/model/QubitAccession.php
mcantelon: This setting doesn't get created by a migration, alas, and the default behaviour IIRC is to use the mask.

jraddaoui: Ah, nice! Thanks

mcantelon:
@jraddaoui I've wrapped the increment in a transaction and tested.

jraddaoui (approved):
Thanks @mcantelon!

PR #682, opened 2018-03-19
Accession import identifier handling, refs #12043

mcantelon:
Changed CSV accession import so if an accession number isn't specified the row
is skipped, unless the assign-id option is used when running the import task.

Fixed issue with generating identifiers based on masks/counters from the CLI.

### DIFF
--- a/lib/task/import/csvAccessionImportTask.class.php
+++ b/lib/task/import/csvAccessionImportTask.class.php
@@ -30,7 +30,7 @@ class csvAccessionImportTask extends csvImportBaseTask
   protected $name                = 'accession-import';
   protected $briefDescription    = 'Import csv acession data';
   protected $detailedDescription = <<<EOF
-Import CSV data
+Import CSV accession data
 EOF;
 
   /**
@@ -51,6 +51,12 @@ EOF;
         null,
         sfCommandOption::PARAMETER_NONE,
         "Index for search during import."
+      ),
+      new sfCommandOption(
+        'assign-id',
+        null,
+        sfCommandOption::PARAMETER_NONE,
+        "Assign identifier, based on mask and counter, if no accession number specified in row."
       )
     ));
   }
@@ -60,6 +66,8 @@ EOF;
    */
   public function execute($arguments = array(), $options = array())
   {
+    parent::execute($arguments, $options);
+
     $this->validateOptions($options);
 
     $skipRows = ($options['skip-rows']) ? $options['skip-rows'] : 0;
@@ -102,7 +110,8 @@ EOF;
         'acquisitionTypes'   => $termData['acquisitionTypes'],
         'resourceTypes'      => $termData['resourceTypes'],
         'processingStatus'   => $termData['processingStatus'],
-        'processingPriority' => $termData['processingPriority']
+        'processingPriority' => $termData['processingPriority'],
+        'assignId'           => $options['assign-id']
       ),
 
       'standardColumns' => array(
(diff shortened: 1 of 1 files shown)


lib/task/import/csvAccessionImportTask.class.php
```diff
@@ -102,7 +108,8 @@ public function execute($arguments = array(), $options = array())
         'acquisitionTypes'   => $termData['acquisitionTypes'],
         'resourceTypes'      => $termData['resourceTypes'],
         'processingStatus'   => $termData['processingStatus'],
-        'processingPriority' => $termData['processingPriority']
+        'processingPriority' => $termData['processingPriority'],
+        'assign-id'          => $options['assign-id']
```
jraddaoui: After getting the value from the options, we should camelCase the var like the others are.

lib/task/import/csvAccessionImportTask.class.php
```diff
@@ -192,23 +199,33 @@ public function execute($arguments = array(), $options = array())
         $result = $statement->fetch(PDO::FETCH_OBJ);
         if ($result)
         {
-          print 'Found '. $result->id ."\n";
+          print $self->logError(sprintf('Found accession ID %d', $result->id));
```
jraddaoui, line 204: Nice.

lib/task/import/csvAccessionImportTask.class.php
```diff
@@ -192,23 +199,33 @@ public function execute($arguments = array(), $options = array())
...
           $self->object = QubitAccession::getById($result->id);
         }
-        else
+        else if (!empty($accessionNumber))
         {
-          $self->object = false;
```
jraddaoui, line 200: hehe

plugins/qtAccessionPlugin/lib/model/QubitAccession.php
```diff
@@ -103,7 +103,9 @@ public static function nextAvailableIdentifier()
       $con->beginTransaction();
 
       // Determine what should be the next identifier
-      $identifier = Qubit::generateIdentifierFromCounterAndMask(self::nextAccessionNumber(), sfConfig::get('app_accession_mask'));
+      $setting = QubitSetting::getByName('accession_mask');
+      $mask = $setting->getValue(array('sourceCulture' => true));
+      $identifier = Qubit::generateIdentifierFromCounterAndMask(self::nextAccessionNumber(), $mask);
```
jraddaoui: Why do you need to do this? `sfConfig` should be populated with the settings in CLI tasks if they extend `arBaseTask` and I believe `csvAccessionImportTask` inherits from it through `csvImportBaseTask`.

jraddaoui (commented):
Looks great @mcantelon. A minor nitpick and a question.

plugins/qtAccessionPlugin/lib/model/QubitAccession.php
jraddaoui: Oh, I see. We're not calling the `parent::execute` from `csvAccessionImportTask`, like we do in:

https://github.com/artefactual/atom/blob/6ad917404acd4f2a8c53f90c034b99d4fb2312e1/lib/task/import/csvImportTask.class.php#L138

If you could add it and keep using sfConfig in here to maintain consistency, that would be great.

jraddaoui (changes requested):
A minor change requested in the end ;)

plugins/qtAccessionPlugin/lib/model/QubitAccession.php
mcantelon: Ah! Good to know! I'll change that.

lib/task/import/csvAccessionImportTask.class.php
mcantelon: Good call.

jraddaoui (approved):
Great, thanks @mcantelon!

PR #684, opened 2018-03-21
Update async. groups of IOs in ES, refs #11855

jraddaoui:
Creates a job to update information object documents in the Elasticsearch
index. This job is launched in places where multiples IOs may be updated,
like updating the descendants of an IO and the related IOs of terms, actors
and repositories. The new job, called `arUpdateEsIoDocumentsJob` will not
be triggered in CLI tasks or other jobs, where the update will be done sync.

Main changes:
- Creates `arUpdateEsIoDocumentsJob`, capable of updating a group of IOs
  and their descendants, skipping ones or the others if requested.
- Update descendants when a higher level IO changes.
- Update related descriptions when an actor changes, including their
  descendants if the relating event is a creation event.
- Update related descriptions and descendants on repository changes.
- Update related descriptions on term label changes of subjects, places and
  genres.

For now, descendants are being updated entirely, but we should use partial
updates in the future to only update the inherited data.

### DIFF
--- a/apps/qubit/modules/term/actions/editAction.class.php
+++ b/apps/qubit/modules/term/actions/editAction.class.php
@@ -577,14 +577,38 @@ class TermEditAction extends DefaultEditAction
 
   protected function updateLinkedInfoObjects()
   {
+    // Only update related IOs of terms that are fully added to the IOs in ES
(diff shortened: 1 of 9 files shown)


sbreker:
Looks great @jraddaoui!  This will be a great feature to have!

jraddaoui:
Thanks @sbreker!!

PR #685, opened 2018-03-22
Fix inherited creators on ES IO save, refs #11788

jraddaoui:
Inherited creators were only being considered if sent as part of the
options, which is the case in the entire collection population but
it isn't in individual updates.


jraddaoui:
I'll cherry-pick to stable/2.4.x when reviewed.

jraddaoui:
Needed in #684

sbreker (approved):

sbreker:
Looks good!

jraddaoui:
5f19171 in stable/2.4.x

PR #686, opened 2018-03-23
Fix granted rights deletion error, refs #11653

mcantelon:
Moved logic processing edit/deletion of granted rights, relating to a right,
until after the right is saved so the right doesn't attempt to, when saved,
save related granted rights that were deleted beforehand.


sbreker (approved):
Looks great @mcantelon!  👍

PR #687, opened 2018-03-23
EAD import emph tag to Parsedown, refs# 12032

sbreker:
Convert <emph> tags to Parsedown on import. Ignore all other EAD
markdown on import.

- Modified QubitXmlImport to take a new 'Filter' param defined in the
object schema files.
- Added a new QubitMarkdown.class.php to contain markdown specific
functionality.

### DIFF
--- a/apps/qubit/modules/object/config/import/ead.yml
+++ b/apps/qubit/modules/object/config/import/ead.yml
@@ -45,10 +45,13 @@ information_object:
       Parameters: ["array('typeId' => QubitTerm::STATUS_TYPE_PUBLICATION_ID, 'statusId' => sfConfig::get('app_defaultPubStatus', QubitTerm::PUBLICATION_STATUS_DRAFT_ID))"]
 
     # each of the following XPath expressions are relative to the current matched node:
-
     title:
       XPath: "did/unittitle[not(@type)]"
       Method:  setTitle
+      Filters:
+        -
+          emph:
+            QubitMarkdown: eadTagToMarkdown
 
     level:
       XPath:  "@level"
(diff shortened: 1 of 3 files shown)


apps/qubit/modules/object/config/import/ead.yml
```diff
@@ -49,6 +49,7 @@ information_object:
     title:
       XPath: "did/unittitle[not(@type)]"
       Method:  setTitle
+      Filters:  ["array('emph' => array('QubitMarkdown' => 'eadTagToMarkdown'))"]
```
mcantelon: What do you think about avoiding the need for `eval` for filters through use of something like the below markup? We use `eval` for some other cases given we need to be able to evaluation some PHP constants, but we probably won't need to with filters specifications.

    Filters:
      -
        emph:
          QubitMarkdown: eadTagToMarkdown

lib/QubitXmlImport.class.php
```diff
@@ -809,31 +815,6 @@ protected function loadXML($xmlFile, $options = array())
       $doc->xpath->registerNamespace($pre, $uri);
     }
 
-/*
```
mcantelon, line 812: Nice. :D

lib/QubitMarkdown.class.php
```diff
@@ -0,0 +1,52 @@
...
+   * @return ead string with markdown replacement
+   */
+  public static function eadTagToMarkdown($eadTag, $node)
+  {
+    switch ($eadTag)
+    {
+      // EAD tags that we want to convert to markdown.
+      case 'emph':
+        // Set default to italic when no render attribute is present.
+        $markdownSymbol = (!empty($node->getAttribute('render'))) ? QubitMarkdown::$markdownMap[$node->getAttribute('render')] : QubitMarkdown::$markdownMap['italic'];
```
mcantelon: Make we could make this multi-line, given it's a long one.

mcantelon (commented):
A couple of ideas/nitpicks, but seems solid!

mcantelon:
👍

PR #688, opened 2018-04-01
Avoid multi-update of IOs in ES, refs #11855

jraddaoui:
Do not update the related IO in ES when events are saved or deleted
from and IO edit page as it will be updated by itself.


mcantelon:
👍

PR #689, opened 2018-04-01
Fix IO inherited creators population, refs #11788

jraddaoui:
Query only for events with an `actor_id` as it may be only dates
creation events.


sbreker (approved):
👍

jraddaoui:
2f7a600 in stable/2.4.x

PR #690, opened 2018-04-03
Added create date updating ability, refs #11820

mcantelon:
* Added ability to update actor history during CSV import
* Added ability to update event during CSV import:
  * Update if CSV date fields are empty
  * Update if CSV date fields are different than stored values

### DIFF
--- a/lib/task/import/csvImportBaseTask.class.php
+++ b/lib/task/import/csvImportBaseTask.class.php
@@ -238,6 +238,35 @@ abstract class csvImportBaseTask extends arBaseTask
       // Add row culture to fetch place term in event creation/update
       $eventData['culture'] = $import->columnValue('culture');
 
+      // If in update mode and CSV event data references an existing actor,
+      // see if the event seems like it should be updated rather than created.
+      if ($import->matchAndUpdate && !empty($eventData['actorName'])
+          && null != $actor = QubitActor::getByAuthorizedFormOfName($eventData['actorName']))
+      {
+        // Check events to determine whether event should be updated rather than created
+        $criteria = new Criteria;
+        $criteria->add(QubitEvent::TYPE_ID, $eventTypeId);
+        $criteria->add(QubitEvent::OBJECT_ID, $import->object->id);
+        $criteria->add(QubitEvent::ACTOR_ID, $actor->id);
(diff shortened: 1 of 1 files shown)


lib/task/import/csvImportBaseTask.class.php
```diff
@@ -238,6 +238,37 @@ static function importEvents(&$import)
       // Add row culture to fetch place term in event creation/update
       $eventData['culture'] = $import->columnValue('culture');
 
+      // If in update mode and event's a creation event with an existing actor
+      // specified, see if the event seems like it should be updated
+      // rather than created.
+      if ($eventTypeId == QubitTerm::CREATION_ID && $import->matchAndUpdate
```
jraddaoui: I think we should look for all event types, not just creation events. I'd remove that condition and add the type to the query directly.

lib/task/import/csvImportBaseTask.class.php
```diff
@@ -238,6 +238,37 @@ static function importEvents(&$import)
...
+          && !empty($eventData['actorName'])
+          && null != $actor = QubitActor::getByAuthorizedFormOfName($eventData['actorName']))
+      {
+        // Check creation events to determine whether event should be updated rather than created
+        $criteria = new Criteria;
+        $criteria->add(QubitEvent::TYPE_ID, QubitTerm::CREATION_ID);
```
jraddaoui: I'd compare with `$eventTypeId` to allow this behavior for all event types.

jraddaoui (changes requested):
Hi @mcantelon,

This is looking great! I feel like we should do it for all event types, I couldn't see anything about the event type in the Redmine ticket.

Also, that ticket is a private one, we should create a public one and change it in the commit message that will be merged.

Thanks!

jraddaoui:
I've created https://projects.artefactual.com/issues/12102

mcantelon:
Good point re: event type and I'll change the issue reference in the commit to the public issue #... thanks!

mcantelon:
I've now removed the logic that prevents it from working with non-creation events.

jraddaoui (approved):
Thanks @mcantelon!

PR #691, opened 2018-04-04
Fix CSV updating of script/language, refs #11990

mcantelon:
Changed import logic so script and language properties, in which values are
stored in properties as serialized data, can be updated as well as created
anew if they don't exist. Without this fix, the properties were being
duplicated during CSV updating which caused issues (after CSV update no values
would show at all).

Also tweaked logic to remove any duplicates from values.


sbreker (approved):
Looks good Mike!  👍

PR #693, opened 2018-04-06
Remove unneeded Docker volumes, refs #12111

jraddaoui:
The 'atom_uploads' and 'atom_downloads' volumes are placed inside another
volume used to share the entire AtoM directory. This causes data loss when
the container is recreated. The entire AtoM volume is shared by the 'atom',
'nginx' and 'atom_worker' services with the same permissions than this two
volumes, so they are actually not needed.

### DIFF
--- a/docker/docker-compose.dev.yml
+++ b/docker/docker-compose.dev.yml
@@ -9,12 +9,6 @@ volumes:
   percona_data:
     driver: "local"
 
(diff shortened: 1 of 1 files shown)


sevein (approved):

jraddaoui:
af106aa in stable/2.4.x

PR #694, opened 2018-04-06
Permissive slug generation, refs #11761

sbreker:
Ensure the characters * : @ = , are represented as literals in the
browser address bar when the 'Permissive Slugs' setting is activated.

- this will prevent these chars from being percent-encoded when
  'permissive slugs' setting is turned on in AtoM
- these chars are allowed in slugs when 'permissive slugs' setting is on.
- if 'permissive slugs' is OFF, use PHP's urlencode as normal.
- this should only affect the PATH segment of an AtoM URI.
- this will affect URI construction for page links and address bar.

### DIFF
--- a/lib/routing/QubitRoute.class.php
+++ b/lib/routing/QubitRoute.class.php
@@ -52,6 +52,25 @@ class QubitRoute extends sfRoute
   }
 
   /**
+  * Used to correctly represent the characters * ; : @ = , in the slug part
+  * of a URI - this is the path segment.
+  *  - these chars are allowed in slugs when 'permissive slugs' setting is on.
+  *  - if 'permissive slugs' is OFF, use urlencode as normal.
+  *  - this should ONLY affect the slug portion of an AtoM URI
+  */
+  public static function urlencode3986($string)
+  {
+    if (QubitSlug::SLUG_PERMISSIVE == sfConfig::get('app_permissive_slug_creation', QubitSlug::SLUG_RESTRICTIVE))
+    {
+      $entities = array('%2A', '%3A', '%40', '%3D', '%2C');
+      $replacements = array('*', ':', '@', '=', ',');
+      return str_replace($entities, $replacements, urlencode($string));
+    }
+
+    return urlencode($string);
+  }
+
+  /**
    * @see sfRoute
    */
   public function matchesParameters($params, $context = array())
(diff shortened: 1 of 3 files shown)


vendor/symfony/lib/routing/sfRoute.class.php
```diff
@@ -283,7 +283,7 @@ protected function generateWithTokens($parameters)
         case 'variable':
           if (!$optional || !isset($this->defaults[$token[3]]) || $parameters[$token[3]] != $this->defaults[$token[3]])
           {
-            $url[] = urlencode($parameters[$token[3]]);
+            $url[] = QubitRoute::urlencode3986($parameters[$token[3]]);
```
jraddaoui: We should try to not make changes in the Symfony classes when possible. I'd overwrite the entire function in the QubitRoute class. I'm not sure if all the other QubitRoute classes extend QubitRoute or sfRoute, so you may need to do the same in others or add an intermediate class.

lib/routing/QubitRoute.class.php
```diff
@@ -51,6 +51,25 @@ protected function filterParams($params)
     return $params;
   }
 
+  /**
+  * Used to correctly represent the characters * ; : @ = , in the slug part
+  * of a URI - this is the path segment.
+  *  - these chars are allowed in slugs when 'permissive slugs' setting is on.
+  *  - if 'permissive slugs' is OFF, use urlencode as normal.
+  *  - this should ONLY affect the slug portion of an AtoM URI
```
jraddaoui: Are all those hyphens intended? I'd prefer a normal paragraph ;)

jraddaoui (changes requested):
Looks great @sbreker!

But I'd try to avoid making changes in the Symfony classes if it's possible.

lib/routing/QubitRoute.class.php
```diff
@@ -85,4 +85,58 @@ public function generate($params, $context = array(), $absolute = false)
...
+          $optional = false;
+          break;
+        case 'separator':
+          if (false === $optional || $first)
+          {
+            $url[] = $token[2];
+          }
+          break;
+        default:
+          // handle custom tokens
```
jraddaoui, line 122: Uppercase.

jraddaoui (approved):
Looks great @sbreker! Just a really minor comment.

PR #695, opened 2018-04-16
Add identifier-related settings page, refs #12139

mcantelon:
Removed identifier-related settings from the global settings page and added
them to a new page.


jraddaoui (approved):
Looks good @mcantelon!

We usually use `$this->$name` instead of `$this->{$name}` but it's not a big deal for me.

mcantelon:
Awesome... I'll change the `$this->{$name}` (quick change so might as well). Thanks Radda!

PR #696, opened 2018-04-17
Digital object lookup enhancement, refs #12066

sbreker:
Adding two new indexes to reduce the number of records scanned.

- New index on column information_object.rgt
Range queries using both LFT and RGT were scanning the whole information
object table to get the RGT value for the comparison.

- New index on column digital_object.path
Queries on digital object path and name were performing whole table
scans. Adding an index on path will narrow the selection to a specific
image and it's derivatives.

### DIFF
--- a/config/schema.yml
+++ b/config/schema.yml
@@ -101,6 +101,8 @@ propel:
     checksum: varchar(255)
     checksum_type: type: varchar(50)
     parent_id: { type: integer, foreignTable: digital_object, foreignReference: id }
+    _indexes:
+      path: [path]
 
   event:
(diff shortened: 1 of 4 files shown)


jraddaoui (approved):
Looks great!

Although I'm not completely sure if this requires to rebuild the models.

sbreker:
Hi Radda - Thanks!

If lib.model.schema.sql is not rebuilt, dev instances are affected negatively: if a dev pulls this commit into their workspace and does a tools:purge, it will set their database to version 161, and will rebuild the database from the current lib.model.schema.sql.  If this commit did not rebuild lib.model.schema.sql, tools:purge will rebuild using an OLD version of the sql, even though their DB says it is version 161.

sbreker:
Ah I see you are referring to propel:build-model.  My mistake.

PR #697, opened 2018-04-18
Physical object EAD culture fallback, refs #12053

mcantelon:
Added culture fallback for physical object-related elements in EAD export.


jraddaoui (approved):
:+1:

PR #698, opened 2018-04-19
Add Markdown support, refs #12148

jraddaoui:
Add a new setting to enable/disable Markdown support. When it's enabled
Markdown content will be parsed and stripped where needed using the
Parsedown and ParsedownExtra libraries, included in this changes. Added
this setting in a new page to add and make clear a notice about the need of
rebuilding the search index after this setting changes.

The Markdown support conflicts with the Symfony escaping strategy so, when
Markdown is enabled the escaping strategy is disabled and we rely in the safe
mode of Parsedown to escape all special chars after parsing the content. The
main consequence of this changes is that now the data can be found scaped
or not in the templates and it needs to be accessed with methods that work in
both sfOutputEscaperObjectDecorator/sfOutputEscaperArrayDecorator and 
objects/arrays.

QubitMarkdown has been converted to an instantiable class for Markdown
operations using Parsedown, it contains parse and strip methods that are being
called from the render helpers in QubitHelper. It is important to use the following
helpers while displaying data that may contain Markdown syntax:

- render_value: uses Parsedown text method in safe mode (adds paragraphs).
- render_value_inline: uses Parsedown line method in safe mode.
- render_value_html: uses Parsedown text method in unsafe mode (for HTML).
- render_title: returns 'Untitled' if render_value_inline returns and empty string. 
- strip_markdown: removes Markdown and HTML tags.

To strip Markdown in the response title meta, QubitWebResponse has been
created extending sfWebResponse and being used in factories.yml.

The new setting will also determine the behavior on the EAD import, converting
`emph` tags to Markdown when it's enabled or just removing them when it's not.

In the Elasticsearch index, the data is stored with the Markdown tags but, to
avoid issues while searching over content like '__test value__', a new char filter
is being added to all analyzers (only if the new setting is enabled). This char
filter replaces the following punctuation chars by spaces: *_#->`+\~:|^=

For the same reason, autocomplete queries executed over the ORM now use two
wildcards ('%query%') instead of only one ('query%'), only when the setting
is enabled.

Parsed Markdown creates HTML and it can break the page look if it's truncated.
To avoid that, the IOs scope and content field in the search results and the links
in the card view of repos and IOs (when no DO/logo exists) are no longer being
truncated. For that reason, the links use now a level five header instead of four
and the scope and content field uses now the same JS expandable method used
in the index pages. Other truncated text added to HTML attributes is passed through
the strip_markdown helper to remove Markdown and HTML tags before it's truncated.

This changes also include:

- Styling fixes to avoid global rules to apply to parsed Markdown.
- Removal of the duplication of flash notices in some pages.
- Order of the setting menu alphabetically.
- sfConfig settings reload on job execution.
- Clear settings cache after upgrade.

### DIFF
--- a/plugins/arDominionPlugin/css/less/scaffolding.less
+++ b/plugins/arDominionPlugin/css/less/scaffolding.less
@@ -767,7 +767,12 @@ div#content > table.table {
   .box-shadow(0 1px 1px rgba(0, 0, 0, .2));
 
   // Section title
-  h2 {
+  .section > h2,
+  .section > a > h2,
+  .section > span > h2,
+  section > h2,
+  section > a > h2,
+  section > span > h2 {
     font-size: @baseFontSize * 1.10;
     line-height: 38px;
     color: @orange;
@@ -1173,7 +1178,7 @@ section .vcard:last-child > .field:last-child > div {
   .creation-details {
     color: @grayLight;
     font-size: @baseFontSize * 0.90;
-    margin-top: -10px !important;
+    margin-top: 10px;
   }
 
   // List of details
@@ -1827,4 +1832,10 @@ body.login #content {
   &+ .multiline-header {
     margin-top: 45px;
   }
+}
+
+// Maintain font size in blockquotes
+
+blockquote p {
+  font-size: inherit;
 }
--- a/COPYRIGHT
+++ b/COPYRIGHT
@@ -218,3 +218,15 @@ elasticsearch-php
 Url: https://github.com/elastic/elasticsearch-php
 Copyright: 2013-2016 Elasticsearch
 License: Apache 2.0 and LGPL v2.1
+
+Parsedown
+---------
+Url: https://github.com/erusev/parsedown
+Copyright: Emanuil Rusev, erusev.com
+License: MIT
+
+Parsedown Extra
+---------------
+Url: https://github.com/erusev/parsedown-extra
+Copyright: Emanuil Rusev, erusev.com
+License: MIT
--- a/apps/qubit/config/autoload.yml
+++ b/apps/qubit/config/autoload.yml
@@ -16,3 +16,6 @@ autoload:
   EasyRdf:
     path: %SF_ROOT_DIR%/vendor/easyrdf/EasyRdf
     recursive: on
+
+  ParseDown:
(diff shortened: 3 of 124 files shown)


jraddaoui:
Hi @sbreker,

I think it's better if you're ignore the individual commits and just read the description and jump to the full changes, as there is quite some back and forward in there.

I have also left the `qubit_auto_link_text` and related stuff in `QubitHelper` as I'll need them for https://projects.artefactual.com/issues/12149, so I'll remove them in the following PR for that feature.

Thanks!

sbreker:
One additional change requested to function in QubitMarkdown::eadTagToMarkdown:

If parsedown setting ON, do what it's doing presently; if parsedown setting OFF, leave EAD tags in place unmodified.

jraddaoui:
Thanks @sbreker,

I have cleaned-up the commit history and sent this for a first round of testing. I'll check your request with Dan and add it alongside his feedback.

jraddaoui:
Hi @sbreker, ready for another review!

The part you already reviewed is squashed in the first commit and all the others are addressing the feedback. This time they should make sense so you could review them one by one. I have also updated the description to reflect the new changes.

Thanks!

plugins/arDominionPlugin/css/less/scaffolding.less
```diff
@@ -767,10 +767,12 @@ div#content > table.table {
...
   // Section title
-  .section > span > h2,
+  .section > h2,
   .section > a > h2,
-  section > span > h2,
-  section > a > h2 {
+  .section > span > h2,
+  section > h2,
+  section > a > h2,
+  section > span > h2 {
```
sbreker: We will need to remember these changes for future client theme/repo upgrades.

sbreker:
We will need to remember the css changes for future client theme/repo upgrades.

These additional changes look great!

sbreker (approved):
Approved

jraddaoui:
Good point, I have noted it in https://internal.artefactual.com/wiki/Development/AtoM/Hosting_Clients_Upgrades_Reminders#Upcoming_upgrades.

Thanks @sbreker!

sbreker:
Looks good @jraddaoui !

👍

PR #699, opened 2018-04-24
Permissive slug uri display, refs #11761

sbreker:
The previous commit to ensure the characters * : @ = , are represented
as literals in the browser address bar when 'Permissive Slugs' setting
is activated had an issue where the class QubitMetadataRoute does need
to extend QubitRoute. This commit corrects this.


jraddaoui (approved):
Asked for a minor change in the comments, but this looks great.

Thanks @sbreker!

PR #700, opened 2018-04-25
New EAD import field mappings, refs #12172

sbreker:
Add new mappings of EAD fields to Archivist's Notes.

EAD import nodes being added to ead.yml (map to Archivist's Notes):
- filedesc/titlestmt/author[@encodinganalog='245$c']
- filedesc/publicationstmt/p


jraddaoui (commented):
Hi @sbreker, this looks good!

But I wonder if, when the three elements exists in a document or when one of them is present multiples times, would it create multiples notes or only one?

sbreker:
Hi @jraddaoui 

Yes, if all three elements exist in a document, this will create multiple Archivist Notes.  I have discussed this consideration with Dan already and he is ok with this.

jraddaoui (approved):
Awesome, thanks @sbreker!

PR #702, opened 2018-05-01
Fix treeview data sorting issue, refs #12179

mcantelon:
Prevent in-memory sorting of treeview data from being done on everything but
the hierarchal browse page. Improve the sorting on the hierarchal browse page
(use natural sort).

### DIFF
--- a/apps/qubit/modules/default/actions/fullTreeViewAction.class.php
+++ b/apps/qubit/modules/default/actions/fullTreeViewAction.class.php
@@ -130,7 +130,6 @@ class DefaultFullTreeViewAction extends sfAction
           // Append result identifier to the base reference code if we're not loading the collection root
           //if ($result['parent'] != QubitInformationObject::ROOT_ID && !empty($result['identifier']))
 
-// TODO what if a child identifer doesn't exist? How does previous version handle this?
           // Append result identifier, if it exists, to the reference passed to the function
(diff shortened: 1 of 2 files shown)


apps/qubit/modules/default/actions/fullTreeViewAction.class.php
```diff
@@ -237,15 +236,15 @@ protected function getNodeOrChildrenNodes($id, $baseReferenceCode, $children = f
     }
 
     // If not ordered in SQL, order in-memory by "text" attribute of each node
-    if (!isset($options['orderColumn']))
+    if ($options['memorySort'])
```
jraddaoui: This throws a PHP warning when it's not set in the options array, I know it's annoying but we should check if it's set first to avoid it, `if (isset($options['memorySort']) && $options['memorySort'])`.

jraddaoui (changes requested):
Hi @mcantelon,

This looks good, just a minor suggestion to avoid a PHP warning. And I wonder, is the move option already disabled in the full collection treeview page?

mcantelon:
Thanks Radda... Drag-and-drop functionality is disabled in the hierarchy browser so it shouldn't AFAIK allow moving of nodes.

jraddaoui (approved):
Awesome, thanks @mcantelon!

PR #704, opened 2018-05-03
Add description change logging, refs #12130

mcantelon:
* Adds optional logging of creation or modification of descriptions
* Modifies description updates page to show this logging data when logging is
  enabled
* Adds option, for autheticated users, to view a particular description's
  creation/modification history
* Allows administrations to view a history of any user's creation/modification
  of descriptions (and allows users to view this data on their profile)

### DIFF
--- a/apps/qubit/modules/informationobject/actions/modificationsAction.class.php
+++ b/apps/qubit/modules/informationobject/actions/modificationsAction.class.php
@@ -1,0 +1,58 @@
+<?php
+
+/*
+ * This file is part of the Access to Memory (AtoM) software.
+ *
+ * Access to Memory (AtoM) is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * Access to Memory (AtoM) is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with Access to Memory (AtoM).  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+class InformationObjectModificationsAction extends sfAction
+{
+  public function execute($request)
+  {
+    $this->resource = $this->getRoute()->resource;
+    $this->abortInvalidRequests();
+
+    // Criteria to fetch user actions
+    $criteria = new Criteria;
+    $criteria->add(QubitAuditLog::OBJECT_ID, $this->resource->id);
+    $criteria->addDescendingOrderByColumn(QubitAuditLog::CREATED_AT);
+
+    // Page results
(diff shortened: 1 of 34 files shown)


apps/qubit/modules/informationobject/actions/modificationsAction.class.php
```diff
@@ -0,0 +1,58 @@
...
+  private function abortInvalidRequests()
+  {
+    // Check that this isn't the root
+    if (!isset($this->resource->parent))
+    {
+      $this->forward404();
+    }
+
+    // Check user authorization
+    if (!QubitAcl::check($this->resource, 'read') || !sfConfig::get('app_audit_log_enabled'))
```
sbreker: Is there any benefit to using sfConfig::get with the second param of default value?  e.g. 

if (sfConfig::get('some_setting', false))

apps/qubit/modules/search/templates/descriptionUpdatesSuccess.php
```diff
@@ -26,7 +26,52 @@
...
+          <td>
+            <?php echo get_component('object', 'clipboardButton', array('slug' => $io->slug, 'wide' => true)) ?>
+          </td>
+        </tr>
+      <?php endforeach; ?>
+      </tbody>
+    </table>
+
+  <?php else: ?>
+  <?php if (isset($pager) && $pager->getNbResults()): ?>
```
sbreker: Indentation formatting needs cleanup.

config/schema.yml
```diff
@@ -44,6 +44,16 @@ propel:
...
 
+  audit_log:
+    id: { type: integer, primaryKey: true, autoIncrement: true, inheritanceKey: false }
+    object_id: { type: integer, required: true, foreignTable: object, foreignReference: id, onDelete: cascade }
+    user_id: { type: integer, required: false, foreignTable: user, foreignReference: id, onDelete: setnull  }
+    user_name: varchar(255)
+    action_type_id: { type: integer, required: false, foreignTable: term, foreignReference: id, onDelete: setnull }
+    created_at: timestamp
+    _indexes:
+      1: [object_id]
```
sbreker: Are there searches on this table that may need an index other than by object_id?  Do we provide a search like "all changes by user"?  This table could get big.

sbreker:
That's a big change Mike!  Looks great!  I've added a couple questions above.

apps/qubit/modules/informationobject/actions/modificationsAction.class.php
mcantelon: Ah cool... there might be if the setting doesn't exist (maybe it generates a PHP notifice), so I'll put that in to be explicit.

config/schema.yml
mcantelon: This index, and an index for user ID, actually get auto-generated so I've removed this from the schema and updated the migration.

sbreker:
There are a few more of these in your code where the default is not supplied: sfConfig::get('app_audit_log_enabled')

It might make things more clear if the default was supplied in all cases especially if the setting is not yet created.  Not sure if it would result in an error if the setting were not present yet?

mcantelon:
Ah cool... I'll update those calls.

sbreker:
Thanks @mcantelon - Looks great!

sbreker (approved):
Approved

PR #705, opened 2018-05-10
Improve event update via CSV ability, refs #11820

mcantelon:
Added ability to update events that don't have actors.


jraddaoui (approved):
Good catch @mcantelon!

PR #706, opened 2018-05-11
Add GA repository dimension, refs #12209

jraddaoui:
- Add config setting in 'app.yml' for repository dimension index:

Added explanatory comments and un-commented the GA settings for a
better understanding.

- Upgrade to Universal Analytics:

Custom dimensions are not available in 'ga.js', which is already
deprecated. See: https://support.google.com/analytics/answer/3450662

- Add GA repository dimension component:

Called from the index pages of IOs, repos and actors, checks the new
config setting, gets the resource's related repository and sets it's
source culture authorized form of name as the value of the dimension
before sending the page view hit.

### DIFF
--- a/config/app.yml
+++ b/config/app.yml
@@ -20,8 +20,15 @@ all:
   # by and environment variable called "ATOM_READ_ONLY"
   read_only: false
 
-  # Google Analitycs
-  # google_analytics_api_key: ...
+  # Google Analitycs:
+  # Set a GA property API key to track page view hits.
+  # E.g.: UA-1234567-89
+  google_analytics_api_key:
+  # Set a GA custom dimension index numver (from the same property)
+  # to track the repository name as a dimension in page view hits
+  # of IOs, actors and repos index pages.
+  # E.g.: 1
+  google_analytics_institutions_dimension_index:
 
   # htmlpurifier is used in static pages, disabled by default
   htmlpurifier_enabled: false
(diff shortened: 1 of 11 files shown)


config/app.yml
```diff
@@ -22,6 +22,8 @@ all:
 
   # Google Analitycs
   # google_analytics_api_key: ...
+  # google_analytics_institutions_dimension_index: 1
```
sbreker: Perhaps a comment indicating what this setting is for:  e.g. a value of zero disables the institution dimension.

sbreker (approved):
This change looks great!  One comment below about adding a comment describing the config options and values in the .yml file.

jraddaoui:
Hi @sbreker,

I've improved a little the GA settings setup and access. GA custom dimensions are created with indexes from 1 to 200, so 0 is not being considered in our current setup. Please, let me know if you think that's an issue.

config/app.yml
```diff
@@ -20,10 +20,15 @@ all:
...
   read_only: false
 
-  # Google Analitycs
-  # google_analytics_api_key: ...
-  # google_analytics_institutions_dimension_index: 1
-
+  # Google Analitycs:
+  # Set a GA property API key to track page view hits.
+  # E.g.: UA-1234567-89
+  google_analytics_api_key: key
```
sbreker: The comments in the block are great!  One question: should these settings be commented out by default?  This placeholder value 'key' appears that it will be assigned to $gaKey and !empty == true thereby running the script in _footer.php with value 'key'?

jraddaoui:
Good catch @sbreker , those where testing values that I forgot to remove. I think it's better to leave them un-commented but without value, to differentiate them from the comments.

sbreker:
Looks great @jraddaoui!  Thanks for that change!  Comments will be a great help to admins setting up the feature.

sbreker (approved):
Approved!  👍

PR #708, opened 2018-05-16
Add descendant date calculation, refs #12115

mcantelon:
This adds functionality that lets you specify an event, associated with an
information object, and calculate a start and end date based on the earliest
and latest, respectively, start and end date of events, of the same type,
associated with descendant information objects.

* Add new job type ("arCalculateDescendantDatesJob"): this job will, given
  the ID of an information object and the ID of an event associated with that
  information object, do the aforementioned date calculation
* Add link, on description viewing pages, that opens a dialog that allows the
  user to select an event whose dates should be replaced then trigger date
  calculation
* Under this link show the last time a date calculation was run for the
  information object

### DIFF
--- a/apps/qubit/modules/informationobject/actions/calculateDatesAction.class.php
+++ b/apps/qubit/modules/informationobject/actions/calculateDatesAction.class.php
@@ -1,0 +1,150 @@
+<?php
+
+/*
+ * This file is part of the Access to Memory (AtoM) software.
+ *
+ * Access to Memory (AtoM) is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * Access to Memory (AtoM) is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with Access to Memory (AtoM).  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+class InformationObjectCalculateDatesAction extends DefaultEditAction
+{
+  // Arrays not allowed in class constants
+  public static
+    $NAMES = array(
+      'eventId');
+
+  protected function addField($name)
+  {
+    switch ($name)
+    {
+      case 'eventId':
+        if (count($this->events) == 1)
+        {
+          // If there's only one event with data range data, display its description
+          // via the form label and relay its ID via a hidden field
+          $this->form->setDefault($name, key($this->events));
+          $this->form->setWidget($name, new sfWidgetFormInputHidden);
+          $label = current($this->events);
+        }
+        else
+        {
+          // Otherwise, display the events using a SELECT element
+          $this->form->setWidget($name, new sfWidgetFormSelect(array('choices' => $this->events)));
+          $label = $this->i18n->__('Event');
+        }
+
+        $this->form->setValidator($name, new sfValidatorInteger);
+        $this->form->getWidgetSchema()->$name->setLabel($label);
+
+        break;
+    }
+  }
+
+  protected function processField($field)
+  {
+    switch ($name = $field->getName())
+    {
+      case 'eventId':
+        $this->eventId = $field->getValue();
+
+        break;
+    }
+  }
+
+  protected function earlyExecute()
+  {
+    $this->i18n = $this->context->i18n;
+    $this->resource = $this->getRoute()->resource;
+    $this->events = self::getResourceEventsWithDateRangeSet($this->resource);
+  }
+
+  public function execute($request)
+  {
+    parent::execute($request);
+
+    // Redirect if unauthorized
+    if (!QubitAcl::check($this->resource, 'update'))
+    {
+      QubitAcl::forwardUnauthorized();
+    }
+
+    // Set response to 403 forbidden if attempting to calculate dates using
+    // non-existant descendants
+    if (!count($this->resource->descendants))
+    {
+      $this->getResponse()->setStatusCode(403);
+      return sfView::NONE;
+    }
+
+    if ($request->isMethod('post'))
+    {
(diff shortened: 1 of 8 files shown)


jraddaoui (commented):
Wow, this is nice @mcantelon!

I didn't take a deep look but I just want to mention that we decided a while ago to use a full page instead of a modal/dialog for this kind of forms to simplify the code. Was that a design requirement or could it be changed?

mcantelon:
Thanks Radda! Ah, forgot about that... I'll check on whether or not it's a requirement.

mcantelon:
(It wasn't a requirement so I've reworked it to use a page rather than dialog.)

apps/qubit/modules/informationobject/actions/calculateDatesAction.class.php
```diff
@@ -0,0 +1,149 @@
...
+
+    $criteria = new Criteria;
+    $criteria->add(QubitEvent::OBJECT_ID, $resource->id);
+
+    // Assemble array of descriptions for any events containing date information
+    foreach(QubitEvent::get($criteria) as $event)
+    {
+      if (!empty($event->date) || !empty($event->startDate) || !empty($event->endDate))
+      {
+        $eventType = QubitTerm::getById($event->typeId)->getName(array('cultureFallback' => true));
```
jraddaoui: `$event->type->getName` should work, but I think we must check the type existence.

apps/qubit/modules/informationobject/actions/calculateDatesAction.class.php
```diff
@@ -0,0 +1,149 @@
...
+        else
+        {
+          // Otherwise, display the events using a SELECT element
+          $this->form->setWidget($name, new sfWidgetFormSelect(array('choices' => $this->events)));
+          $label = $this->i18n->__('Event');
+        }
+
+        $this->form->setValidator($name, new sfValidatorInteger);
+        $this->form->getWidgetSchema()->$name->setLabel($label);
+        break;
```
jraddaoui: Minor nitpick, the `break` bellow has an empty line before and this one doesn't. I think it's better with a preceding empty line, like we use to do with `return`.

apps/qubit/modules/informationobject/templates/_calculateDatesLink.php
```diff
@@ -0,0 +1,13 @@
...
+  <li class="separator"><h4><?php echo __('Tasks') ?></h4></li>
+
+  <li>
+    <a href="<?php echo url_for(array($resource, 'module' => 'informationobject', 'action' => 'calculateDates')) ?>">
+      <i class="fa fa-calendar"></i>
+      <?php echo __('Calculate dates') ?>
+    </a>
+  </li>
+  <li>
+    Last run: <?php echo $lastRun ?>
```
jraddaoui: i18n `Last run:`

apps/qubit/modules/informationobject/templates/calculateDatesSuccess.php
```diff
@@ -0,0 +1,42 @@
+<?php decorate_with('layout_1col.php') ?>
```
jraddaoui: We have most of the IO edit forms with the 2col layout, displaying the repo logo, etc. on the left sidebar.

apps/qubit/modules/informationobject/templates/calculateDatesSuccess.php
```diff
@@ -0,0 +1,42 @@
...
+
+  <?php echo $form->renderFormTag(url_for(array($resource, 'module' => 'informationobject', 'action' => 'calculateDates'))) ?>
+
+    <div id="content">
+
+      <fieldset class="collapsible">
+        <div class="fieldset-wrapper">
+
+          <?php echo $form->eventId
+            ->renderRow() ?>
```
jraddaoui: This could go in one line.

apps/qubit/modules/informationobject/templates/calculateDatesSuccess.php
```diff
@@ -0,0 +1,42 @@
...
+
+          <h6>
+            Note: While the date range update is running, the selected description should not be edited.
+            You can check the job page to determine the current status of the update job.
```
jraddaoui: i18n

jraddaoui (changes requested):
Great job @mcantelon!

Just a few minor changes suggested.

Thanks!

apps/qubit/modules/informationobject/actions/calculateDatesAction.class.php
```diff
@@ -0,0 +1,151 @@
...
+    $events = array();
+
+    $criteria = new Criteria;
+    $criteria->add(QubitEvent::OBJECT_ID, $resource->id);
+
+    // Assemble array of descriptions for any events containing date information
+    foreach(QubitEvent::get($criteria) as $event)
+    {
+      if ((!empty($event->date) || !empty($event->startDate) || !empty($event->endDate))
+        && (null !== $eventTypeTerm = QubitTerm::getById($event->typeId)))
```
jraddaoui: Sorry, I didn't express myself clear enough (or mixed two things in the same feedback) ...

You don't need to get the event by id, you already have it hydrated by the ORM as `type` (without Id) so, `null !== $event->typeId` in here and `$eventTypeName = $event->type->getName(array('cultureFallback' => true));` bellow should work.

jraddaoui (approved):
Thanks @mcantelon!

Just a comment that you could ignore but it will be nice to do, to avoid a double hydration of a term by the ORM. Feel free to do it or not and merge.

Also, for the next time, try to address the feedback in extra commits without amending/squashing the existing ones until they're ready for merge, it's a lot easier for the reviewer than looking again at the whole changes.

apps/qubit/modules/informationobject/actions/calculateDatesAction.class.php
mcantelon: Ah, that makes sense. I'll fix that.

mcantelon:
Thanks Radda! I'll do addition commits next time.

apps/qubit/modules/informationobject/actions/calculateDatesAction.class.php
jraddaoui: Sorry, I meant the type term, not the event, but you got me. Thanks!!

PR #710, opened 2018-05-18
Convert i18n fields to Markdown, refs #12149

jraddaoui:
Convert information object, actor, note, repository, and rights i18n fields.
Modify i18nRemoveHtmlTagsTask to convert HTML links and definition
lists to Markdown syntax instead of to plain text and convert the Redmine
link format to Markdown links in i18nCustomLinkToMarkdownTask, which
is also executed in the upgrade process.

Create abstract task class, i18nTransformBaseTask, to be extended by
the two mentioned tasks and others that may need to loop over and
transform i18n fields.

Remove helpers used to render the Redmine link format.


sbreker:
Looks great @jraddaoui!

sbreker (approved):
Looks great!

PR #711, opened 2018-05-18
EAD emph tag to Parsedown on import, refs #12032

sbreker:
This commit fixes an issue with converting EAD emph tags to Parsedown
where the node was not being parsed correctly when <emph> tags are
nested within other tags (e.g `<emph>` tags within `<p>` tag).

Uses getElementsByTagName instead of traversing each node and testing
the tag name.

### DIFF
--- a/lib/QubitXmlImport.class.php
+++ b/lib/QubitXmlImport.class.php
@@ -604,7 +604,8 @@ class QubitXmlImport
               // Additional node processing. Define filters in the schema file.
               if (!empty($methodMap['Filters']))
               {
-                $nodeValue = self::runFilters($domNode2, $methodMap['Filters']);
+                // Modify $domNode2 by reference.
+                self::runFilters($domNode2, $methodMap['Filters']);
               }
 
               // normalize the node text; NB: this will strip any child elements, eg. HTML tags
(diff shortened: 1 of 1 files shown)


lib/QubitXmlImport.class.php
```diff
@@ -942,23 +942,25 @@ public static function runFilters(&$node, $filterParam)
   {
     foreach ($filterParam as $filters)
     {
-      foreach ($filters as $eadTag => $object)
+      foreach ($filters as $eadTag => $objects)
```
jraddaoui: I'd call it `$tag`, as we could use this function in other imports in the future.

lib/QubitXmlImport.class.php
```diff
@@ -942,23 +942,25 @@ public static function runFilters(&$node, $filterParam)
...
             {
-              if (is_callable(array($object, $method)))
-              {
-                $parameters[] = $eadTag;
-                $parameters[] = $child;
-                $child->nodeValue = call_user_func_array(array( & $object, $method), $parameters);
-              }
+              $parameters[] = $eadTag;
+              $parameters[] = $element;
+              $textValue = call_user_func_array(array( & $object, $method), $parameters);
```
jraddaoui: You don't need to pass `$object` by reference in this case.

jraddaoui: I'd also suggest to change `$objects` and `$object` to `$classes` and `$class` in this function to avoid confusion, but that may just be me.

jraddaoui (changes requested):
Nice @sbreker!

Could you change the comment block on top of the `runFilters` function to explain a little the new procedure and to change the return value to void?

Se also a few comments bellow.

Thanks!

lib/QubitXmlImport.class.php
sbreker: True - good point.  Yes my intent was to be able to use these functions for whatever import filtering we may need to do in the future.

lib/QubitXmlImport.class.php
sbreker: I can see the confusion. Will change these names.

jraddaoui (approved):
Thanks @sbreker!

Please, for the next time, try to address the feedback in extra commits without amending/squashing the existing ones until they're ready for merge, it's a lot easier for the reviewer than looking again at the whole changes.

PR #712, opened 2018-05-20
Fix Dublin Core XML export, refs #12077

jraddaoui:
- Do not include empty elements.
- Use base URL from settings for the repo URL


jraddaoui:
5693412 in stable/2.4.x

PR #713, opened 2018-05-20
Fix siblings order on CSV import, refs #12183

jraddaoui:
Disabling the nested set update on IO creation in the CSV import process
doesn't maintain the order of the siblings from the CSV file. Moreover,
this is disabled based on the `index` option, when it should be using
the `skip-nested-set-build` option.

The `skip-nested-set-build` was only used to not run the build nested
set task at the end of the import, a task that is only needed when the
nested set is not being updated on creation, which doesn't make sense.

The build nested set task execution has been removed from the end of the
import and it should be executed manually after the import task if the
`skip-nested-set-build` option was used.


mcantelon:
👍

jraddaoui:
f9f0f2d in stable/2.4.x

PR #717, opened 2018-05-23
Add creator field to advanced search, refs #12140

mcantelon:
Added ability to search by creator to advanced search.


jraddaoui (commented):
Hi @mcantelon,

This looks great. But we've recently added the inherited creators to the search index and I wonder if this should be querying over both. @fiver-watson, what do you think?

mcantelon:
I've updated the PR to include inherited creators... thanks!

jraddaoui (approved):
Nice! Thank you @mcantelon!

PR #718, opened 2018-05-25
Hide login button for unauth user, refs #12168

mcantelon:
Added a setting that allows the login button to be hidden for unauthenticated
users (via the visible elements setting page).


jraddaoui (approved):
Nice!

PR #721, opened 2018-05-30
Fix render value issues and search pager results count

jraddaoui:
They're different issues that come from the same QA feedback. This could be merged with `--ff-only`.


sbreker (approved):
Looks good @jraddaoui!  

👍

PR #723, opened 2018-05-31
Source culture in contact info form, refs #11358

mcantelon:
When editing an institution's contact information, show source culture values
of translatable fields.

### DIFF
--- a/apps/qubit/modules/contactinformation/actions/indexAction.class.php
+++ b/apps/qubit/modules/contactinformation/actions/indexAction.class.php
@@ -108,6 +108,32 @@ class ContactInformationIndexAction extends sfAction
       $value['contactType'] = $this->resource->contactType;
     }
 
+    // Add source culture information if necessary
+    $culture = sfContext::getInstance()->user->getCulture();
+
+    if ($culture != $this->resource->sourceCulture)
+    {
+      $sourceCultureData = array('fields' => array());
+
+      // Add source culture text direction, if different from user's culture,
+      // so source culture value can be displayed appropriately (some languages
+      // display text from right to left: Arabic and Chinese for example)
+      $sourceCultureDirection = sfCultureInfo::getInstance($this->resource->sourceCulture)->direction;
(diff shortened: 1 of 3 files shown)


apps/qubit/modules/contactinformation/actions/indexAction.class.php
```diff
@@ -108,6 +108,30 @@ public function execute($request)
...
 
+    // Add source culture information if necessary
+    $culture = sfContext::getInstance()->user->getCulture();
+
+    if ($culture != $this->resource->sourceCulture)
+    {
+      $sourceCultureData = array('fields' => array());
+
+      // Add source culture direction, if necessary
+      $sourceCultureInfo = sfCultureInfo::getInstance($this->resource->sourceCulture);
```
jraddaoui: Is this really necessary, why can we get the source culture value in here like it's being done bellow? Could you explain a little why is this needed in a comment?

jraddaoui (changes requested):
Looks great @mcantelon! Just a question and suggestion to add some comments.

apps/qubit/modules/contactinformation/actions/indexAction.class.php
mcantelon: It makes the below line below 40 characters shorter, but I'll get rid of it as it does make the code a bit harder to follow.

mcantelon: Oh, I reference the culture info on line 122 as well... I've expanded on the comment and changed the code so I just take store the source culture direction in a convenience variable, rather than general source culture info, to make the purpose of the convenience variable more explicit.

mcantelon: (I've expanded the comment a bit to clarify function of the direction data a bit. Thanks!)

jraddaoui (approved):
That's perfect. Thanks @mcantelon!

mcantelon:
NP! Thanks!

PR #725, opened 2018-06-01
GDPR message banner, refs #12180

sbreker:
This PR includes work to add the GDPR banner. The GDPR banner can be dismissed with the OK button - when this is performed, a session attribute is created that will hide the banner for the session. This works for both authenticated and unauthenticated visitors. The banner functionality and text have a new settings panel under Admin->Settings called "Privacy Notification" - here the GDPR banner functionality can be deactivated, and the message can be customized. The GDPR will be ON by default. Also added a new privacy message static page which is linked from the Quick Links menu as well as from the notification banner.


apps/qubit/modules/default/actions/privacyMessageComponent.class.php
```diff
@@ -0,0 +1,41 @@
...
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with Access to Memory (AtoM).  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+/**
+ * Check for updates component
```
jraddaoui: ;)

apps/qubit/modules/default/actions/privacyMessageComponent.class.php
```diff
@@ -0,0 +1,41 @@
...
+ *
+ * @package AccesstoMemory
+ * @subpackage default
+ * @author Steven Breker <sbreker@artefactual.com>
+ */
+class DefaultPrivacyMessageComponent extends sfComponent
+{
+  public function execute($request)
+  {
+    if ((null !== $this->privacyMessageDismissed = $this->context->user->getAttribute('privacy_message_dismissed')) ||
```
jraddaoui: No need to save it in `$this->privacyMessageDismissed``, right?

apps/qubit/modules/default/actions/privacyMessageComponent.class.php
```diff
@@ -0,0 +1,41 @@
...
+      !sfConfig::get('app_privacy_notification_enabled', true))
+    {
+      return sfView::NONE;
+    }
+    else
```
jraddaoui: With the `return` in the previous `if` this `else` is not needed and adds unnecessary indentation.

apps/qubit/modules/default/templates/_privacyMessage.php
```diff
@@ -0,0 +1,12 @@
+<script type="text/javascript" charset="utf-8" src="/js/privacyMessage.js"></script>
```
jraddaoui, line 1: I think it's better to add it in the component, so it gets added to the head section instead if in the body. E.g.:

https://github.com/artefactual/atom/blob/qa%2F2.5.x/apps/qubit/modules/informationobject/actions/indexAction.class.php#L122

data/fixtures/settings.yml
```diff
@@ -878,6 +878,16 @@ QubitSetting:
...
+  QubitSetting_privacyNotificationEnabled:
+    name: privacy_notification_enabled
+    value: 1
+  QubitSetting_privacyNotification:
+    name: privacy_notification
+    editable: 1
+    deleteable: 0
+    source_culture: en
+    value:
+      en: 'This website uses cookies to enhance your ability to browse and load content. [More Info.](/privacy)'
```
jraddaoui, line 890: Nice use of the new Markdown feature. Being an editable setting I think it's okay to use it.

apps/qubit/modules/default/actions/privacyMessageComponent.class.php
```diff
@@ -0,0 +1,41 @@
...
+    {
+      $this->notificationMessage = sfConfig::get('app_privacy_notification');
```
jraddaoui: I'm not sure how is this going to work with the settings cache and translations. I think it's better to use the QubitSetting value in this case.

lib/task/migrate/migrations/arMigration0165.class.php
```diff
@@ -0,0 +1,98 @@
...
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with Access to Memory (AtoM).  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+/*
+ * Add taxonomy and terms for user actions
```
jraddaoui: ;)

plugins/arDominionPlugin/css/less/print.less
```diff
@@ -3,6 +3,7 @@
   /* Hide */
 
   #update-check,
+  #privacy-message,
```
jraddaoui, line 6: Good catch.

plugins/arDominionPlugin/css/less/scaffolding.less
```diff
@@ -1552,6 +1552,37 @@ body.login #content {
...
+    margin-left: -20px;
+    margin-right: -20px;
+  }
+}
+
+#privacy-message > .alert {
+  margin-bottom: 0px;
+  margin-top: 0px;
+  text-align: center;
+  padding: 8px 35px 18px 14px !important;
```
jraddaoui: Is this `!important` really necessary, I'd try to find a better selector to see if we can avoid using it.

plugins/arDominionPlugin/css/less/scaffolding.less
```diff
@@ -1552,6 +1552,37 @@ body.login #content {
...
+
+  #privacy-message-content {
+    margin-left: 45px;
+    margin-right: 65px;
+    color: @blueDark;
+    a:not(.btn) {
+      color: @gray;
+    }
+    a:hover {
+      color: darken(@linkColorHover, 10%);
```
jraddaoui, line 1577: Nice! It will help in custom themes.

jraddaoui (changes requested):
Nicely done @sbreker!

A few comments but nothing big.

apps/qubit/modules/default/templates/_privacyMessage.php
sbreker: I have tried to make this work by adding the javascript in the controller but it has no effect even after clearing cache, restarting services etc. I believe that this does not work for classes extending sfComponent. I have tried including as in your example which extends sfAction, but it has no effect - the js is not included in the rendered page. I can find other classes in AtoM which extend sfComponent that have also included the js in the template.

plugins/arDominionPlugin/css/less/scaffolding.less
sbreker: Testing this again, it looks like '!important' is not necessary after all. Removed....

lib/task/migrate/migrations/arMigration0165.class.php
sbreker: Fixed.

data/fixtures/settings.yml
sbreker: 👍

apps/qubit/modules/default/actions/privacyMessageComponent.class.php
sbreker: Correct. Removed.

apps/qubit/modules/default/actions/privacyMessageComponent.class.php
sbreker: Fixed.

apps/qubit/modules/default/templates/_privacyMessage.php
sbreker: I have also tried loading this using the view.yml config file for default module, but this has the side effect of loading the js on every AtoM page access, even when it is not needed. Placing this in the template was the only way I could find to include this js successfully, but only when the banner is actually displayed and the js is required.

jraddaoui (approved):
Great! Thanks Steve!

apps/qubit/modules/default/templates/_privacyMessage.php
jraddaoui: I didn't know that, thanks!

PR #726, opened 2018-06-07
Physical Location report enhancement, refs #12255

sbreker:
This commit corrects a couple issues with the physical location report:

- When physical object records had null values in the type_id field,
or other fields, the report would crash. This should no longer occur.
- Fields for which there was no current translation would be blank.
This should no longer occur.
- Characters with diacritics that occurred in the report title would
be garbled. This has been corrected.
- The box label report was having similar display issues of characters
with diacritics - this should also be fixed.

### DIFF
--- a/lib/job/arGenerateReportJob.class.php
+++ b/lib/job/arGenerateReportJob.class.php
@@ -155,19 +155,7 @@ class arGenerateReportJob extends arBaseJob
     $criteria->addJoin(QubitRelation::OBJECT_ID, QubitInformationObject::ID);
     $criteria->addJoin(QubitRelation::SUBJECT_ID, QubitPhysicalObject::ID);
 
-    $storageLocations = QubitPhysicalObject::get($criteria);
-    $results = array();
-
-    foreach($storageLocations as $item)
-    {
-      $results[] = array(
-        'name' => isset($item->name) ? $item->getName(array('cultureFallback' => true)) : '',
-        'slug' => isset($item->slug) ? $item->slug : '',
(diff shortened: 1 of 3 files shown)


jraddaoui (commented):
Hi @sbreker,

Is the error happening in the render methods when `null` is being sent? If so, I think it should be fixed in there as it may happen in other places.

lib/job/arGenerateReportJob.class.php
```diff
@@ -155,7 +155,19 @@ private function getStorageLocationsResults()
     $criteria->addJoin(QubitRelation::OBJECT_ID, QubitInformationObject::ID);
     $criteria->addJoin(QubitRelation::SUBJECT_ID, QubitPhysicalObject::ID);
 
-    return QubitPhysicalObject::get($criteria);
+    $storageLocations = QubitPhysicalObject::get($criteria);
+    $results = array();
+
+    foreach($storageLocations as $item)
```
jraddaoui: I think some logic in the template in this case would be better, to avoid looping twice over the results. Not sure if this could be fixed in the render methods plus getting the values with source culture fallback in the templates.

sbreker:
I have updated the PR with a second commit. I have removed the fix for dealing with null values - this issue only exists in 2.4.x. Fixes for i18n and utf-8 are still present. 

It will be simplest to review using the "Files Changed" tab so you can view the net change.

sbreker:
@jraddaoui ^^.  No rush!

jraddaoui (approved):
Thanks @sbreker!

PR #732, opened 2018-06-17
Search description changes by user, refs #12130

mcantelon:
Added autocomplete form element to the description updates page, when
description change logging is enabled, that will search the change log
by user.

### DIFF
--- a/apps/qubit/modules/search/templates/_updatesSearch.php
+++ b/apps/qubit/modules/search/templates/_updatesSearch.php
@@ -34,7 +34,7 @@
 
         </div>
 
-        <div class="filter-row double" id="io-options">
+        <div class="filter-row double io-options">
 
           <?php if (sfConfig::get('app_multi_repository')): ?>
             <div class="filter-left">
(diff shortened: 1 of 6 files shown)


apps/qubit/modules/search/templates/_updatesSearch.php
```diff
@@ -59,6 +59,23 @@
...
+            <div class="filter-left">
+              <?php echo $form->user
+                ->label(__('User'))
+                ->renderLabel() ?>
+              <?php echo $form->user->render(array('class' => 'form-autocomplete')) ?>
+              <input class="list" type="hidden" value="<?php echo url_for(array('module' => 'user', 'action' => 'autocomplete')) ?>"/>
+
+              <?php if (isset($user)): ?>
+                <p>
+                  <em><?php echo __('Currently displaying:') ?> <?php echo $user->getUsername() ?></em>
```
sbreker: Nitpick: Should this formatting 'em' tag be done with css so that it can be more easily overridden in a custom theme?

sbreker (commented):
Hi Mike - one comment on the form.  Wondering what your thoughts are with respect to where theme formatting should be: ok in template, or should be restricted to CSS.

sbreker:
Looks great @mcantelon!

👍

PR #733, opened 2018-06-18
Fix event management from actor page, refs #11855

jraddaoui:
The related resource should be updated in the ES index when the events
are managed from the actor edit page.


mcantelon:
Looks good! 👍

PR #734, opened 2018-06-18
Fix render issues in IO rights area, refs #12148

jraddaoui:


sbreker (approved):
Looks good @jraddaoui !

PR #737, opened 2018-06-25
Dev/issue 12180 pick gdpr feature

sbreker:
Changes for pick to 2.4.x.  First commit is the pick from 2.5.x.  Second commit is changes required for 2.4.1 compatibility.


apps/qubit/modules/settings/actions/privacyNotificationAction.class.php
```diff
@@ -0,0 +1,114 @@
...
+
+       case 'privacy_notification_enabled':
+         // Determine default value
+         $defaults = array('privacy_notification_enabled' => 1);
+
+         $default = (null !== $this->$name = QubitSetting::getByName($name))
+           ? $this->$name->getValue(array('sourceCulture' => true))
+           : $defaults[$name];
+
+         $options = array($this->i18n->__('No'), $this->i18n->__('Yes'));
```
jraddaoui, line 57: We should normalize this some day, I prefer them this way but we have them a lot as 'yes' and 'no' (lowercase).

jraddaoui (changes requested):
Hi @sbreker,

Looking great! Good thing you made the 2.5 migration idempotent. I wonder, should we include the GDPR setup script in this PR? To void a different download in the upgrade process.

sbreker:
Not a bad idea including the script. I was going to post it as a gist but perhaps including it is simpler, and will not require downloading it and installing it on an AtoM server as an extra step.

Where would be a good place to put this?  It is essentially a migration, to be run using tools:run.

jraddaoui (approved):
Thanks!!

PR #739, opened 2018-07-04
Fix granted rights saving issues, refs #11653

mcantelon:
Fix side effect, that caused granted rights not to be saved, of earlier
attempt to fix 500 error when deleting granted rights.


sbreker:
Looks good.  Thanks @mcantelon!

sbreker:
Looks good! Thanks @mcantelon

sbreker (approved):
👍

PR #740, opened 2018-07-04
Don't show cached OAI DC if disabled, refs #12272

mcantelon:
Added check to see if XML caching is enabled before showing cached results.

### DIFF
--- a/plugins/arOaiPlugin/lib/arOaiPluginComponent.class.php
+++ b/plugins/arOaiPlugin/lib/arOaiPluginComponent.class.php
@@ -148,6 +148,12 @@ abstract class arOaiPluginComponent extends sfComponent
     return file_exists(QubitInformationObjectXmlCache::resourceExportFilePath($resource, $format, true));
   }
 
+  public static function checkDisplayCachedMetadata($resource, $metadataPrefix)
+  {
+    $xmlCachingEnabled = sfConfig::get('app_cache_xml_on_save', false);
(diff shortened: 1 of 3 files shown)


plugins/arOaiPlugin/lib/arOaiPluginComponent.class.php
```diff
@@ -148,6 +148,12 @@ public static function cachedMetadataExists($resource, $metadataPrefix)
     return file_exists(QubitInformationObjectXmlCache::resourceExportFilePath($resource, $format, true));
   }
 
+  public static function shouldDisplayCachedMetadata($resource, $metadataPrefix)
```
sbreker: Wondering if this should be renamed to make it more clear what it does?  Would just  'displayCachedMetadataIfExists()' work?  Or checkDisplayCachedMetadata()?

I am finding 'shouldDisplayCachedMetadata()' to be unclear.

sbreker (changes requested):
I've added one comment.  Let me know what your thoughts are on that.

plugins/arOaiPlugin/lib/arOaiPluginComponent.class.php
mcantelon: Yeah, I like the latter... more clear.

sbreker (approved):
Looks great!  Thanks @mcantelon !

PR #741, opened 2018-07-05
Fix culture-related permission issue, refs #11410

mcantelon:
Fixed a permissions issue preventing editors from doing certain things if they
change to a different language.


sbreker (approved):
Looks good @mcantelon!

PR #742, opened 2018-07-07
Add tooltip for calculate dates link, refs #12082

mcantelon:
Added a TITLE attribute to the calculate dates link to explain the calculate
dates functionality.


jraddaoui (changes requested):
i18n

mcantelon:
I've updated this PR. Let me know what you think.

jraddaoui (approved):
`echo` is slightly faster.

Thanks @mcantelon!

PR #743, opened 2018-07-07
Inherited repos in DACS validation, refs #12246

mcantelon:
The DACS template was, during description updates, indicating that descriptions
were missing repositories, ignoring repository inheritance.


sbreker (approved):
Looks great @mcantelon!  👍

PR #747, opened 2018-07-10
Login/logout/profile display labels, refs #12167

mcantelon:
Got rid of hardcoded text for "Log in", "Log out", and "Profile" and replaced
with display labels.


jraddaoui (approved):
Nice, thanks @mcantelon!

PR #749, opened 2018-07-11
Pub. status update culture fallback, refs #11507

mcantelon:
Added culture fallback for titles displayed on job management page as part
of the publication status update job.


sbreker (approved):
Looks good @mcantelon

PR #751, opened 2018-07-14
Prevent deletion of profile menu, refs #12167

mcantelon:
Modified QubitMenu class to prevent deletion of the profile menu.


mcantelon:
I ended up implementing this a different way.

PR #752, opened 2018-07-15
Dev/issue 12167 menu locking

mcantelon:
Prevent deletion (adding the myProfile menu: issue 12167) and renaming (11186) of system menus.

### DIFF
--- a/apps/qubit/modules/menu/actions/editAction.class.php
+++ b/apps/qubit/modules/menu/actions/editAction.class.php
@@ -98,7 +98,11 @@ class MenuEditAction extends sfAction
         break;
 
       default:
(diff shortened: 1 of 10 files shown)


jraddaoui (changes requested):
Looks good @mcantelon, but I wonder, should this two columns default to `true`? They will only be `false` in a few menu items and new ones created by the user should be modificable.

mcantelon:
Good call. I'll make that change.

mcantelon:
I made the change. I ended up rebasing given there are no comments on any lines of code in this PR (so I thought it shouldn't interfere with review).

EDIT: I also added some addition content to the menu YML fixture.

jraddaoui (approved):
Awesome, thanks @mcantelon!

PR #754, opened 2018-07-17
Don't append to existing CSV export, refs #8387

mcantelon:
Replace CSV export file if it already exists, yet the export writer hasn't
exported any rows.

This fixes unexpected CSV export writer behavior where, if an export with the
same filename already exists before the writer has exported anything, the
writer will append to it rather than replace it.


sbreker (approved):
Looks great @mcantelon!

PR #755, opened 2018-07-20
Menu admin improvements, refs #11177

mcantelon:
* Prevent renaming/deletion of various menu items
* Add check before saving to prevent accidental use of an existing menu name
* Fix issue preventing new menu items from being named


mcantelon:
**The issue where this is being tracked is actually 12167.**

mcantelon:
I redid this work in another PR.

PR #760, opened 2018-07-27
Fix ISAD event editing issues, refs #10505

mcantelon:
Fixed issues with ISAD event editing related to data either imported in a
way different than ISAD normally stores things or data created using another
format, like RAD.

### DIFF
--- a/plugins/sfIsadPlugin/modules/sfIsadPlugin/actions/editAction.class.php
+++ b/plugins/sfIsadPlugin/modules/sfIsadPlugin/actions/editAction.class.php
@@ -173,7 +173,20 @@ class sfIsadPluginEditAction extends InformationObjectEditAction
           }
           else if (!isset($this->request->sourceId))
           {
-            $item->delete();
+            // Will be indexed when description is saved
+            $item->indexOnSave = false;
+
+            // Only delete event if it has no associated date and start/end date
+            if (null === $item->date && null === $item->startDate && null === $item->endDate)
+            {
+              $item->delete();
+            }
+            else
+            {
+              // Handle specially as data wasn't created using ISAD template
+              $item->actor = null;
+              $item->save();
+            }
           }
         }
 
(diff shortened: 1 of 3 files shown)


plugins/sfIsadPlugin/modules/sfIsadPlugin/actions/editAction.class.php
```diff
@@ -173,7 +173,17 @@ protected function processField($field)
           }
           else if (!isset($this->request->sourceId))
           {
-            $item->delete();
```
jraddaoui, line 176: It looks like I forgot to do it in here, it's important to set the `indexOnSave` flag from the event to `false` before saving or deleting an event when the related description will be saved later, as otherwise it may trigger a duplicated job to update the description descendants in ES.

plugins/sfIsadPlugin/modules/sfIsadPlugin/actions/eventComponent.class.php
```diff
@@ -127,8 +127,20 @@ public function processForm()
...
+        // Only delete event if it has no associated actor
+        if (null === $item->actor)
+        {
+          $item->indexOnSave = false;
+          $item->delete();
+        }
+        else
+        {
+          // Handle specially as data wasn't created using ISAD template
+          $item->startDate = null;
```
jraddaoui, line 142: `$item->indexOnSave = false;` should be set in both cases.

jraddaoui (changes requested):
Nice @mcantelon!

A couple of comments about the `indexOnSave` event flag and I think the same fix is also needed in the DACS plugin:

https://github.com/artefactual/atom/blob/qa%2F2.5.x/plugins/arDacsPlugin/modules/arDacsPlugin/actions/editAction.class.php#L151

Thanks!

mcantelon:
I've implemented the `indexOnSave` flags. Let me know if it seems good to you!

jraddaoui (approved):
Thanks Mike!

PR #761, opened 2018-07-27
Digital object file delete check, refs #11449

mcantelon:
Added check to make sure, before deleting a digital object file asset, that
the same file isn't also used by another digital object.


jraddaoui (approved):
Nicely done @mcantelon!

mcantelon:
Thanks Radda! :D

PR #762, opened 2018-07-27
Made job tasks extend from arBaseTak, refs #8491

mcantelon:
Changed job task classes so they extend from arBaseTask instead of sfBaseTask.


jraddaoui (approved):
:+1:

jraddaoui:
Although, it would have been better using the imperative for the commit subject ("Make") ;)

https://chris.beams.io/posts/git-commit/#imperative

mcantelon:
Ah, good call... I've fixed the commit message.

PR #763, opened 2018-07-27
Allow empty paths in top-level menus, refs #8406

mcantelon:
By default, our top-level menu nodes have the path field empty. If the user
tries to edit a menu node, e.g. change the label of the staticPagesMenu node,
the form won't validate unless the user introduces a path. A workaround is to
type in a slash (/), but the validator should be tweaked in order to allow
empty values for top-level nodes.


sbreker (approved):
Looks great @mcantelon!

PR #765, opened 2018-07-27
Allow deletion of "about" static page, refs #8345

mcantelon:
Make tweak to QubitStaticPage model to allow deletion of the about page.


sbreker (approved):
Looks great @mcantelon!

PR #767, opened 2018-07-30
Change digital object URL in MODS XML, refs #9839

mcantelon:
Change where digital object URLs are stored in MODS XML.


jraddaoui (approved):
Nice!

PR #769, opened 2018-08-05
Allow date calc job to create events, refs #12082

mcantelon:
Added support, in the UI and date calculation job, for a new job parameter,
event type ID, that allows an information object with no events to have one
created for it (rather than replacing date data in an existing event).


jraddaoui (approved):
Looks great! Thanks @mcantelon!

PR #770, opened 2018-08-08
Get confirmation before deleting right, refs #7650

mcantelon:
Get confirmation from the user before deleting a right.


apps/qubit/modules/informationobject/actions/indexAction.class.php
```diff
@@ -112,8 +112,6 @@ public function execute($request)
     // Specified here instead of view.yml so plugins calling
```
sbreker: This comment can go too I think. It relates to the line that was removed?

sbreker (changes requested):
I think that comment can be removed. Everything else looks great!  Thanks @mcantelon !

PR #773, opened 2018-08-15
Improve date calculation, refs #12082

mcantelon:
* Added logic so event types are derived from the types present in descendant
  information object's events

* Removed automatic selection of single events (so their dates can be replaced)
  given there is now an option to add a new event with dates calculated from
  descendant dates

* Changed logic determining whether a date calculation link should be shown so
  it's not shown if no events exist in descendent events

### DIFF
--- a/apps/qubit/modules/informationobject/actions/calculateDatesAction.class.php
+++ b/apps/qubit/modules/informationobject/actions/calculateDatesAction.class.php
@@ -30,37 +30,27 @@ class InformationObjectCalculateDatesAction extends DefaultEditAction
     switch ($name)
     {
       case 'eventId':
-        if (count($this->events) == 1)
+        if (count($this->events))
         {
-          // If there's only one event with data range data, display its description
-          // via the form label and relay its ID via a hidden field
-          $this->form->setDefault($name, key($this->events));
-          $this->form->setWidget($name, new sfWidgetFormInputHidden);
-          $label = current($this->events);
-        }
-        else
-        {
-          // Otherwise, display the events using a SELECT element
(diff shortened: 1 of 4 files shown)


jraddaoui:
Hi @mcantelon!

I can see the commit body formatting issue in here too. It looks to be a "Github" thing and only in some places ...

If you click in the three dots on your commit above (without leaving this page), it looks like Github is taking in consideration the 72 chars rule. However, if you click in the commit link and go to the commit page, the body format is maintained. As you can see, it's a silly thing, I was just wondering how it happened. That being said, if we could keep the commit body lines <= than 72 chars, that would be great.

Reviewing the code now ;)

apps/qubit/modules/informationobject/actions/calculateDatesAction.class.php
```diff
@@ -115,19 +106,34 @@ public function execute($request)
...
     {
-      $message = $this->i18n->__('Warning: Selected date range for the specified event will be overwritten.');
+      $message = $this->i18n->__("Warning: If selecting an event, rather than an event type, the selected event's date range will be overwritten.");
       $this->getUser()->setFlash('notice', $message);
     }
   }
 
+  protected function validCalculationSelectionCheck()
+  {
+    return $this->eventId xor $this->eventTypeId;
```
jraddaoui, line 134: :clap:

jraddaoui (approved):
Nicely done @mcantelon!

mcantelon:
Ah cool, I'll make the commit lines <= 72 chars! Thanks @jraddaoui!

PR #776, opened 2018-08-31
Improve date calc. UI and fix issue, refs #12082

mcantelon:
Simplified and cleared up the data calculation UI.

Fixed issue preventing new events from being generated.


jraddaoui (approved):
:+1:

PR #777, opened 2018-09-05
Fixed term ID issues with migration, refs #12083

mcantelon:
Fixed an issue with the migration that adds two terms used by the audit log
functionality.

### DIFF
--- a/data/fixtures/taxonomyTerms.yml
+++ b/data/fixtures/taxonomyTerms.yml
@@ -3953,6 +3953,20 @@ QubitTerm:
       pt-BR: 'Nota da ocupação do registro de autoridade'
       sr: 'Напомена о занимању ствараоца'
       sv: 'Anmärkning rörande aktörsyrke'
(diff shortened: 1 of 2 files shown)


jraddaoui (approved):
Oh! Thanks @mcantelon! That will definitely reduce the possibility of the ID not belonging to a QubitTerm, I didn't know we could have used lower IDs to "fix" the issue.

However, I still think we should improve the `QubitMigrate` bump functions at some point. For example, in old installations, a lot of the terms without a fixed ID could have been added in the upgrade process by a migration and, if that instance had some resources created in between the install and the upgrade, the terms ID may not have low IDs.

As @djjuhasz said, we don't like to play with IDs and foreign keys, but I think this issue will hit us again at some point.

jraddaoui:
Created https://projects.artefactual.com/issues/12420 with the Slack conversation we've had about this issue.

mcantelon:
Awesome... thanks @radda!

PR #778, opened 2018-09-09
Fix issue with tx pulls, refs #12429

mcantelon:
Fixed an issue where, when running "phing i18n-pull-translations", a
plugin that has no i18n subdirectory would trigger an error if the
i18n:extract task found new translations for it. Fixed by pre-creating
an i18n subdirectory, before running i18n:extract, and then removing it
afterwards if it's empty.

### DIFF
--- a/build.xml
+++ b/build.xml
@@ -201,15 +201,34 @@
   </target>
 
   <target name="i18n-cleanup">
-    <foreach param="lang" target="i18n-cleanup-subtask">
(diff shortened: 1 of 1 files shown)


mcantelon:
I would have added the creation of the subdirectory to the  i18n:extract task, but it's part of Symfony rather than one of our tasks.

jraddaoui (approved):
Thanks @mcantelon!

PR #779, opened 2018-09-09
tx pull

mcantelon:


jraddaoui (approved):
Hi @mcantelon, looks good!

I wonder if you pushed the new strings to Transifex or if you're doing this to move to Weblate.

mcantelon:
Thanks @radda! This is to fix an issue where something in the interface isn't translatable via the translation widget thing in AtoM.

PR #780, opened 2018-09-13
Correct CLI SKOS import issue, refs #12364

sbreker:
The SKOS plugin was updated to use EasyRDF and be run in a background
job launched from the WebUI. This commit updates the CLI task to use the
updated SKOS plugin.


mcantelon:
Looks good! 👍

PR #781, opened 2018-09-18
CLI SKOS plugin activation issue, refs #12364

sbreker:
If the SKOS plugin is turned off, and a purge is performed from the
CLI, importing SKOS files from the CLI import:bulk task will fail.

Errors have to do with the SKOS plugin class sfSkosUniqueRelations
not being available.

This change manually loads the classes req so the CLI task will not
fail. The SKOS plugin activation setting is respected.


mcantelon:
👍

PR #784, opened 2018-09-20
Prevent UI access to lock taxonomies, refs #12421

mcantelon:
Prevent users from viewing the index pages for locked taxonomies and terms
within them.

### DIFF
--- a/apps/qubit/modules/term/actions/indexAction.class.php
+++ b/apps/qubit/modules/term/actions/indexAction.class.php
@@ -99,6 +99,13 @@ class TermIndexAction extends DefaultBrowseAction
       $this->forward404();
     }
 
+    // Disallow access to locked taxonomies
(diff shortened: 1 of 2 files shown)


apps/qubit/modules/term/actions/indexAction.class.php
```diff
@@ -93,10 +93,11 @@ public function execute($request)
       $this->forward404();
     }
 
-    // Check that this isn't the root
-    if (!isset($this->resource->parent))
```
jraddaoui: We should not remove this check, I'd keep it in a different `if` like it was.

jraddaoui (changes requested):

apps/qubit/modules/term/actions/indexAction.class.php
mcantelon: Fixed! Sorry, forgot to not rebase my fix.

jraddaoui (approved):
Thanks @mcantelon!!

PR #786, opened 2018-09-28
Fix date duplication with old headers, refs #12467

djjuhasz:
* Fix strict type comparison bug
* Simplify checks for matching event on update


mcantelon (approved):
I don't see any issues and I did some testing and it seems to work fine... nice work!

djjuhasz:
Merged in commit a770d018fc76370d0a88d498c618b2997f797de0

PR #787, opened 2018-09-28
Fix permission settings form, refs 12434

jraddaoui:
Create `access_statement` settings for new Rights Basis terms when
they are don't exists before setting their value.


mcantelon:
👍

jraddaoui:
5466eb9 in stable/2.4.x.

PR #789, opened 2018-10-05
Change digital object DB column type, refs #12488

mcantelon:
The MySQL database digital object table column for recording the size
of digital object files, in bytes, only allowed for a maximum of 2 GB.
Changed to a BIGINT column.

### DIFF
--- a/config/schema.yml
+++ b/config/schema.yml
@@ -105,7 +105,7 @@ propel:
     name: { type: varchar(1024), required: true }
     path: { type: varchar(1024), required: true }
     sequence: integer
-    byte_size: integer
+    byte_size: bigint
     checksum: varchar(255)
     checksum_type: type: varchar(50)
     parent_id: { type: integer, foreignTable: digital_object, foreignReference: id }
@@ -282,7 +282,7 @@ propel:
     filename: varchar(1024)
     last_modified: timestamp
     date_ingested: bu_date
-    size: integer
+    size: bigint
     mime_type: varchar(255)
 
   property:
--- a/lib/task/migrate/migrations/arMigration0167.class.php
+++ b/lib/task/migrate/migrations/arMigration0167.class.php
@@ -1,0 +1,44 @@
+<?php
+
+/*
+ * This file is part of the Access to Memory (AtoM) software.
+ *
+ * Access to Memory (AtoM) is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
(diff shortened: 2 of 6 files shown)


lib/task/migrate/migrations/arMigration0167.class.php
```diff
@@ -0,0 +1,45 @@
...
+ */
+class arMigration0167
+{
+  const
+    VERSION = 167, // The new database version
+    MIN_MILESTONE = 2; // The minimum milestone required
+
+  public function up($configuration)
+  {
+    try {
```
jraddaoui: New line for the `{`.

lib/task/migrate/migrations/arMigration0167.class.php
```diff
@@ -0,0 +1,45 @@
...
+      // Change byte_size column to BIGINT
+      $sql = "ALTER TABLE `digital_object` CHANGE `byte_size` `byte_size` BIGINT";
+      QubitPdo::modify($sql);
+    }
+    catch (Exception $e)
+    {
+    }
```
jraddaoui: I think we should not catch the exception, or at least show an error message.

config/schema.yml
```diff
@@ -105,7 +105,7 @@ propel:
     name: { type: varchar(1024), required: true }
     path: { type: varchar(1024), required: true }
     sequence: integer
-    byte_size: integer
+    byte_size: bigint
```
jraddaoui, line 108: Could we do the same for https://github.com/artefactual/atom/blob/ad6943d07d99e614fe2636e312c308ec39fefaee/config/schema.yml#L285

jraddaoui (changes requested):
Awesome @mcantelon,

I did a similar fix in Binder but completely forgot to bring it here, it was for the `premis_object` table as that's the one used in there, but we also have it in here (not fully used right now).

Also, this *may* need to rebuild the models, please check if that's true if you have not done it yet ;)

Thanks!

config/schema.yml
mcantelon: Ah, good call. I'll make that tweak.

jraddaoui:
Great! Thanks @mcantelon!

But now we're missing the `php symfony propel:build-sql` and `php symfony propel:build-model` changes (the last one may not be needed).

mcantelon:
Hi @radda... I did run build-sql and the two changes to it (one for digital object and one for PREMIS object) should be part of the PR... also I ran build-model and it only changed two files (it seems there's not much difference in how models handle the two numberic types).

jraddaoui (approved):

jraddaoui:
Sorry, I missed the SQL changes ;)

PR #792, opened 2018-10-12
Improve identifier fields in ES, refs #9290

jraddaoui:
- Use a text type field for identifiers to improve searches and add a
  keyword sub-field for sorting.
- Fix `description` fields in QubitFunction mapping.
- Use `AND` operator in Accessions string query.
- Fix typos.


mcantelon:
Looks good! 👍

PR #793, opened 2018-10-18
Search status CLI task, refs #12521

mcantelon:
* Created "load" function in search model classes to make a count of
  documents available to index available without populating the
  search model

* Created search status task for displaying the number of documents
  available to index, for each type, along with the number currently
  indexed

* Fix comment in search populate task


sbreker (approved):
Looks great @mcantelon!

PR #794, opened 2018-10-25
Performance improvements csv:import, refs #12532

sbreker:
Improve performance of csv:import CLI task:
- change function that checks if note already exists to use PDO instead
of Propel ORM lookup.
- make note, event and relation creation respect the --index CLI option.

### DIFF
--- a/lib/QubitFlatfileImport.class.php
+++ b/lib/QubitFlatfileImport.class.php
@@ -1503,6 +1503,7 @@ class QubitFlatfileImport
 
     $note->content = $this->content($text);
     $note->culture = $this->columnValue('culture');
+    $note->indexOnSave = !$this->searchIndexingDisabled;
     $note->save();
 
     return $note;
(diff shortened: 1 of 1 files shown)


mcantelon:
Look great Steve! 👍

lib/QubitFlatfileImport.class.php
```diff
@@ -1522,15 +1523,14 @@ public function createOrUpdateNote($typeId, $text, $options = array())
...
-
-    return null !== QubitNote::getOne($c);
+    $query = "SELECT n.id FROM note n
+      INNER JOIN note_i18n i ON n.id=i.id
+      WHERE n.object_id=?
+      AND n.type_id=?
+      AND i.content=?
+      AND i.culture=?";
+    $statement = self::sqlQuery($query, array($objectId, $typeId, $content, $culture));
+    return false !== $statement->fetch(PDO::FETCH_OBJ);
```
sbreker, line 1533: I've made an additional change here.  This was comparing to null before which is incorrect - PDO fetch returns 'false' when no records are found.

PR #795, opened 2018-11-08
Remove mcrypt dependent libs, refs #12004

sbreker:
The commit removes the password_compat and random_compat libraries.
These both include calls to mcrypt which is deprecated in PHP 7.1. In
addition, random_compat and password_compat are no longer required in
PHP 7 to provide access to these functions.


jraddaoui (approved):

jraddaoui:
Merged in c0d709f.

PR #796, opened 2018-11-08
Remove repos actor autocomplete, refs# 12322

sbreker:
Do not list repositories in the actor autocomplete. This was occurring
when adding creators to descriptions and when linking actors to one
another. Autocomplete was not filtering types of QubitRepository.


mcantelon:
👍

PR #797, opened 2018-11-15
Update Dockerfile remove mcrypt, refs #12004

sbreker:
Remove reference to mcrypt libs in Dockerfile.

Added specific install of nodejs-npm.

Allow PHP 7.x instead of restricting to 7.0.


sbreker:
I've tested Docker installation using this Dockerfile and all proceeds cleanly. This was originally locked to PHP 7.0 because of the requirement of the mcrypt lib.

jraddaoui (approved):

PR #798, opened 2018-11-20
D.O. representation tmp file cleanup, refs #12564

mcantelon:
Added logic to remove temp file generated during the creation of digital
object representations.


sbreker (approved):
Looks great @mcantelon 👍

PR #799, opened 2018-11-23
Add log option to tools:run task, refs #12580

mcantelon:
Added CLI option to log execution of PHP scripts run via the tools:run task.

The execution time and script filename are logged to log/tools_run.log.

### DIFF
--- a/lib/task/tools/runCustomLogicTask.class.php
+++ b/lib/task/tools/runCustomLogicTask.class.php
@@ -24,7 +24,7 @@
  * @subpackage task
  * @author     Mike Cantelon <mike@artefactual.com>
  */
-class runCustomLogicTask extends sfBaseTask
+class runCustomLogicTask extends arBaseTask
 {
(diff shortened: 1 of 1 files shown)


lib/task/tools/runCustomLogicTask.class.php
```diff
@@ -40,6 +40,7 @@ class runCustomLogicTask extends sfBaseTask
   protected function configure()
   {
     $this->addOptions(array(
+      new sfCommandOption('log', null, sfCommandOption::PARAMETER_NONE, 'Log execution of PHP file'),
```
sbreker, line 45: Should we allow users to optionally specify an alternate log file name (with default of tools_run.log) in case they were testing something a few times?  Otherwise it will blow the log away with each run unless they rename it or copy it to a new location.  Do you think this would be useful/necessary?

sbreker (approved):

PR #801, opened 2018-11-29
Fixed /private redirect in viewAction.class.php

g011um:
When AtoM is not installed in the default location, but is instead being served from a sub-directory of the root, it will have a URL like the following (rather the usual http://example.com/)

- http://example.com/path/to/atom/

In cases like this, when accessing an image/file in the "`uploads/r/`" folder however, the setResponseHeaders() method in viewAction.class.php will redirect to the absolute URL

- http://example.com/private/uploads/r/blah/blah/blah.jpg

instead of 

- http://example.com/path/to/atom/private/uploads/r/blah/blah/blah.jpg

That is, portion of the URL path leading to the location of the installed AtoM instance ("/path/to/atom") is lost.  This will result in a 404 error.

This commit fixes the issue with setResponseHeaders() and ensures the URL path portion is retained. At this point, `$_SERVER['SCRIPT_NAME']` will be `/path/to/atom/index.php`. The fix works by removing the trailing filename (`/index.php`) and using the remainder to construct the correct URL for the `/private` folder.

### DIFF
--- a/apps/qubit/modules/digitalobject/actions/viewAction.class.php
+++ b/apps/qubit/modules/digitalobject/actions/viewAction.class.php
@@ -125,7 +125,8 @@ class DigitalObjectViewAction extends sfAction
     // Using X-Accel-Redirect (Nginx) unless ATOM_XSENDFILE is set
     if (false === filter_var($_SERVER['ATOM_XSENDFILE'], FILTER_VALIDATE_BOOLEAN))
     {
-      $this->response->setHttpHeader('X-Accel-Redirect', '/private'.$this->resource->getFullPath());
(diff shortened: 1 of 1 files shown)


fiver-watson:
Thank you @g011um ! 

This work relates to a discussion in the forum - see: 
* https://groups.google.com/d/msg/ica-atom-users/OzGw9lUoYC4/BqjbEcY5CwAJ

@djjuhasz @sbreker we already have a CLA from Darryl (and from USask) on file :)

sbreker (approved):
Hi @g011um! 

Good catch - and thanks for contributing this fix! I've reviewed the code and looks great - I will be merging the change shortly to qa/2.5.x.

Steve

PR #802, opened 2018-12-04
Dev/issue 12611 full treeview paging

mcantelon:
* Update jsTree library to 3.3.7, refs# 12610
    
  Update to remove a JavaScript console warning about a depricated
  function.

* Add paging to full-width treeview, refs #12611
    
  Add pagination to full-width treeview, sharing code with the hierarchy
  browser, so large collections don't incur a large performance hit
  when displaying.


apps/qubit/modules/informationobject/templates/_actions.php
```diff
@@ -70,4 +70,7 @@
 
   </ul>
 
+  <input type="button" id="fullwidth-treeview-reset-button" class="c-btn c-btn-submit" value="<?php echo __('Reset') ?>" />
+  <input type="button" id="fullwidth-treeview-more-button" class="c-btn c-btn-submit" data-label="<?php echo __('%1% more') ?>" value="" />
+  <span id="fullwidth-treeview-collection-url" data-collection-url="<?php echo sfConfig::get('app_siteBaseUrl') ?>/<?php echo $resource->getCollectionRoot()->getSlug() ?>"></span>
```
jraddaoui: Is there a reason to use the `siteBaseUrl` instead of the normal method to build the links in the templates?

js/treeViewPager.js
```diff
@@ -0,0 +1,105 @@
...
+  // Update count of remaining nodes, etc.
+  Qubit.TreeviewPager.prototype.updateMoreLink = function($moreButton, $resetButton)
+  {
+    var scrollOffset = 0;
+
+    if (this.getRemaining() > 0)
+    {
+      // Update count shown in paging button
+      $moreButton.show();
+      $moreButton.val($moreButton.data('label').replace('%1%', this.getRemaining()));
```
jraddaoui, line 61: Not a big deal but, it's possible to change the value before calling `show()`?

jraddaoui (approved):
Great job @mcantelon! Just a couple of notes.

apps/qubit/modules/informationobject/templates/_actions.php
mcantelon: Hmm... I don't think there is. I'll make that change. ;)

js/treeViewPager.js
mcantelon: Good thinking.

jraddaoui:
Thanks @mcantelon! It's already approved ;)

PR #804, opened 2018-12-06
Fix clipboard clear page HTTP 500, refs #12619

mcantelon:
Prevent clipboard clearing page from being accessed without an entity
type specified, redirecting to the clipboard page and showing an
error notification.

Hide link to clipboard clearing page if the clipboard is empty for the
active entity type.


jraddaoui (approved):
:clap:

PR #805, opened 2018-12-07
Add source notes to SKOS export XML, refs #10854

mcantelon:
Added export of source note data to the XML generated during export to
SKOS.


jraddaoui (approved):
:+1:

PR #806, opened 2018-12-07
Fix site title escaping in OAI XML, refs #12622

mcantelon:
The site title wasn't being escaped and caused errors if it contained
special characters.


sbreker (approved):
Looks great Mike!  👍

PR #807, opened 2018-12-08
Fix global search-related sort issue, refs #12620

mcantelon:
Fix problem where a global search will, after the search is refined
using the advanced search form, change from relevance-based sorting.


sbreker (approved):
Looks great Mike!  👍

PR #808, opened 2018-12-10
Fix search box ampersand key handling, refs #11944

mcantelon:
Fixed issue with Chrome interpreting a press of the ampersand key, in
the global search field, as a press of the up arrow key.


sbreker (approved):
Looks good Mike

PR #809, opened 2018-12-10
Auth subject place access points, refs #12593

sbreker:
This is Feature 1 of a larger project.


lib/model/QubitActor.php
```diff
@@ -607,4 +607,20 @@ public function getOccupations()
 
     return QubitObjectTermRelation::get($criteria);
   }
+
+  public function getSubjectAccessPoints()
+  {
+    return $this->getTermRelations(QubitTaxonomy::SUBJECT_ID);
+  }
+
+  public function getPlaceAccessPoints(/*array $options = array('events' => false)*/)
```
mcantelon: Ghost code. ;)

lib/model/QubitObjectTermRelation.php
```diff
@@ -70,4 +70,40 @@ public static function getOneByObjectId($objectId)
...
     return QubitObjectTermRelation::getOne($c);
   }
+
+  public static function setTermRelationByName($name, $options)
+  {
+    // see if subject term already exists
+    $criteria = new Criteria;
+    $criteria->addJoin(QubitTerm::ID, QubitTermI18n::ID);
+    $criteria->add(QubitTerm::TAXONOMY_ID, $options['taxonomyId']);
+    $criteria->add(QubitTermI18n::NAME, $name);
```
mcantelon: Do we need to worry about culture?

mcantelon (changes requested):
Looks good! One one minor thing found and a question.

lib/model/QubitActor.php
sbreker: Removed this - see latest commit.

lib/model/QubitObjectTermRelation.php
sbreker: I think you're right.  I have added a new commit that adds 'culture' to $options and updates the calls accordingly.

mcantelon:
Looks good! Nice work! 👍

mcantelon (approved):

PR #812, opened 2018-12-18
Physical object delete unlinked task, refs #12630

mcantelon:
EDIT: Issue number in branch is wrong, but one in commit is correct.

Added task to delete any physical objects that aren't linked to an
information object.

### DIFF
--- a/lib/task/physicalobject/physicalObjectDeleteUnlinkedTask.class.php
+++ b/lib/task/physicalobject/physicalObjectDeleteUnlinkedTask.class.php
@@ -1,0 +1,181 @@
+<?php
+
+/*
+ * This file is part of the Access to Memory (AtoM) software.
+ *
+ * Access to Memory (AtoM) is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * Access to Memory (AtoM) is distributed in the hope that it will be useful,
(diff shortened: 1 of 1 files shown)


mcantelon:
Will be adding this development to another branch.

lib/task/physicalobject/physicalObjectDeleteUnlinkedTask.class.php
```diff
@@ -0,0 +1,183 @@
...
+      }
+    }
+
+    // Disable search index
+    QubitSearch::disable();
+
+    // Display initial count of physical objects
+    $physicalObjectsCountBefore = $this->getPhysicalObjectCount();
+
+    $this->log();
```
jraddaoui: Is this to add an empty line in the output? Maybe adding `\n` to the following log looks a little better, but if it's not completely necessary I'd remove the empty line from the logs.

lib/task/physicalobject/physicalObjectDeleteUnlinkedTask.class.php
```diff
@@ -0,0 +1,183 @@
...
+    {
+      $physicalObjectsCountAfter = $this->getPhysicalObjectCount();
+    }
+    else
+    {
+      // Simulate results during dry run
+      $physicalObjectsCountAfter = $physicalObjectsCountBefore - count($toDelete);
+    }
+
+    $this->log();
```
jraddaoui, line 130: Same here ;)

jraddaoui (approved):
Nicely done!

PR #815, opened 2018-12-21
Accessions save deletes relations, refs #12662

sbreker:
All relations are being deleted and re-added every time a change to an
accession is saved regardless of which field is updated. This is causing
all related information objects to be reindexed when an accession is
saved - this can be quite slow if many descriptions are linked.

These lines should be comparing the relation "subjectId" and not
"objectId" - objectId contains the accession.id, subjectId contains the
info_obj.id.

### DIFF
--- a/plugins/qtAccessionPlugin/modules/accession/actions/editAction.class.php
+++ b/plugins/qtAccessionPlugin/modules/accession/actions/editAction.class.php
@@ -330,6 +330,7 @@ class AccessionEditAction extends DefaultEditAction
           $relation = new QubitRelation;
           $relation->subject = $item;
           $relation->typeId = QubitTerm::ACCESSION_ID;
+          $relation->indexOnSave = false;
 
           $this->resource->relationsRelatedByobjectId[] = $relation;
         }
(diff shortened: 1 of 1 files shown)


jraddaoui (changes requested):
Good catch @sbreker!

I think we don't save any data from related accessions in the ES index for IOs, so we could also set `indexOnSave` to false for [the new relations created after the check](https://github.com/artefactual/atom/blob/dff59ac7eebcd2eb9a62449b4697e17294ea7e2d/plugins/qtAccessionPlugin/modules/accession/actions/editAction.class.php#L330), to avoid a few more unnecessary request to ES.

sbreker:
Hi @jraddaoui!  As you suggested, I have made the additional change to prevent re-indexing information objects when relations are added to accessions - good idea!

jraddaoui (approved):
Thanks @sbreker!

jraddaoui:
Merged in https://github.com/artefactual/atom/commit/44b295de2883b8821fc5597ac51e3a7ebc9a961d.

PR #816, opened 2018-12-28
CSV Import note performance, refs #12676

sbreker:
The code in the csv:import logic prevents duplicate notes from being
created if:
- it already exist in the database for this object, culture and note
type.
- the same note content, and type is duplicated in the import file.

Change the logic so that it retrieves all existing notes once only when
checking if notes exist.

### DIFF
--- a/lib/QubitFlatfileImport.class.php
+++ b/lib/QubitFlatfileImport.class.php
@@ -1444,6 +1444,11 @@ class QubitFlatfileImport
       }
     }
 
+    // Get existing notes content as array - do this once per CSV row to reduce DB requests.
+    // Update array with note->content being added so values within CSV are also
+    // checked as they are added.
+    $existingNotes = $this->getExistingNotes($this->object->id, $typeId, $this->columnValue('culture'));
+
     foreach ($textArray as $i => $text)
     {
       $options = array();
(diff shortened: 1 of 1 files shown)


lib/QubitFlatfileImport.class.php
```diff
@@ -1510,27 +1515,63 @@ public function createOrUpdateNote($typeId, $text, $options = array())
...
+   *
+   * @param array $existingNotes  Notes already found in the database for this
+   * @param int $objectId  Object id for object that the note belongs to.
+   * @param int $typeId  Note type id indicating note type.
+   * @param string $content  Note content to check against.
+   * @param string $culture  Note culture to check against.
+   *
+   * @return bool  True if the same note exists, false otherwise.
+   */
+  private function checkNoteExists(&$existingNotes, $objectId, $typeId, $content, $culture)
```
mcantelon: Only $existingNotes and $content seem used in the function body. Should the other parameters be removed?

mcantelon (commented):
Looks good in general, just had one question about the code.

lib/QubitFlatfileImport.class.php
sbreker: Hi @mcantelon!  Thanks for spotting that.  I have updated the function params in the new commit.

mcantelon: Cool... looks good now.

mcantelon:
👍 Looks good!

PR #818, opened 2019-01-02
Physical object normalization task, refs #12618

mcantelon:
Add physical object normalization task, based on logic from
existing script by José Raddaoui Marín.

Physical objects with the same name, and no location, are de-duped.
Physical objects with the same name, location, and type are also
de-duped.


jraddaoui (approved):
Looks great!!

PR #820, opened 2019-01-09
Hide full treeview paging buttons, refs #12695

mcantelon:
Added CSS change to hide full treeview paging buttons when they're not
in use.


sbreker (approved):
Look good @mcantelon !

PR #821, opened 2019-01-11
Authority terms fixes, refs #12647

sbreker:
- XML import works now - fixed var assignment.
- moved processing of access point fields to actor module from isaar
plugin to mirror similar code in ISAD template.
- fixed UI line wrapping of related terms in right sidebar.


mcantelon (approved):
Looks good to me!

PR #823, opened 2019-01-16
Optimize phys. object normalization, refs #12713

mcantelon:
Changed the physical object normalization task to use SQL rather than
the ORM. Also made similar optimization to the task for deleting
unlinked physical objects.


sbreker (approved):
Looks great @mcantelon - thanks!  👍

Issue #822, opened 2019-01-15
Prevent to overwrite existing "settings.yml" every time the atom container restarts


ricardodepinho:
https://github.com/artefactual/atom/blob/1138ae4022dfd7ecc84f8e20997c0a690c87ad00/docker/bootstrap.php#L68
if (!file_exists(_ATOM_DIR.'/apps/qubit/config/settings.yml')) {

jraddaoui:
Great @ricardodepinho, I really like the idea! I've added a similar fix for the other two config files that may be overwritten in `apps/qubit/config`.

Thanks!

fiver-watson:
Radda's PR can be seen here: https://github.com/artefactual/atom/pull/825

We now have an issue ticket filed in our project tracker, here: https://projects.artefactual.com/issues/12721

In light of this, I am going to close this issue now. Thanks again for the report and the proposed fix, @ricardodepinho !

PR #825, opened 2019-01-17
Docker: improve config files setup, refs #12721

jraddaoui:
- Don't overwrite existing config files.
- Use Memcached host and port from config.

Connects to #822.


sbreker (approved):
Looks great!  Thanks @jraddaoui!  👍

PR #826, opened 2019-01-21
Add record group description level, refs #12367

mcantelon:
Add new term representing a new level of description: record group.

### DIFF
--- a/lib/task/migrate/migrations/arMigration0168.class.php
+++ b/lib/task/migrate/migrations/arMigration0168.class.php
@@ -1,0 +1,54 @@
+<?php
+
+/*
+ * This file is part of the Access to Memory (AtoM) software.
+ *
+ * Access to Memory (AtoM) is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as published by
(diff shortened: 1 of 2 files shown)


lib/task/migrate/migrations/arMigration0168.class.php
```diff
@@ -0,0 +1,44 @@
...
+class arMigration0168
+{
+  const
+    VERSION = 168, // The new database version
+    MIN_MILESTONE = 2; // The minimum milestone required
+
+  public function up($configuration)
+  {
+    // Create new term for record group level of description
+    $term = new QubitTerm;
```
sbreker: Do we need to use bumpTerm here or check for existence first?

mcantelon: Good thinking... I'll add a check for existence.

mcantelon: I've now added a check for existence (I don't think bumpTerm is needed as it doesn't need a fixed ID).

sbreker (approved):
Looks great @mcantelon! Thanks!

PR #830, opened 2019-01-25
Actor access point import issue, refs #12593

sbreker:
This commit fixes two issues discovered with EAC import:
- biogHist and existDates now strip leading and trailing whitespace
- actor terms now import correctly


mcantelon (approved):
Looks good to me!

PR #832, opened 2019-01-28
Schema version number bump, refs #12367

mcantelon:
Bump schema version number (a migration was previously added that
neglected to do this).


sbreker (approved):
Looks great @mcantelon!  Thanks!!

PR #835, opened 2019-02-11
Add advanced authority record search, refs #12651

mcantelon:
Adds a form, etc., to perform advanced searches on authority records.

To support this, adds ElasticSearch indexing for maintenance notes and
whether or not an authority record has a digital object associated with
it.

### DIFF
--- a/apps/qubit/modules/actor/actions/browseAction.class.php
+++ b/apps/qubit/modules/actor/actions/browseAction.class.php
@@ -32,16 +32,23 @@ class ActorBrowseAction extends DefaultBrowseAction
 
   // Arrays not allowed in class constants
   public static
+    $NAMES = array(
+      'repository',
+      'hasDigitalObject',
+      'entityType',
+      'emptyField'
+    ),
+
     $AGGS = array(
       'languages' =>
         array('type' => 'term',
               'field' => 'i18n.languages',
               'size' => 10),
-      'types' =>
+      'entityType' =>
         array('type' => 'term',
               'field' => 'entityTypeId',
               'size' => 10),
-      'maintainingRepository' =>
+      'repository' =>
         array('type' => 'term',
               'field' => 'maintainingRepositoryId',
               'size' => 10),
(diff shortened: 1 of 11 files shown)


jraddaoui:
Hi @mcantelon, nice!

I have not reviewed all, but I wonder why did you decide to create `arElasticSearchPluginActorQuery`. So far we use `arElasticSearchPluginQuery` in all ES related browse pages and I think it would be nice to keep it that way unless it brings too much complexity.

mcantelon:
Thanks @radda... yeah, I think you're right and have reworked it. Let me know what you think.

apps/qubit/modules/actor/actions/browseAction.class.php
```diff
@@ -96,38 +188,77 @@ protected function populateAgg($name, $buckets)
...
     {
-      $this->repos = QubitRepository::getById($request->repos);
+      if (!in_array($key, $allowed) || in_array($key, $ignored))
+      {
+        continue;
+      }
+
+      $this->hiddenFields[$key] = $value;
+    }
+  }
```
jraddaoui: We could move this function to the default browse action to avoid duplication. Not a big deal as the function will be pretty small, but it will be nice to put the comments from here (and from IO browse) in the function docs explaining the parameters.

apps/qubit/modules/actor/actions/browseAction.class.php
```diff
@@ -146,16 +277,179 @@ public function execute($request)
...
+  {
+    $this->form = new sfForm;
+    $this->form->getValidatorSchema()->setOption('allow_extra_fields', true);
+
+    foreach ($this::$NAMES as $name)
+    {
+      $this->addField($name);
+    }
+
+    $this->setFilterTags($request);
```
jraddaoui: We could call this function before creating the form, to allow setting the default values in the `addField` function.

apps/qubit/modules/actor/actions/browseAction.class.php
```diff
@@ -146,16 +277,179 @@ public function execute($request)
...
+
+      case 'maintenanceNotes':
+        $esFields = arElasticSearchPluginUtil::getI18nFieldNames('maintenanceNotes.i18n.%s.content');
+        break;
+
+      case 'descriptionIdentifier':
+        $esFields = array('descriptionIdentifier');
+        break;
+
+      case '_all':
```
jraddaoui: I think this case will never be reached as it's not an option of the "emptyField". Also, it probably can't be used for an `exists` query and it will be deprecated in ES 6.x.

apps/qubit/modules/actor/actions/browseAction.class.php
```diff
@@ -146,16 +277,179 @@ public function execute($request)
...
+      if (!empty($request->$fieldName))
+      {
+        $this->form->setDefault($fieldName, $request->$fieldName);
+      }
+    }
+
+    $this->setHiddenFields($request);
+  }
+
+  protected function search($request)
```
jraddaoui: Could we rename this function? There is already a variable with the same name and it may cause confusion.

apps/qubit/modules/actor/templates/browseSuccess.php
```diff
@@ -1,103 +1,172 @@
...
+        <?php endif; ?>
+        <?php $params = $sf_data->getRaw('sf_request')->getGetParameters() ?>
+        <?php unset($params['hasDigitalObject']) ?>
+        <a href="<?php echo url_for(array('module' => 'actor', 'action' => 'browse') + $params) ?>" class="remove-filter"><i class="fa fa-times"></i></a>
+      </span>
+    <?php endif; ?>
+
+    <?php if (isset($repository)): ?>
+      <?php echo get_component('search', 'filter', array('object' => $repository, 'param' => 'repository')) ?>
+    <?php endif; ?>
```
jraddaoui: Nice component!

To make it even better we could move the if to the component itself, to avoid adding one for each component call. The check can be added at the beginning of the execute function in the component and return `sfView::NONE` directly.

It would be great to use it in the IO browse page where possible too, but we can do that later if you think it's out of scope.

jraddaoui (changes requested):
Thanks @mcantelon,

Looking really great! I have added a few comments but nothing big.

apps/qubit/modules/actor/actions/browseAction.class.php
mcantelon: Good thinking.

apps/qubit/modules/actor/templates/browseSuccess.php
mcantelon: Nice... I'll try that. I'll make an issues to incorporate this component into the IO browse page.

mcantelon:
Hi Radda. I've updated the PR with the recommended changes. Let me know what you think.

apps/qubit/modules/actor/templates/browseSuccess.php
```diff
@@ -102,25 +102,11 @@
...
-    <?php endif; ?>
-
-    <?php if (isset($subjects)): ?>
-      <?php echo get_component('search', 'filter', array('object' => $subjects, 'param' => 'subjects')) ?>
-    <?php endif; ?>
+      <?php echo get_component('search', 'filter', array('object' => @$repository, 'param' => 'repository')) ?>
+      <?php echo get_component('search', 'filter', array('object' => @$entityType, 'param' => 'entityType')) ?>
+      <?php echo get_component('search', 'filter', array('object' => @$occupation, 'param' => 'occupation')) ?>
+      <?php echo get_component('search', 'filter', array('object' => @$places, 'param' => 'places')) ?>
+      <?php echo get_component('search', 'filter', array('object' => @$subjects, 'param' => 'subjects')) ?>
```
jraddaoui: I see now why you checked the vars before calling the component and I'm having conflicts about this. I don't like the `@` as it may hide further issues, but in this cases it will only hide a notice. We show many notices like those in debug mode but I like the idea of avoiding as many as possible. I don't like to introduce that the operator but I think it's excessive to initiate all those vars as `null` in the action.

jraddaoui (approved):
Thanks @mcantelon!!

I'm having conflicts about the `@` operator but feel free to merge if you think it's okay.

apps/qubit/modules/actor/templates/browseSuccess.php
jraddaoui: Thinking a little more about this, we could put all the filter objects into a `param => object` array and loop over it to add the components, similar to the hidden fields. That way we could avoid the `@` operator and it will also simplify a little the template.

Would that be possible? If you're running out of time, we could add a note about this to the issue you created to use the component in the IO browse page. Please, add me as a watcher to that issue ;)

Thanks!

mcantelon: Hi @radda... yeah, having a param => object array was what I was thinking too. Given we're tight on time for this I'll add note to the issue about the other filter refactoring. Thanks again for quickly reviewing this!

PR #837, opened 2019-02-15
Authority digtial obj migration fix, refs #12650

sbreker:
Correct the foreign key assignment in migration 169 so FK2 relates to
the object table and not information_object.


mcantelon:
👍

PR #838, opened 2019-02-16
Advanced search improvements, refs #12595

mcantelon:
Make a few tweaks to authority record and information object advanced
search functionality:

* Add a media type facet and filter tag to the advanced authority record
  search

* Add an empty field filter to the advanced authority record search

* Make the advanced information object search form collapsed after a
  page reload

### DIFF
--- a/apps/qubit/modules/actor/actions/browseAction.class.php
+++ b/apps/qubit/modules/actor/actions/browseAction.class.php
@@ -44,6 +44,10 @@ class ActorBrowseAction extends DefaultBrowseAction
         array('type' => 'term',
               'field' => 'i18n.languages',
               'size' => 10),
+      'mediatypes' =>
+        array('type' => 'term',
+              'field' => 'digitalObject.mediaTypeId',
+              'size' => 10),
       'entityType' =>
         array('type' => 'term',
               'field' => 'entityTypeId',
@@ -175,6 +179,7 @@ class ActorBrowseAction extends DefaultBrowseAction
   {
     switch ($name)
     {
+      case 'mediatypes':
       case 'entityType':
       case 'occupation':
       case 'place':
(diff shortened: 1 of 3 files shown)


apps/qubit/modules/informationobject/actions/browseAction.class.php
```diff
@@ -498,13 +498,6 @@ public function execute($request)
       $this->getResponse()->addStylesheet('print-preview', 'last');
     }
 
-    // Default to hide the advanced search panel
-    $this->showAdvanced = false;
-    if (filter_var($request->showAdvanced, FILTER_VALIDATE_BOOLEAN))
-    {
-      $this->showAdvanced = true;
-    }
-
```
jraddaoui: This is here to allow having the adv. search open by default when the "Advanced search" link is clicked in the search-box. I think it would be nice to maintain this bit to allow that behavior.

jraddaoui (changes requested):
Hi @mcantelon and @fiver-watson,

The changes look great! Awesome that you could address the three things. However, I'm not sure the adv. search status change is exactly what Dan wants.

The `showAdvanced` parameter was initially added for three reasons:

1 - Allow the adv. search links in the search-box to open that section by default.
2 - Maintain the section open after a search/reset is performed.
3 - Maintain the section open in following requests after it's opened (may be a consequence of 1 and 2).

From the feedback in the ticket, I think Dan wants to hide the section on page reload but he would like to maintain the behavior of the adv. search links.

In the case we want to fully remove the parameter, we should also remove the checks made over `$show` [in the adv. search template](https://github.com/artefactual/atom/blob/qa/2.5.x/apps/qubit/modules/search/templates/_advancedSearch.php#L3-L5).

mcantelon:
Thanks @radda... I'll restore the adv. search link functionality.

mcantelon:
Okay, I've restored the adv. search link functionality. The empty field filter tag is manually implemented, via markup, as it works slightly differently than the other filter tags, but I'll make a note to look at "componentizing" it and the "has digital object" filter tag in the refactoring issue created earlier.

jraddaoui (approved):
Thanks @mcantelon!!

PR #839, opened 2019-02-16
Dockerfile change PHP version, refs #12821

sbreker:
Memcache will not install correctly from pecl as specified in our
Dockerfile under PHP 7.3. This issue appears that it could be limited
to affecting Windows and Mac hosts and not Linux.

Update the Dockerfile to install Alpine with PHP 7.2.x. Add code to also
load libzip library which is necessary under PHP 7.3 but also works for
PHP 7.0 and up.

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -1,4 +1,4 @@
-FROM php:7-fpm-alpine
+FROM php:7.2-fpm-alpine
 
 ARG GIT_BRANCH=qa/2.5.x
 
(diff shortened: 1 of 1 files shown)


jraddaoui (approved):
Nice, thanks @sbreker!

On my tests it was still installing 7.2, that's probably why I didn't get the same errors.

jraddaoui:
Merged in f8653aa.

jraddaoui:
Hi @sbreker,

Why do your PRs always get closed instead of merged? I'd love to see purple ;)

sbreker:
I... don't... know? :) Sent you an email with more details. Short answer is I am not sure!

PR #840, opened 2019-02-19
Enable qtSwordPlugin by default, refs #12830

jraddaoui:


jraddaoui:
We decided not to do this in https://projects.artefactual.com/issues/12830.

PR #841, opened 2019-02-25
Digital object master discarding, refs #12840

mcantelon:
Added functionality to the digital object load task, and model, to
support a new digital object usage type: local files.

Local files exist outside of AtoM and, when associated with an
information object using the digital:object task, their representations
will be displayed on the information object's detail page but the
master file won't be accessible.


mcantelon:
(Going to make a change then re-open.)

PR #842, opened 2019-02-28
Digital object master discarding, refs #12840

mcantelon:
Added functionality to the digital object load task, and model, to
support a new digital object usage type: local files.

Local files exist outside of AtoM and, when associated with an
information object using the digital:object task, their representations
will be displayed on the information object's detail page but the
master file won't be accessible.

### DIFF
--- a/lib/model/QubitDigitalObject.php
+++ b/lib/model/QubitDigitalObject.php
@@ -1461,6 +1461,23 @@ class QubitDigitalObject extends BaseDigitalObject
     throw new sfException(sprintf('Error downloading "%s" (attempts: %s).', $uri, $i));
   }
 
+  private function file_get_contents_if_not_empty($filepath)
+  {
+    // Only return file contents if file is indeed a file and isn't empty
+    if (is_file($filepath))
+    {
+      $contents = file_get_contents($filepath);
+
+      if (!empty($contents))
+      {
+        return $contents;
+      }
+    }
+
+    // Throw exception on failure
+    throw new sfException(sprintf('Error reading file or file is empty.', $filepath));
+  }
+
   /**
    * Get filename from URI path
    *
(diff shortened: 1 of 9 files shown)


lib/model/QubitDigitalObject.php
```diff
@@ -1577,14 +1631,24 @@ public static function getCount($mediaTypeId, $options = array())
...
   }
 
+  /**
+   * Check whether the master is stored within AtoM
+   *
+   * @return boolean  whether or not master is stored
+   */
+  public function masterStored($usageId)
+  {
+    return QubitTerm::EXTERNAL_URI_ID != $usageId && QubitTerm::EXTERNAL_FILE_ID != $usageId;
```
jraddaoui: Should we check `QubitTerm::OFFLINE_ID` too?

lib/task/migrate/migrations/arMigration0170.class.php
```diff
@@ -0,0 +1,58 @@
...
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with Access to Memory (AtoM).  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+/*
+ * Add new term to the levels of description taxonomy
```
jraddaoui: ;)

lib/task/migrate/migrations/arMigration0170.class.php
```diff
@@ -0,0 +1,58 @@
...
+  public function up($configuration)
+  {
+    // Check if term actually exists before adding
+    $criteria = new Criteria;
+    $criteria->add(QubitTerm::TAXONOMY_ID, QubitTaxonomy::DIGITAL_OBJECT_USAGE_ID);
+    $criteria->addJoin(QubitTerm::ID, QubitTermI18n::ID);
+    $criteria->add(QubitTermI18n::NAME, 'External file');
+    $criteria->add(QubitTermI18n::CULTURE, 'en');
+
+    if (null === QubitTerm::get($criteria))
```
jraddaoui: We're using a fixed id for the term. If it exists, the `QubitMigrate::bumpTerm` should still be called and the id of the existing term changed to the fixed one for this to work properly. Since we don't use `ON UPDATE CASCADE` for foreign keys, it will also require to change the existing relations with that term.

jraddaoui: Good news, the `QubitTaxonomy::DIGITAL_OBJECT_USAGE_ID` seems to be [a locked taxonomy](https://github.com/artefactual/atom/blob/qa/2.5.x/lib/model/QubitTaxonomy.php#L99), so probably we don't need to check for existence.

lib/task/digitalobject/digitalObjectRegenDerivativesTask.class.php
```diff
@@ -213,7 +213,16 @@ public static function regenerateDerivatives(&$digitalObject, $options = array()
...
 
       default:
-        $usageId = $digitalObject->usageId; // MASTER_ID or EXTERNAL_URI_ID
+        $usageId = $digitalObject->usageId; // MASTER_ID or EXTERNAL_URI_ID or EXTERNAL_FILE_ID
+    }
+
+    // If master isn't stored in AtoM, attempt to cache external resource before
+    // deleting existing derivatives
+    //
+    // An unavailable resource will result in an exception
```
jraddaoui: It looks better to me without the empty comment line.

lib/model/QubitDigitalObject.php
jraddaoui: Maybe we could use this function in a few more of the checks you're modifying, including the ones with `QubitTerm::OFFLINE_ID` if we add it.

It's just a note as I'm not sure if the three should have the same behavior in all the cases.

jraddaoui (changes requested):
Looking good @mcantelon! A few comments, only one important.

lib/model/QubitDigitalObject.php
mcantelon: OFFLINE_ID is different in that no derivatives are created at all (only metadata, or a digital object stored elsewhere, is stored in the digital object). I've changed the name of the `masterStored` method to reflect this difference (changed to `derivativesGeneratedFromExternalMaster`).

lib/task/migrate/migrations/arMigration0170.class.php
mcantelon: Nice!

lib/task/digitalobject/digitalObjectRegenDerivativesTask.class.php
mcantelon: Fixed.

jraddaoui (approved):
Thanks @mcantelon!

PR #843, opened 2019-02-28
Fix treeview related visual issue, refs #12695

mcantelon:
Fixed issue where the grey action button container shows even when no
buttons are visible (in the case of logged out users).


sbreker (approved):
Looks good Mike!

PR #844, opened 2019-03-01
CSV Import terms performance, refs #12826

sbreker:
Change logic so that csv:import retrieves existing terms for a
taxonomy once only, while still ensuring that terms that do not yet
exist in the database are still created.

### DIFF
--- a/lib/QubitFlatfileImport.class.php
+++ b/lib/QubitFlatfileImport.class.php
@@ -1211,13 +1211,10 @@ class QubitFlatfileImport
       // Trim whitespace
       $value = trim($value);
 
-      // Create/relate terms
+      // Create/relate terms from array of term names.
       if (isset($self->termRelations) && isset($self->termRelations[$columnName]) && $value)
       {
-        foreach (explode('|', $value) as $name)
-        {
-          $self->createOrFetchTermAndAddRelation($self->termRelations[$columnName], $name);
-        }
+        $self->createOrFetchTermAndAddRelation($self->termRelations[$columnName], explode('|', $value));
       }
 
       // Create/update notes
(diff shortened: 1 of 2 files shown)


sbreker:
Previously, all terms in the taxonomy were being retrieved and then filtered using the term_i18n.name in mysql. This process was being performed for every term contained in the import spreadsheet and was not scaling for large databases where certain taxonomies contain a large number of terms. This change pulls all terms in the taxonomy back to the csv:import task once per CSV row per taxonomy, then uses that data to find or create the new terms being added in the spreadsheet.

lib/QubitFlatfileImport.class.php
```diff
@@ -1917,32 +1914,77 @@ public function createContactInformation($actorId, $options)
...
+
+    if (isset($notArray))
+    {
+      return $terms[0];
+    }
+    return $terms;
+  }
+
+  /**
+   * Retrieve term index for
```
mcantelon: Seems like part of the comment got cut off.

lib/QubitFlatfileImport.class.php
```diff
@@ -1917,32 +1914,77 @@ public function createContactInformation($actorId, $options)
...
+   * Must all be from same taxonomy.
+   *
+   * @param array  $rows  Results array from term query lookup
+   * @param string $name  Term name
+   *
+   * @return integer Index value if $name is found otherwise null
+   */
+  private static function getTermIndex($rows, $name)
+  {
+    foreach ($rows as $index => $row) {
```
mcantelon, line 1983: You could shorten it to this too I think:

`return array_search($name, array_column($rows, 'name'));`

mcantelon (changes requested):
Looks good to me other than a code comment and one possible minor code tweak!

lib/QubitFlatfileImport.class.php
sbreker: I tested the technique you are suggesting and it will not return the correct array key.  array_search will return the key of a _new_ array produced by array_column().  I need the key of the array returned by the PDO db lookup.

lib/QubitFlatfileImport.class.php
sbreker: Fixed in new commit.

lib/QubitFlatfileImport.class.php
mcantelon: D'oh... alright, never might that recommendation. ;)

mcantelon:
Looks good to me!

mcantelon (approved):

PR #845, opened 2019-03-02
Refactor advanced search filter tags, refs #12815

mcantelon:
Refactored filter tags so all of them can be rendered by a component.

Filter tag configuration can be added to an advanced search action
class and, with a call to a method of the parent DefaultBrowseAction
class, a single partial uses to display all filter tags.

### DIFF
--- a/apps/qubit/modules/actor/actions/browseAction.class.php
+++ b/apps/qubit/modules/actor/actions/browseAction.class.php
@@ -39,6 +39,17 @@ class ActorBrowseAction extends DefaultBrowseAction
       'emptyField'
     ),
 
+    $FILTERTAGS = array(
+      'hasDigitalObject' => array(),
+      'repository'       => array('model' => 'QubitRepository'),
+      'entityType'       => array('model' => 'QubitTerm'),
+      'occupation'       => array('model' => 'QubitTerm'),
+      'place'            => array('model' => 'QubitTerm'),
+      'subject'          => array('model' => 'QubitTerm'),
+      'mediatypes'       => array('model' => 'QubitTerm'),
+      'emptyField'       => array()
+    ),
+
     $AGGS = array(
       'languages' =>
         array('type' => 'term',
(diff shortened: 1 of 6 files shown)


mcantelon:
(After merging this I'll refactor other advance search actions using the filter tags partial.)

apps/qubit/modules/actor/templates/browseSuccess.php
```diff
@@ -68,14 +68,14 @@
         <?php echo get_partial('search/aggregation', array(
           'id' => '#facet-places',
           'label' => sfConfig::get('app_ui_label_place'),
-          'name' => 'places',
+          'name' => 'place',
```
jraddaoui, line 71: Why is this changed?

apps/qubit/modules/actor/templates/browseSuccess.php
```diff
@@ -68,14 +68,14 @@
...
           'aggs' => $aggs,
           'filters' => $search->filters)) ?>
 
         <?php echo get_partial('search/aggregation', array(
           'id' => '#facet-subjects',
           'label' => sfConfig::get('app_ui_label_subject'),
-          'name' => 'subjects',
+          'name' => 'subject',
```
jraddaoui, line 78: And this one?

apps/qubit/modules/actor/actions/browseAction.class.php
```diff
@@ -216,46 +227,21 @@ protected function populateAgg($name, $buckets)
...
-    {
-      $this->subject = QubitTerm::getById($request->subject);
-    }
-
-    if (isset($request->mediatypes) && ctype_digit($request->mediatypes))
-    {
-      $this->mediatypes = QubitTerm::getById($request->mediatypes);
-    }
+    // Set label for empty field filter tag
+    $labelText = $this->i18n->__('Empty: %1%', array('%1%' => $this->fieldOptions[$request->emptyField]));
```
jraddaoui: Should we check if `$request->emptyField` is set?

jraddaoui (commented):
Really nice @mcantelon! Just a few comments. Thanks!!

apps/qubit/modules/actor/templates/browseSuccess.php
mcantelon: This facet isn't currently working in qa/2.5.x so I think I misnamed it at some point. Same with subject.

apps/qubit/modules/actor/actions/browseAction.class.php
mcantelon: Good call.

mcantelon:
I made that change... ready for another review.

jraddaoui (approved):

PR #846, opened 2019-03-05
CSV actor import index option fix, refs #12845

sbreker:
The $import variable is reassigned multiple times in the CSV actor
import CLI task. Ensure that the QubitFlatfileImport
searchIndexingDisabled property is reassigned the --index option value
before each call to the QubitFlatfileImport::csv() method.


mcantelon (approved):
Looks good to me!

PR #847, opened 2019-03-06
Add rights field visible elements, refs #12845

sbreker:
Add the RAD Rights notes field to visible elements to control
the visibility of this field when viewing a descriptions using the
RAD template.


mcantelon (approved):
Looks good to me!

PR #849, opened 2019-03-07
CSV terms import performance fix, refs #12826

sbreker:
Ensure that the terms are returned in the same order as the prior
existing logic.


sbreker:
This change tightens this code up a bit, and ensures that the terms array is returned in the order it was received as was with the original behaviour (this impacts the order in which the relations are created, and thus the order that they are displayed when viewing descriptions).

mcantelon (approved):
👍

PR #852, opened 2019-03-19
Increase size of culture column, refs #9051

mcantelon:
Increase size of culture column in i18n tables in order to accommodate
non-standard country codes.


sbreker (approved):
Looks great @mcantelon ! 👍

mcantelon:
Thanks Steve!

PR #853, opened 2019-03-20
Physical object location/type search, refs #12459

mcantelon:
Add ability to, on the physical storage management page, search
physical objects by location and type in addition to by name.


jraddaoui (approved):

mcantelon:
Thanks Radda!

PR #854, opened 2019-03-20
Normalize Dockerfile with stable, refs #12721

jraddaoui:
Use the same Dockerfile as in stable/2.4.x, only changing the
GIT_BRANCH. AtoM 2.5.x doesn't support PHP 7.2 yet (at least not
fully tested) and using two different PHP versions is bringing
different issues while the code difference is not that important.

The only differences between the stable/2.4.x and the qa/2.5.x Docker
Compose environments will be the Git branch and the Elasticsearch
image after this changes.

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -1,4 +1,4 @@
-FROM php:7.2-fpm-alpine
+FROM php:7.0-fpm-alpine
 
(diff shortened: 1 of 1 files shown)


mcantelon (approved):
👍

PR #855, opened 2019-03-20
Fix EAD roundtripping issue, refs #12367

mcantelon:
Fixed EAD roundtripping issue where descriptions exported, that have
the level of description of "Record group", get imported with the
level of description "recordgrp", resulting in a new level of
description term being created.

### DIFF
--- a/lib/model/QubitInformationObject.php
+++ b/lib/model/QubitInformationObject.php
@@ -2286,10 +2286,26 @@ class QubitInformationObject extends BaseInformationObject
 
   public function setLevelOfDescriptionByName($name)
   {
-    // don't proceed with empty $name, or if 'otherlevel' value is passed
+    // Normalize name of "Record group" level of description (for EAD imports)
+    if ($name == 'recordgrp')
+    {
+      // See if Level of Description term already exists, if so link to it
+      $criteria = new Criteria;
+      $criteria->add(QubitTerm::TAXONOMY_ID, QubitTaxonomy::LEVEL_OF_DESCRIPTION_ID);
+      $criteria->addJoin(QubitTerm::ID, QubitTermI18n::ID);
+      $criteria->add(QubitTermI18n::CULTURE, 'en');
+      $criteria->add(QubitTermI18n::NAME, 'Record group');
+
+      if ($term = QubitTermI18n::getOne($criteria))
+      {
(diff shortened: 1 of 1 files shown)


lib/model/QubitInformationObject.php
```diff
@@ -2286,10 +2286,13 @@ public function setHistoryByOrigination($history)
 
   public function setLevelOfDescriptionByName($name)
   {
-    // don't proceed with empty $name, or if 'otherlevel' value is passed
+    // Normalize name of "Record group" level of description (for EAD imports)
+    $name = ($name == 'recordgrp') ? 'Record group' : $name;
```
jraddaoui: What about other cultures? For example, if `recordgrp` is used in a French import.

jraddaoui (commented):
Hi @mcantelon,

I wonder why is this a round-tripping issue and if it could be fixed in the export instead of in the import.

lib/model/QubitInformationObject.php
```diff
@@ -2286,10 +2286,13 @@ public function setHistoryByOrigination($history)
 
   public function setLevelOfDescriptionByName($name)
   {
-    // don't proceed with empty $name, or if 'otherlevel' value is passed
+    // Normalize name of "Record group" level of description (for EAD imports)
+    $name = ($name == 'recordgrp') ? sfContext::getInstance()->i18n->__('Record group') : $name;
```
jraddaoui: The value in the translation messages could be different than the one in the database. Isn't really a way to fix this in the export? Sorry for the insistence.

mcantelon: I think the export has to translate human readable terms into an EAD controlled vocabulary (see: https://www.loc.gov/ead/tglib/elements/archdesc.html).

> LEVEL | #REQUIRED, class, collection, file, fonds, item, otherlevel, recordgrp, series, subfonds, subgrp, subseries

And unfortunately there's no constant term ID for the record group level of description term.

I guess what I'll do is look up the term by name with the understanding that if they change the name of the record group level of description in AtoM it's break EAD roundtripping of any record level descriptions.

mcantelon: I've made name lookup of the record group LOD localized. Let me know if this seems okay.

jraddaoui (approved):
Ok, thanks @mcantelon!

PR #860, opened 2019-03-21
Fixed issue with REST API routing, refs #12829

mcantelon:
Moved REST API routes before a "catch all" route to prevent API
endpoints from 404ing.

Removed redundant route (was in routing.yml twice).


jraddaoui (approved):
Thanks @mcantelon!

PR #863, opened 2019-03-27
Don't set a hardcoded database name in migrations

scollazo:
This is breaking upgrades when the database has a name other than "atom"


sbreker (approved):
Code review complete - Looks good! 👍

PR #864, opened 2019-03-28
Actor CSV alias, etc. export fix, refs #12861

mcantelon:
Fixed issue when, when exporting actors, relations and alias data would
only get exported for a single actor.


sbreker (approved):
Looks great @mcantelon - Nice work!

PR #865, opened 2019-04-01
Custom link to markdown fix, refs 12880

sbreker:
In the i18n custom link to markdown task, fetch row data as associative
array instead of objects to save memory.

This fix along with the change to remove create_function() will help reduce
the memory usage of migration 163 so that large databases can be upgraded
successfully.


mcantelon (approved):
Looks good to me... nice optimization!

PR #866, opened 2019-04-02
Actor name Untitled, refs #12650

sbreker:
The actor name was showing as 'Untitled' when adding and deleting digital
objects. This has been fixed so that the authority record authorized form
of name is shown.


mcantelon (approved):
Commit header should maybe have text "Fix actor name Untitled" instead of "Actor name Untitled" but other than that nitpick looks good to go!

PR #869, opened 2019-04-04
Fix digital object bulk upload, refs #12760

mcantelon:
Fixed issue where thumbnails would overwrite original files when
uploading digital objects using the Flash-based multiple file
uploader.


sbreker (approved):
Looks great Mike!

PR #870, opened 2019-04-05
Fix CSV import issue with PHP 7.1+, refs #12942

mcantelon:
arguments to be provided.


sbreker (approved):
Looks good @mcantelon!  Was wondering originally if it made more sense to make QubitFlatfileExport::loadResourceConfigFile take a variable number of parameters but I think this makes more sense.

mcantelon:
Thanks @sbreker !

PR #871, opened 2019-04-05
Fix deaccession deletion issue, refs #12460

mcantelon:
Fixed issue where deleting a deaccession would remove the
accession/accural related to it from the search index.

### DIFF
--- a/plugins/qtAccessionPlugin/lib/model/QubitDeaccession.php
+++ b/plugins/qtAccessionPlugin/lib/model/QubitDeaccession.php
@@ -38,15 +38,11 @@ class QubitDeaccession extends BaseDeaccession
   {
     parent::save($connection);
 
-    QubitSearch::getInstance()->update($this->accession);
-
     return $this;
   }
 
   public function delete($connection = null)
   {
-    QubitSearch::getInstance()->delete($this->accession);
-
     return parent::delete($connection);
   }
 }


plugins/qtAccessionPlugin/lib/model/QubitDeaccession.php
```diff
@@ -45,8 +45,6 @@ public function save($connection = null)
 
   public function delete($connection = null)
   {
-    QubitSearch::getInstance()->delete($this->accession);
```
sbreker, line 48: If we are updating the index when we save/add an accession, then will we need to also '->update' on a delete? Or is the '->update' in save() not necessary either?

mcantelon: I did some testing and looked at the code and it seems like the parent accession's document exists independently of whether or not there's a deaccession so I've removed both of the ElasticSearch alterations.

mcantelon: (I've gotten rid of the `update` in the save because of this.)

mcantelon (commented):
I've updated the PR.

sbreker (approved):
Looks great @mcantelon! Thanks for looking at that ->update() method!

mcantelon:
Thanks @sbreker!

PR #872, opened 2019-04-08
Fix EAD duplication of country code, refs #12869

mcantelon:
Fix issue when the country code is, in deviation from the EAD standard,
added to the eadid element's mainagencycode attribute.


sbreker (approved):
Looks good @mcantelon !

PR #873, opened 2019-04-08
Fix DIP upload routing issue, refs #12932

sbreker:
The sword/deposit/<slug> route was getting caught by the slug/default
route. Move the sword routes ahead of any :slug routes so they are
caught first and not interpreted as slugs. The api/ routes are also
inserted before the :slug routes.


mcantelon (approved):
Looks good to me!

sbreker:
Modified commit - missed removing a route block I'd moved.

PR #875, opened 2019-04-09
Prevent use of reserved slug names, refs #12947

mcantelon:
Added logic to throw error if slug is reserved for use by an optional
plugin (currently only "api" and "sword").


sbreker (approved):
Looks good @mcantelon!

PR #876, opened 2019-04-10
Add alert ref code setting, refs #12357

sbreker:
Add an alert/warning at the top of the identifier settings page to warn
admins to run the cc and search-populate CLI tasks after changing the
reference code separator setting.


mcantelon (approved):
Looks good to me!

PR #877, opened 2019-04-11
Sitemap file flush fix, refs #12886

sbreker:
There was an issue with the sitemap CLI task where the last chunks of
data were never written to the output files because the end() method
registered with register_shutdown_function() was being called after
both the sitemap index file and the last sitemap file were closed for
writing - end() for the last files was being called after the "DONE"
message was written.

I have removed the call to register_shutdown_function() in favour of
calling end() directly from the sitemap task to complete writing the
last sitemap.#.xml file, and added a call to flush() in the
SitemapWriter class to flush the sitemap index file.

As per AtoM's documentation, a sitemap index file will always be created
and the sitemap file will be broken into 50K line chunks. The sitemap
index file can be ignored if there is only one sitemap chunk created.

A side effect of this fix is that static pages are now included in the
sitemap file as intended, where before they were always in the last
chunk and omitted due to the flush issue on the last file.

### DIFF
--- a/lib/sitemap/SitemapWriter.class.php
+++ b/lib/sitemap/SitemapWriter.class.php
@@ -18,7 +18,7 @@
  */
 
 /**
- * SitemapWriter 
+ * SitemapWriter
  */
 class SitemapWriter
 {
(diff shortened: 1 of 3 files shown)


mcantelon (approved):
Ah, some kind of race condition. That explains a lot.

Nice fix... less voodoo!

PR #879, opened 2019-04-18
Fix APCu cache clearing, refs #12964

mcantelon:
Fixed issue with cache items not being cleared in APCu.


sbreker (approved):
Looks good @mcantelon!

PR #880, opened 2019-04-22
Fix search status issue with PHP 7.1+, refs #12521

mcantelon:
arguments to be provided.


sbreker (approved):
👍 Looks great!

PR #881, opened 2019-04-23
Fix finding aid generation job issue, refs #12969

mcantelon:
Fixed issue, in Centos 7, with finding aid generation job. Made
location of PHP executable explicit when executing task.


sbreker (approved):
Looks good @mcantelon !

PR #882, opened 2019-04-23
Fix unused PO task PHP 7.1+ incompat, refs #12646

mcantelon:
arguments to be provided.


sbreker (approved):
Looks good @mcantelon !

PR #883, opened 2019-04-23
RemoveHtml task fix, refs #12880

sbreker:
Change this task to match changes made to the i18nTransformBaseTask.

Row data is now passed as an associative array.


mcantelon (approved):
Looks good to me!

PR #884, opened 2019-04-24
Revert removehtml task chg, refs #12149

sbreker:
Revert the logic for transforming html description list tags (dl, dt, dd)
to markdown. Old logic removes <dt> tags, and changes <dd> tags to <p>.


mcantelon (approved):
I don't see any issues... looks good!

PR #885, opened 2019-04-26
Create downloads folder, refs #12985

sbreker:
If the downloads folder does not yet exist, allow the 'Upload Finding
Aid' functionality to automatically create it so that the upload will
be successful.


mcantelon (approved):
Looks good to me!

PR #886, opened 2019-04-29
Add downloads directory to manifest, refs #12985

mcantelon:
Add the downloads directory to config/package.xml to the downloads
directory will be included during release packaging.


sbreker (approved):
Looks good @mcantelon!

PR #887, opened 2019-04-30
Make upload/download dirs on install, refs #12985

mcantelon:
Move downloads directory creation code to general class. Add creation
of uploads and downloads directories to installation logic. Add
downloads and uploads directories to list of directories to check (to
make sure they’re writeable).


sbreker (approved):
Looks great @mcantelon !

PR #888, opened 2019-05-02
Escape global search query, refs #12995

mcantelon:
Added entity escaping when showing, as a form input value, the search query.

### DIFF
--- a/apps/qubit/modules/search/templates/_box.php
+++ b/apps/qubit/modules/search/templates/_box.php
@@ -8,9 +8,9 @@
     <input type="hidden" name="sort" value="relevance"/>
 
     <?php if (isset($repository) && !sfConfig::get('app_enable_institutional_scoping')): ?>
-      <input type="text" name="query"<?php if (isset($sf_request->query)) echo ' class="focused"' ?> value="<?php echo $sf_request->query ?>" placeholder="<?php echo __('Search %1%', array('%1%' => strip_markdown($repository))) ?>"/>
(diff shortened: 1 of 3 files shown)


apps/qubit/modules/search/templates/_box.php
```diff
@@ -8,9 +8,9 @@
     <input type="hidden" name="sort" value="relevance"/>
 
     <?php if (isset($repository) && !sfConfig::get('app_enable_institutional_scoping')): ?>
-      <input type="text" name="query"<?php if (isset($sf_request->query)) echo ' class="focused"' ?> value="<?php echo $sf_request->query ?>" placeholder="<?php echo __('Search %1%', array('%1%' => strip_markdown($repository))) ?>"/>
```
sbreker, line 11: Looks good!  I think this change needs to also be made in two additional places:
- In the equivalent file in the Archives Canada plugin
- ~line 30 in: apps/qubit/modules/repository/templates/_holdingsInstitution.php

sbreker (changes requested):

apps/qubit/modules/search/templates/_box.php
mcantelon: Thanks Steve! Will do.

mcantelon: (I've committed escaping to those other templates.)

mcantelon:
@sbreker I've made those changes... thanks!

sbreker (approved):
Looks good @mcantelon - thanks!

mcantelon:
Thanks @sbreker !

PR #890, opened 2019-05-08
Context-sensitive D.O. browse link, refs #12797

mcantelon:
Added a check, in the context menu of information object detail pages,
to see if a digital object exists before displaying a "Browse digital
objects" link.


sbreker (approved):
Looks good @mcantelon ! 👍

PR #894, opened 2019-05-13
Integrate MediaElement player, refs #12902

replaceafill:
This commit replaces the flowplayer library with mediaelement for
playing video and audio digital objects.

It also updates the digital object code to create MPEG4 / H.264
reference derivatives instead of flash video (flv) derivatives.


sbreker (approved):
Looks great @replaceafill! This PR probably could have been broken into smaller commits for clarity (e.g. one commit for removing Flowplayer, one for adding MediaElement, and one for changes to AtoM source) as it is easier for the reviewer to wrap their head around. This one it was pretty clear what is going on though! 👍

PR #897, opened 2019-05-15
Minor fixes to clipboard sending, refs #12800

mcantelon:
Simplified clipboard sending message and hid clipboard send form fields
containing JSON.


sbreker (approved):
Looks great @mcantelon!  👍

PR #898, opened 2019-05-16
Fix issue CLI DO load task, refs #12840

sbreker:
Fixed an issue in the DO load task where if the new --link-source option
is specified and a file in the spreadsheet cannot be found, it was
throwing an exception and stopping the load.

I have updated this so that it behaves the same as when the --link-source
option is not specified: it will log a message including the filename
that was not loaded and it will continue to process the import CSV
without throwing an exception.

### DIFF
--- a/lib/task/digitalobject/digitalObjectLoadTask.class.php
+++ b/lib/task/digitalobject/digitalObjectLoadTask.class.php
@@ -203,6 +203,14 @@ EOF;
           continue;
         }
 
+        if (!file_exists($path = self::getPath($item)))
+        {
+          $this->log(sprintf("Couldn't read file '$item'"));
+          $this->skippedCount++;
+
+          continue;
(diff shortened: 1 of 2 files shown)


lib/task/digitalobject/digitalObjectLoadTask.class.php
```diff
@@ -261,7 +261,12 @@ protected function addDigitalObject($objectId, $path, $options = array())
 
     if ($options['link-source'])
     {
-      $do->importFromFile($path);
+      if(false === $do->importFromFile($path))
```
mcantelon: Nitpick: should have space between "if" and "(".

mcantelon (approved):
One nitpick, but looks good otherwise.

mcantelon (approved):
Looks good to me!

PR #899, opened 2019-05-20
Add clipboard send settings i18n, refs #12792

mcantelon:
Added translation support to settings editing class and implemented for
clipboard settings.


sbreker (approved):
Looks great @mcantelon !

PR #904, opened 2019-05-29
Video player fixes after testing, refs #12902

replaceafill:
This PR:

* Loads the mediaelement JS assets *last* to avoid a problem when the user is authenticated and jQuery is loaded first https://github.com/artefactual/atom/commit/de3395513c5c4bcae178505f9f0f09a239d9b88a
* Sets a fixed and centered 320x240 size for the mediaelement player used previously by the flowplayer https://github.com/artefactual/atom/commit/76df3fd89b990634f49b51b5049deec095796721


sbreker:
Looks good to me @replaceafill!  Make sure that Dan knows about that scaffolding.less change - the docs for 2.6.x will need to include a section for upgrading client themes and will need to be aware of that.

sbreker (approved):
Looks great @replaceafill!  👍

Issue #902, opened 2019-05-28
Uncaught Error: [] operator not supported for strings... is being raised during bulk import


ross-spencer:
During import we are experiencing a type error:
```log
root@server-12851:/usr/share/nginx/server_test/src# php symfony import:bulk /home/artefactual/server/server-migration/migration/archives_migration/
PHP Fatal error:  Uncaught Error: [] operator not supported for strings in /usr/share/nginx/server_test/atom-cefd4f9b59e8d93c16f0b6b8e5625e875f0dc80c/lib/task/import/importBulkTask.class.php:224
Stack trace:
#0 /usr/share/nginx/server_test/atom-cefd4f9b59e8d93c16f0b6b8e5625e875f0dc80c/lib/task/import/importBulkTask.class.php(92): importBulkTask->dir_tree('/home/artefactu...')
#1 /usr/share/nginx/server_test/atom-cefd4f9b59e8d93c16f0b6b8e5625e875f0dc80c/vendor/symfony/lib/task/sfBaseTask.class.php(68): importBulkTask->execute(Array, Array)
#2 /usr/share/nginx/server_test/atom-cefd4f9b59e8d93c16f0b6b8e5625e875f0dc80c/vendor/symfony/lib/task/sfTask.class.php(97): sfBaseTask->doRun(Object(sfCommandManager), NULL)
#3 /usr/share/nginx/server_test/atom-cefd4f9b59e8d93c16f0b6b8e5625e875f0dc80c/vendor/symfony/lib/command/sfSymfonyCommandApplication.class.php(76): sfTask->runFromCLI(Object(sfCommandManager), NULL)
#4 /usr/share/nginx/server_test/atom-cefd4f9b59e8d93c16f0b6b8e5625e875f0dc80c/vendor/symfony/lib/command/cli.php(20): sfSymfonyCommandApplicat in /usr/share/nginx/server_test/atom-cefd4f9b59e8d93c16f0b6b8e5625e875f0dc80c/lib/task/import/importBulkTask.class.php on line 224
```

From @sbreker The issue is happening because of this line in [importBulkTask](https://github.com/artefactual/atom/blob/stable/2.5.x/lib/task/import/importBulkTask.class.php#L205) and more specifically because of a breaking change in PHP 7.1 documented [here](https://www.php.net/manual/en/migration71.incompatible.php#migration71.incompatible.empty-string-index-operator).

jraddaoui:
[...]

PR #906, opened 2019-05-29
Init path as array instead of string, refs #13066

ross-spencer:
This commit corrects the initialization of path in importBulkTask to an array
to counter breaking changes introduced in PHP 7.1 that prevented its previous
initialization to be silently converted to an array.

Connected to #902

### DIFF
--- a/lib/task/import/importBulkTask.class.php
+++ b/lib/task/import/importBulkTask.class.php
@@ -202,7 +202,7 @@ EOF;
 
   protected function dir_tree($dir)
   {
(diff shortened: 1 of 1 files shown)


jraddaoui (approved):

ross-spencer:
Hi @sbreker / @jraddaoui I think we'll verify this in the weekly later (to be cherry-picked into qa/2.5.x), just reading up for myself but can't seem to find it in the public docs, what's the AtoM process for cherry-picking?

PR #907, opened 2019-05-29
Add clipboard send settings i18n, refs #12792

mcantelon:
Fixed issues with translation support in settings editing class.


sbreker (approved):
Looks good @mcantelon - 👍

PR #908, opened 2019-06-05
Set sfRoute->resource explicitly, refs #13069

djjuhasz:
The term/treeViewComponent (and possibly other components)
expect sfRoute->resource to provide the page resource context.


sbreker (approved):
Looks good @djjuhasz !  👍

PR #910, opened 2019-06-06
Fix copyright pop-up HTML, refs #13070

replaceafill:
This prevents the content of the popup to be escaped by using the
Parsedown text method in unsafe mode.


sbreker (approved):
Looks good @replaceafill!  👍

PR #912, opened 2019-06-13
Fix Firefox JS incompatibility, refs #12967

mcantelon:
Fixed problem with navigation of global search form autocomplete
results, via the up and down arrows, in Firefox.


sbreker (approved):
Looks good to me! 👍

PR #913, opened 2019-06-17
Reduce verbosity of ES update job, refs #13085

djjuhasz:
Reduce status updates to reporting after every 100 description updates
to avoid overflowing jobs.output column width.

### DIFF
--- a/lib/job/arUpdateEsIoDocumentsJob.class.php
+++ b/lib/job/arUpdateEsIoDocumentsJob.class.php
@@ -59,14 +59,15 @@ class arUpdateEsIoDocumentsJob extends arBaseJob
     $count = 0;
     foreach ($parameters['ioIds'] as $id)
     {
-      $count++;
-
       if (null === $object = QubitInformationObject::getById($id))
       {
         $this->info($this->i18n->__('Invalid archival description id: %1', array('%1' => $id)));
 
         continue;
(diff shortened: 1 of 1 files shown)


lib/job/arUpdateEsIoDocumentsJob.class.php
```diff
@@ -56,36 +56,44 @@ public function runJob($parameters)
...
-      $this->info($message);
+      // Status update every 100 descriptions
+      if (0 == $count % 100)
+      {
+        $this->info($message);
+      }
     }
 
+    // Final status update
+    $this->info($message);
```
sbreker: This message will be printed twice if the number of objects being indexed is a multiple of 100.

sbreker (commented):
Looks great! One nitpick regarding that last iteration print statement.

lib/job/arUpdateEsIoDocumentsJob.class.php
djjuhasz: Thanks @sbreker, I've fixed this.

sbreker (approved):
Looks good @djjuhasz!

djjuhasz:
Merged to qa/2.6.x with 2976888352adf9b2661e4207e2a038c8fed8af34

PR #915, opened 2019-06-20
Remove unnecessary data from ES, refs #13096

djjuhasz:
* Remove unnecessary repository data from archival description document
* Remove unnecessary name access point data from archival description document

### DIFF
--- a/plugins/arElasticSearchPlugin/config/mapping.yml
+++ b/plugins/arElasticSearchPlugin/config/mapping.yml
@@ -466,8 +466,12 @@ mapping:
           id: { type: integer }
           slug: { type: keyword }
           identifier: { type: keyword }
+      names:
+        _i18nFields: [authorizedFormOfName]
+        dynamic: strict
+        properties:
+          id: { type: integer }
     _foreign_types:
-      names: actor
       creators: actor
       inherited_creators: actor
       subjects: term
--- a/plugins/arElasticSearchPlugin/config/mapping.yml
+++ b/plugins/arElasticSearchPlugin/config/mapping.yml
@@ -459,8 +459,14 @@ mapping:
         properties:
           id: { type: integer }
           slug: { type: keyword }
+      repository:
+        _i18nFields: [authorizedFormOfName]
+        dynamic: strict
+        properties:
+          id: { type: integer }
+          slug: { type: keyword }
+          identifier: { type: keyword }
     _foreign_types:
-      repository: repository
       names: actor
       creators: actor
       inherited_creators: actor
--- a/plugins/arElasticSearchPlugin/config/mapping.yml
+++ b/plugins/arElasticSearchPlugin/config/mapping.yml
@@ -471,6 +471,10 @@ mapping:
         dynamic: strict
         properties:
           id: { type: integer }
+        _foreign_types:
+          other_names: other_name
+          parallel_names: other_name
+          standardized_names: other_name
     _foreign_types:
       creators: actor
       inherited_creators: actor
--- a/plugins/arElasticSearchPlugin/lib/arElasticSearchMapping.class.php
+++ b/plugins/arElasticSearchPlugin/lib/arElasticSearchMapping.class.php
@@ -400,6 +400,11 @@ class arElasticSearchMapping
           'properties' => $nestedI18nObjects));
       }
 
+      if (isset($mapping['_foreign_types']))
+      {
+        $this->processForeignTypes($mapping);
+      }
+
       // Add id of the partial foreign resource
       $mapping['properties']['id'] = array('type' => 'integer', 'include_in_all' => 'false');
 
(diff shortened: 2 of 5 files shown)


plugins/arElasticSearchPlugin/config/mapping.yml
```diff
@@ -459,8 +459,19 @@ mapping:
...
+          slug: { type: keyword }
+          identifier: { type: keyword }
+      names:
+        _i18nFields: [authorizedFormOfName]
+        dynamic: strict
+        properties:
+          id: { type: integer }
     _foreign_types:
-      repository: repository
       names: actor
```
jraddaoui: Should we remove the names from the full foreign types? This is probably adding the old names mapping and overwriting the new one from the partial foreign types.

plugins/arElasticSearchPlugin/lib/model/arElasticSearchInformationObjectPdo.class.php
```diff
@@ -1276,7 +1284,20 @@ public function serialize()
...
+        'id' => $node->id,
+        'i18n' => arElasticSearchModelBase::serializeI18ns(
+          $node->id,
+          array('QubitActor'),
+          array('fields' => array('authorized_form_of_name'))
+        )
+      );
+
+      // Add other names, parallel names, and standardized names
+      $names += $node->serializeAltNames();
```
jraddaoui, line 1298: This will probably not work if you remove the names from the full foreign types as the other names mapping will be missing in the new names mapping.

jraddaoui (changes requested):
It's great to see you back @djjuhasz!

The repository changes look good but I think there may be some issues with the names mapping, see my comments below.

Thanks!

plugins/arElasticSearchPlugin/config/mapping.yml
djjuhasz: Good catch @jraddaoui, thanks.  I've removed the `names: actor` line from the `_foreign_types` definitions.

jraddaoui: It's not ideal but I think it's okay to have the old mapping and only add the partial data if it's not easy to reflect the changes in the names mapping, in which case you could remove the partial foreign type from the mapping instead of the other way around.

jraddaoui:
And a minor formatting issue in the commit message with the line lengths. We try to follow [this rules](https://chris.beams.io/posts/git-commit/#seven-rules) for commit messages ;)

plugins/arElasticSearchPlugin/config/mapping.yml
djjuhasz: @jraddaoui I think you are right in removing the `names: actor` from the full foreign types definitions.  I've checked the resulting data in ES (via the JSON-LD REST endpoint) and the expected "names" data is present without effecting other actor types (e.g. creator, inherited_creator).

djjuhasz:
@jraddaoui thanks, I've updated the commit message. 😄  Okay to merge?

plugins/arElasticSearchPlugin/lib/model/arElasticSearchInformationObjectPdo.class.php
djjuhasz: I've tested with a linked authority record with an "other names" value, and it is stored with the "names" data structure.

e.g.
```
"names": [
            {
              "id": "79776",
              "i18n": {
                "en": {
                  "authorizedFormOfName": "Roy, Flora"
                },
                "languages": [
                  "en"
                ]
              },
              "otherNames": [
                {
                  "sourceCulture": "en",
                  "i18n": {
                    "en": {
                      "name": "Floopy"
                    },
                    "languages": [
                      "en"
                    ]
                  }
                }
              ]
            }
          ],
```

jraddaoui: I wonder when is the `otherNames` mapping inside `names` defined now that the full foreign type is not in there, since the partial foreign type has `dynamic: strict` it should not be possible to add it if it's not defined.

Have you fully recreated the index after the latest changes? Please, could you verify in the mapping definition the final `names` mapping. Just the mapping, as requesting a single IO will only show the populated data in the document.

plugins/arElasticSearchPlugin/lib/arElasticSearchMapping.class.php
```diff
@@ -400,6 +400,11 @@ protected function processPartialForeignTypes(array &$typeProperties)
           'properties' => $nestedI18nObjects));
       }
 
+      if (isset($mapping['_foreign_types']))
+      {
+        $this->processForeignTypes($mapping);
+      }
+
```
jraddaoui, line 407: Great! It wasn't that hard in the end ;)

jraddaoui (approved):
Looks great David, thanks!

djjuhasz:
Merged to qa/2.6.x with commit c08beb6f1e57425a678009a93079621c62124dcd

PR #916, opened 2019-06-21
Truncate PDF text to <64KB, refs #13095

djjuhasz:


jraddaoui (approved):
We may want to do the same for the uploaded finding aid transcript:

https://github.com/artefactual/atom/blob/qa/2.6.x/lib/job/arFindingAidJob.class.php#L247-L263

djjuhasz:
@jraddaoui good call on truncating the finding aid text as well.  I also discovered my 2^16-1 notation was truncating the text to like 8 characters, so I switched to the full decimal representation of 64KB - 1B.

jraddaoui (approved):
Thanks @djjuhasz!

djjuhasz:
Merged to qa/2.6.x with 0934589e264f4440697c4de364608dbc34bb3efb

PR #918, opened 2019-06-29
Handle checkboxes in settings class, refs #13028

mcantelon:
Added logic, to settings editing parent class, to handle checkbox
fields.


sbreker (approved):
Looks good @mcantelon!  👍

PR #920, opened 2019-07-03
Refactored DIP settings page, refs #13028

mcantelon:
Refactored DIP settings page to use settings editing parent class.


sbreker (approved):
Looks great @mcantelon!

PR #921, opened 2019-07-03
Add DACS element/header visibility, refs #13049

mcantelon:
Added settings for visibility of DACS template headers and for the
physical access field element. Added notification when visibility
configuration is saved.


sbreker (approved):
Approved - Looks great @mcantelon!

PR #923, opened 2019-07-04
New digitalobject load task option, refs #13103

sbreker:
Add new option to the CLI digitalobject:load CLI task: --attach-only

When set, this option will force all digital objects in the import CSV
file to be attached via their own information object record.


mcantelon (approved):
Looks good to me!

PR #928, opened 2019-07-18
Fix task plugin loading issue, refs #13109

sevein:
This commit ensures that the action of disabling plugin loading in
tools:upgrade-sql is performed in a way that does not affect other
tasks.

tools:upgrade-sql disables application plugin loading by setting the
$loadPlugins global to false. However, that change was affecting all
tasks because it's performed in the configure method which is always
executed during the CLI bootstrap process, for every task, regardless
the task that the user is running.

A side effect was that jobs:worker was not able to access application
plugins like qtSwordPlugin unless the autoloading config had already
been cached by hitting the website first.

For QA:
1. Clear the cache
2. Do not visit the web application site
3. Start the atom worker
4. Confirm that the qtSwordPlugin ability is loaded

### DIFF
--- a/lib/task/migrate/arUpgradeSqlTask.class.php
+++ b/lib/task/migrate/arUpgradeSqlTask.class.php
@@ -57,7 +57,15 @@ EOF;
     ));
 
     // Disable plugin loading from plugins/ before this task.
-    sfPluginAdminPluginConfiguration::$loadPlugins = false;
+    // Using command.pre_command to ensure that it happens early enough.
+    $this->dispatcher->connect('command.pre_command', function($e) {
+      if (!$e->getSubject() instanceof self)
+      {
+        return;
+      }
+
+      sfPluginAdminPluginConfiguration::$loadPlugins = false;
+    });
   }
 
   /**


jraddaoui (commented):
Wow! Nice one @sevein!

It's great to know that all configure methods are called when any task is executed. I wonder if that's done to have a global `help` message with all tasks or something like that.

The fix looks great and it seems like a long mystery solved! However, I'm having troubles to understand why was it working after hitting the site (before the fix). So, when the project config is cached after a web request, setting `$loadPlugins` to false in `cli` context does nothing?

I also wonder if we could fix it by setting the `$loadPlugins` global to false in the `execute` instead of in the `configure` method, and why do we need to ensure that it happens early enough now.

Thanks!!

jraddaoui:
Also, the length of the commit message lines are a little over the limit, the subject line and some in the body ;)

sevein:
I'm glad you're asking because I don't think I fully understood what was going on but revisiting it today is helping.

> It's great to know that all configure methods are called when any task is executed. I wonder if that's done to have a global `help` message with all tasks or something like that.

I think so. The [`registerTasks`](https://github.com/artefactual/atom/blob/a954dcc96500ce90f618e873a420f7e6e008ed42/vendor/symfony/lib/command/sfCommandApplication.class.php#L113-L131) method in `sfCommandApplication` instantiates our tasks (we extend `sfTask`) during the bootstrap process. The [sfTask constructor calls its `configure` method](https://github.com/artefactual/atom/blob/a954dcc96500ce90f618e873a420f7e6e008ed42/vendor/symfony/lib/task/sfTask.class.php#L32-L43).

> The fix looks great and it seems like a long mystery solved! However, I'm having troubles to understand why was it working after hitting the site (before the fix). So, when the project config is cached after a web request, setting `$loadPlugins` to false in `cli` context does nothing?

Pretty much. `$loadPlugins` will still be evaluated on each task run but at that stage the application has already loaded what's listed in [`cache/project_autoload.cache`](https://gist.github.com/sevein/f6059145b1e8d930eb0dcd3604fc1e06) so it's kind of pointless no wanting to load class X is it's been loaded already - unless it's accretion.

E.g. jobs:worker wasn't loading plugins such qtSwordPlugin (as a result of `tools:upgrade-sql` accidentally leaking that change), so we had two possible outcomes when running the worker task:

- With a clean cache state, qtSwordPlugin was never loaded.
- With the cache populated by the web frontend, qtSwordPlugin was loaded.

> I also wonder if we could fix it by setting the `$loadPlugins` global to false in the `execute` instead of in the `configure` method, and why do we need to ensure that it happens early enough now.

That didn't seem to work. At that point the application bootstrap process has already completed, cache has been leveraged and sfPluginAdminPlugin has been initialized (by config/ProjectConfiguration.class.php).

I guess there's one risk with this PR. Other tasks might break if they were implicitly relying on not having the applications plugins loaded, for example `tools:migrate-sql` wanted sfPluginAdminPlugin disabled because looking up the settings table isn't possible when you don't have a database. I wonder if there are other tasks meant to be executed at similar stages, you probably know better than me!

> Also, the length of the commit message lines are a little over the limit, the subject line and some in the body ;)

Fixed. Subject = 50 cols, Body = 72 cols, right?

jraddaoui (approved):
Amazing @sevein! Thanks a lot for the fix and the clarifications!

I've tested locally using the `atom-bionic` playbook from `deploy-pub` and, when the plugin is enabled/disabled through the GUI, restarting the worker still works like a charm and enables/disables the correspondent ability. In that case, we clear the application cache [in here](https://github.com/artefactual/atom/blob/qa/2.6.x/plugins/sfPluginAdminPlugin/modules/sfPluginAdminPlugin/actions/pluginsAction.class.php#L111-L113). However, if the plugin is enabled/disabled directly in the database or using a CLI task, it's important to note that, if the auto-load is already cached, the cache must be cleared manually before restarting the worker.

I'll add that note in the Redmine ticket, alongside some changes that may be needed in the Ansible role for the service.

jraddaoui:
Some edits done in the latest comment (if you're checking this through email ;).

And a side note, I'd move the QA section to the Redmine ticket. And maybe we should test both cases, with and without the cache active; and change point 4 to confirm that the ability is defined or not in the worker based on the plugin status (enabled/disabled).

Again, thanks a lot for finding and fixing the mystery!!

jraddaoui:
555edf8 in stable/2.5.x

PR #929, opened 2019-07-20
Add warning to security settings, refs #13105

mcantelon:
Added a warning to the security settings page to let users know that
incorrect security settings can result in the web UI becoming
inaccessible.


djjuhasz (approved):
Looks good to me @mcantelon ! 👍

PR #936, opened 2019-07-27
Add warning when user toggles SWORD, refs #13117

mcantelon:
Added a warning, on the plugin admin page, so that when an admin
enables or disables the SWORD plugin they are told to also restart
the AtoM worker.

### DIFF
--- a/plugins/sfPluginAdminPlugin/modules/sfPluginAdminPlugin/actions/pluginsAction.class.php
+++ b/plugins/sfPluginAdminPlugin/modules/sfPluginAdminPlugin/actions/pluginsAction.class.php
@@ -82,6 +82,8 @@ class sfPluginAdminPluginPluginsAction extends sfAction
 
         $settings = unserialize($setting->getValue(array('sourceCulture' => true)));
 
+        $swordEnabled = in_array('qtSwordPlugin', $settings);
+
         foreach (array_keys($this->plugins) as $item)
         {
           if (in_array($item, (array)$this->form->getValue('enabled')))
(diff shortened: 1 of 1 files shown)


jraddaoui (approved):
Nice one!

plugins/sfPluginAdminPlugin/modules/sfPluginAdminPlugin/actions/pluginsAction.class.php
```diff
@@ -112,6 +114,13 @@ public function execute($request)
         $cacheClear = new sfCacheClearTask(sfContext::getInstance()->getEventDispatcher(), new sfFormatter());
         $cacheClear->run();
 
+        // Notify use if SWORD setting has been changed
+        if ($swordEnabled != in_array('qtSwordPlugin', $settings))
+        {
+          $message = $this->context->i18n->__('SWORD plugin setting changed: the AtoM worker must be restarted for this change to take effect.');
+          $this->getUser()->setFlash('error', $message);
```
jraddaoui: Maybe `info` (not sure if that's the actual term used in this flash system) or `warning` are better than `error`. I'd go with `info` in this case ;)

jraddaoui (changes requested):
Actually, a minor detail ...

plugins/sfPluginAdminPlugin/modules/sfPluginAdminPlugin/actions/pluginsAction.class.php
mcantelon: Thanks Radda! Makes sense... I'll change that.

PR #938, opened 2019-07-29
Fix finding aid custom L.O.D. issues, refs #12707

mcantelon:
Fixed issue with finding aid generation where the name of custom
levels of description wouldn't be displayed.

### DIFF
--- a/lib/task/pdf/ead-pdf-inventory-summary.xsl
+++ b/lib/task/pdf/ead-pdf-inventory-summary.xsl
@@ -1451,7 +1451,7 @@
             <xsl:when test="@level='subcollection' or @level='subgrp' or @level='series'
                 or @level='subseries' or @level='collection'or @level='fonds' or
                 @level='recordgrp' or @level='subfonds' or @level='class' or
-                (@level='otherlevel' and not(child::ead:did/ead:container))">
+                (@level='otherlevel' and not(parent::ead:c[@level='series']))">
 
                 <fo:table-row background-color="#ffffff" border-bottom="1pt solid #000"
                               border-top="1pt solid #000"  text-align="left">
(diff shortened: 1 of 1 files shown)


djjuhasz (changes requested):
Please see https://projects.artefactual.com/issues/12707 for feedback on this change.

mcantelon:
I've updated this PR with, hopefully, a fix.

djjuhasz:
@mcantelon the otherlevel value is correctly being displayed on the finding aid now, as well as the "level" values being rendered properly, so that's good. :)

There is still an issue with the display of "otherlevel" descriptions in the finding aid template though which I think will require some further discussion, as the "best" solution is not clear to me.  In the standard finding aid template file and item level descriptions are displayed as a simple table, e.g.
issue_12707_finding_aid_screenshot_2

The "otherlevel" descriptions though are being displayed more like a "Series" description, e.g. 
issue_12707_finding_aid_screenshot_1

In the case of a "box" I would expect the descriptions to be displayed in a table like the file/item table, but this is not necessarily the best solution for all "otherlevel" values. :-/

I can think of a few possible options for how to display "otherlevel" descriptions in the finding aid:

1. All custom levels of description (LODs) are displayed in a table like the file/item list
2. The display of the custom LODs is determined by their position in the descriptive hierarchy, e.g. custom LODs that are a child of a "Series" description are displayed in a table, but custom LODs that are children of a Fonds/collection are displayed like a "Series"
3. We add a UI to configure how custom LODs are displayed in a finding aid

Of the three options I prefer option 2 as the best compromise between work required and functionality.  I believe Option 3 would require sponsorship as it requires significantly more development work than the first two options.

@fiver-watson do you have an opinion on how to display the "otherlevel" custom LODs?

fiver-watson:
I had thought we shelved this, but: of @djjuhasz 's 3 options, I agree that Option 3 would require sponsorship, and that Option 2 is the best compromise available to us. There will be no perfect implementation from the sounds of it - we can't know what levels a user may choose to create or how they will be used, but trying to add some logic based on hierarchy position should at least help address some common use cases.

mcantelon:
I've updated the PR to implement option two. Let me know if further tweaking is needed.

mcantelon:
(Note to self: update commit message before merging.)

djjuhasz (changes requested):
Nope, still not working:
Finding-aid-box-list-2019-09-19

I think your XPath selector is wrong, a series is not normally an "archdesc" element (that's usually the Fonds or Collection).  A series is usually a c-level element (e.g. c, c01, c02).

mcantelon:
Ah, true... nice catch... should be an easy fix.

mcantelon:
I've updated to fix.

djjuhasz (approved):
Looks good @mcantelon!
finding_aid_boxes_c544bda

PR #939, opened 2019-07-29
Escape entities in hidden inputs, refs #13125

jraddaoui:
When Markdown support is enabled, the escaping strategy gets disabled.
This may cause issues in some hidden fields where the user input is
added without validation. The keys added to those inputs are already
validated in the controllers, but the values need to be escaped.


djjuhasz (approved):
Looks good @jraddaoui 👍

jraddaoui:
Merged in a50bc36.

jraddaoui:
c823af0 in stable/2.5.x.

PR #940, opened 2019-07-30
New digitalobject load task option, refs #13126

sbreker:
Add a new option to the digital object load task to disable nested
set value updates on import.

### DIFF
--- a/lib/task/digitalobject/digitalObjectLoadTask.class.php
+++ b/lib/task/digitalobject/digitalObjectLoadTask.class.php
@@ -32,6 +32,7 @@ class digitalObjectLoadTask extends sfBaseTask
   private $curObjNum = 0;
   private $totalObjCount = 0;
   private $skippedCount = 0;
+  private $disableNestedSetUpdating = false;
 
   /**
    * @see sfTask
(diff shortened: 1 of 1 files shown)


lib/task/digitalobject/digitalObjectLoadTask.class.php
```diff
@@ -267,6 +271,12 @@ protected function attachDigitalObject($item, $informationObjectId)
     $informationObject = new QubitInformationObject;
     $informationObject->parent = QubitInformationObject::getById($informationObjectId);
     $informationObject->title = basename($item);
+
+    if (property_exists(get_class($informationObject), 'disableNestedSetUpdating'))
```
jraddaoui: Why do we need to check the property existence?

jraddaoui (approved):
Looks great @sbreker! Just a minor comment ;)

lib/task/digitalobject/digitalObjectLoadTask.class.php
sbreker: We don't I suppose now that I look at it again.  I can remove this.

PR #943, opened 2019-08-01
Fix options digitalobject load task, refs #13103

sbreker:
This fixes an issue with the digitalobject:load task where some options
were not being respected.


mcantelon:
Looks good to me! 👍

mcantelon (approved):

PR #944, opened 2019-08-03
Round trip mode for CSV import, refs #12281

mcantelon:
Added round trip mode to CSV information object import. Round trip mode
treats the values in the legacyID column as internal AtoM IDs and
doesn't create keymap entries.


sbreker (approved):
Looks great @mcantelon!

PR #946, opened 2019-08-06
Get raw parameters in filter tags, refs #13131

jraddaoui:
When the escaping strategy is enabled and Markdown support is disabled,
`getParams` is converted to a `sfOutputEscaperArrayDecorator` instance.
This needs to be converted back using `$sf_data->getRaw()` to be able
to merge it with another array in the template.

### DIFF
--- a/apps/qubit/modules/search/templates/_filterTag.php
+++ b/apps/qubit/modules/search/templates/_filterTag.php
@@ -4,5 +4,5 @@
   <?php else: ?>
     <?php echo render_title($object) ?>
   <?php endif; ?>
-  <a href="<?php echo url_for(array('module' => $module, 'action' => $action) + $getParams) ?>" class="remove-filter"><i class="fa fa-times"></i></a>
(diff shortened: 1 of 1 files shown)


djjuhasz (commented):
@jraddaoui doesn't using `$sf_data->getRaw('getParams')` mean that any user data sent in a GET will be included in the resulting URL, without any sanitization?  Maybe we need to manually do HTML escaping on the array values before outputting the URL?

jraddaoui:
Great point @djjuhasz!

We have to be really careful after un-escaping and in any case, as the escaping strategy may be disabled to enable Markdown. I couldn't find where in the code so I tested it and it looks like `url_for` is taking care of the escaping.

mcantelon (approved):
Looks good to me!

djjuhasz (approved):
Agreed 👍

jraddaoui:
c1e9b3c in stable/2.5.x

PR #947, opened 2019-08-07
Fix arDominionPlugin theme header, refs #13137

jraddaoui:
Make sure the content of the site header is always wrapped within it.
Use a pseudo element and clearance to avoid disrupting themes extending
this one. This also makes the header play nicer with logos higher than
50px and long titles, fixing #3895 and #6293 too.


mcantelon (approved):
Looks good to me!

jraddaoui:
Thanks @mcantelon!

PR #950, opened 2019-08-12
Fix creation of DOs without path, refs #13144

jraddaoui:
This issue came out now probably because of the upgrade to MySQL 5.7
and/or the re-addition of `STRICT_TRANS_TABLES` to the `sql_mode`.


mcantelon (approved):
Looks good to me!

PR #951, opened 2019-08-12
DIP upload fixes, refs #13067 and #13068

jraddaoui:
**Get namespaces from METS file when possible:**

Use `getDocNamespaces` to get the namespaces declared in the METS file
dynamically from the document. To keep backwards compatibility, add the
old URI for main namespaces if they're not found in the document.

Use a key/name mapping to add those namespaces in XPath queries without
modifying the existing ones.

**Allow different namespaces in DC metadata:**

Allow no namespace and `dc` or `dcterms` namespaces.

**Fix main DMD section finding:**

Use always the physical (default) structMap.

**Fix all DMD section finding:**

The DMDID attribute can contain one or more DMD section ids
(e.g.: DMDID="dmdSec_2 dmdSec_3"). When multiple DMD sections
are associated with the same file/dir we'll try to return the
latest one created.

### DIFF
--- a/lib/QubitMetsParser.class.php
+++ b/lib/QubitMetsParser.class.php
@@ -26,10 +26,27 @@ class QubitMetsParser
     // Load document
     $this->document = $document;
 
-    // Register namespaces
-    $this->document->registerXPathNamespace('m', 'http://www.loc.gov/METS/');
-    $this->document->registerXPathNamespace('p', 'info:lc/xmlns/premis-v2');
-    $this->document->registerXPathNamespace('f', 'http://hul.harvard.edu/ois/xml/ns/fits/fits_output');
+    // Get declared namespaces in the document
+    $this->namespaces = $this->document->getDocNamespaces(true);
+
+    // For backwards compatibility, add default namespaces as they
+    // were declared without name in the METS file (fits still is).
+    $defaultNamespaces = array(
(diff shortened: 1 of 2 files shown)


replaceafill (approved):
The code looks great to me @jraddaoui! :+1: 

I tested it with a `DemoTransferCSV` transfer and each object appears in AtoM with its correct metadata. I also arranged a SIP from the backlog with multiple directories, files and levels of description and they showed correctly too. However, I couldn't figure out a way to test a DMDID with multiple values. Can you recommend a way to do that?

Also, out of curiosity, in https://projects.artefactual.com/issues/13067#note-10 you mention:

> it was not creating any archival description before the fix. It should create descriptions now

Could you clarify how can that be tested? Thanks!

jraddaoui:
Thanks @replaceafill!

I appreciate that you gave it a try too. We don't normally do that as this will be tested by the analysts before the release (I did it yesterday on your AM PR because I needed those changes to finalize this work :).

To answer your questions ...

> I couldn't figure out a way to test a DMDID with multiple values. Can you recommend a way to do that?

That requires a partial re-ingest in AM, entering DC metadata on both ingest process.

> Could you clarify how can that be tested?

You already did, no description was being created when the upload was performed from a SIP created in the arrange panel.

PR #952, opened 2019-08-12
Fix DO metadata component, refs #13145

jraddaoui:
When Markdown is disabled the escaping strategy converts
`$resource->object` to an instance of `sfOutputEscaperObjectDecorator`,
which fails to match `QubitInformationObject` when it's compared in the
template. Check the related resource type in the component action
before the transformation happens.


mcantelon (approved):
Looks good to me!

jraddaoui:
Thanks @mcantelon!

PR #953, opened 2019-08-13
Render repo description as Parsedown, refs #13106

mcantelon:
Render the repo description HTML field as Parsedown rather than just
HTML.


jraddaoui (approved):
:+1:

PR #958, opened 2019-08-20
Escape entities, refs #13125

jraddaoui:
- The start and end dates from the adv. search were used to form the
  filter tag label directly from user input.
- The inline-search query was not escaped, this will normally raise an
  ES error before the template is rendered, but it may be an issue if
  the "Escape special chars from searches" setting avoids that error.


mcantelon:
Looks good to me!

jraddaoui:
Thanks @mcantelon!

jraddaoui:
f55a9ee in stable/2.5.x

PR #959, opened 2019-08-20
Update keyboard handling events in JS, refs #12725

mcantelon:
Updates keyboard handling events to remove use of the deprecated
"keypress" event and "keyCode" event properties replacing these,
respectively, with "keydown" events and "which" event properties.
jQuery populates the "which" event property even in browsers that
don't support "which" natively.


sbreker (approved):
Code review complete - looks good! Looks like a good upgrade to this code.

PR #961, opened 2019-08-22
Add authority record CSV export task, refs #13148

mcantelon:
Add task to export authority record data as CSV file(s). If exporting
into a directory then aliases and relations can also be exported.

### DIFF
--- a/lib/task/export/exportAuthorityRecordsTask.class.php
+++ b/lib/task/export/exportAuthorityRecordsTask.class.php
@@ -1,0 +1,84 @@
+<?php
+
+/*
+ * This file is part of the Access to Memory (AtoM) software.
+ *
+ * Access to Memory (AtoM) is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * Access to Memory (AtoM) is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with Access to Memory (AtoM).  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+/**
+ * Export authority records to a CSV file
+ *
+ * @package    symfony
+ * @subpackage task
+ * @author     Mike Cantelon <mike@artefactual.com>
+ */
+class exportAuthorityRecordsTask extends exportBulkBaseTask
+{
+    protected $namespace        = 'csv';
+    protected $name             = 'authority-export';
+    protected $briefDescription = 'Export authority record data as CSV file(s)';
+
+    protected $detailedDescription = <<<EOF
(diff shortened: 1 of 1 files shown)


lib/task/export/exportAuthorityRecordsTask.class.php
```diff
@@ -0,0 +1,99 @@
...
+ */
+
+/**
+ * Export authority records to a CSV file
+ *
+ * @package    symfony
+ * @subpackage task
+ * @author     Mike Cantelon <mike@artefactual.com>
+ */
+class exportAuthorityRecordsTask extends arBaseTask
```
sbreker: Should be using exportBulkBaseTask?

mcantelon: Good point! I've reworked it to inherit form exportBulkBaseTask and simplified the task.

mcantelon:
@sbreker Hi Steve... I've reworked this PR, simplifying it and inheriting from exportBulkBaseTask.

lib/task/export/exportAuthorityRecordsTask.class.php
```diff
@@ -0,0 +1,84 @@
...
+    $this->log('');
+    $this->logSection('csv', "Export complete ($itemsExported authority records exported).");
+  }
+
+  private function getActors()
+  {
+    $sql = "SELECT ai.id, ai.culture FROM actor_i18n ai INNER JOIN object o ON ai.id=o.id \r
+            WHERE o.class_name='QubitActor' AND ai.id <> ?";
+
+    return QubitPdo::fetchAll($sql, array(QubitActor::ROOT_ID));
```
sbreker: Could fetching as associative array work? I think it would use less memory than row objects. e.g.:

QubitPdo::fetchAll($sql, array(QubitActor::ROOT_ID), array('fetchMode' => PDO::FETCH_ASSOC));

lib/task/export/exportAuthorityRecordsTask.class.php
```diff
@@ -0,0 +1,84 @@
...
+      $itemsExported++;
+    }
+
+    $this->log('');
+    $this->logSection('csv', "Export complete ($itemsExported authority records exported).");
+  }
+
+  private function getActors()
+  {
+    $sql = "SELECT ai.id, ai.culture FROM actor_i18n ai INNER JOIN object o ON ai.id=o.id \r
```
sbreker: What does the \r do here in this case? Is it just for mysql query log formatting?

sbreker (changes requested):
I think that query should be made to return an associative array to help reduce mem usage as much as possible since it is selecting a potentially large number of rows.

lib/task/export/exportAuthorityRecordsTask.class.php
mcantelon: Good thinking. I've made that change.

lib/task/export/exportAuthorityRecordsTask.class.php
mcantelon: It's just for log formatting, but not really needed given it's a relatively simple query so I've removed it.

mcantelon:
Hi @sbreker ... I've updated the PR (I'll rebase it before merging).

sbreker (approved):
Looks great @mcantelon! Thanks for making those changes! 👍

PR #962, opened 2019-08-23
Accessions import enhancements, refs# 12799

sbreker:
This commit adds the ability to import a new CSV column donorNotes which
will import to the accession donor notes field which already exists.

Secondly, the import task has been changed such that if a new
accession material type term or acquisition type term is encountered in
the import CSV, the task will now create the new term and log a message
instead of throwing an exception and stopping the import.

I have also updated the log message that is created when a duplicate
accession identifier is found to include the accession identifier.


mcantelon (approved):
Looks good to me!

jraddaoui:
Minor typo, refs #12799. ;)

sbreker:
Will fix typo! :)

PR #964, opened 2019-09-04
CSV translation import: update index, refs #8572

mcantelon:
Added search index updating during CSV import of information object
translations.

### DIFF
--- a/lib/task/import/csvImportTask.class.php
+++ b/lib/task/import/csvImportTask.class.php
@@ -1056,6 +1056,13 @@ EOF;
         {
           $this->handleDigitalObjectPath($self, $path);
         }
+
+        // If importing a translation, re-save entire information object so translation will be indexed
+        if ($self->object instanceof QubitInformationObjectI18n && !$self->searchIndexingDisabled)
+        {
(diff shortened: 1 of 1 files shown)


lib/task/import/csvImportTask.class.php
```diff
@@ -1056,6 +1056,13 @@ public function execute($arguments = array(), $options = array())
         {
           $this->handleDigitalObjectPath($self, $path);
         }
+
+        // If importing a translation, re-save entire information object so translation will be indexed
+        if ($self->object instanceof QubitInformationObjectI18n && !$search->searchIndexingDisabled)
```
sbreker: Does this respect the --index flag? I can't see where $search is getting set.  

The index flag is assigned to the $import object here:
https://github.com/artefactual/atom/blob/qa/2.6.x/lib/task/import/csvImportTask.class.php#L1062

mcantelon: Oops... nice catch! Fixed!

mcantelon:
Updated if you want to take another look... thanks @sbreker !

sbreker (approved):
Looks good @mcantelon!  Thanks for making that change!

PR #965, opened 2019-09-07
Fix OAI base URL in identify XML, refs #13163

mcantelon:
When AtoM's base URL was set to an "https" URL the identify response
was changing this to an "http" URL. Fixed and simplified.

### DIFF
--- a/plugins/arOaiPlugin/modules/arOaiPlugin/actions/identifyComponent.class.php
+++ b/plugins/arOaiPlugin/modules/arOaiPlugin/actions/identifyComponent.class.php
@@ -35,6 +35,7 @@ class arOaiPluginIdentifyComponent extends arOaiPluginComponent
     list ($this->earliestDatestamp) = Propel::getConnection()->query('SELECT MIN('.QubitObject::UPDATED_AT.') FROM '.QubitObject::TABLE_NAME)->fetch();
     $this->earliestDatestamp = date_format(date_create($this->earliestDatestamp), 'Y-m-d\TH:i:s\Z');
 
+    $this->baseUrl = QubitSetting::getByName('siteBaseUrl')->getValue(array('sourceCulture' => true));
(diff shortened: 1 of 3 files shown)


plugins/arOaiPlugin/modules/arOaiPlugin/actions/identifyComponent.class.php
```diff
@@ -35,6 +35,7 @@ public function execute($request)
     list ($this->earliestDatestamp) = Propel::getConnection()->query('SELECT MIN('.QubitObject::UPDATED_AT.') FROM '.QubitObject::TABLE_NAME)->fetch();
     $this->earliestDatestamp = date_format(date_create($this->earliestDatestamp), 'Y-m-d\TH:i:s\Z');
 
+    $this->baseUrl = QubitSetting::getByName('siteBaseUrl')->getValue(array('cultureFallback' => true));
```
djjuhasz: @mcantelon Are you sure you want the translated version of "siteBaseUrl" here (with fallback)?  I would expect:
`$this->baseUrl = QubitSetting::getByName('siteBaseUrl')->getValue(array('sourceCulture' => true));`

djjuhasz (commented):
@mcantelon one question about the code to get the base URL setting value.

plugins/arOaiPlugin/modules/arOaiPlugin/actions/identifyComponent.class.php
mcantelon: Ah, copied that from the existing code and didn't notice it. I'll fix that.

mcantelon: Fixed!

djjuhasz (approved):
Looks good to me 👍

PR #966, opened 2019-09-10
Update storage perms, refs #13169

fiver-watson:
In the 2.4 release, we discovered some inconsistencies in the default permissions, and that authenticated users who were not part of any group could access modules they should not be able to view. We corrected many of these in issue #[11075](https://projects.artefactual.com/issues/11075).

However, in that fix, we implemented too extreme a solution, allowing only administrators to view the physical storage module. There are many cases where Editors may need access, and local translators may also need to add translated container names.

It turns out that the default permissions are contained in a configurable YAML file, located here:

* https://github.com/artefactual/atom/blob/HEAD/apps/qubit/modules/physicalobject/config/security.yml

This fix will change the default permissions so that:

* Administrators, Contributors, Editors, and Translators can view and browse physical storage
* Administrators, Editors, and Translators, can edit physical storage containers
* Administrators and Editors can delete physical storage containers

For any users who wish to restrict these permissions further, they can make local edits to the YAML file listed above.

### DIFF
--- a/apps/qubit/modules/physicalobject/config/security.yml
+++ b/apps/qubit/modules/physicalobject/config/security.yml
@@ -1,13 +1,20 @@
-browse:
-  credentials: administrator
-
-delete:
-  credentials: [[ contributor, editor, administrator ]]
-  is_secure: true
-
-edit:
-  credentials: [[ contributor, editor, administrator, translator ]]
-  is_secure: true
-
 all:
   is_secure: true
+
+autocomplete:
+  credentials: [[ administrator, contributor, editor ]]
+
+boxList:
+  credentials: [[ administrator, editor ]]
+
+browse:
+  credentials: [[ administrator, editor, translator ]]
+
+delete:
+  credentials: [[ administrator, editor ]]
+
+edit:
+  credentials: [[ administrator, editor, translator ]]
+
+index:
+  credentials: [[ administrator, contributor, editor, translator ]]


apps/qubit/modules/physicalobject/config/security.yml
```diff
@@ -1,12 +1,17 @@
+index:
+  credentials: [[ editor, administrator, translator ]]
+  is_secure: true
```
djjuhasz, line 2: It's not necessary to specify `is_secure: true` for each action, it's sufficient that it's specified for the "all:" permissions block.  I would also move the "all:" block to the top of the security list, just for legibility, but it's not necessary.

djjuhasz (changes requested):

apps/qubit/modules/physicalobject/config/security.yml
djjuhasz: Ref: https://symfony.com/legacy/doc/reference/1_4/en/08-Security

apps/qubit/modules/physicalobject/config/security.yml
```diff
@@ -1,13 +1,20 @@
...
-  credentials: [[ contributor, editor, administrator ]]
-  is_secure: true
+delete: 
+   credentials: [[ administrator, editor ]]
 
-edit:
-  credentials: [[ contributor, editor, administrator, translator ]]
-  is_secure: true
+edit: 
+   credentials: [[administrator, editor, translator ]]
```
djjuhasz: Missing a space before "administrator".

apps/qubit/modules/physicalobject/config/security.yml
```diff
@@ -1,13 +1,20 @@
...
 
-all:
-  is_secure: true
+index:
+   credentials: [[administrator, editor, translator ]]
```
djjuhasz: Missing a space before "administrator".

apps/qubit/modules/physicalobject/config/security.yml
```diff
@@ -1,13 +1,20 @@
...
+
+autocomplete: 
+   credentials: [[ administrator, contributor, editor ]]
+
+boxList:
+   credentials: [[ administrator, editor ]]
+
 browse:
-  credentials: administrator
+   credentials: [[ administrator, editor, translator ]]
```
djjuhasz: Indent should be 2 spaces only.  Three spaces is used here and for every subsequent "credentials" line.

djjuhasz (changes requested):
As a contributor I can create a new physical storage location via "Archival description > Link physical storage" and "Create new container" (uses physicalobject/autocomplete permission).  After creating the physical object I see a link to it from the archival description view page, but when I click the link I get a "permission denied" page:

physical_storage_acl_contributor

I think we should grant at least "index" (view) permissions to the contributor to allow viewing existing physical objects.

ALSO, as a contributor I see the "Manage > Physical storage" menu, but if I click the menu item I get a permission denied page.  I think we should either allow contributors to browse physical storage, or hide the menu item from them.

apps/qubit/modules/physicalobject/config/security.yml
```diff
@@ -1,13 +1,20 @@
+all:
+  is_secure: true
+
+autocomplete:
```
djjuhasz: Trailing whitespace

apps/qubit/modules/physicalobject/config/security.yml
```diff
@@ -1,13 +1,20 @@
...
+   credentials: [[ administrator, editor ]]
+
 browse:
-  credentials: administrator
+   credentials: [[ administrator, editor, translator ]]
 
-delete:
-  credentials: [[ contributor, editor, administrator ]]
-  is_secure: true
+delete:
```
djjuhasz: Trailing whitespace

apps/qubit/modules/physicalobject/config/security.yml
```diff
@@ -1,13 +1,20 @@
...
-delete:
-  credentials: [[ contributor, editor, administrator ]]
-  is_secure: true
+delete: 
+   credentials: [[ administrator, editor ]]
 
-edit:
-  credentials: [[ contributor, editor, administrator, translator ]]
-  is_secure: true
+edit:
```
djjuhasz: Trailing whitespace

fiver-watson:
Hi! Ok, the ability for contribs to link or add locations via the descriptions, but not browse the module, was intentional, but I'm open to allowing them browse permissions. I'm not sure what permission controls whether or not the option appears in the Manage menu in any case. I'll give it another go.

djjuhasz (approved):
Looks good!

fiver-watson:
Merged into qa/2.6.x in https://github.com/artefactual/atom/commit/378a77feb80ea9040ad4057fe1388dc6d552a7b3

PR #967, opened 2019-09-12
Truncate stored job output to 64K, refs #13174

mcantelon:
Added logic to limit stored job output to 64K.

### DIFF
--- a/lib/job/arJobLogger.class.php
+++ b/lib/job/arJobLogger.class.php
@@ -47,8 +47,9 @@ class arJobLogger extends sfLogger
       $message);
 
     // TEXT type cannot have a default (i.e. ''), so use CONCAT_WS because it can work with null values
-    // and coerce them into a string with an empty string separater.
-    $sql = 'UPDATE job SET output = CONCAT_WS("", output, ?, "\n") WHERE id = ?';
+    // and coerce them into a string with an empty string separater. Truncate the result of the concat
+    // so it fits in column limit.
+    $sql = 'UPDATE job SET output = SUBSTR(CONCAT_WS("", output, ?, "\n"), 1, 65534) WHERE id = ?';
     QubitPdo::prepareAndExecute($sql, array($fMessage, $this->job->id));
 
     // Forward to `gearman.worker.log` observers, jobWorkerTask will log to console via sfTask.


mcantelon:
(Edge case: 50% it'll leave a � at the end if truncating Unicode, but likely not a significant concern.)

lib/job/arJobLogger.class.php
```diff
@@ -46,6 +46,9 @@ protected function doLog($message, $priority)
       $this->job->name,
       $message);
 
+    // Truncate message to 64K (minus one byte)
+    $fMessage = substr($fMessage, 0, 65534);
```
jraddaoui: Hi @mcantelon  and @djjuhasz,

This is being concatenated bellow to the existing output, so we should check the full result length. Nevertheless, I'm worried about this happening in other columns over the time and I think we should find a proper solution for the long term.

I think the errors were introduced by the MySQL upgrade to 5.7 and that they could be solved in the same manner it worked in 5.6 by removing `STRICT_TRANS_TABLES` from the `sql_mode` ([current docs](https://www.accesstomemory.org/en/docs/2.5/admin-manual/installation/linux/ubuntu-bionic/#create-the-database)).

jraddaoui (changes requested):
Please, see my comment below ...

lib/job/arJobLogger.class.php
djjuhasz: @jraddaoui I would like to move AtoM towards running properly with `STRICT_TRANS_TABLES` enabled - this is the default from MySQL 5.7 on, and if we can avoid changing the default configuration it makes installation easier.   I also think using the STRICT behaviour is better in the long run for data integrity.

I think that we should evaluate each case of data overflow in a column on an individual basis.  The "non-strict" behaviour of MySQL is to silently truncate data that won't fit in a column, but this is not the best choice in all cases.  I think the motivation for the strict behaviour is that silently truncating data can be a bug rather than a feature.   For instance MySQL *has* been truncating job output all along, we just weren't aware of it (at least I wasn't).

mcantelon: Good catch @jraddaoui ... I've updated the PR to incorporate the job ID in the truncation.

jraddaoui: It's not the id but the current output content what is concatenated to the new message in the SQL statement. I'm not sure if `$this->job->output` will be always updated or you'll need to get the value from the database to check its current length.

jraddaoui (approved):
Awesome! Thanks @mcantelon!

PR #972, opened 2019-09-21
Remove identifier Markdown parsing, refs #13173

mcantelon:
Refactored display of title, etc., in a number of description standard plugins
and got rid of Markdown parsing of identifiers.

* Created a component to display level of description, reference code, title,
  and publication status
* In the component the reference code doesn't get parsed as Markdown, but the
  title does
* Refactored description standard plugins to use component
* Removed Markdown parsing for misc. identifiers in index/edit page templates


jraddaoui (approved):
:+1:

PR #973, opened 2019-09-21
Add creator inheritance to DC XML, refs #13130

mcantelon:
Changed the DC XML component so creators will be inherited from ancestors.

### DIFF
--- a/plugins/sfDcPlugin/lib/sfDcPlugin.class.php
+++ b/plugins/sfDcPlugin/lib/sfDcPlugin.class.php
@@ -76,6 +76,9 @@ class sfDcPlugin implements ArrayAccess
   {
     switch ($name)
     {
+      case 'creators':
+        return $this->resource->getActors(array('eventTypeId' => QubitTerm::CREATION_ID, 'inherit' => true));
+
       case '_event':
 
         // Because simple Dublin Core cannot qualify the <date/> or <coverage/>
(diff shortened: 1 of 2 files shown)


plugins/sfDcPlugin/modules/sfDcPlugin/actions/dcComponent.class.php
```diff
@@ -24,5 +24,9 @@ public function execute($request)
     $this->iso639convertor = new fbISO639_Map;
 
     $this->dc = new sfDcPlugin($this->resource);
+
+    // Get creators, inheriting if needed
+    $options = array('eventTypeId' => QubitTerm::CREATION_ID, 'inherit' => true);
```
djjuhasz: I think it would be more consistent to put this logic in https://github.com/artefactual/atom/blob/qa/2.6.x/plugins/sfDcPlugin/lib/sfDcPlugin.class.php in the `__get()` magic accessor method.

djjuhasz (commented):
@mcantelon I just had one suggested change.

plugins/sfDcPlugin/modules/sfDcPlugin/actions/dcComponent.class.php
djjuhasz: Actually, one more thing - I don't think assigning the `$options` array is necessary or enhances readability here.  I would move the `array()` constructor into the `getActors()` call.

mcantelon:
Good recommendations. I've updated the PR.

djjuhasz (approved):
Looks good! 👍

PR #974, opened 2019-09-23
Related term ES index speed, refs #13175

sbreker:
This commit contains an enhancement to speed up ES indexing of
information objects when there is a large number of terms in the terms
table.

Subject, place and genre term ids and parent_ids are preloaded into an
array so a term's parent can be found directly using the term id as the
array index.

For each directly related term, parents are found by searching the term
array recursively instead of using the nested set ancestors query using
the LFT and RGT columns.

This change results in a large performance increase when indexing
information objects when testing with ~350K place and subject terms.

### DIFF
--- a/plugins/arElasticSearchPlugin/lib/model/arElasticSearchInformationObject.class.php
+++ b/plugins/arElasticSearchPlugin/lib/model/arElasticSearchInformationObject.class.php
@@ -22,9 +22,11 @@ class arElasticSearchInformationObject extends arElasticSearchModelBase
   protected static
     $conn,
     $statement,
-    $counter = 0;
+    $counter = 0,
+    $termParentList = array();
 
-  protected $errors = array();
+  protected
+    $errors = array();
 
   public function load()
   {
@@ -36,9 +38,24 @@ class arElasticSearchInformationObject extends arElasticSearchModelBase
     $this->count = QubitPdo::fetchColumn($sql, array(QubitInformationObject::ROOT_ID));
   }
 
+  static function loadTermParentList($options = array())
+  {
+    $sql  = 'SELECT term.id, term.parent_id';
+    $sql .= ' FROM '.QubitTerm::TABLE_NAME;
+
+    if (isset($options['taxonomyIds']) && is_array($options['taxonomyIds']))
+    {
+      $sql .= ' WHERE '.QubitTerm::TABLE_NAME.'.taxonomy_id IN ('.implode(',',$options['taxonomyIds']).')';
+    }
+    $rows = QubitPdo::fetchAll($sql, array() ,array('fetchMode' => PDO::FETCH_KEY_PAIR));
+
+    return $rows;
+  }
+
   public function populate()
   {
     $this->load();
+    self::$termParentList = self::loadTermParentList(array('taxonomyIds' => array(QubitTaxonomy::SUBJECT_ID, QubitTaxonomy::PLACE_ID, QubitTaxonomy::GENRE_ID)));
 
     // Recursively descend down hierarchy
     $this->recursivelyAddInformationObjects(QubitInformationObject::ROOT_ID, $this->count);
@@ -53,7 +70,7 @@ class arElasticSearchInformationObject extends arElasticSearchModelBase
     {
       try
       {
-        $node = new arElasticSearchInformationObjectPdo($item->id, $options);
+        $node = new arElasticSearchInformationObjectPdo($item->id, array_merge($options, array('terms' => self::$termParentList)));
         $data = $node->serialize();
 
         QubitSearch::getInstance()->addDocument($data, 'QubitInformationObject');
@@ -81,8 +98,26 @@ class arElasticSearchInformationObject extends arElasticSearchModelBase
 
   public static function update($object, $options = array())
   {
+    if (!isset(self::$conn))
+    {
+      self::$conn = Propel::getConnection();
+    }
+
+    $sql  = 'SELECT DISTINCT otr.term_id as id';
+    $sql .= ' FROM '.QubitObjectTermRelation::TABLE_NAME.' otr';
+    $sql .= ' WHERE otr.object_id = :id';
+
+    $rows = QubitPdo::fetchAll($sql, array(':id' => $object->id), array('fetchMode' => PDO::FETCH_ASSOC));
+
+    if (count($rows) > 0 && empty(self::$termParentList))
+    {
+      self::$termParentList = self::loadTermParentList(array('taxonomyIds' => array(QubitTaxonomy::SUBJECT_ID, QubitTaxonomy::PLACE_ID, QubitTaxonomy::GENRE_ID)));
+    }
+
     // Update description
-    $node = new arElasticSearchInformationObjectPdo($object->id);
+    $node = new arElasticSearchInformationObjectPdo($object->id,
+      array('terms' => self::$termParentList));
+
     QubitSearch::getInstance()->addDocument($node->serialize(), 'QubitInformationObject');
 
     // Update descendants if requested and they exists
--- a/plugins/arElasticSearchPlugin/lib/model/arElasticSearchInformationObjectPdo.class.php
+++ b/plugins/arElasticSearchPlugin/lib/model/arElasticSearchInformationObjectPdo.class.php
@@ -37,6 +37,7 @@ class arElasticSearchInformationObjectPdo
   protected
     $data = array(),
     $events,
+    $termParentList = array(),
     $languages = array(),
     $scripts = array();
 
@@ -63,6 +64,12 @@ class arElasticSearchInformationObjectPdo
     if (isset($options['ancestors']))
     {
       $this->ancestors = $options['ancestors'];
+    }
+
+    // Get system term id hierarchy
+    if (isset($options['terms']))
+    {
+      $this->termParentList = $options['terms'];
     }
 
     // Get inherited repository, unless a repository is set at current level
@@ -612,32 +619,77 @@ class arElasticSearchInformationObjectPdo
   }
 
   /*
-   * Get related terms and its ancestors
+   * Prepare term query.
+   */
+  protected function prepareRelatedTermsQuery()
+  {
+    if (!isset(self::$statements['relatedTerms']))
+    {
+      $sql  = 'SELECT
+                  DISTINCT current.id';
+      $sql .= ' FROM '.QubitObjectTermRelation::TABLE_NAME.' otr';
+      $sql .= ' JOIN '.QubitTerm::TABLE_NAME.' current
+                  ON otr.term_id = current.id';
+      $sql .= ' WHERE otr.object_id = :id
+                  AND current.taxonomy_id = :taxonomyId';
+
+      self::$statements['relatedTerms'] = self::$conn->prepare($sql);
+    }
+  }
+
+  /*
+   * Get related terms plus any ancestors
    */
   protected function getRelatedTerms($typeId)
   {
-    $sql  = 'SELECT
-                DISTINCT term.id,
-                term.taxonomy_id,
-                term.source_culture,
-                slug.slug,
-                i18n.name';
-    $sql .= ' FROM '.QubitObjectTermRelation::TABLE_NAME.' otr';
-    $sql .= ' JOIN '.QubitTerm::TABLE_NAME.' current
-                ON otr.term_id = current.id';
-    $sql .= ' JOIN '.QubitTerm::TABLE_NAME.' term
-                ON term.lft <= current.lft AND term.rgt >= current.rgt';
-    $sql .= ' JOIN '.QubitTermI18n::TABLE_NAME.' i18n
-                ON term.id = i18n.id';
-    $sql .= ' JOIN '.QubitSlug::TABLE_NAME.' slug
-                ON term.id = slug.object_id';
-    $sql .= ' WHERE otr.object_id = ?
-               AND term.taxonomy_id = ?';
+    $relatedTerms = array();
 
-    self::$statements['relatedTerms'] = self::$conn->prepare($sql);
-    self::$statements['relatedTerms']->execute(array($this->__get('id'), $typeId));
+    $this->prepareRelatedTermsQuery();
 
-    return self::$statements['relatedTerms']->fetchAll(PDO::FETCH_OBJ);
+    self::$statements['relatedTerms']->execute(array(':id' => $this->__get('id'), ':taxonomyId' => $typeId));
+
+    // Get directly related terms.
+    $rows = self::$statements['relatedTerms']->fetchAll(PDO::FETCH_ASSOC);
+
+    if (0 == count($rows))
+    {
+      return $relatedTerms;
+    }
+
+    // Iterate over each directly related term, adding all ancestors of each
+    foreach($rows as $row)
+    {
+      $relatedTerms = array_merge($relatedTerms, $this->recursivelyGetParentTerms($row['id']));
+    }
+
+    $relatedTerms = array_unique($relatedTerms);
+
+    return $relatedTerms;
+  }
+
+  /**
+   * Recursively find all parent terms for any given term. Subject, Place and
+   * Genre terms are preloaded in array $this->termParentList. The import array $ids is
+   * appended to for every parent that is added on each recursive call.
+   *
+   * @param array $id  The term id to find the parents for.
+   * @return array  Array with directly related term first, followed by
+   *                any parents found, in child to top parent order.
+   */
+  private function recursivelyGetParentTerms($id)
+  {
+    if (null === $parent = $this->termParentList[$id])
+    {
+      return array($id);
+    }
+
+    // Do not include root term id.
+    if (QubitTerm::ROOT_ID == $parent)
+    {
+      return array($id);
+    }
+
+    return array_merge(array($id), $this->recursivelyGetParentTerms($parent));
   }
 
   /*
@@ -645,18 +697,10 @@ class arElasticSearchInformationObjectPdo
    */
   protected function getDirectlyRelatedTerms($typeId)
   {
-    $sql  = 'SELECT
-                DISTINCT current.id';
-    $sql .= ' FROM '.QubitObjectTermRelation::TABLE_NAME.' otr';
-    $sql .= ' JOIN '.QubitTerm::TABLE_NAME.' current
-                ON otr.term_id = current.id';
-    $sql .= ' WHERE otr.object_id = ?
-                AND current.taxonomy_id = ?';
+    $this->prepareRelatedTermsQuery();
+    self::$statements['relatedTerms']->execute(array(':id' => $this->__get('id'), ':taxonomyId' => $typeId));
 
-    self::$statements['relatedTerms'] = self::$conn->prepare($sql);
-    self::$statements['relatedTerms']->execute(array($this->__get('id'), $typeId));
-
-    return self::$statements['relatedTerms']->fetchAll(PDO::FETCH_OBJ);
+    return self::$statements['relatedTerms']->fetchAll(PDO::FETCH_ASSOC);
   }
 
   protected function getLanguagesAndScripts()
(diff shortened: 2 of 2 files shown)


plugins/arElasticSearchPlugin/lib/model/arElasticSearchInformationObject.class.php
```diff
@@ -36,12 +38,30 @@ public function load()
     $this->count = QubitPdo::fetchColumn($sql, array(QubitInformationObject::ROOT_ID));
   }
 
+  static function loadTerms($options = array())
```
djjuhasz: This looks like it is just building a lookup list of `term['id'] => term['parent_id']`, in which case I think naming the function "loadTermParentLookups" or "loadTermParentList" or something similar.  From the name "loadTerms()" I would expect it to load term data like names, values, etc.

plugins/arElasticSearchPlugin/lib/model/arElasticSearchInformationObject.class.php
```diff
@@ -36,12 +38,30 @@ public function load()
...
+  }
+
   public function populate()
   {
     $this->load();
-
+    $this->terms = self::loadTerms(array('taxonomyIds' => array(QubitTaxonomy::SUBJECT_ID, QubitTaxonomy::PLACE_ID, QubitTaxonomy::GENRE_ID)));
     // Recursively descend down hierarchy
-    $this->recursivelyAddInformationObjects(QubitInformationObject::ROOT_ID, $this->count);
+    $this->recursivelyAddInformationObjects(QubitInformationObject::ROOT_ID, $this->count, array('terms' => $this->terms));
```
djjuhasz: The value of option['terms'] always seems to be equal to `$this->terms`, so I think it's better to call `$this->terms` directly inside `recursivelyAddInformationObjects()` and not pass the term list as a parameter.

plugins/arElasticSearchPlugin/lib/model/arElasticSearchInformationObject.class.php
```diff
@@ -65,11 +85,12 @@ public function recursivelyAddInformationObjects($parentId, $totalRows, $options
...
         if (1 < ($item->rgt - $item->lft))
         {
-          // Pass ancestors, repository and creators down to descendants
+          // Pass ancestors, repository creators and term lookup table down to descendants
           $this->recursivelyAddInformationObjects($item->id, $totalRows, array(
             'ancestors'  => array_merge($node->getAncestors(), array($node)),
             'repository' => $node->getRepository(),
-            'inheritedCreators' => array_merge($node->inheritedCreators, $node->creators)));
+            'inheritedCreators' => array_merge($node->inheritedCreators, $node->creators),
+            'terms' => $options['terms']));
```
djjuhasz: Again, there seems to be no need to pass `option['terms']` as it's never updated.

plugins/arElasticSearchPlugin/lib/model/arElasticSearchInformationObjectPdo.class.php
```diff
@@ -612,32 +619,80 @@ public function getNameAccessPoints()
...
+   * appended to for every parent that is added on each recursive call.
+   *
+   * @param array $ids  An array with one term id populated.
+   * @return array  The input array with directly related term first, followed by
+   *                any parents found, in child to top parent order.
+   */
+  private function recursivelyGetParentTerms($ids)
+  {
+    // If there were no terms loaded, then there are no ancestors.
+    if (0 >= count($this->terms))
```
djjuhasz: I think `if (0 == count($this->terms))` is sufficient and easier to read.  ;)

plugins/arElasticSearchPlugin/lib/model/arElasticSearchInformationObjectPdo.class.php
```diff
@@ -612,32 +619,80 @@ public function getNameAccessPoints()
...
+
+    // Get last term id found in array.
+    $last_id = end($ids);
+    // Get immediate parent of last_id from $this->terms.
+    $parent = $this->terms[$last_id];
+
+    // Do not include root term id.
+    if (QubitTerm::ROOT_ID != $parent)
+    {
+      $ids[] = $parent;
```
djjuhasz: I think this recursive function could be much simpler with a pattern like:

```
if (QubitTerm::ROOT_ID == $parent)
{
  return array($id);
} 
else 
{
  return array_unshift($this->recursivelyGetParentTerms($parent), $id);
}
```

This would avoid the need to pass an array of $ids to recursivelyGetParentTerms().

djjuhasz (changes requested):
Looks good @sbreker, I just have a few suggestions. :)

plugins/arElasticSearchPlugin/lib/model/arElasticSearchInformationObject.class.php
```diff
@@ -82,7 +103,9 @@ public function recursivelyAddInformationObjects($parentId, $totalRows, $options
   public static function update($object, $options = array())
   {
     // Update description
-    $node = new arElasticSearchInformationObjectPdo($object->id);
+    $node = new arElasticSearchInformationObjectPdo($object->id,
+      array('terms' => self::loadTerms(array('taxonomyIds' => array(QubitTaxonomy::SUBJECT_ID, QubitTaxonomy::PLACE_ID, QubitTaxonomy::GENRE_ID)))));
```
jraddaoui: How much can this take when there is a lot of those terms? This is executed sometimes synchronously in a request cycle and it may be too much for a simple description.

I'm not completely sure, but I think `$object` is always a `QubitInformationObject` in here. If that's the case, we could at least check if it has any object - term relations before loading the terms lookup.

plugins/arElasticSearchPlugin/lib/model/arElasticSearchInformationObjectPdo.class.php
jraddaoui: If there are no terms there won't be any related terms either and this code won't be executed, so this condition will never be true and could be removed.

plugins/arElasticSearchPlugin/lib/model/arElasticSearchInformationObject.class.php
```diff
@@ -65,11 +85,12 @@ public function recursivelyAddInformationObjects($parentId, $totalRows, $options
         // Descend hierarchy
         if (1 < ($item->rgt - $item->lft))
         {
-          // Pass ancestors, repository and creators down to descendants
+          // Pass ancestors, repository creators and term lookup table down to descendants
```
jraddaoui: Missing comma? :P

plugins/arElasticSearchPlugin/lib/model/arElasticSearchInformationObjectPdo.class.php
```diff
@@ -612,32 +619,80 @@ public function getNameAccessPoints()
...
+
+    // Iterate over each directly related term, adding all ancestors of each
+    foreach($rows as $row)
+    {
+      $relatedTerms = array_merge($relatedTerms, $this->recursivelyGetParentTerms(array($row['id'])));
+    }
+
+    // Merge unique related terms to $rows
+    $relatedTerms = array_unique($relatedTerms);
+    sort($relatedTerms, SORT_NUMERIC);
```
jraddaoui: Does sorting really matters? It may saves us some extra milliseconds.

jraddaoui:
This is great @sbreker!! Nicely thought!! 👏

plugins/arElasticSearchPlugin/lib/model/arElasticSearchInformationObject.class.php
sbreker: Agreed - implemented in next PR.

plugins/arElasticSearchPlugin/lib/model/arElasticSearchInformationObject.class.php
sbreker: I have made this change for the populate() method, but not for the update() method as it is static.

plugins/arElasticSearchPlugin/lib/model/arElasticSearchInformationObject.class.php
sbreker: fixed!  :)

plugins/arElasticSearchPlugin/lib/model/arElasticSearchInformationObject.class.php
sbreker: This will be changed as above in the latest PR.

plugins/arElasticSearchPlugin/lib/model/arElasticSearchInformationObjectPdo.class.php
sbreker: I have removed this line.

plugins/arElasticSearchPlugin/lib/model/arElasticSearchInformationObject.class.php
sbreker: I have added logic to not load the terms list if there are no otr records for this object. This will *not* apply if descendants are set to be indexed too - in this case I will load the terms list in case there are terms attached to the descendants.

sbreker: I have added a check here to stop this if there are no otr records for a given object. I have made the object's term list property static, so that if it does encounter a descendant that requires this list, it will be loaded at that time and preserved for later calls.  It should only be loaded once.  

When loading ~360K terms, using the update() method and loading the terms seems to take the same amount of time as the old logic.

plugins/arElasticSearchPlugin/lib/model/arElasticSearchInformationObjectPdo.class.php
sbreker: Good point. I have removed this check.

plugins/arElasticSearchPlugin/lib/model/arElasticSearchInformationObjectPdo.class.php
sbreker: Good idea - this is a better pattern!  I have refactored the code.

plugins/arElasticSearchPlugin/lib/model/arElasticSearchInformationObject.class.php
```diff
@@ -65,7 +85,7 @@ public function recursivelyAddInformationObjects($parentId, $totalRows, $options
         // Descend hierarchy
         if (1 < ($item->rgt - $item->lft))
         {
-          // Pass ancestors, repository and creators down to descendants
+          // Pass ancestors, repository, creators and term lookup table down to descendants
```
jraddaoui: In the end you're passing the term lookup above ;)

plugins/arElasticSearchPlugin/lib/model/arElasticSearchInformationObject.class.php
```diff
@@ -36,11 +38,29 @@ public function load()
...
+
+    return $terms;
+  }
+
   public function populate()
   {
     $this->load();
-
     // Recursively descend down hierarchy
+    self::$termParentList = self::loadTermParentList(array('taxonomyIds' => array(QubitTaxonomy::SUBJECT_ID, QubitTaxonomy::PLACE_ID, QubitTaxonomy::GENRE_ID)));
```
jraddaoui, line 58: I'd move this line above the comment now. You could have done it directly in the `load` function too.

plugins/arElasticSearchPlugin/lib/model/arElasticSearchInformationObject.class.php
```diff
@@ -81,8 +101,29 @@ public function recursivelyAddInformationObjects($parentId, $totalRows, $options
...
+    }
+
+    $sql  = 'SELECT DISTINCT otr.term_id as id';
+    $sql .= ' FROM '.QubitObjectTermRelation::TABLE_NAME.' otr';
+    $sql .= ' WHERE otr.object_id = :id';
+
+    self::$statement = self::$conn->prepare($sql);
+
+    self::$statement->execute(array(':id' => $object->id));
+    $rows = self::$statement->fetchAll(PDO::FETCH_ASSOC);
```
jraddaoui: The prepare statement mechanism improves the overall speed when multiple queries are executed in a row and you avoid its preparation. In this case you could use `QubitPdo` directly.

plugins/arElasticSearchPlugin/lib/model/arElasticSearchInformationObject.class.php
```diff
@@ -81,8 +101,29 @@ public function recursivelyAddInformationObjects($parentId, $totalRows, $options
...
+
+    if (0 < count($rows) && empty(self::$termParentList))
```
jraddaoui: I found it more readable the other way around, `count($rows) > 0`.

plugins/arElasticSearchPlugin/lib/model/arElasticSearchInformationObjectPdo.class.php
```diff
@@ -612,32 +619,67 @@ public function getNameAccessPoints()
...
-    return self::$statements['relatedTerms']->fetchAll(PDO::FETCH_OBJ);
+    // Do not include root term id.
+    if (QubitTerm::ROOT_ID == $parent)
+    {
+      return array($id);
+    }
+    else
+    {
+      return array_merge(array($id), $this->recursivelyGetParentTerms($parent));
+    }
```
jraddaoui: You could remove the `else`:

```php
if (QubitTerm::ROOT_ID == $parent)
{
  return array($id);
}

return array_merge(array($id), $this->recursivelyGetParentTerms($parent));
```

plugins/arElasticSearchPlugin/lib/model/arElasticSearchInformationObjectPdo.class.php
```diff
@@ -650,13 +692,13 @@ protected function getDirectlyRelatedTerms($typeId)
...
-                AND current.taxonomy_id = ?';
+    $sql .= ' WHERE otr.object_id = :id
+                AND current.taxonomy_id = :taxonomyId';
 
     self::$statements['relatedTerms'] = self::$conn->prepare($sql);
-    self::$statements['relatedTerms']->execute(array($this->__get('id'), $typeId));
+    self::$statements['relatedTerms']->execute(array(':id' => $this->__get('id'), ':taxonomyId' => $typeId));
 
-    return self::$statements['relatedTerms']->fetchAll(PDO::FETCH_OBJ);
+    return self::$statements['relatedTerms']->fetchAll(PDO::FETCH_ASSOC);
```
jraddaoui, line 703: The `getDirectlyRelatedTerms` and `getRelatedTerms` functions are now pretty similar and they could be normalized to avoid the duplicated code and query.

Now that I notice, they actually use the same statement name and neither of them check if it exists before preparing it.

jraddaoui (approved):
Thanks a lot @sbreker!

I've added a few minor comments, most of them cosmetic, but I'd try to address the `getDirectlyRelatedTerms` and `getRelatedTerms` unification and proper use of the statement preparation in you have the time.

plugins/arElasticSearchPlugin/lib/model/arElasticSearchInformationObject.class.php
jraddaoui: Awesome, thanks!!

plugins/arElasticSearchPlugin/lib/model/arElasticSearchInformationObject.class.php
djjuhasz: I agree that `count($rows) > 0` is more readable but I've always thought our coding convention was: when an expression has one constant operand, it should be placed on the left side of the operator (i.e. `0 < count($rows)`).  I think this convention is inherited from the [symfony coding standards](https://symfony.com/doc/current/contributing/code/standards.html) which mandates using [Yoda conditions](https://en.wikipedia.org/wiki/Yoda_conditions ).

However, reading the current symfony coding standard (above) and the reasons for Yoda conditions it looks like I have been incorrectly assuming that the Yoda condition rule that is meant strictly for equivalence operators (i.e. ==, ===, !=, !===) should be applied to *all* comparison operators (e.g. '<', '>', '>=', '<=').  Still, putting a constant operand on the left side of the operator for equivalence operators, and on the right for other comparison operators strikes me as confusing and unaesthetic.  :-/

plugins/arElasticSearchPlugin/lib/model/arElasticSearchInformationObjectPdo.class.php
sbreker: Change made in next PR.

plugins/arElasticSearchPlugin/lib/model/arElasticSearchInformationObject.class.php
sbreker: Since the Yoda standard does not apply in this case, I will revert this so that it is: `count($rows) > 0`

plugins/arElasticSearchPlugin/lib/model/arElasticSearchInformationObject.class.php
sbreker: Good point.  Changing to use QubitPdo::fetchAll() for next PR.

plugins/arElasticSearchPlugin/lib/model/arElasticSearchInformationObject.class.php
sbreker: Done.  I did not move this to the load() method though so that is remains to be parallel with how it is called for update().

plugins/arElasticSearchPlugin/lib/model/arElasticSearchInformationObject.class.php
sbreker: True! :)

plugins/arElasticSearchPlugin/lib/model/arElasticSearchInformationObjectPdo.class.php
sbreker: I have factored out the query and prepare statement so they are not repeated. I have added a check to see if the statement has previously been prepared. This will be in the next commit.

plugins/arElasticSearchPlugin/lib/model/arElasticSearchInformationObject.class.php
```diff
@@ -110,12 +111,9 @@ public static function update($object, $options = array())
     $sql .= ' FROM '.QubitObjectTermRelation::TABLE_NAME.' otr';
     $sql .= ' WHERE otr.object_id = :id';
 
-    self::$statement = self::$conn->prepare($sql);
-
-    self::$statement->execute(array(':id' => $object->id));
-    $rows = self::$statement->fetchAll(PDO::FETCH_ASSOC);
+    $rows = QubitPdo::fetchAll($sql, array(':id' => $object->id) ,array('fetchMode' => PDO::FETCH_ASSOC));
```
jraddaoui: Misplaced comma.

plugins/arElasticSearchPlugin/lib/model/arElasticSearchInformationObjectPdo.class.php
```diff
@@ -676,28 +688,18 @@ private function recursivelyGetParentTerms($id)
...
-                ON otr.term_id = current.id';
-    $sql .= ' WHERE otr.object_id = :id
-                AND current.taxonomy_id = :taxonomyId';
-
-    self::$statements['relatedTerms'] = self::$conn->prepare($sql);
-    self::$statements['relatedTerms']->execute(array(':id' => $this->__get('id'), ':taxonomyId' => $typeId));
+    $this->prepareRelatedTermsQuery();
 
+    self::$statements['relatedTerms']->execute(array(':id' => $this->__get('id'), ':taxonomyId' => $typeId));
     return self::$statements['relatedTerms']->fetchAll(PDO::FETCH_ASSOC);
```
jraddaoui, line 703: Empty line before `return`.

jraddaoui (approved):
Thanks @sbreker!!

plugins/arElasticSearchPlugin/lib/model/arElasticSearchInformationObject.class.php
sbreker: Fixed!  Missed that!

plugins/arElasticSearchPlugin/lib/model/arElasticSearchInformationObjectPdo.class.php
sbreker: Fixed.

djjuhasz (approved):
Looks good @sbreker! 👍

jraddaoui:
Awesome, I didn't know about `FETCH_KEY_PAIR` either!

Looks great!

PR #976, opened 2019-09-25
Fix D.O. metadata header display, refs #13190

mcantelon:
Fixed issued where, if Markdown is disabled, digital object metadata section
header doesn't display.


sbreker (approved):
Looks great @mcantelon!

PR #977, opened 2019-09-27
Do not disable escaping strategy, refs #13192

jraddaoui:
To avoid issues rendering Markdown, unescape the content only in the
Markdown parsing process, which is normally used in safe mode. This
reduces performance (escaping and unescaping the content) but increases
security (having all user inputs escaped by default).

Additionally, escape boolean queries inputs. Just in case, as the
escaping strategy could be manually disabled in the configuration.

### DIFF
--- a/apps/qubit/modules/actor/templates/_advancedSearch.php
+++ b/apps/qubit/modules/actor/templates/_advancedSearch.php
@@ -7,7 +7,7 @@
     <?php echo $form->renderFormTag(url_for(array('module' => 'actor', 'action' => 'browse')), array('name' => 'advanced-search-form', 'method' => 'get')) ?>
 
       <?php foreach ($hiddenFields as $name => $value): ?>
-        <input type="hidden" name="<?php echo $name ?>" value="<?php echo esc_entities($value) ?>"/>
+        <input type="hidden" name="<?php echo $name ?>" value="<?php echo $value ?>"/>
       <?php endforeach; ?>
 
       <p><?php echo __('Find results with:') ?></p>
--- a/apps/qubit/modules/actor/templates/autocompleteSuccess.php
+++ b/apps/qubit/modules/actor/templates/autocompleteSuccess.php
@@ -38,7 +38,7 @@
 
 <div class="search">
   <form action="<?php echo url_for(array('module' => 'actor', 'action' => 'list')) ?>">
-    <input name="subquery" value="<?php echo esc_entities($sf_request->subquery) ?>"/>
+    <input name="subquery" value="<?php echo $sf_request->subquery ?>"/>
     <input class="form-submit" type="submit" value="<?php echo __('Search %1%', array('%1%' => sfConfig::get('app_ui_label_actor'))) ?>"/>
   </form>
 </div>
--- a/apps/qubit/modules/default/templates/moveSuccess.php
+++ b/apps/qubit/modules/default/templates/moveSuccess.php
@@ -11,7 +11,7 @@
       <form action="<?php echo url_for(array($resource, 'module' => 'default', 'action' => 'move')) ?>">
         <div class="input-append">
           <?php if (isset($sf_request->query)): ?>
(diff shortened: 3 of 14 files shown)


apps/qubit/modules/actor/templates/_advancedSearch.php
```diff
@@ -26,7 +26,7 @@
                 <option value="not"<?php echo $item['operator'] == 'not' ? ' selected="selected"' : '' ?>><?php echo __('not') ?></option>
               </select>
 
-              <input class="query" type="text" placeholder="<?php echo __('Search') ?>" name="sq<?php echo $key ?>" value="<?php echo $item['query'] ?>"/>
+              <input class="query" type="text" placeholder="<?php echo __('Search') ?>" name="sq<?php echo $key ?>" value="<?php echo esc_entities($item['query']) ?>"/>
```
djjuhasz: Will calling `esc_entities()` explicitly here cause ampersands to be double-escaped (e.g. "&amp;amp;amp;") when global escaping is enabled?

djjuhasz (commented):
@jraddaoui I just have one question about calling `esc_entities()` explicitly.

apps/qubit/modules/actor/templates/_advancedSearch.php
jraddaoui: Yeah, it looks like it ...

http://sandbox.onlinephpfunctions.com/code/4400d1eb438ac3b99720829979e958d2b38fcd3c

djjuhasz:
I agree that we are better off keeping global escaping on when Markdown is enabled, and taking the performance hit on Markdown parsing. 

I think we can assume that symfony global escaping is enabled, as this is the application default, and _not_ explicitly call `esc_entitites()` to avoid double escaping strings.  We may need to remove some explicit calls to `esc_entities()` in the code, but this should be easier than the converse of trying to add `esc_entities()` in the scenario where global escaping is disabled.

We could add a warning to the docs that disabling the symfony escaping strategy in the application setting exposes the site to XSS attacks to make the risk clear to community users installing the application.

jraddaoui:
Sounds good @djjuhasz, I'll remove the `esc_entitites()` calls. Thanks!

jraddaoui:
Hi @djjuhasz,

I've removed the `esc_entities` calls in the direct accesses to `$sf_request` properties and other places where we really know the content should be already escaped (adv. search inputs, etc.).

There are other places where we call it over objects and config settings that are obtained directly in the template and, in those cases, it may still be necessary to escape the content.

Additionally, there are plenty of calls to `esc_specialchars`, most of them in the XML exports (for example in the [DC export](https://github.com/artefactual/atom/blob/qa/2.6.x/plugins/sfDcPlugin/modules/sfDcPlugin/templates/_dc.xml.php)) which I think are causing the double escaping too, but I didn't want to go that far for a minor release.

Please, let me know your thoughts about that and if you think this is ready to squash and merge.

PD. We actually have the escaping strategy forced in [here](https://github.com/artefactual/atom/blob/qa/2.6.x/apps/qubit/config/qubitConfiguration.class.php#L54), so changing it in the `settings.yml` file won't work. Although, you could set the escaping method to `ESC_RAW` to disable it, so the warning in the docs may still be a good idea as that "forcing mechanism" is not strong enough. Alternatively, we could set the escaping method in the same manner to "really force" the escaping strategy and method. [More info](https://symfony.com/legacy/doc/gentle-introduction/1_4/en/07-Inside-the-View-Layer).

djjuhasz:
Looks good to me, thanks Radda! 👍 

I agree with your decision not to take on the work of removing the `esc_specialchars` for the XML exports as part of this pull request or the 2.5.3 patch release.  I think we should document the XML double-escaping issue as a new ticket and address it separately.

I think adding a warning to the documentation not to use `ESC_RAW` due to the risk of XSS attack is a good idea and is showing an abundance of caution.   I don't think changing the code to force output escaping strategy is necessary.

djjuhasz (approved):

jraddaoui:
5b6b6ee in stable/2.5.x.

PR #979, opened 2019-10-02
Add treeview AJAX translation links, refs #8572

mcantelon:
When navigating from description to description, using the full
treeview, if translation links weren't present in the current page
then the translation links weren't getting updated. Fixed.


sbreker (approved):
Looks good @mcantelon !

PR #981, opened 2019-10-07
Fixes to header display component, refs #13173

mcantelon:
Fixed issues in the header display component that was added to fix
Markdown parsing issues.

Issues fixed:

* A function signature in the component was incorrect
* The component's template filename conflicted with with site header
  template filename causing an inappropriate template override


jraddaoui (approved):
I didn't expect that override, looks good @mcantelon!

PR #982, opened 2019-10-09
Fix DC XML creator inheritance, refs #13130

mcantelon:
Changed the DC XML component so creators can be inherited from non-top
level information objects.


djjuhasz (approved):
Looks good to me 👍

PR #983, opened 2019-10-10
Add physical object CSV import, refs #13183

djjuhasz:
- Add csv:physicalobject-import symfony task
- Add `--culture`, `--error-log`, `--header`, `--index`, `--rows-until-update`, `--skip-rows`, and `--source-name` options
- Use composer to manage dependencies
- Use http://csv.thephpleague.com library for CSV file parsing
- Add PHPUnit 8 unit tests for new code
- Add vfsStream library for mocking file system in tests

### DIFF
--- a/lib/PhysicalobjectCsvImporter.class.php
+++ b/lib/PhysicalobjectCsvImporter.class.php
@@ -1,0 +1,657 @@
+<?php
+
+/*
+ * This file is part of the Access to Memory (AtoM) software.
+ *
+ * Access to Memory (AtoM) is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * Access to Memory (AtoM) is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with Access to Memory (AtoM).  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+/**
+ * Importer for Physical Object CSV data
+ *
+ * @package    AccessToMemory
+ * @subpackage PhysicalObject
+ * @author     David Juhasz <djuhasz@artefactual.com>
+ */
+class PhysicalObjectCsvImporter
+{
+  static $columnMap = [
+    'legacyId'         => 'legacyId',
+    'name'             => 'name',
+    'type'             => 'typeId',
+    'location'         => 'location',
+    'culture'          => 'culture',
+    'descriptionSlugs' => 'informationObjectIds',
+  ];
+
+  protected $context;
+  protected $data;
+  protected $dbcon;
+  protected $errorLogHandle;
+  protected $filename;
+  protected $multiValueDelimiter = '|';
+  protected $offset              = 0;
+  protected $options             = [];
+  protected $ormClasses;
+  protected $physicalObjectTypeTaxonomy;
+  protected $progressFrequency   = 1;
+  protected $reader;
+  protected $rowsImported        = 0;
+  protected $rowsTotal           = 0;
+  protected $typeIdLookupTable;
+  protected $updateSearchIndex   = false;
+
+
+  public function __construct(sfContext $context = null, $dbcon = null,
+    $options = array())
+  {
+    if (null === $context)
+    {
+      $context = new sfContext(ProjectConfiguration::getActive());
+    }
+
+    $this->setOrmClasses([
+      'informationObject' => QubitInformationObject::class,
+      'keymap'            => QubitKeymap::class,
+      'physicalObject'    => QubitPhysicalObject::class,
+      'relation'          => QubitRelation::class,
+    ]);
+
+    $this->context = $context;
+    $this->dbcon   = $dbcon;
+    $this->setOptions($options);
+  }
+
+  public function setOrmClasses(Array $classes)
+  {
+    $this->ormClasses = $classes;
+  }
+
+  public function __get($name)
+  {
+    switch ($name)
+    {
+      case 'context':
+      case 'multiValueDelimiter':
+        return $this->$name;
+
+        break;
+
+      case 'dbcon':
+        return $this->getDbConnection();
+
+        break;
+
+      case 'physicalObjectTypeTaxonomy':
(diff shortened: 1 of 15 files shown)


djjuhasz:
This is fully implemented and ready for code review now.

**N.B.**: Using Composer to install http://csv.thephpleague.com/ is going to require changes to our AtoM installation documentation and deployment scripts.  I could just add the code to the AtoM vendor directory, as we've done with other libraries, but I'd like to move to using Composer for managing / installing most AtoM dependencies.

lib/PhysicalobjectCsvImporter.class.php
```diff
@@ -0,0 +1,655 @@
...
+      $msg = <<<EOM
+Invalid header. Please provide a CSV delimited list of column names
+e.g. "name,location,type,culture".
+EOM;
+
+      throw new sfException($msg);
+    }
+
+    // Throw error on unknown column names
+    foreach ($columnNames as $name)
```
sbreker, line 252: I think unknown columns should just be ignored and not trigger an import error. When doing migrations, I will sometimes leave additional client columns in an 'AtoM friendly' CSV for debugging/verification purposes post-import.

lib/PhysicalobjectCsvImporter.class.php
```diff
@@ -0,0 +1,655 @@
...
+      $msg = <<<EOL
+Please use setFilename(\$filename) or doImport(\$filename) to specify the CSV
+file you wish to import.
+EOL;
+      throw new sfException($msg);
+    }
+
+    $this->reader = $this->readCsvFile($this->filename);
+    $stmt = (new \League\Csv\Statement)->offset($this->offset);
+    $records = $this->getRecords($stmt);
```
sbreker, line 322: What is the largest number of rows/csv you've imported?  How was the performance?

lib/PhysicalobjectCsvImporter.class.php
djjuhasz: @sbreker Yeah, I debated whether to throw an error on the column names.  My thinking is because these columns are passed to the importer as a task option (e.g. `--header="name,location"`) that it would be easy to typo the name of a column and none of the data in that column would import. :(   Headers included in the CSV file don't use this code.

lib/PhysicalobjectCsvImporter.class.php
djjuhasz: @sbreker 46 rows.  It was really fast. :P  I agree it would be good to test with a much bigger dataset.

lib/PhysicalobjectCsvImporter.class.php
```diff
@@ -0,0 +1,655 @@
...
+    $ids = [];
+
+    if (null === $str)
+    {
+      return $ids;
+    }
+
+    foreach ($this->processMultiValueColumn($str) as $val)
+    {
+      $class = $this->ormClasses['informationObject'];
```
sbreker, line 580: Is there an advantage to this technique over just using the class name directly?

lib/PhysicalobjectCsvImporter.class.php
sbreker: Yes - curious about memory consumption

lib/PhysicalobjectCsvImporter.class.php
sbreker: Ok - as long as it doesn't affect CSVs with headers included as first row then all is good!  :) Curious what the use case for --header is. I don't think this exists for any of the other tasks?

djjuhasz: The client who is sponsoring this feature is exporting CSV data from a barcode scanning system, and the export doesn't include column headings.

lib/PhysicalobjectCsvImporter.class.php
djjuhasz: Yes, I'm using the ormClasses array to [inject mock ORM objects](https://github.com/artefactual/atom/blob/e973cfd323beb83b1bed8f3991f258ef89b60721/test/phpunit/PhysicalobjectCsvImporterTest.php#L24) when running tests to avoid connecting to the database and avoid side effects from the state of the database data.

sbreker: Ah! I've love a tour of all this and how it fits together - so many interesting changes!

djjuhasz: @sbreker I'm happy to do a tour.  I'm taking care of my kids today (Pro-D day) but maybe on Monday I can take you through it. :)

test/phpunit/PhysicalobjectCsvImporterTest.php
```diff
@@ -0,0 +1,564 @@
+<?php
+
+use org\bovigo\vfs\vfsStream;
+
+class PhysicalObjectCsvImporterTest extends \PHPUnit\Framework\TestCase
+{
+  protected $csvHeader;
+  protected $csvData;
+  protected $daos;
```
djjuhasz: Oops, I seem to have declared $daos and then never used it.  I'll delete before merging.

lib/PhysicalobjectCsvImporter.class.php
```diff
@@ -0,0 +1,655 @@
...
+      case 'descriptionSlugs':
+        $val = $this->processDescriptionSlugs($val);
+
+        break;
+
+      default:
+        $val = trim($val);
+    }
+
+    if ('' !== $val)
```
sbreker, line 563: Quibble: could use empty()

lib/model/QubitTaxonomy.php
```diff
@@ -172,4 +172,54 @@ public static function getTaxonomyTerms($taxonomyId, $options = array())
...
+        term_i18n.culture as `culture`
+      FROM term INNER JOIN term_i18n ON term.id = term_i18n.id
+      WHERE term.taxonomy_id = ?
+      ORDER BY term_i18n.culture ASC, term_i18n.name ASC;
+SQL;
+
+    $statement = $connection->prepare($sql);
+    $statement->execute([$this->id]);
+
+    return  $statement->fetchAll(PDO::FETCH_ASSOC);
```
sbreker, line 199: Good use of FETCH_ASSOC

sbreker (approved):
Great work @djjuhasz! 👍 Some communication will be required for moving forward with new unit tests, use of composer, and new code formatting styles (explicit getters/setters vs __get, array declarations). Integration of Composer into install processes (Vagrant, Ansible, manual installation) to be completed.

lib/PhysicalobjectCsvImporter.class.php
djjuhasz: @sbreker I actually want to return an empty array when `$val = array()` and empty() returns true for an empty array.  I'll add a comment in the code to explain the logic though.

PR #984, opened 2019-10-11
Add actor:move-related-descriptions, refs #13189

jraddaoui:
Add a new task to move the related description from one actor to
another. Including all events and subject access point relations.
Include a `skip-index` option to avoid updating the ES index.

### DIFF
--- a/lib/task/actor/actorMoveDescriptionRelationsTask.class.php
+++ b/lib/task/actor/actorMoveDescriptionRelationsTask.class.php
@@ -1,0 +1,176 @@
+<?php
+
+/*
+ * This file is part of the Access to Memory (AtoM) software.
+ *
+ * Access to Memory (AtoM) is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * Access to Memory (AtoM) is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with Access to Memory (AtoM).  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+class ActorMoveDescriptionRelationsTask extends arBaseTask
+{
+  protected function configure()
+  {
+    $this->addArguments(array(
+      new sfCommandArgument(
+        'source',
+        sfCommandArgument::REQUIRED,
+        'The slug of the source actor'
+      ),
+      new sfCommandArgument(
+        'target',
+        sfCommandArgument::REQUIRED,
+        'The slug of the target actor'
+      ),
+    ));
+
+    $this->addOptions(array(
+      new sfCommandOption(
+        'application',
+        null,
+        sfCommandOption::PARAMETER_OPTIONAL,
+        'The application name',
+        true
+      ),
+      new sfCommandOption(
+        'env',
+        null,
+        sfCommandOption::PARAMETER_REQUIRED,
+        'The environment',
+        'cli'
+      ),
+      new sfCommandOption(
+        'connection',
+        null,
+        sfCommandOption::PARAMETER_REQUIRED,
+        'The connection name',
+        'propel'
+      ),
+      new sfCommandOption(
+        'skip-index',
+        null,
+        sfCommandOption::PARAMETER_NONE,
+        "Skip Elasticsearch indexing"
+      ),
+    ));
+
+    $this->namespace = 'actor';
+    $this->name = 'move-description-relations';
+    $this->briefDescription = 'Move actor-description relations';
+    $this->detailedDescription = <<<EOF
+Move description relations from a source actor to a target actor,
+including all events and name access point relations.
+EOF;
+  }
+
+  protected function execute($arguments = array(), $options = array())
+  {
+    parent::execute($arguments, $options);
+
+    if (null === $source = QubitActor::getBySlug($arguments['source']))
+    {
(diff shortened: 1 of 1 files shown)


lib/task/actor/actorMoveRelatedDescriptionsTask.class.php
```diff
@@ -0,0 +1,121 @@
...
+
+    $this->addOptions(array(
+      new sfCommandOption('application', null, sfCommandOption::PARAMETER_OPTIONAL, 'The application name', true),
+      new sfCommandOption('env', null, sfCommandOption::PARAMETER_REQUIRED, 'The environment', 'cli'),
+      new sfCommandOption('connection', null, sfCommandOption::PARAMETER_REQUIRED, 'The connection name', 'propel'),
+      new sfCommandOption('skip-index', null, sfCommandOption::PARAMETER_NONE, "Skip Elasticsearch indexing"),
+    ));
+
+    $this->namespace = 'actor';
+    $this->name = 'move-related-descriptions';
```
djjuhasz: To me `actor:move-related-descriptions` implies moving descriptions in the descriptive hierarchy.  I would prefer `actor:update-description-relations` or `io:update-actor-relations` or something similar.

lib/task/actor/actorMoveRelatedDescriptionsTask.class.php
```diff
@@ -0,0 +1,121 @@
...
+    $sql = "UPDATE relation
+      JOIN object ON relation.subject_id=object.id
+      SET relation.object_id=:targetId
+      WHERE relation.object_id=:sourceId
+      AND relation.type_id=:typeId
+      AND object.class_name='QubitInformationObject'";
+    $params = array(':targetId' => $target->id, ':sourceId' => $source->id, ':typeId' => QubitTerm::NAME_ACCESS_POINT_ID);
+    $updatedCount += QubitPdo::modify($sql, $params);
+
+    $this->log($updatedCount.' related descriptions moved.');
```
djjuhasz: Again, to me "related descriptions moved." implies the descriptions have been moved in the descriptive hierarchy, which is not the case.  I'd prefer language like "[count] description-actor relationships updated".

lib/task/actor/actorMoveRelatedDescriptionsTask.class.php
```diff
@@ -0,0 +1,121 @@
...
+      $search->update($source);
+      $search->update($target);
+      foreach ($relatedIoIds as $id)
+      {
+        $search->update(QubitInformationObject::getById($id));
+      }
+    }
+    else
+    {
+      $this->log('The Elasticsearch index has not been updated.');
```
djjuhasz: I suggest adding text like "The Elasticsearch index has not been updated. Please run search:populate manually to update the description-actor relationships in the search index."

lib/task/actor/actorMoveRelatedDescriptionsTask.class.php
```diff
@@ -0,0 +1,121 @@
...
+    $updatedCount = QubitPdo::modify($sql, $params);
+
+    // Move name access point relations
+    $sql = "UPDATE relation
+      JOIN object ON relation.subject_id=object.id
+      SET relation.object_id=:targetId
+      WHERE relation.object_id=:sourceId
+      AND relation.type_id=:typeId
+      AND object.class_name='QubitInformationObject'";
+    $params = array(':targetId' => $target->id, ':sourceId' => $source->id, ':typeId' => QubitTerm::NAME_ACCESS_POINT_ID);
```
djjuhasz: This raises a question for me about line lengths and wrapping.  I've been trying to wrap my code lines according to the rules in https://www.php-fig.org/psr/psr-12/ - e.g. aiming for 80 chars per line or less, but allowing 120 chars for legibility, and avoiding >120 chars per line except in extreme cases.   I don't see any recommendations in The [AtoM Coding Standard](https://wiki.accesstomemory.org/wiki/Development/Coding_standard) or [Symfony Coding Standards](https://symfony.com/doc/current/contributing/code/standards.html) for line length, but do we have any consensus among the AtoM devs on line lengths?

djjuhasz (commented):
Code looks good! I've made a few comments about the language around "moving" related descriptions, and had one question about line lengths, but nothing that is a blocker to merging.

lib/task/actor/actorMoveRelatedDescriptionsTask.class.php
jraddaoui: I like that approach @djjuhasz! And I'd love to see a CI process in AtoM that enforces those rules at some point ...

https://github.com/FriendsOfPHP/PHP-CS-Fixer
https://github.com/squizlabs/PHP_CodeSniffer

djjuhasz: @jraddaoui I would also like to add a linter or formatter to our development workflow at some point, but doing so will mean changing some of our current coding standards I think.  I'll make a task for myself to add the line length standard to our coding standard doc as a first step.

lib/task/actor/actorMoveRelatedDescriptionsTask.class.php
jraddaoui: Thanks @djjuhasz!

Since we're using the `source` and `target` parameters, I think "move" is better than "update" but I agree that "description relations" is less confusing, so I went with `move-description-relations` in the end. I also struggle in between "relations" and "relationships" so, please, let me know if you think the later is better.

I'd like to hear @fiver-watson thoughts about this too ...

lib/task/actor/actorMoveRelatedDescriptionsTask.class.php
jraddaoui: I'm not sure how much can be configured with those tools, but we may be able to reflect our current coding standards in them. Although, I don't think it's a bad idea to use an opinionated code formatter, even if we have to change our style.

djjuhasz (approved):
LGTM. 👍

PR #987, opened 2019-10-21
Allow 'add' menu access, Refs #13202

sbreker:
In addition to restricting AtoM's 'add' menu to specific hard coded
groups, check ACL groups to see if the 'add' menu should appear for
users in custom ACL groups. Previously, custom groups were not being
checked to determine if AtoM's 'add' menu should be displayed for
these users.

Removed term, Isdf and Isdiah security.yml files to allow access to
these modules to authenticated users in custom groups.

Co-Authored-By: Mike Cantelon <mcantelon@gmail.com>


mcantelon (approved):
Looks good! 👍

PR #989, opened 2019-10-27
Add warning to CSV I.O. I18n import, refs #8572

mcantelon:
Prevent attempts to add related data during CSV import of information
object translations.

Add a warning to CSV import of information object translations when
columns that aren't supported are populated.


sbreker (approved):
Looks good @mcantelon!

PR #990, opened 2019-10-27
Escape apostrophe in XML parser, refs #13204

mcantelon:
Apostrophes were causing a syntax error during eval in XML parsing
code.

### DIFF
--- a/lib/QubitXmlImport.class.php
+++ b/lib/QubitXmlImport.class.php
@@ -658,7 +658,7 @@ class QubitXmlImport
                         if (null !== ($node = eval('return '.$str.';')))
                         {
                           // Substitute node value for search string
-                          $parameter = str_replace($match, '\''.$node->nodeValue.'\'', $parameter);
+                          $parameter = str_replace($match, '\''.addcslashes($node->nodeValue, "'").'\'', $parameter);
                         }
                         else
                         {


lib/QubitXmlImport.class.php
```diff
@@ -658,7 +658,7 @@ private function processMethods(&$domNode, &$importDOM, $methods, &$currentObjec
                         if (null !== ($node = eval('return '.$str.';')))
                         {
                           // Substitute node value for search string
-                          $parameter = str_replace($match, '\''.$node->nodeValue.'\'', $parameter);
+                          $parameter = str_replace($match, '\''.addcslashes($node->nodeValue, "'").'\'', $parameter);
```
djjuhasz, line 661: As far as I can tell, this allows arbitrary execution (by PHP) of *any* string used for `$node->nodeValue` in line 671 below. :(

djjuhasz (changes requested):
I think the `eval()` statement in line 671 needs to be removed altogether, as it's a security flaw. :(

lib/QubitXmlImport.class.php
mcantelon: True. I think most of our XML importing relies on this mechanism, which is not so good.

sbreker: @mcantelon will be creating a new ticket describing the security issue with the eval() use.

sbreker: Ticket describing eval() issue:
https://projects.artefactual.com/issues/13216

sbreker (approved):
Looks good @mcantelon. Thanks for creating that ticket!

djjuhasz (approved):
Looks good 👍

mcantelon:
Thanks @sbreker !

PR #992, opened 2019-10-29
Move composer autoload to importer, refs #13183

djjuhasz:
Requiring `vendor/composer/autoload.php` in ProjectConfiguration was causing an
application-wide critical error if the autoloader wasn't installed via
composer.


mcantelon (approved):
Looks good to me! 👍

PR #995, opened 2019-11-01
Fix CSV notes import, refs #13209

jraddaoui:
Check `SOURCE_CULTURE` constant instead of `sourceCulture` property
as the latter is accessed through magic methods and `property_exits`
can't find it.


replaceafill (approved):
LGTM :+1: 

Thank you @jraddaoui for all the explanations to help me understand the change!

PR #997, opened 2019-11-06
Support MySQL 8 and fix DB, refs #13220 and #13221

jraddaoui:
**Update Docker environment:**

- Use `percona:8.0` image.
- Use `mysql_native_password` as `default-authentication-plugin`.
- Change default charset to `utf8mb4` and collation to `utf8mb4_0900_ai_ci`.
- Remove `NO_AUTO_CREATE_USER` from `sql_mode`.

**Rename QubitFunction model:**

In MySQL 8, `function` became a reserved word causing errors when
working with QubitFunction due to it's auto-generated table name.

- Change schema declaration to `function_object`.
- Generate SQL and models and remove old ones.
- Change references to old models.
- Update ES mapping and code that depends on the model class name.
- Add migration to change existing database tables and class reference.
- Avoid changing module name and related settings to keep the front-end
  representation (URLs and labels) as they were.
- Change model class reference in old migration classes/process. This
  process should be executed with the 1.1 code to upgrade to 1.1 before
  running the actual upgrade task. Therefore, this change is only
  cosmetic, to avoid finding references to the old model class.

**Normalize database schema:**

- Avoid differences between upgraded and new install database schema.
- Do it before the `utf8mb4` migration to fix NULL columns.
- Improve error management on migrations:
  - Split statements to show errors on arMigration0169.
  - Throw exception when expected indexes or foreign keys are missing
    on arMigration0176 and arMigration0177.

**Migrate to `utf8mb4`:**

The utf8mb3 character set is deprecated and will be removed in a future
MySQL release.

- Change default database and connection encoding.
- Use `utf8mb4_0900_ai_ci` collation.
- Reduce DO path index size in chars to allow the extra byte.
- Convert existing data using MODIFY over char columns.
- Analyze all tables after conversion.


jraddaoui:
About the `function` table name conflict:

Ideally, we would modify the schema and create a migration to rename the table as we wouldn't have use that name knowing it's a reserved word. However, there are several queries in `QubitMigrate` in which the `function` table name may be passed as a variable and those queries can be executed before the new migration is reached. Therefore, we should revisit all raw queries and add backquotes in those where the `function` table is used or can be passed as a variable (PARTIALLY DONE). Additionally, we must remove the use of the `function` word as an alias (DONE?).

After adding the backquotes, changing the table name shouldn't be necessary and I'm having conflicts about doing it or not. On one hand, it can avoid problems and confusions in the future but, on the other, it may not be an easy task as we'll need to modify the schema and model and that could also cause issues on the existing migrations.

docker/etc/mysql/mysqld.cnf
```diff
@@ -1,7 +1,8 @@
 [mysqld]
-character-set-server=utf8
-collation-server=utf8_unicode_ci
+character-set-server=utf8mb4
+collation-server=utf8mb4_unicode_ci
```
djjuhasz: Shouldn't this be `collation-server=utf8mb4_0900_ai_ci`?

djjuhasz:
I'm in favor of sticking with "function" for the table name, unless a good alternative is proposed. I looked at the definition of function in ISDF (quoted below) and I can't see any immediately apparent alternative to "function".

> **Function**.    Any  high  level  purpose,  responsibility  or  task  assigned  to  the  accountability agenda of a corporate body by legislation, policy or mandate. Functions may be decomposed into sets of co-ordinated operations such as subfunctions, business processes, activities, tasks or transactions.

-- https://www.ica.org/sites/default/files/CBPS_2007_Guidelines_ISDF_First-edition_EN.pdf, retrieved 2019-11-06

djjuhasz:
@jraddaoui thanks for providing all of the helpful reference links!  Lots of interesting information and considerations in upgrading. :)

mcantelon:
Yeah, looks like there's about 17 instances of "function." being used to reference the table so seems updateable.

docker/etc/mysql/mysqld.cnf
jraddaoui: Good point @djjuhasz!

I didn't want to change the current collation that much but it may be a good idea to use the latest Unicode version and the case and accent insensitivity.

Extra reading:
https://dev.mysql.com/doc/refman/8.0/en/charset-unicode-sets.html
https://mysqlserverteam.com/new-collations-in-mysql-8-0-0/
https://stackoverflow.com/questions/54885178/whats-the-difference-between-utf8-unicode-ci-and-utf8mb4-0900-ai-ci

jraddaoui:
> I'm in favor of sticking with "function" for the table name, unless a good alternative is proposed. I looked at the definition of function in ISDF (quoted below) and I can't see any immediately apparent alternative to "function".
> 
> > **Function**.    Any  high  level  purpose,  responsibility  or  task  assigned  to  the  accountability agenda of a corporate body by legislation, policy or mandate. Functions may be decomposed into sets of co-ordinated operations such as subfunctions, business processes, activities, tasks or transactions.
> 
> -- https://www.ica.org/sites/default/files/CBPS_2007_Guidelines_ISDF_First-edition_EN.pdf, retrieved 2019-11-06

We wouldn't have to change the visual name, just the model on the backend (`QubitFunctionObject` for example). But I think it's better if we can avoid changing it, and having backquotes in the raw queries table names doesn't hurt and may help us in the future.

jraddaoui:
> Yeah, looks like there's about 17 instances of "function." being used to reference the table so seems updateable.

Thanks @mcantelon! Could you share where are they in the code? Or what did you use to search those occurrences?

mcantelon:
I just used a gigantic grep statement, filtering out stuff in which a reference to "function." didn't seem to be for SQL:

`grep -r "function\." * | grep -v "\.log" | grep -v "\.js" | grep -v "messages\.xml" | grep -v "\.pkg" | grep -v "MailTransport.php" | grep -v "HTMLsymbol.ent" | grep -v "MailInvoker.php" | grep -v "xhtml-symbol.ent" | grep -v "saxon9he.jar" | grep -v "FunctionScore.php" | grep -v "JavascriptHelper.php" | grep -v "sfViewCacheManager.class.php" | grep -v "sfAutoloadAgain.class.php" | grep -v "HTMLModule.php" | grep -v "EasyRdf" | grep -v "htmlpurifier" | grep -v "sfTestFunctionalBase.class.php" | grep -v "Gearman" | grep -v "Doctrine" | grep -v "QubitMarkdown.class.php" | grep -v "DBAdapter.php" | grep -v "sfPropelPlugin" | grep -v "sfSkosUniqueRelations.class.php" | grep -v "sfCultureInfo.class.php" | grep -v "miniColors" | grep -v "swiftmailer" | grep -v "FluentDOM" | grep -v "dtd" | grep -v "Qubit\.class\.php" | grep -v "QubitDigitalObject\.php" | grep -v "QubitMigrate\.class\.php" | grep -v "QubitXmlImport\.class\.php" | grep -v "_relatedFunction\.php" | grep -v "editSuccess\.php" | wc -l`

jraddaoui:
Hi @djjuhasz and @mcantelon,

This is finally ready for code review. I had some back and forwards along the process so I'd suggest that you read the updated PR description and jump into the full code changes instead of going commit by commit.

Also, https://projects.artefactual.com/issues/13221 has more information about some schema differences I found during the upgrade that I addressed in this PR too.

lib/task/migrate/migrations/arMigration0177.class.php
```diff
@@ -0,0 +1,245 @@
...
+    {
+      $baseTable = str_replace('_i18n', '', $i18nTable);
+
+      // Update possible NULL values added in between migrations.
+      // The culture and id together are a unique key on the i18n tables so
+      // there shouldn't be more than a NULL value per object id.
+      $sql = 'UPDATE %s SET culture=:cul WHERE culture IS NULL;';
+      QubitPdo::modify(
+        sprintf($sql, $i18nTable),
+        array(':cul' => $defaultCulture)
```
djjuhasz: I would suggest setting the `source_culture` first (e.g. if source_culture is NULL, set it to default culture) *then* setting the `i18n.culture = source_culture`.   In most cases if there is no culture set, I would assume the culture should be the same as the row source_culture, rather than the default culture.  

Practically, I'm sure in most cases when the culture == null the source_culture == null too but I'm not sure if we can assume it.

djjuhasz (commented):
Looks great @jraddaoui! 😄   I made one suggestion, but it may or may not actually make a difference to the upgrade.

lib/task/migrate/migrations/arMigration0177.class.php
jraddaoui: Good catch, thanks @djjuhasz!

djjuhasz (approved):
Looks great! :shipit:

PR #999, opened 2019-11-13
CSV update of physical object data via, refs #13185

djjuhasz:
* Add --update option to `csv:physicalobject-import` for updating existing physical object data via CSV import
* Add --skip-unmatched option to prevent the addition of new physical objects during update
* Add --multi-match option to control which physical objects to update if more than one matches on name (options: skip, first, all)
* Add --partial-match option to match and update existing physical objects when the import name matches the beginning of an existing physical object
* Add --empty-overwrite option to make an empty CSV value overwrite existing data on update


sbreker (approved):
Looks great @djjuhasz!

djjuhasz:
Merged to qa/2.6.x in 5a234163a667e7de95b18157bb5603ced15226c0

PR #1000, opened 2019-11-13
Improve IO hierarchy deletion, refs #13211

jraddaoui:
Avoid nested set update until the deletion of the top-level description
in the tree. The queries used to update the nested may be time expensive
as they update all the descriptions above the deleted description,
including those outside the deleted tree and those that are inside and
will be deleted after. When the `deleteFromNestedSet` function from
`BaseInformationObject` is called, the delta used to update the LFT and
RGT values is calculated from the resource's RGT -LFT difference.
Considering that this operation is normally run inside a transaction
(otherwise it should), updating the nested set only once at the end
should be enough.

### DIFF
--- a/lib/model/QubitInformationObject.php
+++ b/lib/model/QubitInformationObject.php
@@ -3349,8 +3349,24 @@ class QubitInformationObject extends BaseInformationObject
   public function deleteFullHierarchy()
   {
     $n = 0;
+
     foreach ($this->descendants->andSelf()->orderBy('rgt') as $item)
     {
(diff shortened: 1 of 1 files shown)


jraddaoui:
Merged in d24decb. Sorry for messing up PR 1000 :see_no_evil:

I renamed the branch in the CLI from `qa/2.6.x` to `dev/issue-13211-io-delete` and pushed from there to create this PR, then changed the commit message and force pushed from the Atom editor, it looks like it was still referencing `qa/2.6.x` in the editor and it merged the commit directly.

PR #1001, opened 2019-11-14
Add composer deps when building tarball

djjuhasz:
- Install phing with composer (instead of old PEAR channel) for tarball packaging
- Add composer production dependencies to tarball build

### DIFF
--- a/build.xml
+++ b/build.xml
@@ -72,8 +72,15 @@
   <target name="prepare" depends="debug">
     <delete dir="${build.dir}"/>
     <mkdir dir="${build.dir}"/>
+
+    <echo>Install production dependencies with composer</echo>
+    <exec command="composer install --no-dev" dir="${phing.dir}" checkreturn="true"/>
+
+    <echo>Compile css with less</echo>
     <exec command="make" dir="${phing.dir}/plugins/arDominionPlugin" checkreturn="true"/>
     <exec command="make" dir="${phing.dir}/plugins/arArchivesCanadaPlugin" checkreturn="true"/>
+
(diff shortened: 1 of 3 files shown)


mcantelon:
Seems good. Maybe we want to add something to the build.xml build / prepare target to check to make sure composer is installed?

i.e.
```
    <exec command="which composer" returnProperty="return_code"/>
    <if>
      <equals arg1="${return_code}" arg2="1" />
      <then>
        <fail message="You need to install composer before!"/>
      </then>
    </if>

```

Is "composer" the default command filename for composer, too? Seems like it might be "composer.phar" as per https://getcomposer.org/download/ (but maybe it depends on how you install it).

djjuhasz:
@mcantelon Good thought about testing if composer is installed.  I decided against the "which composer" test ultimately because for AtoM 2.6+ I am recommending installing Phing with Composer rather than PEAR - the [Phing PEAR channel is deprecated](http://pear.phing.info/), and will not work for Phing 3.0+.  Because Phing itself is being installed via composer, I don't think it's necessary to test if composer is installed from within the Phing script.

I did add some extra progress messages to the Phing task to aid in debugging when a build fails.

When composer is installed [globally](https://getcomposer.org/doc/00-intro.md#globally) (recommended) it is available on the PATH as `composer`.  You are right that the local install would require the use of `~/composer.phar` instead.

mcantelon:
>Because phing itself is being installed via composer, I don't think it's necessary to test if composer is installed from within the phing script.

Ah... true that.

mcantelon (approved):
Looks good to me!

PR #1002, opened 2019-11-15
Fix clipboard send escaping issue, refs #12800

mcantelon:
Fixed escaping-related issue with the clipboard send feature.

### DIFF
--- a/apps/qubit/modules/user/actions/clipboardSendAction.class.php
+++ b/apps/qubit/modules/user/actions/clipboardSendAction.class.php
@@ -21,7 +21,7 @@ class UserClipboardSendAction extends sfAction
 {
   public function execute($request)
   {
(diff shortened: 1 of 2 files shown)


djjuhasz (commented):
The new changes look fine to me, but shouldn't the whole thing fail if  `sfConfig::get('app_clipboard_send_url')` is not set?  I can't see how this would work without a destination URL?

mcantelon:
Yeah the "Send" button won't be shown unless `sfConfig::get('app_clipboard_send_url')` is set:

https://github.com/artefactual/atom/blob/qa/2.6.x/apps/qubit/modules/user/templates/clipboardSuccess.php#L55

mcantelon:
I've updated the PR to add a check.

djjuhasz (approved):
LGTM! 👍

PR #1004, opened 2019-11-19
Update password hashing to Argon2, refs #12136

mcantelon:
Updates the hashing logic from using SHA1 to using Argon2. To avoid
users having to re-enter their passwords we use Argon2 to hash the
original SHA1 hash of the user's password (plus salt). As Argon2
automatically generates the salt no additional salt-related logic is
needed.

The hashing algorithm and the options used with it can be changed by
editing the config/app.yml file and changing the
password_hash_algorithm and/or password_hash_algorithm_options options.

The hash algorithm and options are stored as part of the hash string
data so changing hash algorithm and options won't stop previously
stored password hashes from working.

### DIFF
--- a/config/app.yml
+++ b/config/app.yml
@@ -45,3 +45,11 @@ all:
   # to a different string value to make sure the proper workers take the job for
   # their related instance.
   workers_key:
+
+  # Specify password_hash algorithm constant
+  # See: https://www.php.net/manual/en/password.constants.php
+  password_hash_algorithm: PASSWORD_ARGON2I
+
+  # Specify password_hash algorithm options (as JSON)
+  # See: https://www.php.net/manual/en/function.password-hash.php
+  password_hash_algorithm_options: {"memory_cost": "2048", "time_cost": "4", "threads": "3"}
--- a/config/schema.yml
+++ b/config/schema.yml
@@ -417,6 +417,6 @@ propel:
     id: { type: integer, required: true, primaryKey: true, foreignTable: actor, foreignReference: id, onDelete: cascade, inheritanceKey: true }
     username: varchar(255)
     email: varchar(255)
-    sha1_password: varchar(255)
+    password_hash: varchar(255)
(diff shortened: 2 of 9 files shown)


mcantelon:
I've set the hashing algorithm to be PASSWORD_ARGON2I (PHP 7.2 compatible) but I could, if 2.6 is going to be PHP 7.3+, set it to the newer PASSWORD_ARGON2ID algorithm.

config/app.yml
```diff
@@ -45,3 +45,11 @@ all:
...
   # their related instance.
   workers_key:
+
+  # Specify password_hash algorithm constant
+  # See: https://www.php.net/manual/en/password.constants.php
+  password_hash_algorithm: PASSWORD_ARGON2I
+
+  # Specify password_hash algorithm options (as JSON)
+  # See: https://www.php.net/manual/en/function.password-hash.php
+  password_hash_algorithm_options: {"memory_cost": "2048", "time_cost": "4", "threads": "3"}
```
djjuhasz, line 55: Out of curiosity, how did you chose these values for cost? :)

djjuhasz (approved):
Looks good @mcantelon! 👍

djjuhasz:
> I've set the hashing algorithm to be PASSWORD_ARGON2I (PHP 7.2 compatible) but I could, if 2.6 is going to be PHP 7.3+, set it to the newer PASSWORD_ARGON2ID algorithm.

I think ARGON2I is the best choice right now @mcantelon.  If we do decide to target 7.3+ for the AtoM 2.6 release we can easily change the algo for new passwords.

config/app.yml
mcantelon: I just copied them from some blog post. Hopefully it's strong enough.

djjuhasz: Probably fine, I have no idea what reasonable values are. 🤷‍♂ 😄

PR #1005, opened 2019-11-21
Full treeview fix and paging setting, refs #13186

mcantelon:
Fixed issue with full treeview not scrolling to the active treeview
node. Made the number of nodes shown, per treeview page, configurable
via a setting.


sbreker (approved):
Looks good @mcantelon!

PR #1007, opened 2019-11-25
Install composer on Docker, refs #13183

jraddaoui:
- Copy composer executable from composer image.
- Allow to run composer as root.
- Install composer dependencies on build.
- Create named volume on docker-compose to avoid overwrite.


djjuhasz (approved):
Looks good to me :)

PR #1008, opened 2019-11-25
Fix issue with actor event pager, refs #13157

mcantelon:
Fixed an issue with the actor event pager fetching events during
creation of actors rather than when editing.


sbreker (approved):
Looks good @mcantelon!

PR #1010, opened 2019-11-25
Fix PHP version check/bump version, refs #12136

mcantelon:
Fix PHP version check logic. Bump version to 7.2.

### DIFF
--- a/config/package.xml
+++ b/config/package.xml
@@ -57,7 +57,7 @@ The goal is to provide an easy-to-use, flexible toolkit that complies with open 
   <dependencies>
     <required>
       <php>
-        <min>5.3.10</min>
+        <min>7.2.0</min>
       </php>
       <pearinstaller>
         <min>1.4.1</min>
--- a/plugins/sfInstallPlugin/lib/sfInstall.class.php
+++ b/plugins/sfInstallPlugin/lib/sfInstall.class.php
@@ -45,9 +45,16 @@ class sfInstall
 
     // Check if a minimum PHP version is defined, and if it is less than our
     // current version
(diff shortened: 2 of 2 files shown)


config/package.xml
```diff
@@ -57,7 +57,7 @@ The goal is to provide an easy-to-use, flexible toolkit that complies with open
   <dependencies>
     <required>
       <php>
-        <min>5.3.10</min>
+        <min>7.2.0</min>
```
jraddaoui, line 60: Why 7.2? I'm not sure if we ever dropped 5.x support on the latest upgrades (@sbreker?), but I think we should at least allow >= 7.0.

mcantelon: Argon2 hashing support was added in 7.2 (https://wiki.php.net/rfc/argon2_password_hash) and @djjuhasz said 7.2 was going to be the next version, but maybe we should talk further about about possible issues with this jump.

djjuhasz: @jraddaoui We added support for PHP 7.2 in the AtoM 2.5 Release, in part because it's the default version of PHP installed in Ubuntu 18.04 so it's well tested and known to work.   

PHP 5.6 (the final 5.x release) has been unsupported since Dec. 2018 and PHP 7.1 is only receiving security updates until 1 Dec 2019 (2 days from now).    I think we should drop support for both in the AtoM 2.6 release.

I would like to add support for PHP 7.2 - 7.4 with the AtoM 2.6 release, but I'm not sure whether support for >7.2 will require updates to the code.

djjuhasz: Refs: 
* [PHP supported versions](https://www.php.net/supported-versions.php)
* [PHP end-of-life versions](https://www.php.net/eol.php)

PR #1012, opened 2019-12-02
Remove identifier etc. Markdown, refs #13173

mcantelon:
Got rid of more Markdown parsing of identifiers, accession numbers,
and deaccession numbers.

### DIFF
--- a/plugins/qtAccessionPlugin/modules/accession/templates/indexSuccess.php
+++ b/plugins/qtAccessionPlugin/modules/accession/templates/indexSuccess.php
@@ -23,7 +23,7 @@
 
 <?php end_slot() ?>
 
-<?php echo render_show(__('Accession number'), render_value($resource->identifier)) ?>
+<?php echo render_show(__('Accession number'), $resource->identifier) ?>
 
 <?php echo render_show(__('Acquisition date'), render_value(Qubit::renderDate($resource->date))) ?>
 
@@ -160,7 +160,7 @@
     <div class="field">
       <h3><?php echo __('Deaccession') ?></h3>
       <div>
-        <?php echo link_to(render_title($item), array($item, 'module' => 'deaccession')) ?>
+        <?php echo link_to(render_title($item, false), array($item, 'module' => 'deaccession')) ?>
       </div>
     </div>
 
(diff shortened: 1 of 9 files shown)


plugins/qtAccessionPlugin/modules/accession/templates/indexSuccess.php
```diff
@@ -160,7 +160,7 @@
     <div class="field">
       <h3><?php echo __('Deaccession') ?></h3>
       <div>
-        <?php echo link_to(render_title($item), array($item, 'module' => 'deaccession')) ?>
+        <?php echo link_to($item, array($item, 'module' => 'deaccession')) ?>
```
djjuhasz: Removing `render_title()` here is changing the behaviour of the application, beyond just removing markdown support, as render_title substitutes an `<em>Untitled</em>` value when the db value is empty.  Is this your intention in removing `render_title()`?

I question whether we need to allow markdown in titles in general or could remove the `render_inline_value()` here: https://github.com/artefactual/atom/blob/qa/2.6.x/lib/helper/QubitHelper.php#L128 ?  I guess markdown can be handy for emphasis in the title in some cases. If we do want to keep markdown support for `render_title()` maybe we need to introduce an option parameter to the function to allow turning off markdown support?

plugins/qtAccessionPlugin/modules/deaccession/templates/indexSuccess.php
```diff
@@ -26,11 +26,11 @@
 <div class="field">
   <h3>Accession record</h3>
   <div class="value">
-    <?php echo link_to(render_title($resource->accession), array($resource->accession, 'module' => 'accession')) ?>
+    <?php echo link_to($resource->accession, array($resource->accession, 'module' => 'accession')) ?>
```
djjuhasz: See previous note about `render_title()`

djjuhasz (commented):
@mcantelon I've added a question about removing `render_title()` for the accession and deaccession "titles".

plugins/qtAccessionPlugin/modules/accession/templates/indexSuccess.php
mcantelon: Hrm... adding an option to allow turning off Markdown support seems like it might be the way to go. I'll update this PR with that.

mcantelon:
I've updated the PR to address the feedback!

djjuhasz (approved):
Looks good to me 👍

PR #1013, opened 2019-12-03
Fix regen derivatives query logic, refs #13196

mcantelon:
Fixed issue with the regenerate derivatives tasks where an error would
result if one used both the --slug and --no-overwrite CLI options.


sbreker (approved):
Looks good to me @mcantelon! Nice fix!

PR #1014, opened 2019-12-03
Use cached XML in IO XML export job, refs #12998

mcantelon:
Make use of XML cache files when exporting XML via the information
object XML export job.

### DIFF
--- a/lib/job/arInformationObjectXmlExportJob.class.php
+++ b/lib/job/arInformationObjectXmlExportJob.class.php
@@ -101,6 +101,8 @@ class arInformationObjectXmlExportJob extends arBaseJob
 
     $search = QubitSearch::getInstance()->index->getType('QubitInformationObject')->createSearch($this->search->getQuery(false, false));
 
+    $xmlCachingEnabled = sfConfig::get('app_cache_xml_on_save', false);
+
     // Scroll through results then iterate through resulting IDs
     foreach (arElasticSearchPluginUtil::getScrolledSearchResultIdentifiers($search) as $id)
     {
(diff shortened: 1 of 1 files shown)


lib/job/arInformationObjectXmlExportJob.class.php
```diff
@@ -102,6 +102,8 @@ protected function exportResults($path)
     $search = QubitSearch::getInstance()->index->getType('QubitInformationObject')->createSearch($this->search->getQuery(false, false));
 
     // Scroll through results then iterate through resulting IDs
```
sbreker: Nitpick: Comment should be moved down to the foreach line.

lib/job/arInformationObjectXmlExportJob.class.php
```diff
@@ -136,26 +138,36 @@ protected function exportResults($path)
...
+  protected function exportResource($resource, $path, $useCacheFiles)
   {
-    try
-    {
-      // Print warnings/notices here too, as they are often important.
-      $errLevel = error_reporting(E_ALL);
-
-      $rawXml = exportBulkBaseTask::captureResourceExportTemplateOutput($resource, $this->params['format'], $this->params);
-      $xml = Qubit::tidyXml($rawXml);
+    $cachedXmlFilePath = QubitInformationObjectXmlCache::resourceExportFilePath($resource, $this->params['format']);
```
sbreker: Does this need to be done if $useCacheFiles is False?

sbreker (commented):
Looks good @mcantelon! One nitpick and one question.

lib/job/arInformationObjectXmlExportJob.class.php
mcantelon: Hmm yeah it doesn't. I've redone this logic to avoid this.

mcantelon:
Thanks for the feedback @sbreker ... I've updated the PR!

sbreker (approved):
Looks great @mcantelon! Thanks for making those couple changes!

PR #1019, opened 2019-12-18
Fix to EAD repository name import, refs #6317

mcantelon:
Support, as per the EAD standard, the ability to specify a repository
name using text directly inside a "repository" element (rather than in
a "corpname" or "name" element nested in a "repository" element).

Added, in order to support this, an XML import configuration property:
"IgnoreChildElementText".

Normally if an xpath in a format's XML import configuration selects
an element that contains child elements then the text from the child
elements will get added to the selected element's text. This, for
example, enables the EAD configuration's "did/repository/address"
selector to incorporate the text of all child "addressline" elements.
The "IgnoreChildElementText" configuration property prevents this.

### DIFF
--- a/apps/qubit/modules/object/config/import/ead.yml
+++ b/apps/qubit/modules/object/config/import/ead.yml
@@ -62,8 +62,17 @@ information_object:
       Method: SetLevelOfDescriptionByName
 
     repository:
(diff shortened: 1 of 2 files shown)


sbreker (approved):
Looks good @mcantelon!

PR #1022, opened 2019-12-30
Improve hierarchy management queries, refs #13224

jraddaoui:
**Use CTE to fetch ancestors:**

Modify the criteria from `addAncestorsCriteria` to use a recursive
CTE in order the fetch the resource's ancestors without using the
nested set columns.

**Avoid double ancestors query on ACL:**

Calling count and then looping over the results of a criteria causes the
execution of a similar query twice, once for the count and another to
fetch the objects.

In this case, the count is not needed as the ancestors criteria is
followed by `andSelf()`, which will always return at least one result.

**Use only LFT on descendants queries:**

On a nested set implementation, the LFT and RGT values of each record
will not intersect, this means that if the LFT value is in between the
LFT and RGT values of another record you could assume that the RGT
value will also be in between those values.

Using only the LFT column to find descendants on the nested set avoids
the scan over the RGT column, resulting on a performance improvement.

**Avoid query to count descendants:**

Trust on the LFT and RGT values to calculate the descendants count.

---

A full report of this enhancements can be found on the Redmine ticket:

https://projects.artefactual.com/issues/13224

### DIFF
--- a/apps/qubit/modules/term/templates/deleteSuccess.php
+++ b/apps/qubit/modules/term/templates/deleteSuccess.php
@@ -30,8 +30,8 @@
         </div>
       <?php endif; ?>
 
-      <?php if (0 < count($resource->descendants)): ?>
-        <h2><?php echo __('It has %1% descendants that will also be deleted', array('%1%' => count($resource->descendants))) ?><h2>
+      <?php if (0 < $descendantsCount): ?>
+        <h2><?php echo __('It has %1% descendants that will also be deleted', array('%1%' => $descendantsCount)) ?><h2>
         <div class="delete-list">
           <ul>
             <?php foreach ($resource->descendants as $item): ?>
--- a/lib/model/om/BaseActor.php
+++ b/lib/model/om/BaseActor.php
@@ -665,7 +665,18 @@ abstract class BaseActor extends QubitObject implements ArrayAccess
 
   public function addDescendantsCriteria(Criteria $criteria)
   {
-    return $criteria->add(QubitActor::LFT, $this->lft, Criteria::GREATER_THAN)->add(QubitActor::RGT, $this->rgt, Criteria::LESS_THAN);
+    $criterion = $criteria->getNewCriterion(
+      QubitActor::LFT,
+      $this->lft,
+      Criteria::GREATER_THAN
+    );
+    $criterion->addAnd($criteria->getNewCriterion(
+      QubitActor::LFT,
+      $this->rgt,
+      Criteria::LESS_THAN
+    ));
+
+    return $criteria->add($criterion);
   }
 
   protected function updateNestedSet($connection = null)
--- a/lib/model/om/BaseActor.php
+++ b/lib/model/om/BaseActor.php
@@ -641,7 +641,26 @@ abstract class BaseActor extends QubitObject implements ArrayAccess
 
   public function addAncestorsCriteria(Criteria $criteria)
   {
-    return $criteria->add(QubitActor::LFT, $this->lft, Criteria::LESS_THAN)->add(QubitActor::RGT, $this->rgt, Criteria::GREATER_THAN);
+    if (isset($this->parentId))
+    {
+      $condition = '= '.$this->parentId;
+    }
+    else
+    {
+      $condition = 'IS NULL';
+    }
+
+    $subquery = "actor.id IN (
+    	WITH RECURSIVE cte AS
+    	(
+    	  SELECT tb1.id, tb1.parent_id FROM actor tb1 WHERE tb1.id $condition
+    	  UNION ALL
+    	  SELECT tb2.id, tb2.parent_id FROM actor tb2 JOIN cte ON cte.parent_id=tb2.id
+    	)
+    	SELECT id FROM cte
+    )";
+
+    return $criteria->add('', $subquery, Criteria::CUSTOM);
   }
 
   public function addDescendantsCriteria(Criteria $criteria)
--- a/lib/propel/builder/QubitObjectBuilder.php
+++ b/lib/propel/builder/QubitObjectBuilder.php
@@ -1929,11 +1929,35 @@ script;
 
   protected function addAddAncestorsCriteria(&$script)
   {
+    $selfFkTableName = $this->selfFk->getTableName();
+    $selfFkColName = $this->selfFk->getLocalColumns()[0];
+    $selfFkRefColName = $this->selfFk->getForeignColumns()[0];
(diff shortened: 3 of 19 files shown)


apps/qubit/modules/term/templates/deleteSuccess.php
```diff
@@ -30,7 +30,7 @@
         </div>
       <?php endif; ?>
 
-      <?php if (0 < count($resource->descendants)): ?>
+      <?php if (0 < ($resource->rgt - $resource->lft - 1) / 2): ?>
```
djjuhasz: I think `<?php if (0 < ($resource->rgt - $resource->lft - 1)): ?>` (without dividing by 2) is sufficient to test for descendants here.

lib/model/om/BaseActor.php
```diff
@@ -641,12 +641,42 @@ public function hasChildren()
 
   public function addAncestorsCriteria(Criteria $criteria)
   {
-    return $criteria->add(QubitActor::LFT, $this->lft, Criteria::LESS_THAN)->add(QubitActor::RGT, $this->rgt, Criteria::GREATER_THAN);
+    if (isset($this->parentId))
```
djjuhasz: If `!isset($this->parentId)` then the object does not have any ancestors by definition, in which case I think `return $this` (the self node) is sufficient?

lib/model/om/BaseActor.php
```diff
@@ -641,12 +641,42 @@ public function hasChildren()
...
+    {
+      $condition = '= '.$this->parentId;
+    }
+    else
+    {
+      $condition = 'IS NULL';
```
djjuhasz, line 650: If `!isset($this->parentId)` then the object does not have any ancestors by definition, in which case I think `return $this;` (the self node) is sufficient?

lib/propel/builder/QubitObjectBuilder.php
```diff
@@ -1929,11 +1929,35 @@ public function hasChildren()
...
   public function addAncestorsCriteria(Criteria \$criteria)
   {
-    return \$criteria->add({$this->getColumnConstant($this->nestedSetLeftColumn)}, \$this->{$this->getColumnVarName($this->nestedSetLeftColumn)}, Criteria::LESS_THAN)->add({$this->getColumnConstant($this->nestedSetRightColumn)}, \$this->{$this->getColumnVarName($this->nestedSetRightColumn)}, Criteria::GREATER_THAN);
+    if (isset(\$this->{$selfFkPhpName}))
+    {
+      \$condition = '= '.\$this->{$selfFkPhpName};
+    }
+    else
+    {
+      \$condition = 'IS NULL';
```
djjuhasz, line 1947: Again, I think `return \$this;` is better here.

djjuhasz (changes requested):
Looks great @jraddaoui!  I just noted a couple of minor changes that I think would make it even better. :)

apps/qubit/modules/term/templates/deleteSuccess.php
jraddaoui: Good point @djjuhasz, I was going to address that and noticed that the actual count is shown just below, so I've moved the calculation to the action and used it for both cases (like it's done on IO delete action and template).

lib/model/om/BaseActor.php
jraddaoui: Hi @djjuhasz, I struggled a bit whit this too.

We're working with a criteria object on this function, adding a condition and returning that criteria, which is used outside to fetch the resources. At the moment, this function is only used from it's own class but, being public that may change and I didn't want to change its signature.

Returning the criteria directly without applying any condition to it when the resource has no parent will not filter the required resources (all) when the results are fetched. We could use a different condition if the resource has no parent to avoid using the CTE, but the results will be the same.

Moreover, this is quite an edge case as we use a "protected" root resource which should be the parent of all the "top-level" resources on the nested set, so the parent id should only be null for that record.

apps/qubit/modules/term/templates/deleteSuccess.php
djjuhasz: Sounds good @jraddaoui.  I can't see your change to the calculation in the Github UI, I'm not sure why. :(

lib/model/om/BaseActor.php
djjuhasz: Ah okay, then I think you are right that this is okay as is. :)

djjuhasz (approved):
I can't see the one change for some reason, but other than that it all looks great! 👍

apps/qubit/modules/term/templates/deleteSuccess.php
jraddaoui: https://github.com/artefactual/atom/pull/1022/commits/e5fb1283dd99128a589d683b6318b09633efb3e6

plugins/qbAclPlugin/lib/QubitAcl.class.php
```diff
@@ -343,9 +343,9 @@ protected function buildResourceList($resource, $options = array())
     }
 
     // Add resource hierarchy
-    if (is_object($resource) && 0 < count($resources = $resource->ancestors->andSelf()->orderBy('lft')))
+    if (is_object($resource))
     {
-      foreach ($resources as $r)
+      foreach ($resource->ancestors->andSelf()->orderBy('lft') as $r)
```
sbreker, line 348: This is awesome! Great catch @jraddaoui!

lib/propel/builder/QubitObjectBuilder.php
```diff
@@ -1929,11 +1929,35 @@ public function hasChildren()
...
+    	WITH RECURSIVE cte AS
+    	(
+    	  SELECT tb1.{$selfFkRefColName}, tb1.{$selfFkColName} FROM {$selfFkTableName} tb1 WHERE tb1.{$selfFkRefColName} \$condition
+    	  UNION ALL
+    	  SELECT tb2.{$selfFkRefColName}, tb2.{$selfFkColName} FROM {$selfFkTableName} tb2 JOIN cte ON cte.{$selfFkColName}=tb2.{$selfFkRefColName}
+    	)
+    	SELECT {$selfFkRefColName} FROM cte
+    )";
+
+    return \$criteria->add('', \$subquery, Criteria::CUSTOM);
```
sbreker, line 1960: Great work!

sbreker (approved):
This is great @jraddaoui! Looking forward to trying out the ancestors/ACL changes!

jraddaoui:
Thanks @sbreker and @djjuhasz!

PR #1024, opened 2020-01-07
Prevent actor import dupe relations, refs #13148

mcantelon:
Added checks for pre-existing relations during authority record
relation import.

### DIFF
--- a/lib/task/import/csvAuthorityRecordImportTask.class.php
+++ b/lib/task/import/csvAuthorityRecordImportTask.class.php
@@ -535,29 +535,53 @@ EOF;
               }
               else
               {
-                $relation = new QubitRelation;
-                $relation->subjectId = $sourceActorId;
-                $relation->objectId  = $targetActorId;
-                $relation->typeId    = $relationTypeId;
+                // Check for pre-existing relationship before adding
+                $sql = "SELECT id FROM relation \r
+                  WHERE subject_id = :subject_id \r
+                  AND object_id = :object_id \r
+                  AND type_id = :type_id";
 
-                if ($self->rowStatusVars['date'])
-                {
-                  $relation->date = $self->rowStatusVars['date'];
-                }
-                if ($self->rowStatusVars['startDate'])
-                {
-                  $relation->startDate = $self->rowStatusVars['startDate'];
-                }
-                if ($self->rowStatusVars['endDate'])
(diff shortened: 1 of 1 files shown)


lib/task/import/csvAuthorityRecordImportTask.class.php
```diff
@@ -535,29 +535,53 @@ public function execute($arguments = array(), $options = array())
...
-                if ($self->rowStatusVars['startDate'])
-                {
-                  $relation->startDate = $self->rowStatusVars['startDate'];
-                }
-                if ($self->rowStatusVars['endDate'])
+                // Check for pre-existing relationship before adding
+                $sql = "SELECT id FROM relation \r
+                  WHERE subject_id=:subject_id \r
+                  AND object_id=:object_id \r
+                  AND type_id = :type_id";
```
sbreker, line 542: Nitpick: Extra spaces around '=' compared to prev lines.

lib/task/import/csvAuthorityRecordImportTask.class.php
```diff
@@ -535,29 +535,53 @@ public function execute($arguments = array(), $options = array())
...
+                  ':subject_id' => $sourceActorId,
+                  ':object_id' => $targetActorId,
+                  ':type_id' => $relationTypeId
+                );
+
+                $paramsVariant = array(
+                  ':subject_id' => $targetActorId,
+                  ':object_id' => $sourceActorId,
+                  ':type_id' => $relationTypeId
+                );
```
sbreker, line 554: Does it matter that this relation is the opposite of the one being created? Or is this the source of the duped relations?

lib/task/import/csvAuthorityRecordImportTask.class.php
mcantelon: Good spotting... I've added the spaces to all of them as it makes it a bit easier to read.

lib/task/import/csvAuthorityRecordImportTask.class.php
mcantelon: I think a relationship can be specified in two ways, so I'm checking for both.

mcantelon:
I've updated the PR... thanks @sbreker !

sbreker (approved):

PR #1025, opened 2020-01-09
Fix build SQL culture column sizing, refs #13222

mcantelon:
Changed logic, that renders the AtoM schema to XML, so culture columns
have a size of 16 rather than 7.

### DIFF
--- a/vendor/symfony/lib/plugins/sfPropelPlugin/lib/addon/sfPropelDatabaseSchema.class.php
+++ b/vendor/symfony/lib/plugins/sfPropelPlugin/lib/addon/sfPropelDatabaseSchema.class.php
@@ -589,7 +589,7 @@ class sfPropelDatabaseSchema
 	  // Add source_culture column to main table
 	  $this->setIfNotSet($this->database[$main_table], 'source_culture', array(
 	    'required' => true,
-	    'size' => 7,
+	    'size' => 16,
 	    'type' => 'varchar'));
 
           // set id and culture columns for i18n table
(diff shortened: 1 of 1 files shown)


djjuhasz (approved):
Looks good 👍

vendor/symfony/lib/plugins/sfPropelPlugin/lib/addon/sfPropelDatabaseSchema.class.php
```diff
@@ -604,7 +604,7 @@ protected function fixYAMLI18n()
           $this->setIfNotSet($this->database[$i18n_table], 'culture', array(
             'isCulture'  => true,
             'type'       => 'varchar',
-            'size'       => '7',
+            'size'       => 16,
```
jraddaoui, line 607: Don't really know why or if it matters but this size was set as a string (with quotes).

jraddaoui (approved):
Looks good @mcantelon!

As mentioned in the Redmine ticket, I don't really know if this change is the way to go or if it conflicts with Symfony's culture validation and data folders. But we can figure out that on another PR.

Thanks!

vendor/symfony/lib/plugins/sfPropelPlugin/lib/addon/sfPropelDatabaseSchema.class.php
mcantelon: Yeah, one was set to an int an one was set to a string so I went with int. Seems to work.

mcantelon:
Thanks guys!

PR #1026, opened 2020-01-16
Optionally show ref code in DC XML, refs #13244

mcantelon:
Added a setting that allows the user to choose to show the reference
code, instead of the local identifier (default), in Dublin Core XML.


jraddaoui (approved):
Thanks @mcantelon!

PR #1027, opened 2020-01-17
Fix D.O. linking page issue, refs #13243

mcantelon:
Fix D.O. linking page issue, refs #13243
    
    This fix was made to reintroduce functionality that had stopped working
    due to two issues. The functionality lost was display of a resource's
    identifier and title on the AtoM page used to link digital objects to a
    resource.
    
    The two issues that led to the regression:
    
    1. Use of "instanceof" in HTML templates works differently than it used
    to due to the escaping strategy now "wrapping" objects in templates
    (containing them in sfOutputEscaperObjectDecorator objects).
    
    2. Text returned by casting an sfIsadPlugin object as a string was
    changed in a previous commit. Casting an sfIsadPlugin object as a
    string used to result in a string that included the identifier and
    title of a resource, but this has changed. Doing so now doesn't include
    the identifier. This change was made so that title can be parsed as Markdown
    but not the identifier.
    
    The fixes:
    
    Fixing issue 1 involved moving functionality relying on "instanceof"
    from the HTML template into the Symfony comment corresponding to it.
    
    Fixing issue 2 involved adding additional logic to the Symphony
    component to add the identifier to the text description of the resource
    created by casting an sfIsadPlugin object as a string.

### DIFF
--- a/apps/qubit/modules/object/actions/addDigitalObjectAction.class.php
+++ b/apps/qubit/modules/object/actions/addDigitalObjectAction.class.php
@@ -69,6 +69,25 @@ class ObjectAddDigitalObjectAction extends sfAction
       $this->forward404();
     }
 
+    // Assemble resource description
+    sfContext::getInstance()->getConfiguration()->loadHelpers(array('Qubit'));
+
+    if ($this->resource instanceof QubitActor)
+    {
+       $this->resourceDescription = render_title($this->resource);
+    }
+    elseif ($this->resource instanceof QubitInformationObject)
+    {
+      $this->resourceDescription = '';
+
+      if (isset($this->resource->identifier))
+      {
+        $this->resourceDescription .= $this->resource->identifier . ' - ';
+      }
+
+      $this->resourceDescription .= render_title(new sfIsadPlugin($this->resource));
+    }
+
     // Check if already exists a digital object
     if (null !== $digitalObject = $this->resource->getDigitalObjectRelatedByobjectId())
     {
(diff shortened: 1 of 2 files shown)


jraddaoui (commented):
Thanks @mcantelon!

I'm not sure I understand the commit description. It looks to me like the issue in here was the escaping strategy, which is again enabled by default in all cases (Markdown enabled or not), making `$resource` an instance of `sfOutputEscaperObjectDecorator` instead of `QubitActor` or `QubitInformationObject`. If that's the case, I'd suggest you change the commit description for future reference.

mcantelon:
Thank @radda ! Yeah, the commit message was pretty vague... I've added more detail to clarify the nature of the regression that the commit fixes

apps/qubit/modules/object/actions/addDigitalObjectAction.class.php
```diff
@@ -69,6 +69,25 @@ public function execute($request)
...
+    {
+       $this->resourceDescription = render_title($this->resource);
+    }
+    elseif ($this->resource instanceof QubitInformationObject)
+    {
+      $this->resourceDescription = '';
+
+      if (isset($this->resource->identifier))
+      {
+        $this->resourceDescription .= $this->resource->identifier .' - ';
```
jraddaoui: I think we adopted Symfony's style for the `.` separator (no spaces around it) but I personally prefer spaces across all operators. Either way, one or the other ;)

jraddaoui (approved):
Thanks @mcantelon, I see the identifier thing now ;)

I added a minor comment below and, in the commit message:

> ... from the HTML template into the Symfony comment corresponding to it.

I think you mean "Symfony action".

apps/qubit/modules/object/actions/addDigitalObjectAction.class.php
mcantelon: Cool... I'll do the spaces around them.

PR #1028, opened 2020-01-17
Copy file on DO load task, refs #13236 and #13245

jraddaoui:
Using the file path to create the QubitAsset avoids loading the entire
file on memory with `file_get_contents` and directly copies the file to
the uploads folder. It also avoids the use of different hashing
functions to validate the copied file.

---

We'll use https://projects.artefactual.com/issues/13283 to fix the issue in other places.


mcantelon (approved):
Looks good to me! 👍

jraddaoui:
Thanks @mcantelon!

PR #1029, opened 2020-01-20
Add limit option to XML caching task, refs #13246

mcantelon:
Add a limit option to the cache:xml-representations task so the number
of information objetcs to be exported to XML, as cache files, can be
limited.


mcantelon:
I'm going to add something to this PR.

PR #1032, opened 2020-01-22
Fix IO date sorting, refs #13230

jraddaoui:
Add `startDateSort` and `enDateSort` fields to the ES index for IOs
and use them to sort where needed.


mcantelon (approved):
Looks good to me!

jraddaoui:
Thanks @mcantelon!

PR #1034, opened 2020-01-23
Add limit option to XML caching task, refs #13246

mcantelon:
Add a limit option to the cache:xml-representations task so the number
of information objects to be exported to XML, as cache files, can be
limited.

Fetch information objects to export using SQL, rather than the ORM,
in order to speed up export and minimize memory usage.

Keep memory caches clear to prevent excessive memory use.

### DIFF
--- a/lib/QubitInformationObjectXmlCache.class.php
+++ b/lib/QubitInformationObjectXmlCache.class.php
@@ -84,27 +84,43 @@ class QubitInformationObjectXmlCache
   public function exportAll($options = array())
   {
     $skip = isset($options['skip']) ? $options['skip'] : 0;
+    $limit = isset($options['limit']) ? $options['limit'] : 0;
 
-    // Get not-root and published information objects
-    $criteria = new Criteria;
-    $criteria->add(QubitInformationObject::ID, QubitInformationObject::ROOT_ID, Criteria::NOT_EQUAL);
-    $criteria = QubitAcl::addFilterDraftsCriteria($criteria);
-    $criteria->addAscendingOrderByColumn(QubitInformationObject::ID); // sort so optional skip will be consistent
-    $criteria->setOffset($skip);
-
-    $results = QubitInformationObject::get($criteria);
+    // Get not-root and published information objects (sorted by ID so optional
+    // skip/limit will be consistent)
+    $sql = "SELECT i.id FROM ". QubitInformationObject::TABLE_NAME ." i \r
+      INNER JOIN status s ON i.id=s.object_id \r
+      WHERE i.id != :root_id \r
+      AND s.status_id=:status_id AND s.type_id=:type_id \r
+      ORDER BY i.id";
+ 
+    $params = array(
+      ':status_id' => QubitTerm::PUBLICATION_STATUS_PUBLISHED_ID,
+      ':type_id' => QubitTerm::STATUS_TYPE_PUBLICATION_ID,
+      ':root_id' => QubitInformationObject::ROOT_ID,
+    );
 
     $exporting = 0;
+
+    // Apply skip/limit and fetch results
+    Propel::getDB()->applyLimit($sql, $skip, $limit);
+    $results = QubitPdo::fetchAll($sql, $params);
 
     if (count($results))
     {
       $this->logger->info($this->i18n->__('%1% published information objects found.', array('%1%' => count($results))));
 
-      foreach ($results as $io)
+      // Export each information object
+      foreach ($results as $row)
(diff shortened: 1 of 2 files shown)


lib/QubitInformationObjectXmlCache.class.php
```diff
@@ -84,27 +84,43 @@ public function export($resource)
...
+    $params = array(
+      ':status_id' => QubitTerm::PUBLICATION_STATUS_PUBLISHED_ID,
+      ':type_id' => QubitTerm::STATUS_TYPE_PUBLICATION_ID,
+      ':root_id' => QubitInformationObject::ROOT_ID,
+    );
 
     $exporting = 0;
 
+    // Apply skip/limit and fetch results
+    Propel::getDB()->applyLimit($sql, $skip, $limit);
```
jraddaoui, line 106: I didn't know about this ;)

lib/QubitInformationObjectXmlCache.class.php
```diff
@@ -84,27 +84,43 @@ public function export($resource)
...
-
-    // Get not-root and published information objects
-    $criteria = new Criteria;
-    $criteria->add(QubitInformationObject::ID, QubitInformationObject::ROOT_ID, Criteria::NOT_EQUAL);
-    $criteria = QubitAcl::addFilterDraftsCriteria($criteria);
-    $criteria->addAscendingOrderByColumn(QubitInformationObject::ID); // sort so optional skip will be consistent
-    $criteria->setOffset($skip);
-
-    $results = QubitInformationObject::get($criteria);
+    $limit = isset($options['limit']) ? $options['limit'] : 0;
```
jraddaoui, line 87: This is probably unnecessary since you're adding a default value to the option.

lib/task/arCacheDescriptionXmlTask.class.php
```diff
@@ -32,7 +32,8 @@ protected function configure()
       new sfCommandOption('application', null, sfCommandOption::PARAMETER_OPTIONAL, 'The application name', 'qubit'),
       new sfCommandOption('env', null, sfCommandOption::PARAMETER_REQUIRED, 'The environment', 'cli'),
       new sfCommandOption('connection', null, sfCommandOption::PARAMETER_REQUIRED, 'The connection name', 'propel'),
-      new sfCommandOption('skip', null, sfCommandOption::PARAMETER_OPTIONAL, 'Number of information objects to skip', 0)
+      new sfCommandOption('skip', null, sfCommandOption::PARAMETER_OPTIONAL, 'Number of information objects to skip', 0),
+      new sfCommandOption('limit', null, sfCommandOption::PARAMETER_OPTIONAL, 'Number of information objects to export', 0)
```
jraddaoui: Does `limit = 0` mean no limit using `Propel::getDB()->applyLimit`? Otherwise, I'd change the default.

jraddaoui (changes requested):
Thanks @mcantelon! See my comments below.

lib/task/arCacheDescriptionXmlTask.class.php
mcantelon: It does, yeah. Maybe I'll try changing the default for it to null as that would be a bit clearer.

lib/QubitInformationObjectXmlCache.class.php
mcantelon: Ah true... I'll remove.

jraddaoui: Oh, sorry, I just noticed it's on a different class, it may be a good idea to leave it just in case.

lib/task/arCacheDescriptionXmlTask.class.php
jraddaoui: Maybe it's better to give the option of no limit like it's then.

jraddaoui (approved):

lib/QubitInformationObjectXmlCache.class.php
mcantelon: Yeah, it's new to me too... kind of handy.

lib/QubitInformationObjectXmlCache.class.php
mcantelon: Sounds good.

lib/task/arCacheDescriptionXmlTask.class.php
mcantelon: Seems like it works the same with null as a default.

PR #1038, opened 2020-01-29
Optimize repo bulk updating of I.O.s, refs #13250

mcantelon:
Fetch related information objects using SQL, rather than the ORM, in
order to speed up bulk operations on them and minimize memory usage.
Also keep memory caches clear, during these bulk operations, to
prevent excessive memory use.

### DIFF
--- a/lib/model/QubitRepository.php
+++ b/lib/model/QubitRepository.php
@@ -124,43 +124,54 @@ class QubitRepository extends BaseRepository
     return $this;
   }
 
+  public function getRelatedInformationObjectIds()
+  {
+    $sql = "SELECT id FROM ". QubitInformationObject::TABLE_NAME ." WHERE repository_id=:repository_id";
+
+    $params = array(':repository_id' => $this->id);
+
+    return QubitPdo::fetchAll($sql, $params, array('fetchMode' => PDO::FETCH_COLUMN));
+  }
+
   public function updateRelatedIos()
   {
-    $ioIds = array();
-    $context = sfContext::getInstance();
-    $env = $context->getConfiguration()->getEnvironment();
+    $ioIds = $this->getRelatedInformationObjectIds();
 
-    foreach ($this->informationObjects as $io)
-    {
-      // Update synchronously in CLI tasks and jobs
-      if (in_array($env, array('cli', 'worker')))
-      {
-        QubitSearch::getInstance()->update($io, array('updateDescendants' => true));
-      }
-      else
-      {
-        // Otherwise save id to update asynchronously
-        $ioIds[] = $io->id;
-      }
-    }
-
-    if (count($ioIds) == 0)
+    if (empty($ioIds))
     {
       return;
     }
 
-    // Update asynchronously the saved IOs ids
-    $jobOptions = array(
-      'ioIds' => $ioIds,
-      'updateIos' => true,
-      'updateDescendants' => true
-    );
-    QubitJob::runJob('arUpdateEsIoDocumentsJob', $jobOptions);
+    // Handle web request asynchronously
+    $context = sfContext::getInstance();
 
-    // Let user know related descriptions update has started
-    $jobsUrl = $context->routing->generate(null, array('module' => 'jobs', 'action' => 'browse'));
-    $message = $context->i18n->__('Your repository has been updated. Its related descriptions are being updated asynchronously – check the <a href="%1">job scheduler page</a> for status and details.', array('%1' => $jobsUrl));
-    $context->user->setFlash('notice', $message);
+    if (!in_array($context->getConfiguration()->getEnvironment(), array('cli', 'worker')))
+    {
+      // Let user know related descriptions update has started
+      $jobsUrl = $context->routing->generate(null, array('module' => 'jobs', 'action' => 'browse'));
+      $message = $context->i18n->__('Your repository has been updated. Its related descriptions are being updated asynchronously – check the <a href="%1">job scheduler page</a> for status and details.', array('%1' => $jobsUrl));
+      $context->user->setFlash('notice', $message);
+
+      // Update asynchronously the saved IOs ids
+      $jobOptions = array(
+        'ioIds' => $ioIds,
+        'updateIos' => true,
+        'updateDescendants' => true
+      );
+      QubitJob::runJob('arUpdateEsIoDocumentsJob', $jobOptions);
+
+      return;
+    }
+
+    // Handle CLI and worker requests synchronously
+    foreach ($ioIds as $id)
+    {
+      $io = QubitInformationObject::getById($id);
+      QubitSearch::getInstance()->update($io, array('updateDescendants' => true));
+
+      // Keep caches clear to prevent memory use from ballooning
+      Qubit::clearClassCaches();
+    }
   }
 
   /**
@@ -192,11 +203,16 @@ class QubitRepository extends BaseRepository
     // Remove adv. search repository options from cache
     QubitCache::getInstance()->removePattern('search:list-of-repositories:*');
 
-    foreach ($this->informationObjects as $item)
+    foreach ($this->getRelatedInformationObjectIds() as $id)
     {
+      $item = QubitInformationObject::getById($id);
+
       unset($item->repository);
 
       $item->save();
+
+      // Keep caches clear to prevent memory use from ballooning
+      Qubit::clearClassCaches();
     }
 
     // Events, relations and the Elasticsearch document are deleted in QubitActor


lib/model/QubitRepository.php
```diff
@@ -124,14 +124,32 @@ public function save($connection = null)
     return $this;
   }
 
+  public function getRelatedInformationObjectIds()
+  {
+    $sql = "SELECT id FROM ". QubitInformationObject::TABLE_NAME ." WHERE repository_id=:repository_id";
+
+    $results = QubitPdo::fetchAll($sql, array(':repository_id' => $this->id));
```
sbreker: If you don't need the rows returned as objects, memory can be saved here by using `$results = QubitPdo::fetchAll($sql, $params, array('fetchMode' => PDO::FETCH_ASSOC));`.

lib/model/QubitRepository.php
```diff
@@ -124,14 +124,32 @@ public function save($connection = null)
...
   public function updateRelatedIos()
   {
     $ioIds = array();
     $context = sfContext::getInstance();
     $env = $context->getConfiguration()->getEnvironment();
 
-    foreach ($this->informationObjects as $io)
+    foreach ($this->getRelatedInformationObjectIds() as $id)
     {
+      $io = QubitInformationObject::getById($id);
```
sbreker: In the case where we are going to update asynchronously (a WebUI call that is going to create a list of IO id's to pass to the async job) it looks like we could omit doing $io = QubitInformationObject::getById($id); to rebuild the array. Can we pass the output of getRelatedInformationObjectIds straight into jobOptions?

lib/model/QubitRepository.php
```diff
@@ -142,6 +160,9 @@ public function updateRelatedIos()
         // Otherwise save id to update asynchronously
         $ioIds[] = $io->id;
```
sbreker: This seems unnecessary if we already have the id value.

sbreker (changes requested):
Looks great so far @mcantelon! Added some notes where we might be able to avoid some extra processing.

lib/model/QubitRepository.php
mcantelon: Good call!

mcantelon:
I've updated the PR... thanks!

lib/model/QubitRepository.php
```diff
@@ -124,24 +124,45 @@ public function save($connection = null)
...
+    $results = QubitPdo::fetchAll($sql, $params, array('fetchMode' => PDO::FETCH_ASSOC));
+
+    $ioIds = array();
+
+    foreach ($results as $row)
+    {
+      array_push($ioIds, $row['id']);
+    }
+
+    return $ioIds;
```
jraddaoui: You could get a one-dimensional array directly using `PDO::FETCH_COLUMN` to avoid the loop. I'd guess it returns an empty array if there are no results, but I'd check that out to be sure, in case it returns `false`.

```php
return QubitPdo::fetchAll($sql, $params, array('fetchMode' => PDO::FETCH_COLUMN));
```

lib/model/QubitRepository.php
```diff
@@ -124,24 +124,45 @@ public function save($connection = null)
...
       if (in_array($env, array('cli', 'worker')))
       {
+        $io = QubitInformationObject::getById($id);
         QubitSearch::getInstance()->update($io, array('updateDescendants' => true));
       }
       else
       {
         // Otherwise save id to update asynchronously
-        $ioIds[] = $io->id;
+        $ioIds[] = $id;
```
jraddaoui: As @sbreker said, this is really unnecessary. I introduced this logic, but it's far from ideal ...  🤦

I'd suggest to check the environment first and trigger the job directly as you already have the ids array. In pseudocode ...

```
if not related io ids
{
  return
}

if not in cli and not in worker
{
  launch job
  set flash
  return
}

foreach related io ids
{
  get io
  update
  clear cache
}
```

lib/model/QubitRepository.php
```diff
@@ -192,11 +213,16 @@ public function delete($connection = null)
...
+    foreach ($this->getRelatedInformationObjectIds() as $id)
     {
+      $item = QubitInformationObject::getById($id);
+
       unset($item->repository);
 
       $item->save();
+
+      // Keep caches clear to prevent memory use from ballooning
+      Qubit::clearClassCaches();
```
jraddaoui, line 215: @mcantelon, it's great that you spotted the same issue on the deletion, thanks!!

While this changes will prevent the memory problem, the process is still prone to timeouts. Also, it looks like this will trigger a job for each related IO with descendants, when it could be done in a single one (like on the save method).

I understand there are time constrains for this fix and the repository deletion (specially with a lot of related IOs) is not as normal as its modification, so I'll create a related ticket for this particular matter.

jraddaoui (changes requested):
Thanks for the quick turnaround @mcantelon and @sbreker!

Please, see my comments below.

lib/model/QubitRepository.php
jraddaoui: Actually, we could use the same one so I've added a note and modified the title of https://projects.artefactual.com/issues/13250, but no need to do it on this PR.

lib/model/QubitRepository.php
djjuhasz: @jraddaoui awesome! I don't think I've used PDO::FETCH_COLUMN before, but this seems like an ideal use case. 😁

mcantelon: Nice trick. I'll use that.

lib/model/QubitRepository.php
mcantelon: Sounds good!

lib/model/QubitRepository.php
```diff
@@ -124,24 +124,45 @@ public function save($connection = null)
...
     $env = $context->getConfiguration()->getEnvironment();
 
-    foreach ($this->informationObjects as $io)
+    foreach ($this->getRelatedInformationObjectIds() as $id)
     {
       // Update synchronously in CLI tasks and jobs
       if (in_array($env, array('cli', 'worker')))
       {
+        $io = QubitInformationObject::getById($id);
         QubitSearch::getInstance()->update($io, array('updateDescendants' => true));
```
djjuhasz: Could [arElasticSearchPlugin::partialUpdate()](https://github.com/artefactual/atom/blob/qa/2.6.x/plugins/arElasticSearchPlugin/lib/arElasticSearchPlugin.class.php#L473) be used in this context? It may be more efficient then a full update if it's updating a small number of fields.

lib/model/QubitRepository.php
mcantelon: Nice!

lib/model/QubitRepository.php
jraddaoui: That's a great point @djjuhasz. At the moment the partial update function doesn't consider descendants and that's required here. I believe this is kind of urgent, so maybe that fits better on the work you started to reduce the amount of data from related resources in the IO documents. I hope we can improve that soon in the projects that are going on.

mcantelon: Good call. I'm going to make a separate issue for that, though, given it'll take more time to test.

lib/model/QubitRepository.php
```diff
@@ -124,43 +124,54 @@ public function save($connection = null)
     return $this;
   }
 
+  public function getRelatedInformationObjectIds()
+  {
+    $sql = "SELECT id FROM ". QubitInformationObject::TABLE_NAME ." WHERE repository_id=:repository_id";
+
+    $params = array(':repository_id' => $this->id);
+    return QubitPdo::fetchAll($sql, $params, array('fetchMode' => PDO::FETCH_COLUMN));
```
jraddaoui, line 133: Blank line before return.

lib/model/QubitRepository.php
```diff
@@ -124,43 +124,54 @@ public function save($connection = null)
...
+  }
+
   public function updateRelatedIos()
   {
-    $ioIds = array();
     $context = sfContext::getInstance();
     $env = $context->getConfiguration()->getEnvironment();
```
jraddaoui: Real nitpick: the context and env could be obtained after checking the IO ids, to avoid it if there are no related IOs.

jraddaoui (approved):
Awesome @mcantelon! Just a couple of cosmetic nitpicks ;)

djjuhasz (approved):
Looks good! 👍

lib/model/QubitRepository.php
mcantelon: Fixed!

lib/model/QubitRepository.php
mcantelon: True. Moved those and got rid of $env variable as env is only referred to once.

mcantelon:
Thanks folks!

sbreker (approved):
Looks great @mcantelon!

PR #1041, opened 2020-01-31
Make repo bulk delete logic async, refs #13250

mcantelon:
Use the I.O. updating job to handle the updating part of removing the
association between a repository and all related information objects.


jraddaoui (approved):
Looks great, thanks @mcantelon!

PR #1044, opened 2020-02-05
Allow updating in search index task, refs #13253

mcantelon:
Added command line option, to the search index populate task, to allow
updating of existing ElasticSearch documents rather than deleting first
then indexing.


jraddaoui (approved):
Thanks @mcantelon! Great idea! :)

I find a bit confusing the logic around this options. It's more about code readability than its functionality, so not a big deal.

mcantelon:
Thanks @jraddaoui ! Was @sbreker 's idea and definitely a good one!

PR #1046, opened 2020-02-06
Optimize updates via physical object import

djjuhasz:
Improve preformance when updating physical object data via csv importer

- Add --debug (-d) flag to importer and CLI task
- Output timing data for various subroutines when debug flag set
- Add quickUpdate() method to QubitPhysicalObject model
- Improve unit tests and code testability

### DIFF
--- a/lib/PhysicalobjectCsvImporter.class.php
+++ b/lib/PhysicalobjectCsvImporter.class.php
@@ -37,8 +37,6 @@ class PhysicalObjectCsvImporter
     'descriptionSlugs' => 'informationObjectIds',
   ];
 
-  public $timer;
-
   protected $context;
   protected $data;
   protected $dbcon;
@@ -51,10 +49,12 @@ class PhysicalObjectCsvImporter
   protected $reader;
   protected $rowsImported       = 0;
   protected $rowsTotal          = 0;
+  protected $timers;
   protected $typeIdLookupTable;
 
   // Default options
   protected $options = [
+    'debug'               => false,
     'defaultCulture'      => 'en',
     'errorLog'            => null,
     'header'              => null,
@@ -69,6 +69,11 @@ class PhysicalObjectCsvImporter
     'updateExisting'      => false,
     'updateSearchIndex'   => false,
   ];
+
+
+  #
+  # Public methods
+  #
 
   public function __construct(sfContext $context = null, $dbcon = null,
     $options = array())
@@ -150,6 +155,11 @@ class PhysicalObjectCsvImporter
 
   public function validateFilename($filename)
   {
+    if (empty($filename))
+    {
+      throw new sfException('Please specify a filename for import');
+    }
+
     if (!file_exists($filename))
     {
       throw new sfException("Can not find file $filename");
@@ -201,6 +211,7 @@ class PhysicalObjectCsvImporter
         break;
 
       // boolean options
+      case 'debug':
       case 'insertNew':
       case 'overwriteWithEmpty':
       case 'partialMatches':
@@ -304,13 +315,6 @@ EOM;
     }
   }
 
-  public function getRow($offset)
-  {
-    $stmt = (new \League\Csv\Statement)->offset($offset);
-
-    return $this->getRecords($stmt)->fetchOne();
-  }
-
   public function countRowsImported()
   {
     return $this->rowsImported;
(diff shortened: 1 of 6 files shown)


lib/PhysicalobjectCsvImporter.class.php
```diff
@@ -421,56 +441,191 @@ public function savePhysicalobjects($data)
...
+
+    return $output;
+  }
+
+  public function reportTimes()
+  {
+    if (!$this->getOption('debug'))
+    {
+      return sprintf('Total import time: %01.2fs',
+        $this->getElapsedTime('total')).PHP_EOL;
```
mcantelon: Nitpick: I've been putting spaces around the `.` concat operator... not sure if we have a standard for that or not though.

lib/model/QubitPhysicalObject.php
```diff
@@ -326,4 +326,86 @@ public function getRelationsAsArray(int $typeId = null)
...
+    {
+      $connection = QubitTransactionFilter::getConnection();
+    }
+
+    foreach ($updateCols as $name)
+    {
+      if (!empty($data[$name]) && $data[$name] != $this->$name)
+      {
+        $doUpdate = true;
+        $newvals[$name] = $data[$name];
```
mcantelon, line 365: Can't see where this array gets initialized...

lib/model/QubitPhysicalObject.php
```diff
@@ -326,4 +326,86 @@ public function getRelationsAsArray(int $typeId = null)
 
     return $results->fetchAll(PDO::FETCH_ASSOC);
   }
+
+  /**
+   * An quick update of an existing physical_object database row
+   *
+   * If passed data wouldn't alter the current db data, return false without
```
mcantelon, line 334: Nice!

lib/model/QubitPhysicalObject.php
```diff
@@ -326,4 +326,86 @@ public function getRelationsAsArray(int $typeId = null)
...
+        $doUpdate = true;
+        $newvals[$name] = $data[$name];
+      }
+      else
+      {
+        $newvals[$name] = $this->$name;
+      }
+    }
+
+    if (!$doUpdate)
```
mcantelon, line 373: Does this indicate there was an error updating? If so maybe should be an exception? Dunno...

test/phpunit/PhysicalobjectCsvImporterTest.php
```diff
@@ -423,102 +425,29 @@ public function testDoImportNoFilenameException()
...
     $this->assertSame(explode(',', $this->csvHeader), $importer->getHeader());
-    $this->assertSame($this->getCsvRowAsAssocArray(1), $importer->getRow(1));
-    $this->assertSame(2, $importer->countRowsImported());
-    $this->assertSame(4, $importer->countRowsTotal());
+    $this->assertSame($this->getCsvRowAsAssocArray(1), $records->fetchOne());
+    $this->assertSame(3, $importer->countRowsTotal());
   }
 
-  public function testDoImportWithSetHeader()
+  public function testLoadCsvDatWithSetHeader()
```
mcantelon: Nitpick: Is "Dat" short for data or is it something else?

test/phpunit/PhysicalobjectCsvImporterTest.php
```diff
@@ -645,6 +574,67 @@ public function testMatchExistingRecordsThrowsExceptionOnMultiMatch()
...
+      Save physical object: 0.00s
+      Save keymap:          0.00s
+      Update IO relations:  0.00s
+  Progress reporting:       0.00s
+---------------------------------
+Total import time:          0.00s
+
+EOM;
+
+    $this->assertSame($expectedOutput, $importer->reportTimes());
```
mcantelon, line 610: Slick!

mcantelon (approved):
A few nitpicks/questions, but otherwise looks good!

lib/PhysicalobjectCsvImporter.class.php
djjuhasz: [PSR-2-R](https://github.com/php-fig-rectified/fig-rectified-standards/blob/master/PSR-2-R-coding-style-guide-additions.md) recommends a space before and after the concat operator, so I agree we should use that convention.

test/phpunit/PhysicalobjectCsvImporterTest.php
djjuhasz: Yeah, that's a typo. 😊  It should be "Data". :)

lib/model/QubitPhysicalObject.php
djjuhasz: Good catch, I'll add a line to initialize `$newvals` it at the top of the method.

lib/model/QubitPhysicalObject.php
djjuhasz: No, not an error. If `$doUpdate` is false the incoming data matches the data already in the database, so there's no need to do an update.  I'll add a comment to the code to make this clearer.

djjuhasz:
Merged to qa/2.6.x in 8d6ca45b249d5c929f66453b43ad23c60c759aff

PR #1049, opened 2020-02-16
Enhance enabled languages access, refs #13238

jraddaoui:
Accessing the enabled languages through QubitSetting causes an expensive
query because the settings have to be filtered by the `scope` column.
As all the settings, these languages get added to sfConfig for quick
access, but they are added with different keys (e.g.:
`app_i18n_languages_en__source`, `app_i18n_languages_es__source`).

Whit this changes, the enabled languages are added directly as an array
to a single sfConfig key (`app_i18n_languages`), which is now used in
all the places that were using QubitSetting before.

### DIFF
--- a/lib/model/QubitSetting.php
+++ b/lib/model/QubitSetting.php
@@ -84,13 +84,21 @@ class QubitSetting extends BaseSetting
       LEFT JOIN '.QubitSettingI18n::TABLE_NAME.' source
         ON (setting.ID = source.id AND source.CULTURE = setting.SOURCE_CULTURE)';
 
-    $settings = array();
+    $settings = $i18nLanguages = array();
     $culture = sfContext::getInstance()->user->getCulture();
 
     foreach (QubitPdo::fetchAll($sql, array($culture)) as $qubitSetting)
     {
       if ($qubitSetting->scope)
       {
+        // Collect enabled languages into a single setting
+        if ($qubitSetting->scope == 'i18n_languages')
+        {
+          $i18nLanguages[] = $qubitSetting->value_source;
+
+          continue;
+        }
+
         $key = 'app_'.$qubitSetting->scope.'_'.$qubitSetting->name;
       }
       else
@@ -102,6 +110,8 @@ class QubitSetting extends BaseSetting
 
       $settings[$key.'__source'] = $qubitSetting->value_source;
     }
+
+    $settings['app_i18n_languages'] = $i18nLanguages;
 
     return $settings;
   }
--- a/apps/qubit/modules/menu/actions/changeLanguageMenuComponent.class.php
+++ b/apps/qubit/modules/menu/actions/changeLanguageMenuComponent.class.php
@@ -21,34 +21,10 @@ class MenuChangeLanguageMenuComponent extends sfComponent
 {
   public function execute($request)
   {
-    // The following snippet populates a list of language codes based on the
-    // languages that the administrator have made available via settings. We're
-    // using sfConfig as it's a cached source. An alternative would be to cache
-    // the component and use QubitSetting instead but we discarded that option
-    // in the past for some reason that I can't remember now.
-
-    $this->langCodes = array();
-
-    $prefix = 'app_i18n_languages';
-    $prefixLength = strlen($prefix);
-    $suffix = '__source';
-    $suffixLength = strlen($suffix);
-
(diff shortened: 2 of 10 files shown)


lib/model/QubitSetting.php
```diff
@@ -84,13 +84,21 @@ public static function getSettingsArray()
       LEFT JOIN '.QubitSettingI18n::TABLE_NAME.' source
         ON (setting.ID = source.id AND source.CULTURE = setting.SOURCE_CULTURE)';
 
-    $settings = array();
+    $settings = $i18nLanguages = array();
     $culture = sfContext::getInstance()->user->getCulture();
 
     foreach (QubitPdo::fetchAll($sql, array($culture)) as $qubitSetting)
     {
       if ($qubitSetting->scope)
```
djjuhasz, line 92: This looks like compensation for some bad data modelling to me.  @jraddaoui do you think it's practical to change the i18n language setting data from this:

```
mysql> select * from setting where scope='i18n_languages';
+------+----------------+----------+------------+----------------+----+---------------+
| name | scope          | editable | deleteable | source_culture | id | serial_number |
+------+----------------+----------+------------+----------------+----+---------------+
| en   | i18n_languages |        1 |          0 | en             | 27 |             0 |
| fr   | i18n_languages |        1 |          1 | en             | 28 |             0 |
| es   | i18n_languages |        1 |          1 | en             | 29 |             0 |
| nl   | i18n_languages |        1 |          1 | en             | 30 |             0 |
| pt   | i18n_languages |        1 |          1 | en             | 31 |             0 |
| fi   | i18n_languages |        1 |          1 | en             | 48 |             0 |
| uk   | i18n_languages |        1 |          1 | en             | 49 |             0 |
| sk   | i18n_languages |        1 |          1 | en             | 94 |             0 |
+------+----------------+----------+------------+----------------+----+---------------+
8 rows in set (0.00 sec)

mysql> select * from setting_i18n where id in (select id from setting where scope='i18n_languages');
+-------+----+---------+
| value | id | culture |
+-------+----+---------+
| en    | 27 | en      |
| fr    | 28 | en      |
| es    | 29 | en      |
| nl    | 30 | en      |
| pt    | 31 | en      |
| fi    | 48 | en      |
| uk    | 49 | en      |
| sk    | 94 | en      |
+-------+----+---------+
8 rows in set (0.00 sec)
```

to something sensible like:

```
mysql> select * from setting where name='i18n_languages';
+----------------+-------+----------+------------+----------------+----+---------------+
| name           | scope | editable | deleteable | source_culture | id | serial_number |
+----------------+-------+----------+------------+----------------+----+---------------+
| i18n_languages | null  |        1 |          0 | en             | 27 |             0 |
+----------------+-------+----------+------------+----------------+----+---------------+
1 rows in set (0.00 sec)

mysql> select * from setting_i18n where id =27;
+-------------------------+----+---------+
| value                   | id | culture |
+-------------------------+----+---------+
| en,es,fi,fr,nl,pt,sk,uk | 27 | en      |
+-------------------------+----+---------+
1 rows in set (0.00 sec)
```

Or does changing the data fixture require too much work?

djjuhasz (commented):
Please see my comment about changing the way the i18n_languages are stored in the db.

lib/model/QubitSetting.php
djjuhasz: The i18n_language `value` could be a serialized PHP array if that it easier and/or more resistant to bad data.

jraddaoui: Thanks @djjuhasz!

That's a great idea but, as you say, I think it requires too much work (fixtures, migration, form, etc.?) for the benefits it will provide (AFAIK). It'd be clearer if we go that way but it won't affect a lot the sfConfig population (as we already loop over all settings) and we'll access the value in the same manner.

Please, let me know if you strongly think we should do it and I'll give it a try ;)

djjuhasz: @jraddaoui you are probably right that the benefits of fixing the data fixtures are not enough to justify the the effort required. 😦  I would personally *like* to see the data fixed, but I don't have the authority to make that call.

djjuhasz (approved):
Looks good @jraddaoui!

jraddaoui:
Thanks @djjuhasz! I've crated https://projects.artefactual.com/issues/13263 to follow up about the multiple settings.

PR #1050, opened 2020-02-18
Add accession alternative IDs, refs #13254

mcantelon:
Add support for alternative identifiers to the accession plugin:

* Add new taxonomy for accession alternative identifier types
* Allow adding/editing of accession alternative identifiers during
  editing
* Display alternative identifiers of accession detail page

### DIFF
--- a/plugins/qtAccessionPlugin/modules/accession/templates/_alternativeIdentifiersIndex.php
+++ b/plugins/qtAccessionPlugin/modules/accession/templates/_alternativeIdentifiersIndex.php
@@ -1,0 +1,14 @@
+<div class="field">
+  
+  <h3><?php echo __('Alternative identifier(s)') ?></h3>
+
+  <div>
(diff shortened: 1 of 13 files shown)


plugins/qtAccessionPlugin/modules/accession/templates/_alternativeIdentifiersIndex.php
```diff
@@ -0,0 +1,14 @@
+<div class="field">
+  
+  <h3><?php echo __('Alternative identifier(s)') ?></h3>
```
sbreker, line 3: We may want to review the CAAIS wording around the identifier field so that AtoM's WebUI is close. http://archivescanada.ca/CWG_AccessionStandard#1.2_e

Maybe just "Identifier(s)" here?

sbreker (approved):
Looks great @mcantelon!

PR #1051, opened 2020-02-19
Improve ES indexing process, refs #13261

jraddaoui:
See commit messages for more information.

### DIFF
--- a/plugins/arElasticSearchPlugin/lib/model/arElasticSearchAccession.class.php
+++ b/plugins/arElasticSearchPlugin/lib/model/arElasticSearchAccession.class.php
@@ -19,28 +19,34 @@
 
 class arElasticSearchAccession extends arElasticSearchModelBase
 {
+  protected static $statements;
+
   public function load()
   {
-    $accessions = QubitAccession::getAll();
+    $accessionIds = QubitPdo::fetchAll(
+      'SELECT id FROM '.QubitAccession::TABLE_NAME,
+      array(),
+      array('fetchMode' => PDO::FETCH_COLUMN)
+    );
 
-    $this->count = count($accessions);
+    $this->count = count($accessionIds);
 
-    return $accessions;
+    return $accessionIds;
   }
 
   public function populate()
   {
     $errors = array();
 
-    foreach ($this->load() as $key => $accession)
+    foreach ($this->load() as $key => $id)
     {
       try
       {
-        $data = self::serialize($accession);
+        $data = self::serialize($id);
 
         $this->search->addDocument($data, 'QubitAccession');
 
-        $this->logEntry($accession->__toString(), $key + 1);
+        $this->logEntry($data['identifier'], $key + 1);
       }
       catch (sfException $e)
       {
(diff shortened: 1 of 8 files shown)


plugins/arElasticSearchPlugin/lib/model/arElasticSearchAccession.class.php
```diff
@@ -51,38 +57,57 @@ public function populate()
...
-    $serialized['identifier'] = $object->identifier;
+    self::$statements['accession']->execute(array(':id' => $id));
+    $data = self::$statements['accession']->fetch(PDO::FETCH_ASSOC);
 
-    $serialized['date'] = arElasticSearchPluginUtil::convertDate($object->date);
-    $serialized['createdAt'] = arElasticSearchPluginUtil::convertDate($object->createdAt);
-    $serialized['updatedAt'] = arElasticSearchPluginUtil::convertDate($object->updatedAt);
+    if (false === $data)
+    {
+      throw new sfException("Couldn't find accession (id: $id)");
```
sbreker, line 82: This probably should never occur, but if it does, is it better to throw an exception and stop the index process, or log the error and keep going?

plugins/arElasticSearchPlugin/lib/model/arElasticSearchAccession.class.php
```diff
@@ -51,38 +57,57 @@ public function populate()
...
+        WHERE acc.id = :id';
 
-    $serialized['id'] = $object->id;
-    $serialized['slug'] = $object->slug;
+      self::$statements['accession'] = self::$conn->prepare($sql);
+    }
 
-    $serialized['identifier'] = $object->identifier;
+    self::$statements['accession']->execute(array(':id' => $id));
+    $data = self::$statements['accession']->fetch(PDO::FETCH_ASSOC);
```
sbreker, line 78: Like the PDO options!  👍

sbreker (commented):
Looks really great @jraddaoui! I've made one comment regarding the throwing of an exception vs logging the error and continuing on.

plugins/arElasticSearchPlugin/lib/model/arElasticSearchAccession.class.php
jraddaoui: Thanks @sbreker, I've implemented this like we do for the other resources. If the exception is thrown in the populate task it's caught (in the populate method above) and logged at the end. In a single accession indexing (update method below) the exception won't be caught and it will cause an error.

I'm not a big fan of the implementation either, specially the exception message in the ES IO PDO class, where it can be thrown for various reasons, but I think we should addressed it in a different ticket/PR for all resources if it's needed.

sbreker: Ok that's great - Thanks @jraddaoui!

plugins/arElasticSearchPlugin/lib/model/arElasticSearchModelBase.class.php
```diff
@@ -83,27 +88,35 @@ public static function serializeI18ns($id, array $classes, $options = array())
...
     }
 
-    // Tableize class name
-    foreach ($classes as &$class)
+    foreach ($classes as $class)
     {
-      $class = str_replace('Qubit', '', $class);
-      $class = sfInflector::tableize($class);
-      $class .= '_i18n';
+      // Tableize class name
```
djjuhasz: Alternately I think you could get the table name via:

```
$class_i18n = $class."I18n";
$table = $class_i18n ::TABLE_NAME;
```

plugins/arElasticSearchPlugin/lib/model/arElasticSearchModelBase.class.php
```diff
@@ -83,27 +88,35 @@ public static function serializeI18ns($id, array $classes, $options = array())
...
+      $table = str_replace('Qubit', '', $class);
+      $table = sfInflector::tableize($table);
+      $table .= '_i18n';
```
djjuhasz: Alternately I think you could get the table name via:

```
$class_i18n = $class."I18n";
$table = $class_i18n::TABLE_NAME;
```

djjuhasz (commented):
Looks good. :)

plugins/arElasticSearchPlugin/lib/model/arElasticSearchModelBase.class.php
jraddaoui: Thanks @djjuhasz, addressed in 0652427.

sbreker (approved):
Looks great @jraddaoui!!

PR #1052, opened 2020-02-20
Fix error in purge task, refs #8491

mcantelon:
The purge task was prematurely aborting due to an error caused by it
extending the wrong base class.


jraddaoui (approved):
Thanks @mcantelon!

PR #1057, opened 2020-02-25
Support accession alt. ID CSV import, refs #13262

mcantelon:
Add support for accession alternative identifiers to accession CSV
import.


lib/task/import/csvAccessionImportTask.class.php
```diff
@@ -238,6 +243,58 @@ public function execute($arguments = array(), $options = array())
...
+
+            for ($index = 0; $index < max(count($identifiers), count($identifierNotes)); $index++)
+            {
+              $identifier = (empty($identifiers[$index])) ? null : $identifiers[$index];
+
+              if (!empty($identifier) || !empty($identifierNotes[$index]))
+              {
+                $otherName = new QubitOtherName;
+                $otherName->object = $self->object;
+                $otherName->name = $identifier;
```
sbreker, line 262: If the alt Identifier column is blank for a row but altIdNotes is populated, will it create othername records with blank names?

sbreker (commented):
Hi @mcantelon - Looks great. One question below regarding blank identifiers.

lib/task/import/csvAccessionImportTask.class.php
mcantelon: It will, luckily! The archival spec makes the identifiers optional (so I guess you could put in a note that an alternative identifier of some sort is needed).

sbreker: CAAIS spec allows for the actual identifier to be optional.

sbreker (approved):
Looks good @mcantelon!

PR #1060, opened 2020-03-17
Add physical object storage report, refs #13269

mcantelon:
Add task and job to export, as CSV, a report on physical object
holdings.


sbreker (approved):
Looks great @mcantelon! Nice work!

PR #1061, opened 2020-03-18
Suppress accession location error, refs #13268

mcantelon:
On the accession view page an error is normally displayed if the
location information field is empty. Added logic to suppress this error
if physical storage is linked to the accession.


sbreker (approved):
Good catch @mcantelon! 👍

PR #1063, opened 2020-03-22
Simplify auth. record alias import, refs #13277

mcantelon:
Remove the need to use a separate CSV file to import authority record
alternative names. Move alternatve name importing code to CSV import
base class and share with repository import.


sbreker (commented):
This change will also need to be made to the current authority export which will still generate a separate authority and relation file? Are there plans to make these same changes to combine the aliases file for the Auth export?

mcantelon:
Thanks Steve... good point about the export! I'll close this PR and add that.

I'm working on a separate PR to move the relation import out into a separate task to simplify.

PR #1064, opened 2020-03-24
Simplify auth. record alias CSV, refs #13277

mcantelon:
Remove the need to use a separate CSV file to import authority record
alternative names. Remove creation of a separate CSV file for these
names during CSV export. Move alternatve name importing code to CSV
import base class and share with repository import. Add population of
alternatve name columns to CSV export.


sbreker (approved):
Looks great @mcantelon!  👍

PR #1066, opened 2020-03-28
Physical object holdings report fixes, refs #13269

mcantelon:
* Change holding type filtering to be inclusive rather than exclusive
* Sort export rows by physical object's source culture name
* Add level of description column
* Improve CLI task help
* Wording changes
* Update unit tests


sbreker (approved):
Loogs great @mcantelon! 👍

PR #1068, opened 2020-04-02
Add Google Tag Manager tracking, refs #13284

replaceafill:
- Add config setting in 'config/app.yml':
    
Enables tracking through a container ID. See
https://developers.google.com/tag-manager/quickstart
    
- Add tag manager component:
    
Called from all the layout templates (except install) to render the
script and noscript snippets at the suggested points in the HTML.

### DIFF
--- a/apps/qubit/modules/default/actions/tagManagerComponent.class.php
+++ b/apps/qubit/modules/default/actions/tagManagerComponent.class.php
@@ -1,0 +1,30 @@
+<?php
+
+/*
+ * This file is part of the Access to Memory (AtoM) software.
+ *
+ * Access to Memory (AtoM) is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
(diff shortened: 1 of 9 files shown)


apps/qubit/modules/default/actions/tagManagerComponent.class.php
```diff
@@ -0,0 +1,30 @@
...
+ * You should have received a copy of the GNU General Public License
+ * along with Access to Memory (AtoM).  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+class DefaultTagManagerComponent extends sfComponent
+{
+  public function execute($request)
+  {
+    $config = sfConfig::get('app_google_tag_manager_container_id', '');
+    if (empty($config) || !isset($this->code)) {
```
mcantelon: Nitpick: Can code be an empty string? If not, using `!empty($this->code)` would guard again that.

mcantelon (approved):
One nitpick, but looks good otherwise!

replaceafill:
Thank you @mcantelon!

PR #1069, opened 2020-04-02
Digital object load --replace, refs #13152

sbreker:
Add '--replace' option to digital object load task. DO's and their
derivatives will be deleted and replaced with new ones based on CSV contents.

Allow 'slug' column to be used to identify descriptions in digital object
load CSV.

### DIFF
--- a/lib/task/digitalobject/digitalObjectLoadTask.class.php
+++ b/lib/task/digitalobject/digitalObjectLoadTask.class.php
@@ -50,7 +50,8 @@ class digitalObjectLoadTask extends sfBaseTask
       new sfCommandOption('link-source', 's', sfCommandOption::PARAMETER_NONE, 'Link source', null),
       new sfCommandOption('path', 'p', sfCommandOption::PARAMETER_OPTIONAL, 'Path prefix for digital objects', null),
       new sfCommandOption('limit', 'l', sfCommandOption::PARAMETER_OPTIONAL, 'Limit number of digital objects imported to n', null),
-      new sfCommandOption('attach-only', 'a', sfCommandOption::PARAMETER_NONE, 'Always attach digital objects instead of linking', null),
+      new sfCommandOption('attach-only', 'a', sfCommandOption::PARAMETER_NONE, 'Always attach digital objects to a new child description', null),
+      new sfCommandOption('replace', 'r', sfCommandOption::PARAMETER_NONE, 'Delete and replace digital objects', null),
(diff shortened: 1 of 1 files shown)


lib/task/digitalobject/digitalObjectLoadTask.class.php
```diff
@@ -96,30 +102,32 @@ public function execute($arguments = array(), $options = array())
 
     $this->disableNestedSetUpdating = ($options['skip-nested-set-build']) ? true : false;
 
-    $this->logSection('digital-object', "Load digital objects from {$arguments['filename']}...");
+    $this->logSection('digital-object', ($options['replace'] ? "Replace" : "Load") . " digital objects from {$arguments['filename']}...");
```
mcantelon: Nitpick: might be cleaner to use sprintf, something like the below, rather than string concatenation but either way works.
```
$operation = $options['replace'] ? "Replace" : "Load";
$this->logSection('digital-object', sprintf('%s digital objects from %s...', $operation, $arguments['filename']));
```

lib/task/digitalobject/digitalObjectLoadTask.class.php
```diff
@@ -162,16 +170,24 @@ public function execute($arguments = array(), $options = array())
     $this->curObjNum = 0;
 
     // Set up prepared query based on identifier type
-    $sql = 'SELECT io.id, do.id FROM '.QubitInformationObject::TABLE_NAME.' io
-      LEFT JOIN '.QubitDigitalObject::TABLE_NAME.' do ON io.id = do.object_id';
+    $sql = 'SELECT io.id, do.id FROM '.QubitInformationObject::TABLE_NAME.' io ';
```
mcantelon: Nitpick: Could be good to have spaces around string concat operator.

mcantelon (approved):
Looks good other than a couple nitpicks! There's a line in the commit that's a couple chars over 72 characters too I think. Good to go otherwise!

lib/task/digitalobject/digitalObjectLoadTask.class.php
sbreker: Change completed. Good idea - definitely cleaner!

lib/task/digitalobject/digitalObjectLoadTask.class.php
sbreker: Done.

PR #1070, opened 2020-04-05
Add actor relation import to web UI, refs #13286

mcantelon:
* Refactor/clean up relationship CSV import logic
* Move actor relationship CSV import into new CLI task
* Make label text for authority record relationships user configurable
* Add ability to import actor relationships via the web UI

### DIFF
--- a/lib/QubitCsvImport.class.php
+++ b/lib/QubitCsvImport.class.php
@@ -78,6 +78,11 @@ class QubitCsvImport
 
         break;
 
+      case 'authorityRecordRelationship':
+        $taskClassName = 'csv:authority-relation-import';
+
+        break;
+
       case 'event':
         $taskClassName = 'csv:event-import';
 
--- a/lib/flatfile/config/QubitRelation-actor.yml
+++ b/lib/flatfile/config/QubitRelation-actor.yml
@@ -1,0 +1,10 @@
+# Columns that should be included in the actor relation import
+columnNames:
+ - legacyId
+ - objectAuthorizedFormOfName
+ - subjectAuthorizedFormOfName
+ - relationType
+ - description
+ - date
+ - startDate
+ - endDate
--- a/lib/task/import/csvAuthorityRecordRelationImportTask.class.php
+++ b/lib/task/import/csvAuthorityRecordRelationImportTask.class.php
@@ -1,0 +1,198 @@
+<?php
+
+/*
+ * This file is part of the Access to Memory (AtoM) software.
+ *
+ * Access to Memory (AtoM) is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * Access to Memory (AtoM) is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with Access to Memory (AtoM).  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+/**
+ * Import authority record relations using CSV
+ *
+ * @package    AccessToMemory
+ * @subpackage task
+ * @author     Mike Cantelon <mike@artefactual.com>
+ */
+class csvAuthorityRecordRelationImportTask extends arBaseTask
+{
+  /**
+   * @see sfTask
+   */
+  protected function configure()
+  {
+    $this->addArguments([
+      new sfCommandArgument('filename', sfCommandArgument::REQUIRED, 'Output filename')
+    ]);
+
+    $this->addOptions([
+      new sfCommandOption('application', null,
+        sfCommandOption::PARAMETER_OPTIONAL, 'The application name', true),
+      new sfCommandOption('env', null,
+        sfCommandOption::PARAMETER_REQUIRED, 'The environment', 'cli'),
+      new sfCommandOption('connection', null,
+        sfCommandOption::PARAMETER_REQUIRED, 'The connection name', 'propel'),
+
+      new sfCommandOption('source-name', null,
+        sfCommandOption::PARAMETER_OPTIONAL, 'Source name to use when inserting keymap entries.'),
+      new sfCommandOption('index', null,
+        sfCommandOption::PARAMETER_NONE, 'Index for search during import.'),
+    ]);
+
+    $this->namespace = 'csv';
+    $this->name = 'authority-relation-import';
+    $this->briefDescription = 'Import authority record relations using CSV data.';
+    $this->detailedDescription = <<<EOF
+      Import authority record relations using CSV data
+EOF;
+  }
+
+  /**
+   * @see sfTask
+   */
+  public function execute($arguments = [], $options = [])
+  {
+    parent::execute($arguments, $options);
+
+    $this->log('Importing relations...');
+
+    $sourceName = $options['source-name'] ?: basename($arguments['filename']);
+    $this->import($arguments['filename'], $sourceName, $options['index']);
+
+    $this->log('Done.');
+  }
+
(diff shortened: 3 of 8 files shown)


lib/task/migrate/migrations/arMigration0180.class.php
```diff
@@ -0,0 +1,53 @@
...
+   * Upgrade
+   *
+   * @return bool True if the upgrade succeeded, False otherwise
+   */
+  public function up($configuration)
+  {
+    if (null === QubitSetting::getByName('authority_record_relationships'))
+    {
+      $setting = new QubitSetting;
+      $setting->name  = 'Authority Record Relationships';
```
djjuhasz: This should be the the `$setting_i18n->value` value.  The name should match the name in the fixture: `authority_record_relationships`.

lib/task/migrate/migrations/arMigration0180.class.php
```diff
@@ -0,0 +1,53 @@
...
+      $setting->scope = 'ui_label';
+      $setting->editable = 1;
+      $setting->deleteable = 0;
+      $setting->source_culture = 'en';
+      $setting->setValue($settingValue, array('culture' => 'en'));
```
djjuhasz: Where does `$settingValue` get set?

lib/QubitCsvImport.class.php
```diff
@@ -101,6 +106,9 @@ public function import($csvFile, $type = null, $csvOrigFileName = null)
     // Figure out whether user option should be added to command
     $commandUser = ($taskClassName == 'csv:import') ? sprintf('--user-id="%s"', sfContext::getInstance()->getUser()->getUserId()) : '';
 
+    // Figure out whether a source name should be added to command
+    $commandSourceName = ($taskClassName != 'csv:authority-relation-import') ? sprintf('--source-name="%s"', escapeshellarg($csvOrigFileName)) : '';
```
djjuhasz: Why is `source_name` specifically excluded for `csv:authority-relation-import` and no other import types?

lib/QubitCsvImport.class.php
```diff
@@ -135,14 +143,14 @@ public function import($csvFile, $type = null, $csvOrigFileName = null)
...
-        escapeshellarg($csvOrigFileName),
+        $commandSourceName,
         escapeshellarg($this->parent->slug),
         escapeshellarg($transformedFile ? $transformedFile : $csvFile));
     }
     else
     {
       // Example: php symfony csv:import /tmp/foobar
-      $command = sprintf('php %s %s %s %s %s %s %s --quiet %s --source-name=%s %s',
+      $command = sprintf('php %s %s %s %s %s %s %s %s %s --quiet %s',
```
djjuhasz: Having a whole separate `$command` assignment for the `--default-parent-slug` case seems like unnecessary repetition that could lead to mistakes where a new option is added to one case but not the other.  Can this if/else be combined into a single statement?

lib/flatfile/config/QubitRelation-actor.yml
```diff
@@ -0,0 +1,9 @@
+# Columns that should be included in the actor relation iport
+columnNames:
+ - sourceAuthorizedFormOfName
+ - targetAuthorizedFormOfName
+ - category
```
djjuhasz: I think the relationship "type" needs to be added to this list.
atom_250_authority_record_relationship_modal

lib/task/import/csvAuthorityRecordRelationImportTask.class.php
```diff
@@ -0,0 +1,185 @@
...
+              ? $self->rowStatusVars['sourceAuthorizedFormOfName']
+              : $self->rowStatusVars['targetAuthorizedFormOfName'];
+
+            $error = sprintf('Actor "%s" does not exist', $badActor);
+            print $self->logError($error);
+          }
+          else
+          {
+            // Add relationship if it doesn't yet exist
+            if (!$this->checkIfRelationshipExists($sourceActor->id, $targetActor->id, $relationTypeId))
```
djjuhasz: Nitpick: imho `if (!$this->relationshipExists(...` reads better, and makes the method name shorter.

lib/task/import/csvAuthorityRecordRelationImportTask.class.php
```diff
@@ -0,0 +1,185 @@
...
+            {
+              $relation = new QubitRelation;
+              $relation->subjectId = $sourceActor->id;
+              $relation->objectId  = $targetActor->id;
+              $relation->typeId    = $relationTypeId;
+
+              // Set relationship properties from column values
+              foreach (['date', 'startDate', 'endDate', 'description'] as $property)
```
djjuhasz, line 149: Nice 👍

lib/task/import/csvAuthorityRecordRelationImportTask.class.php
```diff
@@ -0,0 +1,185 @@
...
+      AND type_id = :type_id";
+
+    $params = [
+      ':subject_id' => $sourceActorId,
+      ':object_id' => $targetActorId,
+      ':type_id' => $relationTypeId
+    ];
+
+    $paramsVariant = [
+      ':subject_id' => $targetActorId,
```
djjuhasz, line 190: Clever to check the reciprocal case 😄 👍

djjuhasz (changes requested):
I've requested a few changes @mcantelon :)

lib/task/migrate/migrations/arMigration0180.class.php
mcantelon: D'oh! Thanks... should have tested this migration.

lib/QubitCsvImport.class.php
mcantelon: It wasn't supported when this logic was baked into the main authority import task, but yeah it make be good to add it for consistency and in case it's needed so I'll add it.

lib/flatfile/config/QubitRelation-actor.yml
mcantelon: The "category" in the UI dialog's context is a nicety given it's a parent term of the type term... once you select the "category" it populates the select element for "type". So just specifying the type (labelled "category" in the existing CSV) works for importing.

djjuhasz: @mcantelon In that case I think calling the CSV column "relationship_type" (or similar) is better than calling it "category", to avoid confusing it with the UI "category" field (which isn't represented in the CSV template)

djjuhasz: I'm assuming here that the "Relationship type" UI field actually sets the `relation.type_id` column in the database, and is the value we want in the CSV doc.

mcantelon: Yeah, that does make sense. I've renamed it to "relationType".

mcantelon:
Okay I've updated the PR (I'll squash the commits before I merge)! Let me know if you spot any further possible issues.

lib/flatfile/config/QubitRelation-actor.yml
```diff
@@ -0,0 +1,9 @@
+# Columns that should be included in the actor relation iport
```
djjuhasz: Nitpick: Typo "iport" should be "import" :)

lib/flatfile/config/QubitRelation-actor.yml
djjuhasz: @mcantelon I'm not sure if Github is just showing me an old version of this file but I'm still seeing "- category" and not "- relationType" in this column list.

lib/flatfile/config/QubitRelation-actor.yml
mcantelon: D'oh! I'll fix. Just realized I've got to update the columns in this file too.

djjuhasz (changes requested):

lib/flatfile/config/QubitRelation-actor.yml
mcantelon: I think renaming "sourceAuthorizedFormOfName" to "objectAuthorizedFormOfName" and "targetAuthorizedFormOfName" to "subjectAuthorizedFormOfName" might be clearer too...

djjuhasz (approved):
Looks good @mcantelon ! 👍

PR #1072, opened 2020-04-15
Add config/app.yml to .gitignore

scollazo:
This file needs to be modified in centos deployments to use
PASSWORD_BCRYPT instead of de the default ARGON1, as that isn't
supported in the PHP packages available in RHEL/Centos.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,6 +8,7 @@
 /config/config.php
 /config/databases.yml
 /config/propel.ini
+/config/app.yml
 
 # Internal use
 /cache/


jraddaoui:
Hi @scollazo,

You can overwrite the default value from `config/app.yml` in `apps/qubit/config/app.yml`, which is already added to `.gitignore`.

jraddaoui:
The new password hash variables are not considered in the AtoM role but you could implement them in there ...

https://github.com/artefactual-labs/ansible-atom/blob/master/templates/atom/apps/qubit/config/app.yml

scollazo:
Works for me, thanks for the tip. 

Closing....

PR #1074, opened 2020-04-18
Authority record search by relation, refs #13302

mcantelon:
Add the ability to filter authority record searches by relationship
type.

* Add Elasticsearch indexing of authority record relations
* Add related type field to authority record search form
* Add related authority record field to authority record search form
* Add filter tag for relation types
* Code cleanup/validation tightening

### DIFF
--- a/apps/qubit/modules/actor/actions/browseAction.class.php
+++ b/apps/qubit/modules/actor/actions/browseAction.class.php
@@ -36,13 +36,16 @@ class ActorBrowseAction extends DefaultBrowseAction
       'repository',
       'hasDigitalObject',
       'entityType',
-      'emptyField'
+      'emptyField',
+      'relatedType',
+      'relatedAuthority'
     ),
 
     $FILTERTAGS = array(
       'hasDigitalObject' => array(),
       'repository'       => array('model' => 'QubitRepository'),
       'entityType'       => array('model' => 'QubitTerm'),
+      'relatedType'      => array('model' => 'QubitTerm'),
       'occupation'       => array('model' => 'QubitTerm'),
       'place'            => array('model' => 'QubitTerm'),
       'subject'          => array('model' => 'QubitTerm'),
@@ -104,8 +107,8 @@ class ActorBrowseAction extends DefaultBrowseAction
         }
         else
         {
-          $choices = array();
-          $choices[null] = null;
+          $choices = array(null => null);
+
           foreach (QubitRepository::get($criteria) as $repository)
           {
             $choices[$repository->id] = $repository->__toString();
@@ -143,17 +146,14 @@ class ActorBrowseAction extends DefaultBrowseAction
         break;
 
       case 'entityType':
-        $this->form->setValidator($name, new sfValidatorString);
-
-        $choices = array();
-        $choices[null] = null;
+        $choices = array(null => null);
 
         foreach (QubitTaxonomy::getTaxonomyTerms(QubitTaxonomy::ACTOR_ENTITY_TYPE_ID) as $item)
         {
           $choices[$item->id] = $item->__toString();
         }
 
-        $this->form->setValidator($name, new sfValidatorString);
+        $this->form->setValidator($name, new sfValidatorChoice(array('choices' => array_keys($choices))));
         $this->form->setWidget($name, new sfWidgetFormSelect(array('choices' => $choices)));
 
         if (!empty($this->getFilterTagObject('entityType')))
@@ -164,23 +164,55 @@ class ActorBrowseAction extends DefaultBrowseAction
         break;
 
       case 'emptyField':
-        $this->form->setValidator($name, new sfValidatorString);
-
-        $choices = array();
-        $choices[null] = null;
+        $choices = array(null => null);
 
         foreach ($this->fieldOptions as $field => $label)
         {
           $choices[$field] = $label;
         }
 
-        $this->form->setValidator($name, new sfValidatorString);
+        $this->form->setValidator($name, new sfValidatorChoice(array('choices' => array_keys($choices))));
         $this->form->setWidget($name, new sfWidgetFormSelect(array('choices' => $choices)));
 
         if (!empty($request->emptyField))
         {
           $this->form->setDefault('emptyField', $request->emptyField);
         }
+
+        break;
+
+      case 'relatedType':
+        $choices = array(null => null);
+
+        foreach (QubitTaxonomy::getTaxonomyTerms(QubitTaxonomy::ACTOR_RELATION_TYPE_ID) as $item)
+        {
+          // Omit category terms
+          if ($item->parentId != QubitTerm::ROOT_ID)
+          {
+            $choices[$item->id] = $item->__toString();
+          }
+        }
+
+        $this->form->setValidator($name, new sfValidatorChoice(array('choices' => array_keys($choices))));
+        $this->form->setWidget($name, new sfWidgetFormSelect(array('choices' => $choices)));
+
+        if (!empty($this->getFilterTagObject($name)))
+        {
+          $this->form->setDefault($name, $this->getFilterTagObject($name)->id);
+        }
+
+        break;
+
+      case 'relatedAuthority':
+        $defaultChoices = array();
+
+        if (!empty($this->relatedAuthority))
+        {
+          $defaultChoices = array($request->$name => $this->relatedAuthority->getAuthorizedFormOfName(array('cultureFallback' => true)));
+        }
+
+        $this->form->setValidator($name, new sfValidatorString);
+        $this->form->setWidget($name, new sfWidgetFormSelect(array('choices' => $defaultChoices)));
 
         break;
     }
@@ -297,15 +329,26 @@ class ActorBrowseAction extends DefaultBrowseAction
       $this->addField($name, $request);
     }
 
-    // Set which values will be relayed via the advanced search form
+    // Set which values will be relayed as hidden fields via the advanced search form
     $allowed = array_merge(
       array_keys($this::$AGGS),
       array('sort', 'sortDir')
     );
 
-    $ignored = array('repository', 'entityType');
+    // These get relayed as form field default values
+    $ignored = array('repository', 'entityType', 'relatedType', 'relatedAuthority');
 
     $this->setHiddenFields($request, $allowed, $ignored);
+  }
+
+  private function getRelatedAuthorityUsingSlug($slug)
+  {
+    if (!empty($slug))
+    {
+      $params = $this->context->routing->parse(Qubit::pathInfo($slug));
+
+      return get_class($params['_sf_route']->resource) == 'QubitActor' ? $params['_sf_route']->resource : null;
+    }
   }
 
   protected function doSearch($request)
@@ -323,6 +366,43 @@ class ActorBrowseAction extends DefaultBrowseAction
       $this->search->queryBool->addMust($queryField);
     }
 
+    // Filter by related authority and/or relation type
+    if (!empty($this->relatedAuthority) && !empty($request->relatedType))
+    {
+      // Include actors that relate to the specified actor and have a relation
+      // of the specified type
+      $queryBool = new \Elastica\Query\BoolQuery;
+
+      $queryBool->addMust($this->actorRelationsQueryForActor($this->relatedAuthority->id));
+      $queryBool->addMust($this->actorRelationsQueryForType($this->request->relatedType));
+
+      $this->search->queryBool->addMust($this->actorRelationsNestedQuery($queryBool));
+
+      // Omit the specified actor
+      $this->search->queryBool->addMust($this->actorExcludeQuery($this->relatedAuthority->id));
+    }
+    else if (!empty($this->relatedAuthority))
+    {
+      // Include actors that relate to the specified actor
+      $queryBool = new \Elastica\Query\BoolQuery;
+
+      $queryBool->addMust($this->actorRelationsQueryForActor($this->relatedAuthority->id));
+
+      $this->search->queryBool->addMust($this->actorRelationsNestedQuery($queryBool));
+
+      // Omit the specified actor
+      $this->search->queryBool->addMust($this->actorExcludeQuery($this->relatedAuthority->id));
+    }
+    else if (!empty($this->request->relatedType))
+    {
+      // Include actors with a relation of a specified type
+      $queryBool = new \Elastica\Query\BoolQuery;
+
+      $queryBool->addMust($this->actorRelationsQueryForType($this->request->relatedType));
+
+      $this->search->queryBool->addMust($this->actorRelationsNestedQuery($queryBool));
+    }
+
     // Filter out results if a specific field isn't empty
     if (!empty($request->emptyField))
     {
@@ -335,6 +415,66 @@ class ActorBrowseAction extends DefaultBrowseAction
     $this->search->query->setQuery($this->search->queryBool);
 
     return QubitSearch::getInstance()->index->getType('QubitActor')->search($this->search->getQuery(false));
+  }
+
+  private function actorRelationsQueryForActor($actorId)
+  {
+    // Result relations must have either a related object or subject ID
+    $queryRelatedBool = new \Elastica\Query\BoolQuery;
+
+    $queryField = new \Elastica\Query\Term;
+    $queryField->setTerm('actorRelations.objectId', $actorId);
+    $queryRelatedBool->addShould($queryField);
+
+    $queryField = new \Elastica\Query\Term;
+    $queryField->setTerm('actorRelations.subjectId', $actorId);
+    $queryRelatedBool->addShould($queryField);
+
+    return $queryRelatedBool;
+  }
+
+  private function actorRelationsQueryForType($typeId)
+  {
+    $queryTypeBool = new \Elastica\Query\BoolQuery;
+
+    $queryField = new \Elastica\Query\Term;
+    $queryField->setTerm('actorRelations.typeId', $typeId);
+    $queryTypeBool->addShould($queryField);
+
+    $relationTerm = QubitTerm::getById($typeId);
+
+    // Look for converse term as well if one exists
+    $converseTerm = $relationTerm->getConverseActorRelationTerm();
+
+    if ($converseTerm !== null && $typeId != $converseTerm->id)
+    {
+      $queryField = new \Elastica\Query\Term;
+      $queryField->setTerm('actorRelations.typeId', $converseTerm->id);
+      $queryTypeBool->addShould($queryField);
+    }
+
+    return $queryTypeBool;
+  }
+
+  private function actorRelationsNestedQuery($query)
+  {
+    $queryNested = new \Elastica\Query\Nested();
+    $queryNested->setPath('actorRelations');
+    $queryNested->setQuery($query);
+
+    return $queryNested;
+  }
+
+  private function actorExcludeQuery($actorId)
+  {
+    // Omit the actor that the others are related to
+    $queryBool = new \Elastica\Query\BoolQuery;
+
+    $queryField = new \Elastica\Query\Term;
+    $queryField->setTerm('_id', $actorId);
+    $queryBool->addMustNot($queryField);
+
+    return $queryBool;
   }
 
   /**
@@ -422,6 +562,9 @@ class ActorBrowseAction extends DefaultBrowseAction
     }
 
     parent::execute($request);
+
+    // Look up related authority, if specified
+    $this->relatedAuthority = $this->getRelatedAuthorityUsingSlug($request->relatedAuthority);
 
     // Prepare filter tags, form, and hidden fields/values
     $this->setFilterTagsAndForm($request);
--- a/apps/qubit/modules/actor/templates/_advancedSearch.php
+++ b/apps/qubit/modules/actor/templates/_advancedSearch.php
@@ -127,6 +127,29 @@
         </div>
       </div>
 
+      <p><?php echo __('Find results where:') ?></p>
+
+      <div class="criteria">
+
+        <div class="filter-row">
+
+          <div class="filter-left relation">
+            <?php echo $form->relatedType
+              ->label(__('Relationship'))
+              ->renderRow() ?>
+          </div>
+
+          <div class="filter-right relation">
+            <?php echo $form->relatedAuthority
+              ->label(__('Related [%1%]', array('%1%' => sfConfig::get('app_ui_label_actor'))))
+              ->renderLabel() ?>
+            <?php echo $form->relatedAuthority->render(array('class' => 'form-autocomplete')) ?>
+            <input class="list" type="hidden" value="<?php echo url_for(array('module' => 'actor', 'action' => 'autocomplete')) ?>"/>
+          </div>
+
+        </div>
+      </div>
+
       <section class="actions">
         <input type="submit" class="c-btn c-btn-submit" value="<?php echo __('Search') ?>"/>
         <input type="button" class="reset c-btn c-btn-delete" value="<?php echo __('Reset') ?>"/>
--- a/lib/job/arUpdateEsActorRelationsJob.class.php
+++ b/lib/job/arUpdateEsActorRelationsJob.class.php
@@ -1,0 +1,121 @@
+<?php
+
+/*
+ * This file is part of the Access to Memory (AtoM) software.
+ *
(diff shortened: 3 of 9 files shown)


apps/qubit/modules/actor/actions/browseAction.class.php
```diff
@@ -297,13 +329,14 @@ protected function setFilterTagsAndForm($request)
...
-    // Set which values will be relayed via the advanced search form
+    // Set which values will be relayed as hidden fields via the advanced search form
     $allowed = array_merge(
       array_keys($this::$AGGS),
       array('sort', 'sortDir')
     );
 
-    $ignored = array('repository', 'entityType');
+    // These get relayed as form field default values
+    $ignored = array('repository', 'entityType', 'relatedType');
```
jraddaoui: Should this include `relatedAuthority`?

apps/qubit/modules/actor/actions/browseAction.class.php
```diff
@@ -442,4 +570,13 @@ public function execute($request)
       $request->subquery = $request->sq0;
     }
   }
+
+  private function getRelatedAuthority()
+  {
+    if (!empty($this->request->relatedAuthority))
+    {
+      $params = $this->context->routing->parse(Qubit::pathInfo($this->request->relatedAuthority));
+      return $params['_sf_route']->resource;
```
jraddaoui: Blank line before return.

apps/qubit/modules/actor/templates/_advancedSearch.php
```diff
@@ -127,6 +127,29 @@
...
+
+          <div class="filter-left relation">
+            <?php echo $form->relatedType
+              ->label(__('Relationship'))
+              ->renderRow() ?>
+          </div>
+
+          <div class="filter-right relation">
+            <?php echo $form->relatedAuthority
+              ->label(__('Related [%1%]', array('%1%' => sfConfig::get('app_ui_label_digitalobject'))))
```
jraddaoui: `app_ui_label_actor`?

apps/qubit/modules/actor/actions/browseAction.class.php
```diff
@@ -323,6 +356,101 @@ protected function doSearch($request)
...
+        $queryBool->addMust($queryRelatedBool);
+
+        // Result relations must be between two actors
+        $queryField = new \Elastica\Query\Term;
+        $queryField->setTerm('relations.objectClass', 'QubitActor');
+        $queryBool->addMust($queryField);
+
+        $queryField = new \Elastica\Query\Term;
+        $queryField->setTerm('relations.subjectClass', 'QubitActor');
+        $queryBool->addMust($queryField);
```
jraddaoui: I'd do this kind of filtering at indexing time. We only need ACTOR_RELATION_TYPE_ID relations, which should already be limited to actor <-> actor.

apps/qubit/modules/actor/actions/browseAction.class.php
```diff
@@ -323,6 +356,101 @@ protected function doSearch($request)
...
+        $queryNested->setQuery($queryBool);
+
+        $this->search->queryBool->addMust($queryNested);
+
+        // Omit the actor that the others are related to
+        $queryField = new \Elastica\Query\Term;
+        $queryField->setTerm('_id', $actor->id);
+        $queryBool->addMustNot($queryField);
+
+        $this->search->queryBool->addMust($queryNested);
```
jraddaoui: Nice trick to solve the reciprocal relations issue!

apps/qubit/modules/actor/actions/browseAction.class.php
```diff
@@ -323,6 +356,101 @@ protected function doSearch($request)
...
+      $queryBool->addMust($queryTypeBool);
+
+      // Result relations must be between two actors
+      $queryField = new \Elastica\Query\Term;
+      $queryField->setTerm('relations.objectClass', 'QubitActor');
+      $queryBool->addMust($queryField);
+
+      $queryField = new \Elastica\Query\Term;
+      $queryField->setTerm('relations.subjectClass', 'QubitActor');
+      $queryBool->addMust($queryField);
```
jraddaoui: Same in here, this could be filtered at indexing time.

plugins/arElasticSearchPlugin/config/mapping.yml
```diff
@@ -338,6 +350,7 @@ mapping:
       standardized_names: other_name
       subjects: term
       places: term
+      relations: relation
```
jraddaoui: As mentioned, I'd remove the classes from this field and call it `actor_relations`.

plugins/arElasticSearchPlugin/lib/model/arElasticSearchActorPdo.class.php
```diff
@@ -349,6 +355,31 @@ public function serializeAltNames()
...
+  public function serializeObjectRelations()
+  {
+    $relations = array();
+
+    $sql = "SELECT r.type_id, r.object_id, r.subject_id, oo.class_name \r
+              AS object_class, so.class_name AS subject_class \r
+              FROM ".QubitRelation::TABLE_NAME." r \r
+              INNER JOIN ".QubitObject::TABLE_NAME." oo ON r.object_id=oo.id \r
+              INNER JOIN ".QubitObject::TABLE_NAME." so ON r.subject_id=so.id \r
+              WHERE r.object_id=? OR r.subject_id=?";
```
jraddaoui: Remove the JOINS to the `object` table, add a JOIN to the `term` table on `type_id` and filter by `taxonomy_id` (ACTOR_RELATION_TYPE_ID). This should speed up the SQL query and avoid the filtering in the ES query.

Nitpick, the `\r` are not really needed.

plugins/arElasticSearchPlugin/lib/model/arElasticSearchActorPdo.class.php
```diff
@@ -349,6 +355,31 @@ public function serializeAltNames()
...
+      array_push($relations, array(
+        'typeId'       => $relation->type_id,
+        'objectId'     => $relation->object_id,
+        'objectClass'  => $relation->object_class,
+        'subjectId'    => $relation->subject_id,
+        'subjectClass' => $relation->subject_class
+      ));
+    }
+
+    return $relations;
```
jraddaoui: If you change the SELECT statement to `SELECT r.type_id as typeId, r.object_id as objectId, r.subject_id as subjectId FROM ...`, you could return the array directly with:

```php
return QubitPdo::fetchAll($sql, array($this->id, $this->id), array('fetchMode' => PDO::FETCH_ASSOC));
```

Not a big deal in here but `FETCH_ASSOC` reduces memory usage and it will return quicker without the loop.

plugins/arElasticSearchPlugin/lib/model/arElasticSearchActorPdo.class.php
```diff
@@ -262,6 +262,12 @@ public function serialize()
       array(QubitTaxonomy::PLACE_ID, QubitTaxonomy::SUBJECT_ID)
     );
 
+    // Related objects
+    if (count($relations = $this->serializeObjectRelations()))
```
jraddaoui: In this case, where the count is not actually needed, `!empty()` is a bit faster.

jraddaoui (changes requested):
Looking good @mcantelon!

This is quite a complex feature to implement and I think we're not considering a couple of important issues:

**Elasticsearch update of related resources:**

When a relation is created in an actor, the other end of the relation should also be updated in Elasticsearch to reflect that relation. It should also be updated when the relation is deleted or modified. We have some properties and processes to update related resources on save, but I doubt this types of relation are considered.

**Search on both, relation type and actor:**

The current search, while using nested queries, it uses two different ones for the type and the actor. Consider the following relations:

- Actor A _is student of_ Actor B.
- Actor A _is the parent of_ Actor C.

In the current implementation, if you search for type: _is student of_ and actor: _Actor C_, Actor A will appear as a hit while it shouldn't. To solve that you'll need to include both clauses on the same nested query, in pseudo code ...

```
if (request->relatedAuthority and request->relatedType)
{
  // one nested query with all clauses
}
else if (request->relatedAuthority)
{
  // as it is (see comments below)
}
else if (request->relatedType)
{
  // as it is (see comments below)
}
```

See also a few less important comments below.

Thanks!

mcantelon:
Thanks for all the great suggestions @jraddaoui !

Maybe I'll add a bulk update actor job, then, so we can, before saving or deleting an actor, take stock of related actors so they can be updated after an actor is saved or deleted.

jraddaoui:
Thanks @mcantelon, 

I'm not sure that's going to work in this case. You could amalgamate all the related actor ids that need to be updated on the edit actor post request ...

- Deletions: https://github.com/artefactual/atom/blob/qa/2.6.x/plugins/sfIsaarPlugin/modules/sfIsaarPlugin/actions/relatedAuthorityRecordComponent.class.php#L129
- Creations: https://github.com/artefactual/atom/blob/qa/2.6.x/apps/qubit/modules/relation/actions/editComponent.class.php#L140
- Editions: https://github.com/artefactual/atom/blob/qa/2.6.x/apps/qubit/modules/relation/actions/editComponent.class.php#L136

But that won't consider changes made through other mechanisms like EAC/CSV import, even if we don't import this relations at the moment, we may do it at some point, and those imports now perform updates and deletes too. However, looking at the components linked above, you may not be able to amalgamate the previously related actor ids in the actor save method.

We have been studying to use ES' update by query API ([Redmine ticket](https://projects.artefactual.com/issues/13273) with an example), but the Painless script needed for this update will be complex, specially in the actor save. I think this would be the ideal solution (still triggered from a background job as that update by query request takes some time and you want to verify it worked properly) but I'm not sure how much time you have for this feature and if you'd like to give it a try.

Another option could be to do the same more manually. Trigger a job with the saved/deleted actor id, query ES to get the existing related actors and combine them with the currently related, then partially update the `actor_relations` field in all of them.

Please, let me know if you want to chat or have a quick call about this.

jraddaoui:
Thinking it through, I'd go with the last option. The Painless script needed for the update by query request can be over-complicated while the partial updates should be pretty solid and not require a lot of resources.

After getting all the actor ids, previously and currently related, if you make `serializeObjectRelations` a static function with the actor id as a parameter, you could easily generate the partial update requests needed for all the actors without using the ORM, and those updates will be executed in bulk requests by default.

mcantelon:
Sounds good... thanks!

peterVG:
Note: referring Redmine issue has been changed to https://projects.artefactual.com/issues/13302 (after accidental deletion of #13276)

mcantelon:
Note to self: change issue number in commit and use new getConverseActorRelationTerm in qa/2.6.x in lib/model/QubitTerm.

mcantelon:
Hi @jraddaoui ... I've updated the PR! Let me know what you think!

apps/qubit/modules/actor/actions/browseAction.class.php
```diff
@@ -164,24 +164,55 @@ protected function addField($name, $request)
...
+        if (!empty($this->getFilterTagObject($name)))
+        {
+          $this->form->setDefault($name, $this->getFilterTagObject($name)->id);
+        }
+
+        break;
+
+      case 'relatedAuthority':
+        $defaultChoices = array();
+        if (!empty($request->$name) && !empty($actor = $this->getRelatedAuthorityUsingSlug($this->request->relatedAuthority)))
```
jraddaoui: Really minor details ... I'd be clearer if you access the request in the same manner, both `$request->$name` and `$this->request->relatedAuthority` are the same, and empty line before the `if`.

apps/qubit/modules/actor/actions/browseAction.class.php
```diff
@@ -297,17 +328,28 @@ protected function setFilterTagsAndForm($request)
...
     $this->setHiddenFields($request, $allowed, $ignored);
   }
 
+  private function getRelatedAuthorityUsingSlug($relatedAuthority)
+  {
+    if (!empty($relatedAuthority))
+    {
+      $params = $this->context->routing->parse(Qubit::pathInfo($relatedAuthority));
+
+      return (get_class($params['_sf_route']->resource) == 'QubitActor') ? $params['_sf_route']->resource : null;
```
jraddaoui: You shouldn't need the parenthesis, right?

apps/qubit/modules/actor/actions/browseAction.class.php
```diff
@@ -323,6 +365,45 @@ protected function doSearch($request)
       $this->search->queryBool->addMust($queryField);
     }
 
+    // Parse actor ID from route (if a related actor has been specified)
+    $actor = $this->getRelatedAuthorityUsingSlug($this->request->relatedAuthority);
```
jraddaoui: Not a big deal again, but you could use a global variable (`$this->relatedActor`) to avoid fetching it twice (here and in the form processing).

apps/qubit/modules/actor/actions/browseAction.class.php
```diff
@@ -337,6 +418,66 @@ protected function doSearch($request)
...
+    $queryField = new \Elastica\Query\Term;
+    $queryField->setTerm('actorRelations.typeId', $typeId);
+    $queryTypeBool->addShould($queryField);
+
+    $relationTerm = QubitTerm::getById($typeId);
+
+    // Look for converse term as well if one exists
+    $converseTerm = $relationTerm->getConverseActorRelationTerm();
+
+    if ($converseTerm !== null && $this->request->relatedType != $converseTerm->id)
```
jraddaoui: Another not important matter, I'd use the `$typeId` function parameter instead of accessing the global request.

plugins/arElasticSearchPlugin/lib/model/arElasticSearchActorPdo.class.php
```diff
@@ -349,6 +355,28 @@ public function serializeAltNames()
     return $serialized;
   }
 
+  static public function serializeObjectRelations($actorId)
```
jraddaoui: Cosmetic, `public static` (if we'd use [PSR-2](https://github.com/php-fig/fig-standards/blob/master/accepted/PSR-2-coding-style-guide.md#1-overview)).

plugins/arElasticSearchPlugin/lib/model/arElasticSearchActorPdo.class.php
```diff
@@ -349,6 +355,28 @@ public function serializeAltNames()
...
+             AND object_id=? OR r.subject_id=?";
+
+    foreach (QubitPdo::fetchAll($sql, array($actorId, $actorId), array('fetchMode' => PDO::FETCH_ASSOC)) as $relation)
+    {
+      array_push($relations, array(
+        'objectId'     => $relation['object_id'],
+        'subjectId'    => $relation['subject_id'],
+        'typeId'       => $relation['type_id']
+      ));
+    }
```
jraddaoui: Did you try the "select .. as .." tip to avoid the loop?

plugins/sfIsaarPlugin/modules/sfIsaarPlugin/actions/relatedAuthorityRecordComponent.class.php
```diff
@@ -121,6 +121,12 @@ protected function processField($field)
 
   public function processForm()
   {
+    if ($this->resource instanceof QubitActor)
```
jraddaoui: It will always be the case in here, no?

jraddaoui: There is a similar component in the sfIsdfPlugin, but it's not extending this one.

plugins/sfIsaarPlugin/modules/sfIsaarPlugin/actions/relatedAuthorityRecordComponent.class.php
```diff
@@ -130,7 +136,18 @@ public function processForm()
       }
     }
 
-    return parent::processForm();
+    parent::processForm();
+
+    if ($this->resource instanceof QubitActor)
```
jraddaoui: Same in here.

lib/job/arUpdateEsActorRelationsJob.class.php
```diff
@@ -0,0 +1,114 @@
...
+    $this->job->setStatusCompleted();
+    $this->job->save();
+
+    return true;
+  }
+
+  private function updateActorRelationships($actor)
+  {
+    $relationData = arElasticSearchActorPdo::serializeObjectRelations($actor->id);
+    QubitSearch::getInstance()->partialUpdate($actor, ['actorRelations' => $relationData]);
```
jraddaoui, line 89: It's a shame that we need to pass the entire object to `partialUpdate` ... I created https://projects.artefactual.com/issues/13272 a while ago to deal with it.

jraddaoui: Also, we use [updates with partial documents](https://www.elastic.co/guide/en/elasticsearch/reference/5.6/docs-update.html#_updates_with_a_partial_document) in this process, which does a simple recursive merge, inner merging of objects, replacing core "keys/values" and arrays.

I'm not sure if that update is going to work like we're expecting, specially deleting and updating existing values. I think it will, using the "replacing core "keys/values" and arrays", but it may not if these nested fields are considered as objects ("inner merging of objects").

plugins/arElasticSearchPlugin/lib/model/arElasticSearchActorPdo.class.php
```diff
@@ -262,6 +262,12 @@ public function serialize()
       array(QubitTaxonomy::PLACE_ID, QubitTaxonomy::SUBJECT_ID)
     );
 
+    // Related objects
+    if (!empty($relations = self::serializeObjectRelations($this->id)))
+    {
+      $serialized['actorRelations'] = $relations;
+    }
+
```
jraddaoui: We may want to initialize this field in all cases for consistency with the partial updates, similar to what we do with the IO's finding aid fields. I think it's not needed for the current implementation but it may be a problem if we ever target this field with ES' missing filter.

lib/job/arUpdateEsActorRelationsJob.class.php
```diff
@@ -0,0 +1,114 @@
...
+      $actors[$relation['subjectId']] = true;
+    }
+
+    // Omit ID of actor ID itself if not in inclusive mode (for deletes)
+    if (!$inclusive)
+    {
+      unset($actors[$actorId]);
+    }
+
+    return array_keys($actors);
```
jraddaoui: Maybe a one-dimensional array + array_unique + array_diff would have been clearer.

jraddaoui (changes requested):
This is looking amazing @mcantelon, great work!

I have added some comments below, mostly cosmetic changes (feel free to ignore them) and a few questions.

My biggest concern is that we're only updating the related resources from the actors' edit and delete forms. I believe we're adding CSV import for this kind of relations, right? And, can that import be triggered from the GUI? If we're developing that import only as a CLI task, adding a note about re-indexing all actors in the documentation could be enough. But, if the relations import can be triggered from the GUI, it's not ideal to ask for a re-index. That's why I suggested to launch the job from the actor save method and obtain the previously related actors from ES, mix them with the currently related form the DB and update all of them.

Please, let me know a bit more about that import and if you think we should address that concern. Otherwise, this is really close to approval.

lib/job/arUpdateEsActorRelationsJob.class.php
mcantelon: Interesting... hopefully we'll catch it if it works differently.

plugins/arElasticSearchPlugin/lib/model/arElasticSearchActorPdo.class.php
mcantelon: Good thinking!

mcantelon:
Hi @jraddaoui ... thanks for the great suggestions! I've updated this PR to incorporate them... let me know what you think!

plugins/arElasticSearchPlugin/lib/model/arElasticSearchActorPdo.class.php
mcantelon: Overlooked that... you're right... I can just return the result... good suggestion!

lib/model/QubitActor.php
```diff
@@ -132,6 +132,14 @@ protected function insert($connection = null)
 
   public function save($connection = null)
   {
+    if ('QubitActor' == $this->className)
+    {
+      // Take note of which actors are related to the actor about to be updated
+      $previouslyRelatedActorIds = !empty($this->id)
+        ? arUpdateEsActorRelationsJob::relationActorIds($this->id)
```
jraddaoui: At this point the previous relations may have been deleted in the component, that's why I suggested to get the previously related actors querying Elasticsearch.

lib/model/QubitActor.php
```diff
@@ -232,13 +240,27 @@ public function save($connection = null)
     if ('QubitActor' == $this->className)
     {
       QubitSearch::getInstance()->update($this);
+
+      // Update, in Elasticsearch, the actors previously or currently related to the actor
+      $actorsToUpdate = $previouslyRelatedActorIds + arUpdateEsActorRelationsJob::relationActorIds($this->id);
```
jraddaoui: We should avoid duplicates in here with `array_unique` after the merging both arrays.

lib/model/QubitActor.php
```diff
@@ -232,13 +240,27 @@ public function save($connection = null)
...
+
+      if (!empty($actorsToUpdate))
+      {
+        QubitJob::runJob('arUpdateEsActorRelationsJob', ['actorIds' => $actorsToUpdate]);
```
jraddaoui: This point could be reached from different environments (cli, worker, etc.). We usually check that and run the process synchronously in the cli and worker environments and async. in all the others (like we do in the QubitRepository save).

I wonder if that's really needed. The Gearman server and the AtoM worker should be available in those environments too, the job(s) triggered from another job will need to wait in a single worker environment, but that shouldn't be a big deal in here. What do you think?

lib/job/arUpdateEsActorRelationsJob.class.php
```diff
@@ -0,0 +1,116 @@
...
+      $actors[] = $relation['objectId'];
+      $actors[] = $relation['subjectId'];
+    }
+
+    $actors = array_unique($actors);
+
+    // Omit ID of actor ID itself if not in inclusive mode (for deletes)
+    if (!$inclusive)
+    {
+      $actors = array_unique($actors, [$actorId]);
```
jraddaoui: Should this be `array_diff`? Also, is the inclusive flag really needed? The actor should be saved/deleted in all cases outside the job.

jraddaoui (changes requested):
Thanks @mcantelon!!

There may be an edge case where the relation is made directly using [`QubitRelation`](https://github.com/artefactual/atom/blob/qa/2.6.x/lib/model/QubitRelation.php#L53-L74) but I think we can disregard it as this is getting quite complicated (as expected).

I've added a few comments below, the most important one about getting the previously related actors from ES and not the DB and how they are merged to avoid duplicated updates.

lib/model/QubitActor.php
mcantelon: Ah, that makes sense. Fixed.

lib/job/arUpdateEsActorRelationsJob.class.php
mcantelon: I've simplified the logic and gotten rid of the inclusive flag. When deleting one needs to remove the ID of the actor being deleted from the array of related IDs, but I've just done this outside the function.

lib/model/QubitActor.php
mcantelon: Seems like it's good to do (and didn't take much to implement).

mcantelon:
Hi @jraddaoui ... I've updated. Let me know if you spot anything that needs tweaking!

jraddaoui (approved):
Thanks @mcantelon! Quite a feature to implement, great work!

lib/model/QubitActor.php
jraddaoui: I just noticed this was not addressed. It's not a huge deal but it would be nice to avoid the duplicated updates.

jraddaoui:
@mcantelon, I just noticed one unresolved comment above. It's not a huge deal but it would be nice to avoid the duplicated updates.

mcantelon:
Ah cool... I'll fix. Thanks @jraddaoui !

PR #1076, opened 2020-04-23
Fix date range in description updates, refs #13292

replaceafill:
When description change logging is enabled the criteria for
QubitAuditLog::CREATED_AT is first set with the startDate but gets
immediately overwritten with the endDate.

This changes the criteria to result in an AND in the WHERE clause.


jraddaoui (approved):
Thanks @replaceafill!

jraddaoui (changes requested):
Missing the ref in the commit message now that I notice ;)

replaceafill:
> Missing the ref in the commit message now that I notice ;)

Thanks @jraddaoui! Just added the "refs ..." and fixed a typo :)

PR #1077, opened 2020-04-23
Remove mediatype short option, refs #13288

sbreker:
Removed the media type short option '-m' from both the digital object
delete CLI task and the digital object regenerate derivatives CLI task
as it was not working. Filter by media type using existing
"--media-type" option.


jraddaoui (approved):
Thanks @sbreker!

Would you know why? It's used by a default option or something like that?

PR #1078, opened 2020-04-24
Tweak hierarchy browse pagination, refs #12389

jraddaoui:
- Always show the more button.
- Consider item per pages setting.

### DIFF
--- a/apps/qubit/modules/default/actions/fullTreeViewAction.class.php
+++ b/apps/qubit/modules/default/actions/fullTreeViewAction.class.php
@@ -271,7 +271,7 @@ class DefaultFullTreeViewAction extends sfAction
         $titles[$key] = $node['text'];
       }
 
-      usort($data, function($el1, $el2) { return strnatcmp( $el1[‘text’], $el2[‘text’]); });
+      usort($data, function($el1, $el2) { return strnatcmp($el1['text'], $el2['text']); });
     }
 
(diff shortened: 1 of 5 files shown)


apps/qubit/modules/default/actions/fullTreeViewAction.class.php
```diff
@@ -271,7 +271,7 @@ protected function getNodeOrChildrenNodes($id, $baseReferenceCode, $children = f
         $titles[$key] = $node['text'];
       }
 
-      usort($data, function($el1, $el2) { return strnatcmp( $el1[‘text’], $el2[‘text’]); });
+      usort($data, function($el1, $el2) { return strnatcmp($el1['text'], $el2['text']); });
```
jraddaoui, line 274: Fixes a PHP warning that will become an error:

```
PHP Warning:  Use of undefined constant ‘text’ - assumed '‘text’' (this will throw an Error in a future version of PHP) in /atom/src/apps/qubit/modules/default/actions/fullTreeViewAction.class.php on line 274"
```

sbreker: Good catch!

sbreker (approved):
Looks great @jraddaoui! 👍

jraddaoui:
Partially added (only the more button change) to stable 2.5.x in 140e8ef.

PR #1079, opened 2020-04-24
Fix purge task warnings, refs #13290

jraddaoui:
Configure the context with the task options to avoid session warnings.

### DIFF
--- a/lib/task/tools/purgeTask.class.php
+++ b/lib/task/tools/purgeTask.class.php
@@ -104,7 +104,10 @@ EOF;
       }
     }
 
-    sfContext::createInstance($this->configuration);
+    $configuration = ProjectConfiguration::getApplicationConfiguration(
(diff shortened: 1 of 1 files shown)


lib/task/tools/purgeTask.class.php
```diff
@@ -104,7 +104,10 @@ public function execute($arguments = array(), $options = array())
       }
     }
 
-    sfContext::createInstance($this->configuration);
+    $configuration = ProjectConfiguration::getApplicationConfiguration(
```
djjuhasz, line 107: Wow, this is a good pattern to know. The way symfony builds the sfContext has always been pretty mysterious to me. 👍

djjuhasz (approved):
Looks good @jraddaoui!

jraddaoui:
Thanks @djjuhasz!

PR #1080, opened 2020-04-24
Fix hierarchy browse pagination, refs #13294

jraddaoui:
Only update the pager total count in the first tree-view load.


sbreker (approved):
Looks good @jraddaoui!

jraddaoui:
697010e in stable/2.5.x.

PR #1081, opened 2020-04-28
Title tooltip for search facets, refs #12471

mcantelon:
Makes it so hovering over a search facet shows the full title.

### DIFF
--- a/apps/qubit/modules/search/templates/_aggregation.php
+++ b/apps/qubit/modules/search/templates/_aggregation.php
@@ -15,7 +15,7 @@
         <li <?php if (!isset($filters[$name])) echo 'class="active"' ?>>
           <?php echo link_to(__('All'),
             array($name => null,'page' => null) +
-            $sf_data->getRaw('sf_request')->getParameterHolder()->getAll(), array('title' => '')) ?>
+            $sf_data->getRaw('sf_request')->getParameterHolder()->getAll(), array('title' => __('All'))) ?>
         </li>
       <?php endif; ?>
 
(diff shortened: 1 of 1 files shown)


apps/qubit/modules/search/templates/_aggregation.php
```diff
@@ -26,7 +26,7 @@
         <li <?php if ($active) echo 'class="active"' ?>>
           <?php echo link_to(__(render_value_inline($bucket['display'])) . '<span>, ' . $bucket['doc_count'] . ' ' . __('results') . '</span>',
             array('page' => null, $name => $bucket['key'] == 'unique_language' ? null : $bucket['key']) +
-            $sf_data->getRaw('sf_request')->getParameterHolder()->getAll(), array('title' => '')) ?>
+            $sf_data->getRaw('sf_request')->getParameterHolder()->getAll(), array('title' => $bucket['display'])) ?>
```
djjuhasz: I _think_ it would be good to wrap the display text in the i18n function, e.g. `array('title' => __($bucket['display'])))`?

I don't know how the `$bucket['display']` value is set though, so I'm not 100% sure.

djjuhasz (commented):
I've added one possible change @mcantelon

djjuhasz:
It might also be good to set the "title" value at https://github.com/artefactual/atom/blob/965367980dec871373fee04401b4c5d33de47d12/apps/qubit/modules/search/templates/_aggregation.php#L18 while you are at it?

apps/qubit/modules/search/templates/_aggregation.php
mcantelon: D'oh. Good point. I've updated it.

apps/qubit/modules/search/templates/_aggregation.php
```diff
@@ -26,7 +26,7 @@
         <li <?php if ($active) echo 'class="active"' ?>>
           <?php echo link_to(__(render_value_inline($bucket['display'])) . '<span>, ' . $bucket['doc_count'] . ' ' . __('results') . '</span>',
             array('page' => null, $name => $bucket['key'] == 'unique_language' ? null : $bucket['key']) +
-            $sf_data->getRaw('sf_request')->getParameterHolder()->getAll(), array('title' => '')) ?>
+            $sf_data->getRaw('sf_request')->getParameterHolder()->getAll(), array('title' => __(render_value_inline($bucket['display'])))) ?>
```
djjuhasz: I think it's probably better to strip any markdown and just use the plain text value, e.g. `'title' => __(strip_markdown($bucket['display'])`.  I'm not sure if it's best to strip markdown from the link value on line 28 too, e.g. `link_to(__(strip_markdown($bucket['display']))`?

djjuhasz (commented):
Sorry @mcantelon, one more suggestion.

apps/qubit/modules/search/templates/_aggregation.php
mcantelon: Didn't know we had such a function! Yeah, seems like doing it on line 28 makes too. Updated!

djjuhasz (approved):
Looks good to me! :)

PR #1082, opened 2020-05-01
Add lang facet fix, refs #13299

fiver-watson:
This fix has been collaboratively prepared by Monica Wood (Library Systems, University of Tasmania) and Artefactual Systems (input from @jraddaoui). The relevant discussion can be seen in the user forum, here: 

* https://groups.google.com/d/msg/ica-atom-users/134crsm9gEs/Ppi70Z1nBAAJ

Currently with other facets in AtoM's information object search/browse page, any facet filter with less than 2 populated facets will be automatically hidden, since 0 or 1 facet results does not in fact provide users with a useful means of limiting the results. However, currently the language facet filter will always be displayed (even in a monolingual site), because it includes a count of unique records in addition to any language count - meaning there are always at least 2 facets available, and the facet filter is not hidden. 

This fix, propposed by Monica Wood and refined by @jraddaoui, inserts a new custom condition for displaying the language facet into line 1 of `apps/qubit/modules/search/templates/_aggregation.php`, raising the required facet count for displaying the language facet filter from 2 to 3 results.


jraddaoui (approved):
Thanks @fiver-watson!

PR #1083, opened 2020-05-01
Accession import Alt Id duplicates, refs #13262

sbreker:
Fixes an issue where new alt id type terms were being created in the
database even if they already existed.


djjuhasz (approved):
Looks good to me @sbreker :)

PR #1084, opened 2020-05-02
Don't suppress errors with "@", refs #13307

djjuhasz:
The @ operator on the first argument of the end() function was throwing
the following error in PHP 7.3.17

PHP Fatal error:  Only variables can be passed by reference in
/usr/share/nginx/atom/plugins/qtSwordPlugin/lib/qtPackageExtractorBase.class.php
on line 237


mcantelon (approved):
Looks good! One line of the commit message is longer than 72 chars but given it's a literal it seems inavoidable so 🤷 .

PR #1085, opened 2020-05-02
Actor relation CSV import fix, refs #13303

mcantelon:
Changed actor relation CSV import so import doesn't halt, and instead
skips the current row, if a non-existent relation type is encountered.

### DIFF
--- a/lib/task/import/csvAuthorityRecordRelationImportTask.class.php
+++ b/lib/task/import/csvAuthorityRecordRelationImportTask.class.php
@@ -113,14 +113,15 @@ EOF;
           $self->columnValue('subjectAuthorizedFormOfName'), ['culture' => $self->columnValue('culture')]);
 
         // Determine type ID of relationship type
-        $relationTypeId = array_search(
+        $relationTypeId = array_search_case_insensitive(
           $self->columnValue('relationType'),
           $self->status['actorRelationTypes'][$self->columnValue('culture')]
         );
 
         if (!$relationTypeId)
         {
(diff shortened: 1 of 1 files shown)


lib/task/import/csvAuthorityRecordRelationImportTask.class.php
```diff
@@ -113,14 +113,15 @@ private function import($filepath, $sourceName, $indexDuringImport = false)
...
+        $relationTypeId = array_search_case_insensitive(
           $self->columnValue('relationType'),
           $self->status['actorRelationTypes'][$self->columnValue('culture')]
         );
 
         if (!$relationTypeId)
         {
-          throw new sfException(sprintf('Unknown relationship type %s:', $self->columnValue('relationType')));
+          $error = sprintf('Unknown relationship type "%s"... skipping row.', $self->columnValue('relationType'));
+          print $self->logError($error);
```
jraddaoui, line 124: Is `print` really needed in here and in L137?

jraddaoui (approved):
Looks good @mcantelon!

lib/task/import/csvAuthorityRecordRelationImportTask.class.php
mcantelon: Yeah print or echo are needed as the log error function returns the argument prepended with the current CSV row number of the import (kind of unintuitive, alas).

mcantelon:
Thanks @jraddaoui !

PR #1086, opened 2020-05-03
Actor relation CSV export fixes, refs #13305

mcantelon:
Fixes incorrect column names in actor relation CSV export (issue 13306)
and issue with relation type column not being set properly.

### DIFF
--- a/lib/flatfile/csvActorExport.class.php
+++ b/lib/flatfile/csvActorExport.class.php
@@ -57,35 +57,47 @@ class csvActorExport extends QubitFlatfileExport
   {
     $rows = array();
 
-    foreach ($resource->getActorRelations() as $item)
+    foreach ($resource->getActorRelations() as $relation)
     {
-      $relatedEntity = $item->getOpposedObject($resource->id);
+      $relatedEntity = $relation->getOpposedObject($resource->id);
 
-      if (QubitTerm::ROOT_ID == $item->type->parentId)
+      // Take note of relationship type
+      $relationType = $relation->type;
+
+      /* If the current actor being exported is the subject, rather than object, of a
+         relation then we check for a converse relationship type, if any, to put in
+         the "relationType" column.
+
+         For example if the current actor is the subject of a "controls" type relation
+         (where the object "controls" the subject) then the converse relationship type
+         would be "is controlled by".
+
+         Some relation types like "is the sibling of", however, are reciprical and
+         don't have converse types. A lookup of the converse type for this relation type
+         will return null.
+      */
+      if ($relation->subjectId == $resource->id)
       {
-        $category = $item->type;
-      }
-      else
-      {
-        $category = $item->type->parent;
+        $converseRelation = $relationType->getConverseActorRelationTerm();
+        $relationType = (empty($converseRelation)) ? $relationType : $converseRelation;
       }
 
       $rows[] = array(
-        'sourceAuthorizedFormOfName' => $resource->authorizedFormOfName,
-        'targetAuthorizedFormOfName' => $relatedEntity->authorizedFormOfName,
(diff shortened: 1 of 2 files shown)


lib/flatfile/csvActorExport.class.php
```diff
@@ -57,28 +57,27 @@ private function exportRelations($filename, $resource)
...
-        'targetAuthorizedFormOfName' => $relatedEntity->authorizedFormOfName,
-        'category'                   => (string)$category, // Return string representation for QubitTerm
-        'description'                => $item->description,
-        'date'                       => $item->date,
-        'startDate'                  => $item->startDate,
-        'endDate'                    => $item->endDate,
-        'culture'                    => $resource->culture
+        'objectAuthorizedFormOfName'  => $resource->authorizedFormOfName,
+        'subjectAuthorizedFormOfName' => $relatedEntity->authorizedFormOfName,
+        'relationType'                => (string)$relationType, // Return string representation for QubitTerm
```
djjuhasz, line 88: Does `(string)$relationType` return the correct i18n value if the culture is not 'en'?

lib/flatfile/csvActorExport.class.php
```diff
@@ -57,28 +57,27 @@ private function exportRelations($filename, $resource)
...
+      // Determine appropriate relation type term
+      $relationType = $relation->type;
+      if ($relation->subjectId == $resource->id)
       {
-        $category = $item->type;
-      }
-      else
-      {
-        $category = $item->type->parent;
+        $converseRelation = $relationType->getConverseActorRelationTerm();
```
djjuhasz, line 81: Nitpick: I think it would be helpful to add a comment here about why you want the converse relation term, if there is one.  I assume it's because the "forward" direction for the relation type is information_object -> actor, and this is going the other direction?

djjuhasz (commented):
Looks pretty good @mcantelon - I just had one question and one nitpick. :)

lib/flatfile/csvActorExport.class.php
mcantelon: Hmm... good question! I'll test...

lib/flatfile/csvActorExport.class.php
mcantelon: That's the reason, yeah... I'll add a comment to clarify that.

lib/flatfile/csvActorExport.class.php
mcantelon: Basically how CSV exports seem to work, based on how the I.O. export works, is that everything gets exported in the site's default language (set in `apps/qubit/config/settings.yml`... usually English) so there's no cultural fallback when exporting something that doesn't exist in the default language (otherwise if you round-tripped the data you'd get translations showing up as the default language).

So seems like it's working correctly? It seems like a possible oversight that our CSV export tools don't have a CLI option to specify an alternative culture.

lib/flatfile/csvActorExport.class.php
mcantelon: (Added. Also renamed a function and added another comment to make export functionality clearer.)

djjuhasz: Wow, great comment! 👍

lib/flatfile/csvActorExport.class.php
djjuhasz: Yeah, now that you mention it I recalling rumors that the CSV export doesn't work in other cultures... :(

djjuhasz (approved):
Looks good @mcantelon! 👍

lib/flatfile/csvActorExport.class.php
mcantelon: Thanks mister! :D

PR #1087, opened 2020-05-06
Accession import donor fax field, refs #13310

sbreker:
Add donor fax field to accession CSV import.


mcantelon (approved):
Looks good to me!

PR #1088, opened 2020-05-12
Authority relation ES indexing fix, refs #13302

mcantelon:
Fixed issue where looking up actor relations in Elasticsearch, for an
actor that hasn't been indexed, would throw an exception.


mcantelon:
(Without this an exception gets thrown, when running `tools:purge`.)

jraddaoui (approved):
Thanks @mcantelon!

I'm currently doing some CI experiments with Github Actions that would have helped spotting this issue ... https://github.com/jraddaoui/atom/pull/1/checks?check_run_id=666158687

mcantelon:
Thanks @jraddaoui ! And very cool re: the CI!

PR #1093, opened 2020-05-15
 Fix multi-value autocomplete, refs #12709

djjuhasz:
When submitting an information object that adds more than one new related
resources (e.g. two subject access point terms) a 500 error would
usually occur because one (or more) of the related resources (e.g. one
of the subject terms) was not created successfully before trying to link
it to the information_object, causing a foreign key constraint failure.

This fix delays the submit of main (e.g. information_object) form until after
all new related resources have been completed (via iframe submit) and have
returned a response.

In addition I normalized the code for adding new rows to a multi-row
autocomplete, and for adding a new iframe to the DOM for creating new
related resources.

### DIFF
--- a/js/autocomplete.js
+++ b/js/autocomplete.js
@@ -13,17 +13,196 @@
           if (1 > --count) {
             $form
 
-              // Unbind submit listeners to avoid triggering again
+              // Unbind submit listener so it doesn't preventDefault()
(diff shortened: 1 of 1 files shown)


djjuhasz:
Needs work to fix transaction deadlock issue.

PR #1095, opened 2020-05-20
Nested set and foreign keys

jraddaoui:
Refs #13222, #13237, #13313 and #10306.

See commit messages for more info.

### DIFF
--- a/config/schema.yml
+++ b/config/schema.yml
@@ -15,7 +15,7 @@ propel:
     description_detail_id: { type: integer, foreignTable: term, foreignReference: id, onDelete: setnull }
     description_identifier: varchar(1024)
     source_standard: varchar(1024)
-    parent_id: { type: integer, foreignTable: actor, foreignReference: id }
+    parent_id: { type: integer, foreignTable: actor, foreignReference: id, onDelete: cascade }
 
   actor_i18n:
     authorized_form_of_name: varchar(1024)
@@ -106,7 +106,7 @@ propel:
     byte_size: bigint
     checksum: varchar(255)
     checksum_type: type: varchar(50)
-    parent_id: { type: integer, foreignTable: digital_object, foreignReference: id }
+    parent_id: { type: integer, foreignTable: digital_object, foreignReference: id, onDelete: cascade }
     _indexes:
       path: [path(768)]
 
@@ -118,7 +118,7 @@ propel:
     end_time: time
     type_id: { type: integer, required: true, foreignTable: term, foreignReference: id, onDelete: cascade }
     object_id: { type: integer, foreignTable: object, foreignReference: id, onDelete: cascade }
-    actor_id:
+    actor_id: { type: integer, foreignTable: actor, foreignReference: id, onDelete: setnull }
 
   event_i18n:
     name: varchar(1024)
@@ -127,9 +127,9 @@ propel:
 
   function_object:
     id: { type: integer, required: true, primaryKey: true, foreignTable: object, foreignReference: id, onDelete: cascade, inheritanceKey: true }
-    type_id: { type: integer, foreignTable: term, foreignReference: id }
-    description_status_id: { type: integer, foreignTable: term, foreignReference: id }
-    description_detail_id: { type: integer, foreignTable: term, foreignReference: id }
+    type_id: { type: integer, foreignTable: term, foreignReference: id, onDelete: setnull }
+    description_status_id: { type: integer, foreignTable: term, foreignReference: id, onDelete: setnull }
+    description_detail_id: { type: integer, foreignTable: term, foreignReference: id, onDelete: setnull }
     description_identifier: varchar(1024)
     source_standard: varchar(1024)
 
@@ -150,9 +150,9 @@ propel:
     identifier: varchar(1024)
     oai_local_identifier: { type: integer, index: unique, required: true, autoIncrement: true }
     level_of_description_id: { type: integer, foreignTable: term, foreignReference: id, onDelete: setnull }
-    collection_type_id: { type: integer, foreignTable: term, foreignReference: id }
-    repository_id:
-    parent_id: { type: integer, foreignTable: information_object, foreignReference: id }
+    collection_type_id: { type: integer, foreignTable: term, foreignReference: id, onDelete: setnull }
+    repository_id: { type: integer, foreignTable: repository, foreignReference: id, onDelete: setnull }
+    parent_id: { type: integer, foreignTable: information_object, foreignReference: id, onDelete: cascade }
     description_status_id: { type: integer, foreignTable: term, foreignReference: id, onDelete: setnull }
     description_detail_id: { type: integer, foreignTable: term, foreignReference: id, onDelete: setnull }
     description_identifier: varchar(1024)
@@ -210,7 +210,7 @@ propel:
     object_id: { type: integer, required: true, foreignTable: object, foreignReference: id, onDelete: cascade }
     type_id: { type: integer, foreignTable: term, foreignReference: id, onDelete: setnull }
     scope: varchar(1024)
-    user_id:
+    user_id: { type: integer, foreignTable: user, foreignReference: id, onDelete: setnull }
 
   note_i18n:
     content: longvarchar
@@ -269,7 +269,7 @@ propel:
 
   premis_object:
     id: { type: integer, required: true, primaryKey: true, foreignTable: object, foreignReference: id, onDelete: cascade, inheritanceKey: true }
-    information_object_id:
+    information_object_id: { type: integer, foreignTable: information_object, foreignReference: id, onDelete: cascade }
     puid: varchar(255)
     filename: varchar(1024)
     last_modified: timestamp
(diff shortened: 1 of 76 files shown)


jraddaoui:
Hi @sbreker and @djjuhasz, while the PR is still in progress (I'll address #13239 in here too) the current commits relate to other issues and you could start reviewing if you have some time. Thanks!

config/schema.yml
```diff
@@ -287,9 +287,9 @@ propel:
 
   relation:
     id: { type: integer, required: true, primaryKey: true, foreignTable: object, foreignReference: id, onDelete: cascade, inheritanceKey: true }
-    subject_id: { type: integer, required: true, foreignTable: object, foreignReference: id }
-    object_id: { type: integer, required: true, foreignTable: object, foreignReference: id }
-    type_id: { type: integer, foreignTable: term, foreignReference: id }
+    subject_id: { type: integer, required: true, foreignTable: object, foreignReference: id, onDelete: cascade }
+    object_id: { type: integer, required: true, foreignTable: object, foreignReference: id, onDelete: cascade }
+    type_id: { type: integer, foreignTable: term, foreignReference: id, onDelete: cascade }
```
djjuhasz, line 292: This one worries me a bit. In principle I think it's true that a relation without at `type_id` doesn't mean much, my concern is that a user deletes a term without realizing it will also delete all related relation rows.  It's a tough choice but I lean slightly towards "onDelete: setnull" for type_id to avoid unintended relation deletes.

lib/task/migrate/migrations/arMigration0177.class.php
```diff
@@ -75,102 +75,27 @@ public function up($configuration)
...
-        'tableName' => 'digital_object',
-        'columnName' => 'object_id',
-        'refTableName' => 'object',
-        'newConstraintName' => 'digital_object_FK_2',
-        'onDelete' => 'ON DELETE CASCADE',
+        'table' => 'digital_object',
+        'column' => 'object_id',
+        'refTable' => 'object',
+        'constraint' => 'digital_object_FK_2',
+        'onDelete' => ' ON DELETE CASCADE',
```
djjuhasz: I'd prefer to see the leading space in the SQL templates, rather than having it in the 'onDelete' value.  To me it seems too easy to forgot to add the leading space when adding new migrations.

lib/propel/builder/QubitObjectBuilder.php
```diff
@@ -1924,6 +1929,46 @@ public function hasChildren()
...
+      	  UNION ALL
+      	  SELECT tb2.{$selfFkRefColName}, tb2.{$selfFkColName}, aas.lev + 1
+          FROM {$selfFkTableName} tb2
+          JOIN aas ON aas.{$selfFkColName}=tb2.{$selfFkRefColName}
+      	)
+      	SELECT {$selfFkRefColName}, lev FROM aas
+      )";
+
+      \$criteria = new Criteria;
+      \$criteria->addJoin({$selfFkRefColCons}, 'cte.id', "RIGHT JOIN \$cte");
```
djjuhasz, line 1963: Huh, this is cool - I didn't realize you could use raw SQL in a Propel Criteria join like this. 😮

lib/task/propel/propelBuildNestedSetTask.class.php
```diff
@@ -61,8 +61,8 @@ public function execute($arguments = array(), $options = array())
 
     $tables = array(
       'information_object' => 'QubitInformationObject',
-      'actor' => 'QubitActor',
-      'term' => 'QubitTerm'
+      'term' => 'QubitTerm',
+      'menu' => 'QubitMenu'
```
djjuhasz, line 65: lol, I can't believe QubitMenu was left out of this task for so long! 😆

plugins/qbAclPlugin/lib/QubitAcl.class.php
```diff
@@ -269,7 +269,7 @@ protected function addRole($role)
     {
       if (!in_array($role->id, $this->_roles))
       {
-        foreach ($role->ancestors->andSelf()->orderBy('lft') as $ancestor)
+        foreach ($role->getAncestorsAndSelfForAcl() as $ancestor)
```
djjuhasz, line 272: Awesome! 😁

djjuhasz (commented):
This is looking great so far @jraddaoui!  A few questions/comments but no blockers. 👍

jraddaoui:
Thanks for the quick and great feedback @djjuhasz!

config/schema.yml
jraddaoui: I kind of agree with @djjuhasz but I'd like to hear your opinion in this one, @fiver-watson.

lib/task/migrate/migrations/arMigration0177.class.php
jraddaoui: Great point!

lib/propel/builder/QubitObjectBuilder.php
jraddaoui: Me neither :smile:! It wasn't easy to figure out and I feel like I'm hacking it, but it works :stuck_out_tongue:.

config/schema.yml
fiver-watson: Not sure I'm totally understanding this... What would be the advantage of leaving a relation row that points to a NULL, @djjuhasz  - can you think of real-world examples where someone would want to delete a term but still have references to the missing term in the related entities? And, would the user have any option for deleting that relation once the term is gone? Thanks1

lib/propel/builder/QubitObjectBuilder.php
sbreker: This is going to really help the digital object browse page!

config/schema.yml
sbreker: Current functionality appears to be 'ON DELETE RESTRICT' which will prevent deletion of terms for which relations exist.

Dan raises a good point - if relation.type_id is NULL, then the relation will be ignored in any queries that filter on types of relations.

sbreker:
Great work so far @jraddaoui!

config/schema.yml
djjuhasz: @fiver-watson I can't think of any use case where having a relation row with no type_id is desirable.  

My concern is that somebody deleting a term (especially via direct MySQL access rather than the UI) that is used for a relation.type_id, without realizing that it's going to delete every relationship that uses that term.  E.g. a user naively deletes the "controls" term, and wipes out all "controls" relationships.  At least if the relation.type_id is set to null, the user can recover by adding a new type_id to the row. After deleting the term, the user could delete all relation rows where relation.type_id = NULL if that's the objective.

It's a tough call though, allowing relation.type_id to be null does have the potential to hide those relation rows from the UI and in effect make them hidden data.

fiver-watson: I've talked about this a bit more with @jraddaoui and we are both still kind of leaning towards ON DELETE CASCADE. In the UI, we at least give users a warning page before a deletion (though with terms we don't give a count of affected relations). In the DB, we **always** recommend making a backup before directly manipulating the data using SQL - i'm not sure that leaving empty rows to account for potentially undoing user error is a good tradeoff in terms of overall database consistency. 

We could potentially consider rewording the warning screen that appears when a term is deleted, to add a clearer warning? Even without a count, that might be useful. 

Ultimately, i think this is a decision that I trust you all to come to consensus around, as it feels more about project maintenance than end user experience and usage. I think your argument is an interesting one David, but what would it mean if we applied it consistently throughout AtoM?

djjuhasz: Okay, I'm fine with ON DELETE CASCADE - I agree that it's better for database consistency and I'm probably being over-cautious in worrying about unanticipated deletes.

djjuhasz: On the topic of database consistency, I wonder if it's possible to do an ON DELETE CASCADE of the `object` row when a related object is deleted?  I think it would require putting a foreign key constraint on e.g. `information_object.id` to `object.id` which may cause a circular dependency, but it would sure be nice to avoid leaving behind useless `object` table rows.

jraddaoui: Hi @djjuhasz, I'm not sure how that would work in the ORM but I think @fiver-watson added some notes about this in the SQL work he did. I'd prefer not to go that way and ask to do the deletes from the object row, otherwise we may want to do the same in other inheritances (QubitRepository and others from QubitActor for example) and it could get messy.

jraddaoui: And thanks all for the discussion.

lib/propel/builder/QubitObjectBuilder.php
jraddaoui: Thanks @sbreker!

We were already using CTE but it was in a subquery so this should be a bit better. Additionally, it removes the extra processing from `andSelf()` and `orderBy()`:

https://github.com/artefactual/atom/blob/qa/2.6.x/lib/QubitQuery.class.php#L137-L170

:rocket:

jraddaoui:
Hi @djjuhasz and @sbreker,

Please, let me know if you think this is ready for merge. I'll address #13239 in a different PR in the end to allow testing this changes in the meantime.

Thanks!

PS: I'll squash only the last two commits if it's ready and merge with `--ff-only`.

djjuhasz (approved):
Looks good to me. :)

PR #1096, opened 2020-05-20
Fix relation type autocomplete issue, refs #13325

mcantelon:
Fixed issue, on the page for adding a new authority record, where
adding a new relationship would add inappropriate text to the
autocomplete results for relationship type.


djjuhasz (approved):
Looks okay to me. 👍

PR #1097, opened 2020-05-21
Fix error when adding two subjects, refs #12709

djjuhasz:
When submitting an information object that adds more than one new related resources (e.g. two subject access point terms) a 500 error would usually occur because one (or more) of the related resources (e.g. one of the subject terms) was not created successfully before trying to link it to the information_object, causing a foreign key constraint failure.

This fix involved multiple changes:
- Refactored the autocomplete.js code to reduce the amount of duplicate code, and clarify the flow,
- Changed how autcomplete.js submit logic works to ensure the primary form is submitted only after all iframes forms have been submitted returned a response
- Added logic to the QubitTransactionFilter to re-execute the code and SQL writes when MySQL returns a transaction deadlock error
- Removed some unnecessary and confusing calls to `QubitTransactionFilter::getConnection()`

### DIFF
--- a/apps/qubit/config/filters.yml
+++ b/apps/qubit/config/filters.yml
@@ -1,9 +1,6 @@
 # You can find more information about this file on the symfony website:
 # http://www.symfony-project.org/reference/1_4/en/12-Filters
 
-transaction:
-  class: QubitTransactionFilter
-
 history:
   class: sfHistoryPluginFilter
 
@@ -24,6 +21,9 @@ QubitMeta:
 
 QubitLimitResults:
   class: QubitLimitResults
+
+QubitTransaction:
+  class: QubitTransactionFilter
 
 cache: ~
 execution: ~
--- a/lib/filter/QubitTransactionFilter.class.php
+++ b/lib/filter/QubitTransactionFilter.class.php
@@ -20,19 +20,8 @@
 class QubitTransactionFilter extends sfFilter
 {
   protected static
-    $connection = null,
     $retry = 0,
     $retryLimit = 3;
-
-  public static function getConnection()
-  {
-    if (!isset(self::$connection))
-    {
-      self::$connection = Propel::getConnection();
-    }
-
-    return self::$connection;
-  }
 
   protected function retry()
   {
@@ -63,10 +52,6 @@ class QubitTransactionFilter extends sfFilter
     $filterChain = new sfFilterChain();
     $filterChain->loadConfiguration($actionInstance);
 
-    $this->context->getEventDispatcher()->notify(
-      new sfEvent($this, 'transaction.retry')
-    );
-
     // Execute whole filter chain again
     $filterChain->execute();
 
@@ -77,7 +62,7 @@ class QubitTransactionFilter extends sfFilter
   {
     try
     {
-      $conn = self::getConnection();
+      $conn = Propel::getConnection();
       $conn->beginTransaction();
     }
     catch (PropelException $e)
--- a/lib/filter/QubitTransactionFilter.class.php
+++ b/lib/filter/QubitTransactionFilter.class.php
@@ -20,7 +20,9 @@
 class QubitTransactionFilter extends sfFilter
 {
   protected static
-    $connection = null;
+    $connection = null,
+    $retry = 0,
+    $retryLimit = 3;
 
   public static function getConnection()
   {
@@ -30,6 +32,45 @@ class QubitTransactionFilter extends sfFilter
     }
 
     return self::$connection;
+  }
+
+  protected function retry()
+  {
+    // If we've hit the retry limit, abort and return false
+    if (self::$retry++ > self::$retryLimit)
+    {
+      return false;
+    }
+
+    // Log a warning
+    if (sfConfig::get('sf_logging_enabled'))
+    {
+      $this->context->getLogger()->warning(
+        sprintf('Encountered a SQL transaction deadlock, retry %d of %d',
+          self::$retry, self::$retryLimit
+        )
+      );
+    }
+
+    // Get the current action instance
+    $actionInstance = $this->context
+      ->getController()
+      ->getActionStack()
+      ->getLastEntry()
+      ->getActionInstance();
+
+    // Create a new filter chain and reload config
+    $filterChain = new sfFilterChain();
+    $filterChain->loadConfiguration($actionInstance);
+
+    $this->context->getEventDispatcher()->notify(
+      new sfEvent($this, 'transaction.retry')
+    );
+
+    // Execute whole filter chain again
+    $filterChain->execute();
+
+    return true;
   }
 
   public function execute($filterChain)
@@ -50,6 +91,27 @@ class QubitTransactionFilter extends sfFilter
       if (isset($conn))
       {
         $conn->commit();
+      }
+    }
+    catch (PDOException $e)
+    {
+      // Rollback the transaction
+      $conn->rollBack();
+
+      // If there was a transaction deadlock error (MySQL error code 1213)
+      if (isset($conn) && $e->errorInfo[1] == 1213)
+      {
+        // Retry the current action (returns false when out of retries)
+        if (!$this->retry())
+        {
+          // If we've hit the retry limit, re-throw the exception
+          throw $e;
+        }
+      }
+      else
+      {
+        // Re-throw any other PDOExceptions
+        throw $e;
       }
     }
     catch (Exception $e)
--- a/lib/model/QubitJob.php
+++ b/lib/model/QubitJob.php
@@ -275,7 +275,7 @@ class QubitJob extends BaseJob
       // so the resources modified are persisted before the assigned worker starts
       // processing the task. If we don't do this now the transaction will be committed
       // once this request is processed but not before the worker hits the database.
-      $connection = QubitTransactionFilter::getConnection();
+      $connection = Propel::getConnection();
       $connection->commit();
 
       // Start a new transaction as there might be more database work within the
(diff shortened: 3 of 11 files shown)


apps/qubit/config/filters.yml
```diff
@@ -25,5 +22,8 @@ QubitMeta:
 QubitLimitResults:
   class: QubitLimitResults
 
+QubitTransaction:
+  class: QubitTransactionFilter
+
```
jraddaoui, line 27: This filters can be extended/overwritten in the plugins, we should make sure those reference `QubitTransaction` now. I'm not sure if the reordering affects those extensions too but we should check that out too if the order is important.

lib/model/QubitJob.php
```diff
@@ -275,7 +275,7 @@ public static function runJob($jobName, $jobParams = array())
       // so the resources modified are persisted before the assigned worker starts
       // processing the task. If we don't do this now the transaction will be committed
       // once this request is processed but not before the worker hits the database.
-      $connection = QubitTransactionFilter::getConnection();
+      $connection = Propel::getConnection();
```
jraddaoui, line 278: I think both return the same connection, but in this case using `QubitTransaction` makes clearer from where the transaction comes. The comment may be good enough, but I'd leave this one like it was, just in case.

lib/model/QubitObject.php
```diff
@@ -150,23 +150,18 @@ public function save($connection = null)
 
   protected function insert($connection = null)
   {
-    if (!isset($connection))
-    {
-      $connection = QubitTransactionFilter::getConnection(QubitObject::DATABASE_NAME);
-    }
-
```
jraddaoui, line 157: Why removing this one entirely? I'd do the same change that is done below.

lib/filter/QubitTransactionFilter.class.php
```diff
@@ -32,6 +34,45 @@ public static function getConnection()
...
+    // Create a new filter chain and reload config
+    $filterChain = new sfFilterChain();
+    $filterChain->loadConfiguration($actionInstance);
+
+    $this->context->getEventDispatcher()->notify(
+      new sfEvent($this, 'transaction.retry')
+    );
+
+    // Execute whole filter chain again
+    $filterChain->execute();
```
jraddaoui: :mage:

That's a nice one @djjuhasz! I didn't know you could something like this and I have a few questions ...

- It's rare to see actions chained together and I wonder if it even matters but, should we get the first action entry to fully retry the request?
- What would happen when the request cycle involves other processes apart from the DB transaction? I'm thinking in DO uploads for example, would the file be copied again, etc.?

jraddaoui (changes requested):
Thanks a lot for looking into this @djjuhasz!

Quite some magic going on in that filter :smile: Please, see my comments and questions below, the only blocker would be the plugin filters.

PS. The actual changes from the autocomplete Javascript were quite hard to find. It's nice that you moved the cosmetic changes to a different commit, but I'd have appreciated another one for the reformatting part.

lib/filter/QubitTransactionFilter.class.php
jraddaoui: Another minor thing, why the event dispatcher? I'd not add it unless it's listened somewhere else.

apps/qubit/config/filters.yml
djjuhasz: Huh, okay I'll have a look at the plugins for filter config.

lib/model/QubitJob.php
djjuhasz: When I saw `QubitTransactionFilter::getConnection()` my assumption was that transaction is started by the `getConnection()` method call, whereas `Propel::getConnection()` doesn't start a transaction, which is false.  I think having a separate `QubitTransactionFilter::getConnection()` method that is basically just a wrapper for `Propel::getConnection()` is misleading and unnecessary.

I'm tempted to get rid of `QubitTransactionFilter::getConnection()` altogether, but I think it does help legibility and factoring in `QubitTransactionFilter::execute()`.  Maybe it's best to make `QubitTransactionFilter::getConnection()` a protected method to make it clear that it's not part of the public signature.

jraddaoui: The transaction is started after getting the connection, when `$conn->beginTransaction();` is called. In the end `QubitTransactionFilter::getConnection()` calls `Propel::getConnection()` if it doesn't have a connection already. In both cases (I am almost sure) the same connection is obtained so I think it's not a big deal to use one or the other but, to be sure, calling `QubitTransactionFilter::getConnection()` gives me a bit more confident about having the same connection that started the transaction.

lib/model/QubitObject.php
djjuhasz: `$connection` is only used in the local scope as an argument to `BaseObject::insert()` and `self::insertSlug()`, and both methods already use the same `$connection = Propel::getConnection()` pattern, so it's unnecessary here.

lib/filter/QubitTransactionFilter.class.php
djjuhasz: > It's rare to see actions chained together and I wonder if it even matters but, should we get the first action entry to fully retry the request?

My current theory is that the whole filter chain is executed for *every* action, and that only the current (last) action is relevant to $filter->execute(). This theory is based on:

1. The `$filterChain->loadConfiguration($actionInstance)` requires a single `$actionInstance`
2. The `sfController::forward()` method adds a new action to the action stack AND [creates and executes a new $filterChain](https://github.com/artefactual/atom/blob/qa/2.6.x/vendor/symfony/lib/controller/sfController.class.php#L224)
3. The `sfExecutionFilter()` only executes the [last action on the stack](https://github.com/artefactual/atom/blob/qa/2.6.x/vendor/symfony/lib/filter/sfExecutionFilter.class.php#L35) (it's where I got the code to get the $actionInstance in QubitTransactionFilter)

djjuhasz: >What would happen when the request cycle involves other processes apart from the DB transaction? I'm thinking in DO uploads for example, would the file be copied again, etc.?

The request is processed before the filter chain is executed, but all of the code in the filter chain and action execution stages will be re-run.  I think it's necessary to run the filter/action code again when redoing the transaction to avoid problems like the nested set values being stale, and because there is no way (that I know of) with symfony or Propel to *just* replay all of the database read/writes in isolation from the action code execution.

I found this summary of the front controller execution from [Inside the Controller Layer](https://symfony.com/legacy/doc/gentle-introduction/1_4/en/06-Inside-the-Controller-Layer) "The Front Controller's Job in Detail" to be helpful:

1. Load the project configuration class and the symfony libraries.
2. Create an application configuration and a symfony context.
3. Load and initiate the core framework classes.
4. Load the configuration.
5. Decode the request URL to determine the action to execute and the request parameters.
6. If the action does not exist, redirect to the 404 error action.
7. Activate filters (for instance, if the request needs authentication).
8. Execute the filters, first pass.
9. Execute the action and render the view.
10. Execute the filters, second pass.
11. Output the response.

By re-executing the filter chain, I believe steps 7-10 are re-run, then the response is output (step 11).

djjuhasz: > Another minor thing, why the event dispatcher? I'd not add it unless it's listened somewhere else.

Yes, fair enough.  I was copying the pattern from https://github.com/artefactual/atom/blob/qa/2.6.x/vendor/symfony/lib/controller/sfController.class.php#L224-L238 but in the absence of listeners, the event dispatcher notification is not doing anything.

djjuhasz:
> PS. The actual changes from the autocomplete Javascript were quite hard to find. It's nice that you moved the cosmetic changes to a different commit, but I'd have appreciated another one for the reformatting part.

Sorry about that.  My development process for the autocomplete.js changes was not methodical - there was no "refactor" stage before changing the functionality, it was all mixed together.  I don't know any practical way to separate out a clean "refactor" commit after the fact in a case like this.

lib/filter/QubitTransactionFilter.class.php
jraddaoui: > By re-executing the filter chain, I believe steps 7-10 are re-run, then the response is output (step 11).

Yeah, what worries me the most is the re-execution of step 9 but lets hope the actions that may cause a DB deadlock don't do a lot of not idempotent work outside the DB changes.

lib/filter/QubitTransactionFilter.class.php
```diff
@@ -20,20 +20,9 @@
...
-  public static function getConnection()
-  {
-    if (!isset(self::$connection))
-    {
-      self::$connection = Propel::getConnection();
-    }
-
-    return self::$connection;
-  }
-
```
jraddaoui, line 34: (╯°□°）╯︵ ┻━┻

jraddaoui (commented):
Thanks @djjuhasz, great work!

> Sorry about that. My development process for the autocomplete.js changes was not methodical - there was no "refactor" stage before changing the functionality, it was all mixed together. I don't know any practical way to separate out a clean "refactor" commit after the fact in a case like this.

Np David, maybe doing the reorganization into functions in a different commit would have helped, but not a big deal at this point.

jraddaoui (approved):

lib/filter/QubitTransactionFilter.class.php
djjuhasz: Yes, there's a lot of magic going on behind the curtain here.  I wish I knew a practical way to determine if this approach is creating regressions, but without good test coverage I can't see how. 🤷 As you say, hopefully the code execution is effectively idempotent outside of the DB transaction boundaries.

djjuhasz: I guess worst-case scenario it won't take a lot of work to undo the transaction retry code, so if testing does turn up regressive behaviour, reversing the changes can be the exit strategy.

djjuhasz:
Merged to qa/2.6.x: 0f22ac4520629bae5123a6c3953982ffa6b9fc77

PR #1098, opened 2020-05-21
Add ID to actor relation export, refs #13329

mcantelon:
Changed the authority record relation export logic, in the
csv:authority-export task, so there's a "legacyId" column and it gets
set to a relation's ID.


peterVG:
A question that's arisen, do we even need a legacyId column on relations data? If not, it should be removed from the template and from the import logic that's currently looking for it. Otherwise, we should go ahead with this PR as is.

mcantelon:
Changing this to "work in progress" until we figure out legacyId issue.

mcantelon:
Closed as we decided not to do this.

PR #1100, opened 2020-05-21
Index actor relations after import, refs #13328

mcantelon:
Added logic, to CSV authority record relation import, to keep track
of which actors were related and to, after import, update their
relation data in Elasticsearch.

### DIFF
--- a/lib/task/import/csvAuthorityRecordRelationImportTask.class.php
+++ b/lib/task/import/csvAuthorityRecordRelationImportTask.class.php
@@ -118,6 +118,7 @@ EOF;
           $self->status['actorRelationTypes'][$self->columnValue('culture')]
         );
 
+        // Skip invalid relationships
         if (!$relationTypeId)
         {
           $error = sprintf('Unknown relationship type "%s"... skipping row.', $self->columnValue('relationType'));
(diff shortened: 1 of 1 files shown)


lib/task/import/csvAuthorityRecordRelationImportTask.class.php
```diff
@@ -172,6 +184,19 @@ private function import($filepath, $sourceName, $indexDuringImport = false)
...
+    {
+      $this->log('Updating Elasticsearch actor relation data...');
+
+      foreach ($import->status['actorIds'] as $actorId)
+      {
+        $actor = QubitActor::getById($actorId);
+        arUpdateEsActorRelationsJob::updateActorRelationships($actor);
+        Qubit::clearClassCaches();
+      }
+    }
```
jraddaoui, line 199: Hi @mcantelon, why is this not performed from QubitActor's `save()` method when the actors are created?

mcantelon: Hi @jraddaoui ... With the authority record relation import we're just importing relations between existing actors and not modifying the actors themselves. We could though, in the above code, simply re-save the actors instead of explicitly calling ` arUpdateEsActorRelationsJob::updateActorRelationships` to get the same outcome but I think that might end up doing other stuff as well (updating related information objects, etc.) and incurring a performance hit.

jraddaoui: Oh! I see, actually what I mentioned in https://github.com/artefactual/atom/pull/1074#pullrequestreview-408761222 and it looks like it wasn't a good idea to disregard it, sorry about that.

jraddaoui (approved):
Thanks @mcantelon!

lib/task/import/csvAuthorityRecordRelationImportTask.class.php
mcantelon: No worries... I'd meant to love at that too and ended up forgetting about it too!

mcantelon:
Thanks @jraddaoui !

PR #1102, opened 2020-05-26
Normalize hierarchy deletion, refs #13239

jraddaoui:
Avoid nested set update until the end of the full tree deletion to avoid
unnecessary nested set updates.

- Normalize `deleteFullHierarchy()` in QubitObject.
- Check that it's only called from QubitInformationObject and QubitTerm.
- Run hierarchy deletion within transaction in delete description task.
- Use `deleteFullHierarchy()` on term deletion.
- Don't show all descendants on term deletion page.

### DIFF
--- a/apps/qubit/modules/term/actions/deleteAction.class.php
+++ b/apps/qubit/modules/term/actions/deleteAction.class.php
@@ -45,13 +45,7 @@ class TermDeleteAction extends sfAction
 
     if ($request->isMethod('delete'))
     {
-      foreach ($this->resource->descendants->andSelf()->orderBy('rgt') as $item)
-      {
-        if (QubitAcl::check($item, 'delete'))
-        {
-          $item->delete();
-        }
-      }
+      $this->resource->deleteFullHierarchy();
 
       if (isset($this->resource->taxonomy))
       {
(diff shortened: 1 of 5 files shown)


jraddaoui:
See https://projects.artefactual.com/issues/13239#note-5 for a bit more info.

apps/qubit/modules/term/actions/deleteAction.class.php
```diff
@@ -45,13 +45,7 @@ public function execute($request)
 
     if ($request->isMethod('delete'))
     {
-      foreach ($this->resource->descendants->andSelf()->orderBy('rgt') as $item)
-      {
-        if (QubitAcl::check($item, 'delete'))
```
sbreker, line 50: Is this ACL check still necessary? It looks like it has been removed.

jraddaoui: Good point @sbreker! Sorry, I forgot to add a note about this ...

The ACL permissions for terms are set at the taxonomy level so, if the user has permission to delete a term, it will also have it to delete its descendants. But, now that you say it, we didn't have that check for IOs either, where the permissions can be set at the IO level. However, I'm not sure if we want to add it as it would be a rare case and it will seriously affect the performance of this hierarchy deletions.

At least this is something @fiver-watson should know, maybe we want to add a note in the documentation if we don't have it already or address it in a different issue.

sbreker (approved):
Looks great @jraddaoui! 👍

PR #1103, opened 2020-05-26
CSV actor relation import updating, refs #13280

mcantelon:
Cleaned up the CSV authority record relationship import code and added
updating functionality.

### DIFF
--- a/lib/task/import/csvAuthorityRecordRelationImportTask.class.php
+++ b/lib/task/import/csvAuthorityRecordRelationImportTask.class.php
@@ -24,8 +24,10 @@
  * @subpackage task
  * @author     Mike Cantelon <mike@artefactual.com>
  */
-class csvAuthorityRecordRelationImportTask extends arBaseTask
+class csvAuthorityRecordRelationImportTask extends csvImportBaseTask
 {
+  private $import;
+
   /**
    * @see sfTask
    */
@@ -47,6 +49,9 @@ class csvAuthorityRecordRelationImportTask extends arBaseTask
         sfCommandOption::PARAMETER_OPTIONAL, 'Source name to use when inserting keymap entries.'),
       new sfCommandOption('index', null,
         sfCommandOption::PARAMETER_NONE, 'Index for search during import.'),
+      new sfCommandOption('update', null,
+        sfCommandOption::PARAMETER_REQUIRED, 'Attempt to update if relation has already been imported. Valid option values are "match-and-update" & "delete-and-replace".'
+      ),
     ]);
 
     $this->namespace = 'csv';
@@ -64,15 +69,17 @@ EOF;
   {
     parent::execute($arguments, $options);
 
+    $this->validateUpdateOptions($options);
+
     $this->log('Importing relations...');
 
     $sourceName = $options['source-name'] ?: basename($arguments['filename']);
-    $this->import($arguments['filename'], $sourceName, $options['index']);
+    $this->import($arguments['filename'], $sourceName, $options['index'], $options['update']);
 
     $this->log('Done.');
   }
 
-  private function import($filepath, $sourceName, $indexDuringImport = false)
+  private function import($filepath, $sourceName, $indexDuringImport = false, $updateMode = false)
   {
     if (false === $fh = fopen($filepath, 'rb'))
     {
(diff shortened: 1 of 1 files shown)


lib/task/import/csvAuthorityRecordRelationImportTask.class.php
```diff
@@ -218,7 +232,96 @@ private function relationshipExists($sourceActorId, $targetActorId, $relationTyp
...
-    return QubitPdo::fetchOne($sql, $params) !== false
-      || QubitPdo::fetchOne($sql, $paramsVariant) !== false;
+    if ($results = QubitPdo::fetchAll($sql, $params, ['fetchMode' => PDO::FETCH_ASSOC]))
+    {
+      return $results[0]['id'];
+    }
+
+    $results = QubitPdo::fetchAll($sql, $paramsVariant, ['fetchMode' => PDO::FETCH_ASSOC]);
+
+    return count($results) ? $results[0]['id'] : false;
```
djjuhasz: I think you could use [QubitPdo::fetchColumn()](https://github.com/artefactual/atom/blob/qa/2.6.x/lib/QubitPdo.class.php#L62) to return the first row, first column (should be "id"?) value, or false if there are no results.

mcantelon: Ah good call... I always forget that exists.

mcantelon:
@djjuhasz I've updated the PR. Thanks for looking at it!

lib/task/import/csvAuthorityRecordRelationImportTask.class.php
```diff
@@ -218,7 +232,96 @@ private function relationshipExists($sourceActorId, $targetActorId, $relationTyp
...
-    return QubitPdo::fetchOne($sql, $params) !== false
-      || QubitPdo::fetchOne($sql, $paramsVariant) !== false;
+    if ($relationId = QubitPdo::fetchColumn($sql, $params))
+    {
+      return $relationId;
+    }
+
+    $relationId = QubitPdo::fetchColumn($sql, $paramsVariant);
+
+    return !empty($relationId) ? $relationId : false;
```
djjuhasz: I think `return $relationId;` is equivalent to `!empty($relationId) ? $relationId : false;` here, since Qubit::fetchOne() will return false if no results are found.

lib/task/import/csvAuthorityRecordRelationImportTask.class.php
```diff
@@ -218,7 +232,96 @@ private function relationshipExists($sourceActorId, $targetActorId, $relationTyp
...
+    }
+  }
+
+  /**
+   * Note actor IDs whose relationship data has possibly changed
+   *
+   * @param int $sourceActorId  ID of an actor
+   * @param int $targetActorId  ID of an actor
+   */
+  private function noteActorIds($sourceActorId, $targetActorId)
```
djjuhasz: This would be more concise if the method took an array of actor ids as an argument (e.g. `noteActorIds(array $actorIds)`) then iterated through the array. The logic is the same for source or target ids, so you could just pass them both as array members.

Nitpick: When I saw "noteActorIds" I thought this was related to the 'note' table. 😆 Maybe a name like `addUpdatedActorIds()` or `addUpdatedIds()` would avoid potential confusion?

djjuhasz (changes requested):
Hi @mcantelon, I added another few suggestions but looking good overall. :)

mcantelon:
@djjuhasz I've made those tweaks... thanks!

lib/task/import/csvAuthorityRecordRelationImportTask.class.php
djjuhasz: Hi @mcantelon, sorry if I wasn't clear with my first comment but I think you can just use `return $relationId;` for line 242.  QubitPdo::fetchColumn() returns false if their are no rows in the resultset, so there's no need to check `!empty()`, $relationId will either have value, or be false.

djjuhasz: Oh, nevermind. You fixed this in the next commit. 😊

djjuhasz (approved):
Looks good!

PR #1104, opened 2020-05-27
Fix non-use of UI label settings, refs #13332

mcantelon:
On the taxonomy index page select terms have extra columns added to the
results to show associated descriptions and authority records. The
column names, however, couldn't be changed by label settings. Fixed so
now they can be.


djjuhasz (approved):
Looks good 👍

PR #1105, opened 2020-05-28
Fix actor subject/place term links, refs #13275

mcantelon:
Changed subject and place term links on authority record view pages so
they link to the term's related authorities page rather than the term's
view page.


djjuhasz (approved):
Looks good to me 👍

PR #1106, opened 2020-06-01
Fix actor relation type search, refs #13326

mcantelon:
Changed how authority record search by relation type works so only the
relation type directly related to an authority record is returned.

### DIFF
--- a/plugins/arElasticSearchPlugin/lib/model/arElasticSearchActorPdo.class.php
+++ b/plugins/arElasticSearchPlugin/lib/model/arElasticSearchActorPdo.class.php
@@ -36,7 +36,8 @@ class arElasticSearchActorPdo
   protected static
     $conn,
     $lookups,
-    $statements;
+    $statements,
+    $converseTermIds;
 
   /**
    * METHODS
@@ -264,6 +265,7 @@ class arElasticSearchActorPdo
 
     // Related objects
     $serialized['actorRelations'] = self::serializeObjectRelations($this->id);
+    $serialized['actorDirectRelationTypes'] = self::serializeObjectDirectRelationTypes($this->id, $serialized['actorRelations']);
 
     // Places
     if (isset($relatedTerms[QubitTaxonomy::PLACE_ID]))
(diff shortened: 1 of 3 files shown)


plugins/arElasticSearchPlugin/lib/model/arElasticSearchActorPdo.class.php
```diff
@@ -352,6 +354,59 @@ public function serializeAltNames()
...
+          $converseTerm = $relationTerm->getConverseActorRelationTerm();
+
+          if ($converseTerm !== null)
+          {
+            $converseTermId = $converseTerm->id;
+          }
+
+          // Cache result
+          self::$converseTermIds[$typeId] = $converseTermId;
+        };
```
sbreker: stray ';'?

mcantelon: Good catch!

mcantelon:
Thanks @sbreker ... I've updated the PR.

apps/qubit/modules/term/actions/navigateRelatedComponent.class.php
```diff
@@ -54,6 +54,7 @@ static function getEsDocsRelatedToTerm($relatedModelClass, $term, $options = [])
 
     // Allow for options to override default behavior
     $search = !empty($options['search']) ? $options['search'] : new arElasticSearchPluginQuery();
+    $search->query->setSize(10000);
```
jraddaoui: Is this really needed only to get the count?

plugins/arElasticSearchPlugin/lib/model/arElasticSearchActorPdo.class.php
```diff
@@ -352,6 +354,59 @@ public function serializeAltNames()
     return $serialized;
   }
 
+  public static function serializeObjectDirectRelationTypes($actorId, $relationData)
```
jraddaoui: Why public and static?

plugins/arElasticSearchPlugin/lib/model/arElasticSearchActorPdo.class.php
```diff
@@ -352,6 +354,59 @@ public function serializeAltNames()
...
+          // Look up converse term, if any
+          $converseTermId = $typeId;
+
+          $relationTerm = QubitTerm::getById($typeId);
+          $converseTerm = $relationTerm->getConverseActorRelationTerm();
+
+          if ($converseTerm !== null)
+          {
+            $converseTermId = $converseTerm->id;
+          }
```
jraddaoui: It's great that you cached them but I think we should use raw SQL in here, even if they are a few it should be a pretty easy query.

jraddaoui (changes requested):

jraddaoui:
Thanks @mcantelon!

Just a couple of questions and a suggestion, but feel free to disregard.

PS. Somehow this wasn't included in the initial review.

apps/qubit/modules/term/actions/navigateRelatedComponent.class.php
mcantelon: Removed!

plugins/arElasticSearchPlugin/lib/model/arElasticSearchActorPdo.class.php
mcantelon: Good point. Made it "protected".

plugins/arElasticSearchPlugin/lib/model/arElasticSearchActorPdo.class.php
mcantelon: Good call.

mcantelon:
Thanks @radda ! I've updated the PR.

jraddaoui (approved):
Thanks @mcantelon!

PR #1110, opened 2020-06-03
Do ACL check sooner in OAI GetRecord, refs #12387

jraddaoui:
Also, do not display drafts for consistency with other verbs and remove
redundant draft filter from IO's getUpdatedRecords method.


sbreker (approved):
Looks good @jraddaoui! 👍

PR #1111, opened 2020-06-03
Fix actor relation indexing check, refs #13326

jraddaoui:


mcantelon (approved):
Thanks @radda ! Good catch!

PR #1117, opened 2020-06-11
Weblate plugin string export, refs #13344

sbreker:
This commit updates AtoM's build.xml file so that old Transifex export
features integrate with the Weblate export scripts. Updates to build.xml
include:

- Removed all Transifex references and config file
- If uncommitted changes found, ask user if they want to proceed instead
of hard stop
- Changed when the contents of .tx folder are deleted so contents can be
sent to Weblate

The build.xml targets 'i18n-push-sources', 'i18n-push-translations', and
'i18n-pull-translations' can now be used with Weblate as they were with
Transifex. The artefactual-labs:weblate-migration-scripts use these
targets to automate moving the XLIFF files to Weblate - using these
specific targets ensures that the fixture and plugin strings are
included.

Also modified regex in i18nUpdateFixturesTask.class.php so ca@valencia
is correctly parsed.

There is a second commit that will ensure that HTML entities that are not compatible with XML are encoded before consolidating the XLIFF files in AtoM's .tx folder.

### DIFF
--- a/.tx/config
+++ b/.tx/config
@@ -1,8 +1,0 @@
-[main]
-host = https://www.transifex.com
-
-[atom.ui-atom-25x]
-file_filter = .tx/i18n/<lang>/messages.xml
-source_lang = en
-source_file = .tx/i18n/en/messages.xml
-type = XLIFF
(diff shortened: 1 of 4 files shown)


mcantelon (approved):
Nice!

PR #1118, opened 2020-06-11
Legacy DB upgrade fixes, refs #13350

jraddaoui:
See [RM#13350](https://projects.artefactual.com/issues/13350) and commit messages for more information.

### DIFF
--- a/lib/task/migrate/arUpgrader120.class.php
+++ b/lib/task/migrate/arUpgrader120.class.php
@@ -229,7 +229,7 @@ class arUpgrader120
         {
           foreach ($cols as $col)
           {
-            $sql = sprintf('ALTER TABLE %s MODIFY `%s` VARCHAR(1024);', $tablename, $col);
+            $sql = sprintf('ALTER TABLE `%s` MODIFY `%s` VARCHAR(1024);', $tablename, $col);
             QubitPdo::modify($sql);
           }
         }
@@ -244,7 +244,7 @@ class arUpgrader120
         {
           foreach ($cols as $col)
           {
-            $sql = sprintf('ALTER TABLE %s MODIFY %s VARCHAR(1024) NOT NULL;', $tablename, $col);
+            $sql = sprintf('ALTER TABLE `%s` MODIFY `%s` VARCHAR(1024) NOT NULL;', $tablename, $col);
             QubitPdo::modify($sql);
           }
         }
--- a/lib/task/migrate/migrations/arMigration0098.class.php
+++ b/lib/task/migrate/migrations/arMigration0098.class.php
@@ -36,47 +36,21 @@ class arMigration0098
    */
   public function up($configuration)
   {
-    // Remove existing constraint and index
-    $sql = "ALTER TABLE `event` DROP FOREIGN KEY `event_FK_3`";
-    QubitPdo::modify($sql);
+    $defaultCulture = sfConfig::get('sf_default_culture', 'en');
 
-    $sql = "DROP INDEX `event_FI_3` ON `event`";
-    QubitPdo::modify($sql);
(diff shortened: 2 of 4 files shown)


lib/task/migrate/arUpgrader120.class.php
```diff
@@ -244,7 +244,7 @@ public function up($version, $configuration, $options)
         {
           foreach ($cols as $col)
           {
-            $sql = sprintf('ALTER TABLE %s MODIFY %s VARCHAR(1024) NOT NULL;', $tablename, $col);
+            $sql = sprintf('ALTER TABLE `%s` MODIFY %s VARCHAR(1024) NOT NULL;', $tablename, $col);
```
djjuhasz: Nitpick: I'd add backticks around the column name here too, for consistency

lib/task/migrate/migrations/arMigration0098.class.php
```diff
@@ -36,47 +36,21 @@ class arMigration0098
...
+    // - Ignore nested set values. At the point we're fixing this issue,
+    //   the nested set will be removed from the actor table in arMigration0181.
+    $sql = "INSERT INTO object (class_name, created_at, updated_at, id, serial_number)
+            VALUES ('QubitRepository', now(), now(), ?, 0);";
+    QubitPdo::modify($sql, array(QubitRepository::ROOT_ID));
 
-    // Attempt to rename old column name to new column name so newer
-    // ORM will work with it
-    try {
-      $sql = "ALTER TABLE `event` CHANGE `information_object_id` `object_id` INT(11) DEFAULT NULL";
```
djjuhasz, line 49: Why is this no longer required?

djjuhasz (commented):
@jraddaoui I don't understand why the arMigration0098 doesn't need the code to change the `event.information_object_id` column and foreign key relationship to `event.object_id` 😕

lib/task/migrate/migrations/arMigration0098.class.php
jraddaoui: This comes from a system we had to do in times of 1.4. Some of this migrations (3 or 4 only) have a `MIN_MILESTONE = 1` which indicates they are needed for a 1.x to 1.4 upgrade. To be fair, I never understood this process entirely, but you can see the migrations included in qa/1.4.x:

https://github.com/artefactual/atom/tree/qa/1.4.x/lib/task/migrate/migrations

What I noticed now is that, if you're upgrading from 1.x to 2.6, those migrations get executed alongside all the others with `MIN_MILESTONE = 2`. Which causes the modification of this column twice (in here and in arMigration0121) and it's probably the reason of having the try/catch blocks.

Since we never released 1.4, all upgrades from 1.x pass through both migrations and with this changes, apart from avoiding the duplication, we can remove the try/catch as this is an important change and it's probably better to raise the exception.

PS. In the end, I think the error mentioned in the ticket about this migration was a non issue (probably caused by some changes along the process). However, I did all the final tests with this fix and I thought it will be nice to have it for the reasons explained above.

lib/task/migrate/arUpgrader120.class.php
jraddaoui: Lets hope they don't become reserved too :P

jraddaoui:
Thanks for the quick review @djjuhasz!

lib/task/migrate/migrations/arMigration0098.class.php
djjuhasz: Ohh, interesting.  I forgot all about the 1.4/2.0 fork. 😮  Thanks for the explanation. :)

djjuhasz (approved):
LGTM! 👍

PR #1119, opened 2020-06-11
Fix add authority relations GUI, refs #13342

djjuhasz:
The ability to add authority record relationships via the GUI was broken
in commit f1eba0b41dc2fde049324681bc2b92f7958ebd12 due to the changes to
autocomplete.js.

The fix removes the autocomplete.js code that removed all onsubmit listeners
from the target form.

### DIFF
--- a/js/autocomplete.js
+++ b/js/autocomplete.js
@@ -5,34 +5,44 @@
         // Share <form/> with nested scopes
         var $form = $(this);
 
-        // Support multiple submit listeners which must all complete
-        // before form is submitted
-        var count = 0;
-        function done() {
-          // Decrement count of listeners and submit if all done
-          if (1 > --count) {
-            $form
+        // Keep a list of new releated resource forms that must be submitted
+        // before $form so we can get the URIs after they are created
+        var newRelatedResourceForms = [];
 
-              // Unbind submit listener so it doesn't preventDefault()
-              .unbind('submit')
+        function done($iframeForm) {
+          var i = newRelatedResourceForms.indexOf($iframeForm);
 
-              .submit();
+          // Remove $form from newRelatedResourceForms stack
+          if (i > -1) {
+            newRelatedResourceForms.splice(i, 1);
+          }
+
+          // If stack is empty, submit the main $form
+          if (0 == newRelatedResourceForms.length) {
+            $form.submit();
           }
         }
 
-        // Fire preSubmit events, then let done() call the actual form submit
-        $form.submit(function (event) {
-          event.preventDefault();
-
-          // Trigger all "preSubmit" listeners
-          $form.trigger('preSubmit');
-
-          // If no preSubmits fired, then unbind this submit handler, and fire a
-          // new submit event
-          if (count == 0) {
-            $form.off('submit');
-            $form.submit();
+        // Submit forms to create new related resources (e.g. subjects)
+        function submitRelatedResourceForms() {
+          if (0 == newRelatedResourceForms.length) {
+            // Return true to continue with form submit
+            return true;
           }
+
+          for (var i = 0; i < newRelatedResourceForms.length; i++) {
+            newRelatedResourceForms[i].submit();
+          }
+
+          // Prevent $form.submit() from continuing - final submit will be
+          // triggered by done() function above
+          return false;
+        }
+
+        // Submit any related resource forms and wait for response before
+        // submitting the main $form
+        $form.submit(function (event) {
+          return submitRelatedResourceForms();
         });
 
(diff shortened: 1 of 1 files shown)


js/autocomplete.js
```diff
@@ -13,26 +13,27 @@
...
           $form.trigger('preSubmit');
 
-          // If no preSubmits fired, then unbind this submit handler, and fire a
-          // new submit event
+          // If no preSubmits fired, then continue with form submit as usual
           if (count == 0) {
-            $form.off('submit');
-            $form.submit();
+            return true;  // Go ahead and submit form
           }
```
jraddaoui: Could it be a race condition while the preSubmit listeners increment the count? Should we do this synchronously? It's hard to tell, but looking at the stable/2.5.x version, it seems like we only change the main form submit after interactions are made over the autocomplete fields. I really like that you called it `preSubmit`, because reading the stable/2.5.x version is quite a puzzle.

djjuhasz: Yeah, the logic is confusing.  The basic code path is this:

1. When a new related resource (e.g. a subject access point) is added via an autocomplete dialog (e.g. the user types a new subject name into the autocomplete), add an iframe to the DOM which contains a form to add the new related resource (e.g. subject),
2. On form submit (main page), fire the preSubmit event, which submits each iframe form and increments the counter,
3. When an iframe returns a response with the new related resource (e.g. new subject) URI  copy the URI into the main form and decrement the counter,
4. If the counter is less than 1 (e.g. all iframes responses are complete), then fire the main form submission,

If you are only linking existing resources, than there should be no iframes, no preSubmit listeners, and the form submit should proceed as normal.

But, this is all complicated with the authority record relations, because the dialog.js code attaches it's own "onSubmit" listeners, and does it's own DOM manipulation before allowing the form submit (HTTP POST).

I don't know if race conditions are possible when the preSubmit listeners increment the count, but I haven't had problems so far with the form submit not happening, or with the submit firing before all of the iframes responses return - i.e. it seems to be working. :)

When you say "should we do this synchronously" do you mean firing the iframe submits in serial, or are you suggesting using promises (w/ async/await?) to serialize the callback behaviour?

jraddaoui: Yeah, that makes sense, thanks @djjuhasz! 

> When you say "should we do this synchronously" do you mean firing the iframe submits in serial, or are you suggesting using promises (w/ async/await?) to serialize the callback behaviour?

It doesn't have to be in serial, you could submit all at once. I think the issue would be to use the event trigger/listener as that happens async. If you could submit the iframes directly in here, without using the event, that will avoid the race condition. You could probably remove the use of the count going this way, so I'm not sure if this is the way to go. I think the biggest change between 2.5 and now is how the main form submit is modified.

djjuhasz: Okay, I've made the iframe submit() callers synchronous.

js/autocomplete.js
```diff
@@ -5,83 +5,102 @@
...
+          $iframe.one('load', function () {
+            // Get iframe > form > input using targetInputId, and parent form
+            var $input = $iframe.contents().find(targetInputId);
+
+            // $($input[0].form) should be faster than $input.closest("form")
+            var $iframeForm = $($input[0].form);
+
+            // Add one submit handler for each new choice. Use a named function
+            // so it can be unbound if the choice is removed
+            var submit = function () {
```
jraddaoui: It is a bit confusing for me to call the function `submit`, but it's just cosmetic and it may be only me :P

js/autocomplete.js
```diff
@@ -5,83 +5,102 @@
...
-          if (1 > --count) {
-            $form
-
-              // Unbind preSubmit listeners so they don't re-fire on submit
-              .off('preSubmit')
-
-              .submit();
+        // Keep a list of new releated resource forms that must be submitted
+        // before $form so we can get the URIs after they are created
+        newRelatedResourceForms = [];
```
jraddaoui: Should this have `var` in front?

jraddaoui (approved):
Thanks a lot @djjuhasz!

js/autocomplete.js
djjuhasz: Oh, probably 😝 😆

js/autocomplete.js
djjuhasz: Well, it is a function that's bound to the form submit event... 😆 

I'm open to changing the name. Do you find it confusing because it the same handle as the "submit" event?  Would "submitCallback" be less confusing?

jraddaoui: Heheh, yeah, I think submitCallback is better ;)

PR #1121, opened 2020-06-15
Add accession events, refs #13315

mcantelon:
This adds support for accession events, based on the Canadian Archival
Accession Information Standard (CAAIS).

* Add new AtoM object type: accession events
* Add accession event type taxonomy
* Add "Physical transfer" accession event type term
* Add method to accession model to get an accession's accession events
* Add ability to add/edit accession events via the web UI
* Add Elasticsearch document type for accession events
* Update accession search/indexing to incorporate accession event
  types, dates, agents, and notes

### DIFF
--- a/data/sql/plugins.qtAccessionPlugin.lib.model.schema.sql
+++ b/data/sql/plugins.qtAccessionPlugin.lib.model.schema.sql
@@ -78,6 +78,56 @@ CREATE TABLE `accession_i18n`
 )Engine=InnoDB;
 
 #-----------------------------------------------------------------------------
+#-- accession_event
+#-----------------------------------------------------------------------------
+
+DROP TABLE IF EXISTS `accession_event`;
+
+
+CREATE TABLE `accession_event`
+(
+	`id` INTEGER  NOT NULL,
+	`type_id` INTEGER,
+	`accession_id` INTEGER,
+	`date` DATE,
+	`source_culture` VARCHAR(16)  NOT NULL,
+	PRIMARY KEY (`id`),
+	CONSTRAINT `accession_event_FK_1`
+		FOREIGN KEY (`id`)
+		REFERENCES `object` (`id`)
+		ON DELETE CASCADE,
+	INDEX `accession_event_FI_2` (`type_id`),
+	CONSTRAINT `accession_event_FK_2`
+		FOREIGN KEY (`type_id`)
+		REFERENCES `term` (`id`)
+		ON DELETE SET NULL,
+	INDEX `accession_event_FI_3` (`accession_id`),
+	CONSTRAINT `accession_event_FK_3`
+		FOREIGN KEY (`accession_id`)
+		REFERENCES `accession` (`id`)
+		ON DELETE CASCADE
+)Engine=InnoDB;
+
+#-----------------------------------------------------------------------------
+#-- accession_event_i18n
+#-----------------------------------------------------------------------------
+
+DROP TABLE IF EXISTS `accession_event_i18n`;
+
+
+CREATE TABLE `accession_event_i18n`
+(
+	`agent` VARCHAR(255),
+	`id` INTEGER  NOT NULL,
+	`culture` VARCHAR(16)  NOT NULL,
+	PRIMARY KEY (`id`,`culture`),
+	CONSTRAINT `accession_event_i18n_FK_1`
+		FOREIGN KEY (`id`)
+		REFERENCES `accession_event` (`id`)
+		ON DELETE CASCADE
+)Engine=InnoDB;
+
+#-----------------------------------------------------------------------------
 #-- deaccession
 #-----------------------------------------------------------------------------
 
--- a/plugins/arElasticSearchPlugin/config/mapping.yml
+++ b/plugins/arElasticSearchPlugin/config/mapping.yml
@@ -78,6 +78,22 @@ mapping:
       subject_id: { type: integer, include_in_all: false }
       type_id: { type: integer, include_in_all: false }
 
+  accession_event:
+    _attributes:
+      i18n: true
+      nested_only: true
+    _partial_foreign_types:
+      type:
+        _i18nFields: [name]
+        dynamic: strict
+      notes:
+        _i18nFields: [content]
+        dynamic: strict
+    dynamic: strict
+    properties:
+      date: { type: date, include_in_all: false }
+      date_string: { type: keyword }
+
   donor:
     _attributes:
       i18nExtra: [actor]
@@ -390,6 +406,7 @@ mapping:
       rawFields:  [title]
       sortFields: [title]
     _foreign_types:
+      accession_events: accession_event
       alternative_identifiers: other_name
       donors: donor
       creators: actor
--- a/plugins/arElasticSearchPlugin/lib/model/arElasticSearchAccession.class.php
+++ b/plugins/arElasticSearchPlugin/lib/model/arElasticSearchAccession.class.php
@@ -25,8 +25,8 @@ class arElasticSearchAccession extends arElasticSearchModelBase
   {
     $accessionIds = QubitPdo::fetchAll(
       'SELECT id FROM '.QubitAccession::TABLE_NAME,
-      array(),
-      array('fetchMode' => PDO::FETCH_COLUMN)
+      [],
+      ['fetchMode' => PDO::FETCH_COLUMN]
     );
 
     $this->count = count($accessionIds);
@@ -36,7 +36,7 @@ class arElasticSearchAccession extends arElasticSearchModelBase
 
   public function populate()
   {
-    $errors = array();
+    $errors = [];
 
     foreach ($this->load() as $key => $id)
     {
@@ -74,7 +74,7 @@ class arElasticSearchAccession extends arElasticSearchModelBase
       self::$statements['accession'] = self::$conn->prepare($sql);
     }
 
-    self::$statements['accession']->execute(array(':id' => $id));
+    self::$statements['accession']->execute([':id' => $id]);
     $data = self::$statements['accession']->fetch(PDO::FETCH_ASSOC);
 
     if (false === $data)
@@ -82,7 +82,7 @@ class arElasticSearchAccession extends arElasticSearchModelBase
       throw new sfException("Couldn't find accession (id: $id)");
     }
 
-    $serialized = array();
+    $serialized = [];
     $serialized['id'] = $id;
     $serialized['slug'] = $data['slug'];
     $serialized['identifier'] = $data['identifier'];
@@ -90,28 +90,30 @@ class arElasticSearchAccession extends arElasticSearchModelBase
     $serialized['createdAt'] = arElasticSearchPluginUtil::convertDate($data['created_at']);
     $serialized['updatedAt'] = arElasticSearchPluginUtil::convertDate($data['updated_at']);
     $serialized['sourceCulture'] = $data['source_culture'];
-    $serialized['i18n'] = self::serializeI18ns($id, array('QubitAccession'));
+    $serialized['i18n'] = self::serializeI18ns($id, ['QubitAccession']);
 
     $sql = "SELECT o.id, o.source_culture, o.type_id FROM ".QubitOtherName::TABLE_NAME." o \r
               INNER JOIN ".QubitTerm::TABLE_NAME." t ON o.type_id=t.id \r
               WHERE o.object_id = ? AND t.taxonomy_id= ?";
-    $params = array($id, QubitTaxonomy::ACCESSION_ALTERNATIVE_IDENTIFIER_TYPE_ID);
+    $params = [$id, QubitTaxonomy::ACCESSION_ALTERNATIVE_IDENTIFIER_TYPE_ID];
 
     foreach (QubitPdo::fetchAll($sql, $params) as $item)
     {
       $serialized['alternativeIdentifiers'][] = arElasticSearchOtherName::serialize($item);
     }
 
-    foreach (QubitRelation::getRelationsBySubjectId($id, array('typeId' => QubitTerm::DONOR_ID)) as $item)
+    foreach (QubitRelation::getRelationsBySubjectId($id, ['typeId' => QubitTerm::DONOR_ID]) as $item)
     {
       $serialized['donors'][] = arElasticSearchDonor::serialize($item->object);
     }
 
-    foreach (QubitRelation::getRelationsByObjectId($id, array('typeId' => QubitTerm::CREATION_ID)) as $item)
+    foreach (QubitRelation::getRelationsByObjectId($id, ['typeId' => QubitTerm::CREATION_ID]) as $item)
     {
       $node = new arElasticSearchActorPdo($item->subject->id);
       $serialized['creators'][] = $node->serialize();
     }
+
+    $serialized['accessionEvents'] = self::getAccessionEvents($id);
 
     return $serialized;
   }
@@ -123,5 +125,61 @@ class arElasticSearchAccession extends arElasticSearchModelBase
     QubitSearch::getInstance()->addDocument($data, 'QubitAccession');
 
     return true;
+  }
+
+  public static function getAccessionEvents($accessionId)
+  {
+    if (!isset(self::$statements['event']))
+    {
+      $sql  = 'SELECT
+                 event.id,
+                 event.type_id,
+                 event.date,
+                 note.id AS note_id
+               FROM '.QubitAccessionEvent::TABLE_NAME.' event
+               INNER JOIN '.QubitNote::TABLE_NAME.' note
+               ON (event.id=note.object_id AND note.type_id=?)
+               WHERE event.accession_id = ?';
+
+      self::$statements['event'] = self::$conn->prepare($sql);
+    }
+
+    self::$statements['event']->execute([QubitTerm::ACCESSION_EVENT_NOTE_ID, $accessionId]);
+
+    $events = [];
+    foreach (self::$statements['event']->fetchAll(PDO::FETCH_ASSOC) as $item)
+    {
+      $event = ['id' => $item['id']];
+
+      $event['date'] = arElasticSearchPluginUtil::normalizeDateWithoutMonthOrDay($item['date']);
+      $event['dateString'] = Qubit::renderDate($item['date']);
+
+      // Serialize type term
+      if ($item['type_id'])
+      {
+        $event['type'] = ['i18n' => arElasticSearchModelBase::serializeI18ns(
+          $item['type_id'],
+          array('QubitTerm'),
+          array('fields' => array('name'))
+        )];
+      }
+
+      // Serialize note
+      if ($item['note_id'])
+      {
+        $event['notes'] = ['i18n' => arElasticSearchModelBase::serializeI18ns(
+          $item['note_id'],
+          array('QubitNote'),
+          array('fields' => array('content'))
+        )];
+      }
+
+      // Serialize accession event i18n data
+      $event['i18n'] = arElasticSearchModelBase::serializeI18ns($item['id'], ['QubitAccessionEvent']);
+
+      $events[] = $event;
+    }
+
+    return $events;
   }
 }
--- a/plugins/arElasticSearchPlugin/lib/model/arElasticSearchModelBase.class.php
+++ b/plugins/arElasticSearchPlugin/lib/model/arElasticSearchModelBase.class.php
@@ -126,8 +126,8 @@ abstract class arElasticSearchModelBase
       }
     }
 
-    // Remove duplicated cultures
-    $i18ns['languages'] = array_unique($i18ns['languages']);
+    // Remove duplicated cultures from language values
+    $i18ns['languages'] = array_unique(array_values($i18ns['languages']));
 
     return $i18ns;
   }
--- a/plugins/qtAccessionPlugin/lib/model/QubitAccessionEvent.php
+++ b/plugins/qtAccessionPlugin/lib/model/QubitAccessionEvent.php
@@ -1,0 +1,31 @@
+<?php
+
+/*
+ * This file is part of the Access to Memory (AtoM) software.
+ *
+ * Access to Memory (AtoM) is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * Access to Memory (AtoM) is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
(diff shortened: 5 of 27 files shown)


data/sql/plugins.qtAccessionPlugin.lib.model.schema.sql
```diff
@@ -77,6 +77,58 @@ CREATE TABLE `accession_i18n`
...
 )Engine=InnoDB;
 
+#-----------------------------------------------------------------------------
+#-- accession_event
+#-----------------------------------------------------------------------------
+
+DROP TABLE IF EXISTS `accession_event`;
+
+
+CREATE TABLE `accession_event`
```
djjuhasz, line 87: Is there a reason the existing event & event_i18n tables couldn't be used instead of adding new tables?

data/sql/plugins.qtAccessionPlugin.lib.model.schema.sql
```diff
@@ -77,6 +77,58 @@ CREATE TABLE `accession_i18n`
...
+(
+	`id` INTEGER  NOT NULL,
+	`type_id` INTEGER,
+	`accession_id` INTEGER,
+	`date` DATE,
+	`created_at` DATETIME  NOT NULL,
```
djjuhasz: The object table already contains `created_at` and `updated_at` columns, so it seems unnecessary to repeat them here.

plugins/arElasticSearchPlugin/lib/model/arElasticSearchModelBase.class.php
```diff
@@ -126,8 +126,8 @@ public static function serializeI18ns($id, array $classes, $options = array())
       }
     }
 
-    // Remove duplicated cultures
-    $i18ns['languages'] = array_unique($i18ns['languages']);
+    // Remove duplicated cultures from language values
+    $i18ns['languages'] = array_unique(array_values($i18ns['languages']));
```
djjuhasz, line 130: Is `array_values()` necessary here? `array_unique()` only compares array values anyway.

plugins/qtAccessionPlugin/lib/model/QubitAccession.php
```diff
@@ -226,4 +226,22 @@ public function getAlternativeIdentifiers()
 
     return $otherNames;
   }
+
+  /**
+   * Get events
+   */
+  public function getEvents()
+  {
+    $events = [];
```
djjuhasz: I think you can just call `return $this->_get('accessionEvents')` here - it's implemented as part of the Propel model in the `BaseAccession::_get()` method.

plugins/qtAccessionPlugin/lib/model/om/BaseAccession.php
```diff
@@ -159,6 +164,23 @@ public function __get($name)
       return $this->refFkValues['accessionI18ns'];
     }
 
+    if ('accessionEvents' == $name)
+    {
+      if (!isset($this->refFkValues['accessionEvents']))
```
djjuhasz, line 169: Here's the `__get('accessionEvents')` auto-generated code.

plugins/qtAccessionPlugin/modules/accession/actions/eventsComponent.class.php
```diff
@@ -0,0 +1,147 @@
...
+        }
+
+        $event->accessionId = $this->resource->id;
+        $event->typeId = $item['eventType'];
+        $event->date = empty($item['date']) ? null : $item['date'];
+        $event->agent = $item['agent'];
+        $event->save();
+
+        // Store note as property
+        $property = $event->getPropertyByName('accessionEventNote');
```
djjuhasz: Why not use the `note` table here instead of the `property` table?

plugins/qtAccessionPlugin/modules/accession/actions/eventsComponent.class.php
```diff
@@ -0,0 +1,147 @@
...
+        $property->objectId = $event->id;
+        $property->name = 'accessionEventNote';
+        $property->value = $item['note'];
+        $property->save();
+      }
+    }
+
+    // Delete the old alternative identifiers if they don't appear in the table (removed by multiRow.js)
```
djjuhasz: Nitpick: This comment needs to be updated ;)

plugins/qtAccessionPlugin/modules/accession/templates/_events.php
```diff
@@ -0,0 +1,89 @@
...
+  <tfoot>
+    <tr>
+      <td colspan="4"><a href="#" class="multiRowAddButton"><?php echo __('Add new') ?></a></td>
+    </tr>
+  </tfoot>
+
+</table>
+
+<div class="description">
+  <?php echo __('<strong>Type:</strong> Enter a name for the alternative identifier field that indicates its purpose and usage.<br/><strong>Agent:</strong> Enter the agent associated with the event.') ?>
```
djjuhasz: This needs to be updated.

djjuhasz (changes requested):
A couple of questions and comments for you @mcantelon.

data/sql/plugins.qtAccessionPlugin.lib.model.schema.sql
mcantelon: The event model represents a date range, rather than just a single date, and agents aren't necessarily AtoM actors or users. I think Steve also came to the conclusion that it wouldn't be good to try to use the existing event table.

data/sql/plugins.qtAccessionPlugin.lib.model.schema.sql
mcantelon: Good call. I'll nix those.

plugins/arElasticSearchPlugin/lib/model/arElasticSearchModelBase.class.php
mcantelon: There's a weird issue I ran into that necessitated this...

The TLDR is that not having that would result, in some cases, in data like this for languages:

`array(0 => 'en', 2 => 'fr')`

rather than

`array(0 => 'en', 1 => 'fr')`

And Elastica would index the former differently than the latter and an ES parse error would result.

plugins/qtAccessionPlugin/lib/model/QubitAccession.php
mcantelon: Nice!

plugins/qtAccessionPlugin/modules/accession/actions/eventsComponent.class.php
mcantelon: I think the CAAIS spec allows for just one note (http://archivescanada.ca/CWG_AccessionStandard#5.1_e) so I was using property to keep it simple, but I could do it with note too. I guess using note would be more future proof in case they update the spec.

data/sql/plugins.qtAccessionPlugin.lib.model.schema.sql
djjuhasz: Hmm, I do wonder if it's a good idea to make the agent column a varchar rather than a related entity like authority records, donors, and rights holders. 🤔 There are pros and cons with either design of course, but I'm not clear why agent doesn't follow the established pattern.

Anyway, I don't consider it a blocker if you and Steve decided a new table is necessary, and agent should be varchar column.

plugins/arElasticSearchPlugin/lib/model/arElasticSearchModelBase.class.php
djjuhasz: Ahh, okay.  Thanks for the explanation. :)

plugins/arElasticSearchPlugin/config/mapping.yml
```diff
@@ -74,6 +74,16 @@ mapping:
       subject_id: { type: integer, include_in_all: false }
       type_id: { type: integer, include_in_all: false }
 
+  accession_event:
+    _attributes:
+      i18n: true
+      i18nExtra: [property, term]
+      nested_only: true
+    dynamic: true
```
jraddaoui: Why dynamic? I think it's better to have full control over the mapping fields.

PS. I don't know why we have it on the `donors` too but all others are set to strict.

plugins/arElasticSearchPlugin/config/mapping.yml
```diff
@@ -74,6 +74,16 @@ mapping:
       subject_id: { type: integer, include_in_all: false }
       type_id: { type: integer, include_in_all: false }
 
+  accession_event:
+    _attributes:
+      i18n: true
+      i18nExtra: [property, term]
```
jraddaoui: Not a big deal but this extra i18n fields are normally used to get the values from inheritance tables, like the actor row for a repository. In the end it works but it's a bit confusing to mix the i18n fields in this case, where they are different entities. It also disables the option to add different configuration to those fields, for example, you may need to make the property value a nested field to search over two fields from the same property (similar to the actor relations issue we discussed a while ago).

In the end, it's probably not an issue in this case, so maybe it's enough to clarify the note in the initial comments to reflect this kind of behavior:

> - i18nExtra: to append i18n fields from other tables (used for inheritance and related resources)

plugins/arElasticSearchPlugin/lib/model/arElasticSearchModelBase.class.php
jraddaoui: That's my bad, thanks @mcantelon!

plugins/arElasticSearchPlugin/lib/model/arElasticSearchAccession.class.php
```diff
@@ -124,4 +126,52 @@ public static function update($object)
...
+    self::$statements['event']->execute(array($accessionId));
+
+    $events = array();
+    foreach (self::$statements['event']->fetchAll(PDO::FETCH_ASSOC) as $item)
+    {
+      $event = array(
+        'id' => $item['id'],
+        'typeId' => $item['type_id'],
+        'date' => $item['date'],
+        'dateString' => $item['date']);
```
jraddaoui: I think we need to normalize this dates, like we do for events ...

```php
$serialized['startDate'] = arElasticSearchPluginUtil::normalizeDateWithoutMonthOrDay($event->start_date);
$serialized['startDateString'] = Qubit::renderDate($event->start_date);
```

plugins/qtAccessionPlugin/lib/model/QubitAccessionEvent.php
```diff
@@ -0,0 +1,37 @@
...
+ *
+ * 
+ *
+ * You should add additional methods to this class to meet the
+ * application requirements.  This class will only be generated as
+ * long as it does not already exist in the output directory.
+ *
+ * @package    plugins.qtAccessionPlugin.lib.model
+ */
+class QubitAccessionEvent extends BaseAccessionEvent {
```
jraddaoui: New line for the bracket and no empty line before the first function.

plugins/qtAccessionPlugin/lib/model/QubitAccessionEvent.php
```diff
@@ -0,0 +1,37 @@
...
+      return $note;
+    }
+    else
+    {
+      $note = new QubitNote;
+      $note->objectId = $this->id;
+      $note->typeId = QubitTerm::ACCESSION_EVENT_NOTE_ID;
+
+      return $note;
+    }
```
jraddaoui: Returning a placeholder note is not something we usually do. Not a big deal but if you need it, you could remove the else block since you're returning in the first condition.

```php
if ($note = QubitNote::getOne($criteria))
{
  return $note;
}

$note = new QubitNote;
$note->objectId = $this->id;
$note->typeId = QubitTerm::ACCESSION_EVENT_NOTE_ID;

return $note;
```

plugins/qtAccessionPlugin/lib/model/QubitAccessionEvent.php
```diff
@@ -0,0 +1,37 @@
+<?php
+
+
```
jraddaoui: One extra empty line.

jraddaoui: Also, missing the ` * This file is part of the Access to Memory (AtoM) software...` comment.

plugins/qtAccessionPlugin/lib/model/QubitAccessionEvent.php
```diff
@@ -0,0 +1,37 @@
...
+/**
+ * Skeleton subclass for representing a row from the 'accession_event' table.
+ *
+ * 
+ *
```
jraddaoui, line 10: Two extra empty lines.

plugins/qtAccessionPlugin/lib/model/QubitAccessionEventI18n.php
```diff
@@ -0,0 +1,17 @@
+<?php
+
+
+/**
+ * Skeleton subclass for representing a row from the 'accession_event_i18n' table.
+ *
+ * 
+ *
```
jraddaoui: Same issues than in QubitAccessionEvent.php.

plugins/qtAccessionPlugin/lib/model/QubitAccessionEventI18n.php
```diff
@@ -0,0 +1,17 @@
...
+ * You should add additional methods to this class to meet the
+ * application requirements.  This class will only be generated as
+ * long as it does not already exist in the output directory.
+ *
+ * @package    plugins.qtAccessionPlugin.lib.model
+ */
+class QubitAccessionEventI18n extends BaseAccessionEventI18n {
```
jraddaoui: New line for the bracket. No empty line inside the empty class?

plugins/qtAccessionPlugin/lib/model/QubitAccessionEventI18n.php
jraddaoui: I think you copied the comments from the auto-generated table map class. Check the other custom model classes, as we add the other comment at the beginning.

plugins/arElasticSearchPlugin/config/mapping.yml
jraddaoui: In the end, it looks like you only need the term name and the property value. I think `_partial_foreign_types` is a better fit than `i18nExtra` for this fields.

jraddaoui (changes requested):
Great work @mcantelon!

lib/task/migrate/migrations/arMigration0183.class.php
```diff
@@ -0,0 +1,65 @@
...
+    $term->id = QubitTerm::ACCESSION_EVENT_NOTE_ID;
+    $term->parentId = QubitTerm::ROOT_ID;
+    $term->taxonomyId = QubitTaxonomy::NOTE_TYPE_ID;
+    $term->sourceCulture = 'en';
+    $term->setName('Accession event note', array('culture' => 'en'));
+    $term->save();
+
+    return true;
+  }
+}
```
jraddaoui: I just thought, we're missing the new tables creation on the migration process!

plugins/arElasticSearchPlugin/config/mapping.yml
mcantelon: Nice... I've reworked it with `_partial_foreign_types` and it's definitely cleaner.

plugins/arElasticSearchPlugin/config/mapping.yml
mcantelon: Fixed!

mcantelon:
Hi @djjuhasz and @jraddaoui ... I've reworked things incorporating your feedback. Let me know if you spot anything else that might use tweaking!

plugins/arElasticSearchPlugin/lib/model/arElasticSearchAccession.class.php
```diff
@@ -135,39 +135,43 @@ public static function getAccessionEvents($accessionId)
...
+      $event['dateString'] = Qubit::renderDate($item['date']);
 
-      // Serialize event type term's i18n data
-      $i18nFields = arElasticSearchModelBase::serializeI18ns($item['type_id'], array('QubitTerm'));
+      // Serialize type term
+      if ($item['type_id'])
+      {
+        $node = new arElasticSearchTermPdo($item['type_id']);
+        $serialized = $node->serialize();
+        $event['type'] = ['i18n' => $serialized['i18n']];
```
jraddaoui: I think this could add other fields apart from the name and there is no need to serialize the entire term. Now that you're using partial foreign types, you could serialize only the required field with ...

```php
$event['type'] = ['i18n' => arElasticSearchModelBase::serializeI18ns(
  $item['type_id'],
  array('QubitTerm'),
  array('fields' => array('name'))
)];
```

plugins/arElasticSearchPlugin/lib/model/arElasticSearchAccession.class.php
```diff
@@ -135,39 +135,43 @@ public static function getAccessionEvents($accessionId)
...
 
-      // Serialize note property's i18n data and merge in event type i18n data
-      if ($item['property_id'])
+      // Serialize note
+      if ($item['note_id'])
       {
-        $i18nFields = arElasticSearchModelBase::serializeI18ns($item['property_id'], array('QubitProperty'), array('merge' => $i18nFields));
+        $object = new stdClass;
+        $object->id = $item['note_id'];
+        $event['notes'] = arElasticSearchNote::serialize($object);
```
jraddaoui: Same in here.

jraddaoui (changes requested):
Thanks @mcantelon!

jraddaoui:
@mcantelon, also, don't forget about https://github.com/artefactual/atom/pull/1121#discussion_r441058290.

lib/task/migrate/migrations/arMigration0183.class.php
sbreker: In case we need to pick this feature to another branch that is < 2.6.x, this migration perhaps should also check if these terms and DB changes exist already?

jraddaoui:
@mcantelon, I took your migration number :stuck_out_tongue_closed_eyes:

Please, move it to 184, thanks!

lib/task/migrate/migrations/arMigration0183.class.php
mcantelon: Thanks guys... I've made the migration idempotent.

mcantelon:
Thanks @jraddaoui ! I've updated the PR.

lib/task/migrate/migrations/arMigration0184.class.php
```diff
@@ -0,0 +1,132 @@
...
+                REFERENCES `term` (`id`)
+                ON DELETE SET NULL,
+        INDEX `accession_event_FI_3` (`accession_id`),
+        CONSTRAINT `accession_event_FK_3`
+                FOREIGN KEY (`accession_id`)
+                REFERENCES `accession` (`id`)
+                ON DELETE CASCADE
+)Engine=InnoDB;
+
+CREATE TABLE `accession_event_i18n`
```
jraddaoui: CREATE TABLE IF NOT EXISTS

lib/task/migrate/migrations/arMigration0184.class.php
```diff
@@ -0,0 +1,132 @@
...
+
+sql;
+
+    // Add accession event type taxonomy if it doesn't exist
+    $sql = "SELECT * FROM ".QubitTaxonomyI18n::TABLE_NAME."
+              WHERE name='Accession event type'
+              AND culture='en'
+              AND id=?";
+
+    if (null == QubitPdo::fetchOne($sql, array(QubitTaxonomy::ACCESSION_EVENT_TYPE_ID)))
```
jraddaoui: Optimizations getting into your head, I know the feeling lately, hehehe. I think it would have been okay with `QubitTaxonomy::getById()`. Also, really unlikely but the name may have changed, I'd use only the id to query over the taxonomy table.

jraddaoui: And similar for the terms below ;)

plugins/qtAccessionPlugin/lib/model/map/AccessionEventI18nTableMap.php
```diff
@@ -1,18 +1,22 @@
...
- * (i.e. if it's a text column type).
+ * Access to Memory (AtoM) is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
  *
- * @package    plugins.qtAccessionPlugin.lib.model.map
+ * You should have received a copy of the GNU General Public License
+ * along with Access to Memory (AtoM).  If not, see <http://www.gnu.org/licenses/>.
  */
```
jraddaoui: This will probably be removed the next time we build the models. I think we don't add the default comment in this classes for that reason.

jraddaoui (changes requested):
Thanks @mcantelon!!

lib/task/migrate/migrations/arMigration0184.class.php
mcantelon: Yeah, you're right... `getById` makes sense for QubitTaxonomy given users can add to it. For the terms, however, wouldn't there be, without a name check, the risk that a user had added a custom term using the ID?

mcantelon:
Hi @jraddaoui ... I've updated the PR except for https://github.com/artefactual/atom/pull/1121#discussion_r442533444 which I wanted to get your take on!

lib/task/migrate/migrations/arMigration0184.class.php
jraddaoui: Oh, I didn't think about that, great point @mcantelon. Moreover, without being protected terms, the name could have changed. Should they be protected? We're getting quite edgy in here knowing that this will be partially cherry picked to a 2.5.4 version, so it's probably okay as it was.

@sbreker, if we don't protect the terms in the end, we should at least notify that the term names shouldn't be changed for the 2.6 upgrade to work properly.

jraddaoui: @mcantelon, you could add an extra condition over the `taxonomy_id` to those term queries to reduce a bit the risk.

Also, do you know if we need the same checks on arMigration0179? (@sbreker and @djjuhasz too)

https://github.com/artefactual/atom/blob/qa/2.6.x/lib/task/migrate/migrations/arMigration0179.class.php

mcantelon: Yeah, might be good to add those checks to arMigration0179. I'll make a separate PR for that.

I've made the two terms added by migration 184 protected and added taxonomy_id checks for both.

mcantelon:
Hi @jraddaoui ... I've updated the PR. Thanks for the recommendations!

fiver-watson:
Hey @mcantelon  - just a quick reminder... We've reached code freeze for new features in qa/2.6.x, so we are likely going to target this enhancement for 2.7. In which case, please remember to keep this on a branch until we release 2.6 and create a new qa/2.7.x branch? Thanks!

mcantelon:
@fiver-watson Sounds good... thanks!

mcantelon:
I'm rebasing this so I can cherry-pick it to an internal repo so QA can be done on it (given that we can't merge anything into the AtoM 2.6 branch given we're in code freeze).

PR #1122, opened 2020-06-16
Do not index IO lft value, refs #13356

jraddaoui:
It's not needed and may be outdated in some cases (CSV import with
nested set update disabled). Also, do not use it to order the ancestors.


djjuhasz (approved):
Looks good to me 👍

PR #1123, opened 2020-06-16
Nested set build task changes, refs #13357 and #13358

jraddaoui:


lib/task/propel/propelBuildNestedSetTask.class.php
```diff
@@ -65,10 +66,26 @@ public function execute($arguments = array(), $options = array())
       'menu' => 'QubitMenu'
     );
 
+    $excludeTables = array();
+
+    if (!empty($options['exclude-tables']))
+    {
+      $excludeTables = explode(',', $options['exclude-tables']);
```
djjuhasz: Suggestion: It might be worth wrapping this in a `trim()` to remove leading or trailing whitespace in the list.

djjuhasz: Oh, it's an array - so I guess it would have to use `array_map()` or similar to remove white space.

djjuhasz (approved):
Looks good!  I made one suggestion, but it's not a blocker.

lib/task/propel/propelBuildNestedSetTask.class.php
jraddaoui: Good call @djjuhasz, thanks!

PR #1124, opened 2020-06-17
Add option to cache only DC or EAD, refs #13361

tw4l:
Implemented with --format flag for cache:xml-representations task.

This option is permissive, so --format=dc or --format=ead will produce only DC or EAD XML, respectively, and omitting the flag or using an invalid value like --format=nonesuch will produce both DC and EAD XML, as before.

### DIFF
--- a/lib/QubitInformationObjectXmlCache.class.php
+++ b/lib/QubitInformationObjectXmlCache.class.php
@@ -59,25 +59,29 @@ class QubitInformationObjectXmlCache
   }
 
   /**
-   * Export information object to EAD (if top-level) and DC.
+   * Export information object to EAD (if top-level) and/or DC.
    *
    * @param object  information object instance to export
+   * @param string  format of XML ('dc' or 'ead')
    *
    * @return null
    */
-  public function export($resource)
+  public function export($resource, $format = null)
   {
     // Only cache top-level information object's EAD XML
-    if ($resource->parentId == QubitInformationObject::ROOT_ID)
+    if ($resource->parentId == QubitInformationObject::ROOT_ID && $format !== 'dc')
     {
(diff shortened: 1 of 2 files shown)


lib/task/arCacheDescriptionXmlTask.class.php
```diff
@@ -57,7 +58,7 @@ private function exportAll($options)
     $logger->log('Caching XML representations of information objects...');
 
     $cache = new QubitInformationObjectXmlCache(array('logger' => $logger));
-    $cache->exportAll(array('skip' => $options['skip'], 'limit' => $options['limit']));
+    $cache->exportAll(array('skip' => $options['skip'], 'limit' => $options['limit'], 'format' => $options['format']));
```
jraddaoui, line 61: I think the other options won't affect the `exportAll` function so it may be better to use `$options` directly. Not a big deal for sure.

lib/QubitInformationObjectXmlCache.class.php
```diff
@@ -59,25 +59,29 @@ protected function createExportDestinationDirs()
...
-   * Export information object to EAD (if top-level) and DC.
+   * Export information object to EAD (if top-level) and/or DC.
    *
    * @param object  information object instance to export
+   * @param string  format of XML ('dc', 'ead', or 'all')
    *
    * @return null
    */
-  public function export($resource)
+  public function export($resource, $format = 'all')
```
jraddaoui: I'd prefer `null` instead of `'all'`, I think it would have been easier to process. Same, not a big deal at this point.

jraddaoui (approved):
Thanks @tw4l!

This looks good to me, just a couple of comments (feel free to ignore them). I'd wait to merge until the ticket gets confirmation for the 2.6 release.

PR #1125, opened 2020-06-17
CSV import fixes, refs #13352, #13354 and #13355

jraddaoui:
Requires https://github.com/artefactual/atom/pull/1123 and https://github.com/artefactual/atom/pull/1127, the last one only if STRICT_TRANS_TABLES is enabled in the `sql_mode`.

Read commit messages for more info.

### DIFF
--- a/lib/QubitFlatfileImport.class.php
+++ b/lib/QubitFlatfileImport.class.php
@@ -1542,7 +1542,7 @@ class QubitFlatfileImport
 
     $note->content = $this->content($text);
     $note->culture = $this->columnValue('culture');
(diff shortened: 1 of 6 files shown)


mcantelon:
Looks good to me! Maybe at some point we should get rid of the indexing/nested set uping logic in `lib/QubitFlatfileImport.class.php` if it's suboptimal.

mcantelon (approved):
Looks good to me! Maybe at some point we should get rid of the indexing/nested set uping logic in `lib/QubitFlatfileImport.class.php` if it's suboptimal.

djjuhasz (approved):
Looks great @jraddaoui! :shipit:

PR #1126, opened 2020-06-17
Increase treeview items per page max, refs #13320

djjuhasz:
- Add config/app.yml variable for maximum allowed "items per page" value
  for the fullwidth treeview with a default value of 10 000
- Use the "treeview_items_per_page_max" config variable for the treeview
  settings page help text and form validation
- Use the symfony number_format() function to localize digit grouping
- If "treeview_items_per_page_max" var is not set, default to a max of 10 000


jraddaoui (approved):
Nice!

PR #1127, opened 2020-06-17
Allow null values on lft/rgt columns, refs #13331

jraddaoui:
Needed when STRICT_TRANS_TABLES is enabled in the sql_mode
to run CSV import and other processes that create IOs, terms or
menus with the nested set updates disabled.


djjuhasz (approved):
Oof, good fix @jraddaoui.  LGTM!

jraddaoui:
Thanks @djjuhasz! :)

PR #1130, opened 2020-06-18
Fix actor relation import via web UI, refs #13329

mcantelon:
* Remove --source-name option from authority record relation import
  task
* Modified QubitCsvImport class so the --source-name option isn't added
  for authority record relation imports
* Remove keymap entry creation during authority record relation import
* Remove legacyId column from authority record relation example CSV


djjuhasz (approved):
Looks good to me! :)

PR #1131, opened 2020-06-18
Fix actor rel. CSV export col. order, ref #13363

mcantelon:
Swaps the order of the subjectAuthorizedFormOfName and
objectAuthorizedFormOfName columns so things are clearer.


mcantelon:
Need to re-test.

PR #1134, opened 2020-06-19
Normalize DB constraints, refs #13221

jraddaoui:
Rename index and constraint not named during upgrades.


djjuhasz (approved):
Looks good @jraddaoui

PR #1135, opened 2020-06-23
Specify db name when looking up fks, refs #13365

djjuhasz:
- Use `QubitMigrate::updateForeignKeys()` method for updating foreign keys in arMigration0176
- Limit foreign key name lookup to the current database by specifying the "CONSTRAINT_SCHEMA" value

### DIFF
--- a/lib/task/migrate/QubitMigrate.class.php
+++ b/lib/task/migrate/QubitMigrate.class.php
@@ -930,16 +930,20 @@ class QubitMigrate
 
   public static function updateForeignKeys($foreignKeys)
   {
+    $dbname = QubitPdo::fetchColumn('select database();');
+
     foreach ($foreignKeys as $foreignKey)
     {
       // Get actual contraint name
       $sql = 'SELECT CONSTRAINT_NAME FROM INFORMATION_SCHEMA.KEY_COLUMN_USAGE ';
       $sql .= 'WHERE TABLE_NAME=:table AND COLUMN_NAME=:column ';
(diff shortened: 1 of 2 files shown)


lib/task/migrate/QubitMigrate.class.php
```diff
@@ -930,16 +930,20 @@ public static function updateIndexes($indexes)
 
   public static function updateForeignKeys($foreignKeys)
   {
+    $dbname = QubitPdo::fetchColumn('select database();');
```
jraddaoui, line 933: Maybe we can get it from the config to avoid the query on each foreign key update, not a big deal though.

lib/task/migrate/migrations/arMigration0176.class.php
```diff
@@ -83,86 +83,50 @@ public function up($configuration)
...
-        $sql,
-        $foreignKey['tableName'],
-        $oldConstraintName,
-        $foreignKey['newConstraintName'],
-        $foreignKey['columnName'],
-        $foreignKey['refTableName'],
-        $foreignKey['onDelete']
-      ));
-    }
+    ));
```
jraddaoui, line 129: 👍

jraddaoui (approved):
Thanks a lot @djjuhasz, it's really great that you found and fixed this before release!!

lib/task/migrate/QubitMigrate.class.php
djjuhasz: @jraddaoui I looked for a way to get the DB name from the config, but I think it requires parsing the DSN string :(

PR #1138, opened 2020-06-27
Option to run specific migration(s), refs #13368

mcantelon:
Added option, to the tools:upgrade-sql task, to run specific migrations
rather than a full upgrade. When running a specific migration the
schema version, in the database, won't be changed.

### DIFF
--- a/lib/task/migrate/arUpgradeSqlTask.class.php
+++ b/lib/task/migrate/arUpgradeSqlTask.class.php
@@ -54,6 +54,7 @@ EOF;
       new sfCommandOption('connection', null, sfCommandOption::PARAMETER_REQUIRED, 'The connection name', 'propel'),
       new sfCommandOption('no-confirmation', 'B', sfCommandOption::PARAMETER_NONE, 'Do not ask for confirmation'),
       new sfCommandOption('verbose', 'v', sfCommandOption::PARAMETER_NONE, 'Verbose mode', null),
+      new sfCommandOption('number', 'n', sfCommandOption::PARAMETER_OPTIONAL, 'Run only specific migration number(s) (separated by commas if multiple)', null),
     ));
 
     // Disable plugin loading from plugins/ before this task.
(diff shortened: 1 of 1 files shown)


lib/task/migrate/arUpgradeSqlTask.class.php
```diff
@@ -54,6 +54,7 @@ protected function configure()
       new sfCommandOption('connection', null, sfCommandOption::PARAMETER_REQUIRED, 'The connection name', 'propel'),
       new sfCommandOption('no-confirmation', 'B', sfCommandOption::PARAMETER_NONE, 'Do not ask for confirmation'),
       new sfCommandOption('verbose', 'v', sfCommandOption::PARAMETER_NONE, 'Verbose mode', null),
+      new sfCommandOption('number', 'n', sfCommandOption::PARAMETER_OPTIONAL, 'Run only specific migration number(s)', null),
```
djjuhasz: I think it would be helpful to indicate in the help text that multiple numbers should be comma-separated.

lib/task/migrate/arUpgradeSqlTask.class.php
```diff
@@ -149,88 +150,117 @@ protected function execute($arguments = array(), $options = array())
...
-      $className = preg_replace('/.*(arUpgrader\d+).*/', '$1', $filename);
-      $class = new $className;
-
-      if ($class::INIT_VERSION > $version)
-      {
-        continue;
-      }
-
-      try
+      foreach (explode(',', $options['number']) as $number)
```
djjuhasz: I think it would be better to factor the `$number` option code into a separate method to aid readability.

lib/task/migrate/arUpgradeSqlTask.class.php
```diff
@@ -149,88 +150,117 @@ protected function execute($arguments = array(), $options = array())
...
+            $this->logSection('upgrade-sql', sprintf('Migration %d not found', $number));
+
+            continue;
+          }
+
+          $this->logSection('upgrade-sql', sprintf('Running migration %d', $number));
+
+          $class = new $className;
+
+          if (true !== $class->up($this->configuration))
```
djjuhasz: I think this should use a `try ... catch` block like the main execution path.

lib/task/migrate/arUpgradeSqlTask.class.php
```diff
@@ -149,88 +150,117 @@ protected function execute($arguments = array(), $options = array())
...
+          {
+            throw new sfException(sprintf('Failed to apply upgrade %s', get_class($class)));
+          }
         }
       }
-      catch (Exception $e)
+    }
+    else
+    {
+      foreach (sfFinder::type('file')
```
djjuhasz, line 159: I think it would be better to factor this (main) execution branch into a separate method to aid code readability.

djjuhasz (changes requested):
Looks pretty good @mcantelon, I just have a few suggestions. :)

mcantelon:
@djjuhasz Thanks for the review... got sidetracked, but have updated the branch!

lib/task/migrate/arUpgradeSqlTask.class.php
```diff
@@ -340,6 +329,49 @@ protected function updateDatabaseVersion($version)
...
+      $class = new $className;
+
+      // This upgrade should have been applied already
+      if ($class::VERSION <= $version)
+      {
+        // Unless the user is moving from 1.x to 2.x
+        if (2 == $class::MIN_MILESTONE && 1 == $previousMilestone && 2 == $currentMilestone)
+        {
+          // Run migration but don't bump dbversion
+          if (true !== $class->up($this->configuration)) throw new sfException('Failed to apply upgrade '.get_class($class));
```
djjuhasz: I think this is a good opportunity to make this "if () throw ..." statement (and the one below) follow our coding standard and add the parentheses and newlines, e.g. 

```
if (true !== $class->up($this->configuration)) 
{
  throw new sfException('Failed to apply upgrade ' . get_class($class));
}
```

djjuhasz (approved):
Looks good @mcantelon. I just had one request to bring the legacy code in line with our coding standard. 👍

PR #1139, opened 2020-06-28
add digitalObjectPath to CSV template, ref  #13370

peterVG:
Umm, so I ran into some Git issues on my Vagrant box. I couldn't push to the dev branch I had created for this. I was getting `error: src refspec master does not match any`
`error: failed to push some refs to 'git@git.artefactual.com:atom.git'`
So then I got a little too aggressive and somehow pushed this commit to the `Master` branch, sorry. 
Anyway, this is a tiny commit, fixing an example CSV template that had left out a column.


mcantelon (approved):
Looks good to me!

mcantelon:
Only issue is that commit message header (first line) is more than 50 characters.

We use this standard for commit messages: https://tbaggery.com/2008/04/19/a-note-about-git-commit-messages.html

peterVG:
50 characters exactly now :-)

PR #1142, opened 2020-06-29
Allow deleting browse menu nodes, refs #13371

replaceafill:
Removes the `browseDigitalObjects` and `browseSubjects` options from
the `menu_locking_info` setting that prevents deleting them in the
user interface.

A new migration has also been added to update the setting value.

### DIFF
--- a/lib/task/migrate/migrations/arMigration0184.class.php
+++ b/lib/task/migrate/migrations/arMigration0184.class.php
@@ -1,0 +1,79 @@
+<?php
+
+/*
+ * This file is part of the Access to Memory (AtoM) software.
+ *
+ * Access to Memory (AtoM) is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * Access to Memory (AtoM) is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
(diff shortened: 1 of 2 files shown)


mcantelon (commented):
Looks good to me!

mcantelon (approved):

replaceafill:
Thank you @mcantelon!

lib/task/migrate/migrations/arMigration0184.class.php
```diff
@@ -0,0 +1,79 @@
...
+        QubitMenu::QUICK_LINKS_ID,
+        QubitMenu::ADD_EDIT_ID,
+        QubitMenu::ADMIN_ID
+      )
+    );
+
+    $setting = QubitSetting::getByName('menu_locking_info');
+    if (isset($setting))
+    {
+      $setting->value = serialize($lockedMenus);
```
jraddaoui, line 73:
```php
$setting->setValue(serialize($lockedMenus), array('sourceCulture' => true));
```

To be sure the proper value is updated, just in case.

lib/task/migrate/migrations/arMigration0184.class.php
```diff
@@ -0,0 +1,79 @@
...
+        QubitMenu::MAIN_MENU_ID,
+        QubitMenu::MANAGE_ID,
+        QubitMenu::QUICK_LINKS_ID,
+        QubitMenu::ADD_EDIT_ID,
+        QubitMenu::ADMIN_ID
+      )
+    );
+
+    $setting = QubitSetting::getByName('menu_locking_info');
+    if (isset($setting))
```
jraddaoui, line 71: An edge case, since you're talking about adding this migration to a non fully upgraded site, if that site is on 2.4.x the setting won't exist. Not a problem is they are in 2.5.x already but you could create the setting if it doesn't exist to make the migration a bit more idempotent.

jraddaoui (commented):
Thanks @replaceafill! A couple of comment below.

jraddaoui:
Oh, I just saw it is already merged ... sorry, a bit late.

PR #1143, opened 2020-06-29
Fix storage report cancel button, refs #13372

mcantelon:
Make the export storage report cancel button go to the physical storage
browse page rather than the referrer.


replaceafill (approved):
LGTM :+1:

PR #1144, opened 2020-06-30
Actor relation import update change, refs #13280

mcantelon:
Changed actor relation import so the delete-and-replace update mode
works differently. Before it would check for an identical relation
and delete it then add a new one. Now it deletes any relations between
the two actors involved first then adds a new one.

### DIFF
--- a/lib/task/import/csvAuthorityRecordRelationImportTask.class.php
+++ b/lib/task/import/csvAuthorityRecordRelationImportTask.class.php
@@ -169,37 +169,41 @@ EOF;
 
   private function importRow($sourceActorId, $targetActorId, $relationTypeId)
   {
-    // Check to see if a relation of the same type already exists
-    if ($relationId = $this->getRelation($sourceActorId, $targetActorId, $relationTypeId))
+    if (!empty($this->import->status['updateMode']))
     {
-      // Ignore this data if not updating
-      if (empty($this->import->status['updateMode']))
-      {
-        return;
-      }
-
-      // Delete existing relationship, then add new, if in "delete and replace" update mode
       if ($this->import->status['updateMode'] == 'delete-and-replace')
       {
-        $relation = QubitRelation::getById($relationId);
-        $relation->delete();
+        // Handle delete-and-replace updating by deleting all existing relations between source
+        // and target then adding new relation
+        $relations = $this->getRelations($targetActorId, $sourceActorId);
+        $relationsAlternate = $this->getRelations($sourceActorId, $targetActorId);
 
-        $this->addRelation($sourceActorId, $targetActorId, $relationTypeId);
+        foreach (array_unique(array_merge($relations, $relationsAlternate)) as $relationId)
+        {
+          $relation = QubitRelation::getById($relationId);
+          $relation->delete();
+        }
       }
-      else
+      else if ($relationId = $this->getRelationByType($sourceActorId, $targetActorId, $relationTypeId))
       {
-        // In "match and update" mode then update the relation
+        // Handle match-and-update when a relation already exists
         $this->updateRelation($relationId, $sourceActorId, $targetActorId, $relationTypeId);
+
+        return;
       }
     }
-    else
+    else if (!empty($this->getRelationByType($sourceActorId, $targetActorId, $relationTypeId)))
     {
-      $this->addRelation($sourceActorId, $targetActorId, $relationTypeId);
+      // If not updating, but relation already exists, then don't create new relation
+
+      return;
     }
+
+    $this->addRelation($sourceActorId, $targetActorId, $relationTypeId);
   }
 
   /**
-   * Use SQL to fetch the ID of a relation if it exists
+   * Use SQL to fetch the ID of a relation, of a certain type, if it exists
    *
    * @param int $sourceActorId  ID of an actor that possibly relates to others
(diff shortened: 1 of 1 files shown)


lib/task/import/csvAuthorityRecordRelationImportTask.class.php
```diff
@@ -235,6 +241,30 @@ private function getRelation($sourceActorId, $targetActorId, $relationTypeId)
...
+   *
+   * @return array  array of relation IDs
+   */
+  private function getRelations($sourceActorId, $targetActorId)
+  {
+    $sql = "SELECT id FROM relation
+      WHERE subject_id = :subject_id
+      AND object_id = :object_id";
+
+    $params = [
```
djjuhasz: Nitpick: You could avoid the $params variable assignment and define the parameter array in the `QubitPdo::fetchAll()` call.

lib/task/import/csvAuthorityRecordRelationImportTask.class.php
```diff
@@ -169,45 +169,51 @@ private function import($filepath, $indexDuringImport = false, $updateMode = fal
...
+        }
 
         $this->addRelation($sourceActorId, $targetActorId, $relationTypeId);
       }
-      else
+      else if ($relationId = $this->getRelationByType($sourceActorId, $targetActorId, $relationTypeId))
       {
-        // In "match and update" mode then update the relation
+        // Handle match-and-update when a relation already exists
         $this->updateRelation($relationId, $sourceActorId, $targetActorId, $relationTypeId);
```
djjuhasz, line 190: This won't change the existing data at all, will it?  The match requires an exact match on source, target, and relation type so the update will just update the database with the exact same three values I think.

lib/task/import/csvAuthorityRecordRelationImportTask.class.php
```diff
@@ -169,45 +169,51 @@ private function import($filepath, $indexDuringImport = false, $updateMode = fal
...
+        $relations = $this->getRelations($targetActorId, $sourceActorId);
+        $relationsAlternate = $this->getRelations($sourceActorId, $targetActorId);
+
+        foreach (array_unique(array_merge($relations, $relationsAlternate)) as $relationId)
+        {
+          $relation = QubitRelation::getById($relationId);
+          $relation->delete();
+        }
 
         $this->addRelation($sourceActorId, $targetActorId, $relationTypeId);
```
djjuhasz: Suggestion: Since the delete-and-replace, match-and-update with no match, and "no update" conditionals all result in adding a new row, you _could_ remove this call to `$this->addRelation()` and call it once after evaluating all the conditionals.  In this case, you could `return` from the match-and-update with match conditional to avoid calling `addRelation()`.  This isn't a required change imho, but would reduce the code a bit.

djjuhasz (changes requested):
@mcantelon I added one question about the need to call `updateRelation()` in the match and update scenario, and a couple of suggestions that are optional.

lib/task/import/csvAuthorityRecordRelationImportTask.class.php
mcantelon: The call to `setRelationFields` in `updateRelation` looks at current column values and uses those to set relation properties.

lib/task/import/csvAuthorityRecordRelationImportTask.class.php
mcantelon: Nice... done!

lib/task/import/csvAuthorityRecordRelationImportTask.class.php
mcantelon: Nice... yeah, that does reduce the code.

mcantelon:
@djjuhasz I've updated the PR. Thanks!

lib/task/import/csvAuthorityRecordRelationImportTask.class.php
djjuhasz: Ohhhh, I missed that.  Looks good then. :)

lib/task/import/csvAuthorityRecordRelationImportTask.class.php
```diff
@@ -235,6 +239,25 @@ private function getRelation($sourceActorId, $targetActorId, $relationTypeId)
...
+   *
+   * @return array  array of relation IDs
+   */
+  private function getRelations($sourceActorId, $targetActorId)
+  {
+    $sql = "SELECT id FROM relation
+      WHERE subject_id = :subject_id
+      AND object_id = :object_id";
+
+    $results = QubitPdo::fetchAll($sql, [':subject_id' => $sourceActorId, ':object_id' => $targetActorId], array('fetchMode' => PDO::FETCH_ASSOC));
```
djjuhasz: Okay, I feel bad now but I have another nitpick: the line is too long. 😊   Could you please split the method parameters into separate lines. 🙏

djjuhasz (changes requested):
Hi @mcantelon looks good over all, but I had one more nitpick. Sorry! 😊

mcantelon:
@djjuhasz I've updated the PR!

djjuhasz (approved):
Looks good! :shipit:

PR #1145, opened 2020-06-30
Remove actor import relation import, refs #13369

mcantelon:
From CSV authority record import remove the option to also import
relations. We now have a separate task to import authority record
relations.


djjuhasz (approved):
Looks good! 👍

PR #1146, opened 2020-07-01
Actor relation import update fix, refs #13375

mcantelon:
Added logic so newly added relations aren't deleted during CSV import
of actor relations in delete-and-replace mode.

Added display of "skipping row" warning when not in update mode and
attempting to import a relationship that already exists.

### DIFF
--- a/lib/task/import/csvAuthorityRecordRelationImportTask.class.php
+++ b/lib/task/import/csvAuthorityRecordRelationImportTask.class.php
@@ -27,6 +27,7 @@
 class csvAuthorityRecordRelationImportTask extends csvImportBaseTask
 {
   private $import;
+  private $newlyAdded = [];
 
   /**
    * @see sfTask
@@ -169,9 +170,13 @@ EOF;
 
   private function importRow($sourceActorId, $targetActorId, $relationTypeId)
   {
-    if (!empty($this->import->status['updateMode']))
+    $updateMode = !empty($this->import->status['updateMode'])
+                    ? $this->import->status['updateMode']
+                    : false;
+
+    if ($updateMode)
     {
-      if ($this->import->status['updateMode'] == 'delete-and-replace')
+      if ($updateMode == 'delete-and-replace')
       {
         // Handle delete-and-replace updating by deleting all existing relations between source
         // and target then adding new relation
@@ -180,8 +185,12 @@ EOF;
 
         foreach (array_unique(array_merge($relations, $relationsAlternate)) as $relationId)
         {
-          $relation = QubitRelation::getById($relationId);
-          $relation->delete();
+          // Don't delete relations that have been added during this import
+          if (!in_array($relationId, $this->newlyAdded))
+          {
+            $relation = QubitRelation::getById($relationId);
+            $relation->delete();
+          }
         }
       }
       else if ($relationId = $this->getRelationByType($sourceActorId, $targetActorId, $relationTypeId))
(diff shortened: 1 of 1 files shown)


lib/task/import/csvAuthorityRecordRelationImportTask.class.php
```diff
@@ -169,9 +170,13 @@ private function import($filepath, $indexDuringImport = false, $updateMode = fal
 
   private function importRow($sourceActorId, $targetActorId, $relationTypeId)
   {
-    if (!empty($this->import->status['updateMode']))
+    $updateMode = !empty($this->import->status['updateMode'])
+                    ? $this->import->status['updateMode']
+                    : false;
+
+    if (!empty($updateMode))
```
djjuhasz: This can be be simplified to `if ($updateMode)` now that you are assigning `false` when the flag is empty above.

lib/task/import/csvAuthorityRecordRelationImportTask.class.php
```diff
@@ -181,7 +186,12 @@ private function importRow($sourceActorId, $targetActorId, $relationTypeId)
         foreach (array_unique(array_merge($relations, $relationsAlternate)) as $relationId)
         {
           $relation = QubitRelation::getById($relationId);
```
djjuhasz: I think this should be moved into the conditional below (with `$relation->delete()`), to avoid the cost of querying the database unnecessarily.

lib/task/import/csvAuthorityRecordRelationImportTask.class.php
```diff
@@ -195,11 +205,16 @@ private function importRow($sourceActorId, $targetActorId, $relationTypeId)
     else if (!empty($this->getRelationByType($sourceActorId, $targetActorId, $relationTypeId)))
     {
       // If not updating, but relation already exists, then don't create new relation
+      echo $this->import->logError('Skipping row as relationship already exists');
```
djjuhasz, line 207: nice :)

lib/task/import/csvAuthorityRecordRelationImportTask.class.php
```diff
@@ -195,11 +205,16 @@ private function importRow($sourceActorId, $targetActorId, $relationTypeId)
...
 
       return;
     }
 
-    $this->addRelation($sourceActorId, $targetActorId, $relationTypeId);
+    $relation = $this->addRelation($sourceActorId, $targetActorId, $relationTypeId);
+    if (!empty($updateMode) && $updateMode == 'delete-and-replace')
```
djjuhasz: `!empty($updateMode)` is unnecessary here.

djjuhasz (changes requested):
I've flagged a couple of minor changes @mcantelon, but looks good overall.

lib/task/import/csvAuthorityRecordRelationImportTask.class.php
mcantelon: Good call!

mcantelon:
Thanks for the review @djjuhasz ! I've update the PR.

lib/task/import/csvAuthorityRecordRelationImportTask.class.php
```diff
@@ -195,11 +204,16 @@ private function importRow($sourceActorId, $targetActorId, $relationTypeId)
...
     {
       // If not updating, but relation already exists, then don't create new relation
+      echo $this->import->logError('Skipping row as relationship already exists');
 
       return;
     }
 
-    $this->addRelation($sourceActorId, $targetActorId, $relationTypeId);
+    $relation = $this->addRelation($sourceActorId, $targetActorId, $relationTypeId);
+    if ($updateMode == 'delete-and-replace')
```
djjuhasz, line 214: Nitpick: there should be a blank line before the `if` conditional.

djjuhasz (changes requested):
Sorry @mcantelon, one last stylistic nitpick, then it should be good to go.

PR #1147, opened 2020-07-02
add digitalObjectPath to example CSV, ref 13370

peterVG:
A previous commit for the same issue# was not merged properly due to some gitolite issues. Re-doing it here now. 

This change adds a digitalObjectPath column to the `example authority_records.csv` file.


mcantelon (approved):
Looks good!

PR #1150, opened 2020-07-07
Warn of duplicate actor identifiers, refs #13373

mcantelon:
Added, in the web UI and during authority record CSV import/updating,
warnings when an authority record identifier has already been used.

### DIFF
--- a/lib/task/import/csvAuthorityRecordImportTask.class.php
+++ b/lib/task/import/csvAuthorityRecordImportTask.class.php
@@ -234,6 +234,13 @@ EOF;
       {
         if ($self->object)
         {
+          // Warn if identifier's already been used
+          if (!empty($identifier = $self->columnValue('descriptionIdentifier'))
+              && QubitValidatorActorDescriptionIdentifier::identifierUsedByAnotherActor($identifier, $self->object))
+          {
+            print $self->logError(sprintf('Authority record identifier "%s" not unique.', $identifier));
+          }
+
           if (
             isset($self->rowStatusVars['typeOfEntity'])
             && $self->rowStatusVars['typeOfEntity']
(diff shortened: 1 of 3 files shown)


lib/task/import/csvAuthorityRecordImportTask.class.php
```diff
@@ -234,6 +234,15 @@ public function execute($arguments = array(), $options = array())
       {
         if ($self->object)
         {
+          // Warn if identifier's already been used
+          if (!empty($identifier = $self->columnValue('descriptionIdentifier')))
+          {
+            if (QubitValidatorActorDescriptionIdentifier::identifierUsedByAnotherActor($identifier, $self->object))
```
jraddaoui: Nitpick: it could be one `if`.

plugins/sfIsaarPlugin/lib/QubitValidatorActorDescriptionIdentifier.class.php
```diff
@@ -0,0 +1,57 @@
...
+                   '%1%Authority record identifier%2% - value not unique.',
+                   array('%1%' => '<a href="http://ica-atom.org/doc/RS-2#5.4.1">', '%2%' => '</a>'));
+
+      throw new sfValidatorError($this, $message, array('value' => $value));
+    }
+
+    return $value;
+  }
+
+  public static function identifierUsedByAnotherActor($identifier, $byResource = null)
```
jraddaoui: Since the resource is a required option, why do we make it optional in this function?

jraddaoui (commented):
Looking good @mcantelon!

Shouldn't we reinforce this in the DB column making it unique and checking there isn't already duplicated values?

Also, I guess this has to wait for 2.7, right?

mcantelon:
Thanks @jraddaoui ... I've updated the PR! This is for 2.7 and the work order was specifically to warn about identifier reuse rather than prevent it (so they can add/import things with clashing identifiers and fix them later I guess).

jraddaoui (approved):
Thanks @mcantelon!

Ready for when we create the new qa/2.x branch ;)

PR #1153, opened 2020-07-07
Avoid warning when no PREMIS results, refs #13348

djjuhasz:
Prevent a PHP Warning when adding an information object to ES that has no
related PREMIS metadata.


sbreker (approved):
Looks great! Thanks @djjuhasz!

PR #1155, opened 2020-07-09
Add DO option to clipboard export, refs #13395

PeteVox:
Add a new system setting to enable export of digital objects attached to
the clipboard (information object and actor records only). All digital
object types can be exported, but only masters are exported for any
image type.

Digital objects are included in the export ZIP along with associated
metadata.

Incorporate basic user permissions checks prior to export, with further
checks for rights as part of the background job/task. Add a new method
to the digital object model to check for any conditional copyright
attached. If set, then do not export the affected digital object.

Modify the export page interface to rebuild options panel depending on
controller-based permissions checks. Also move help and other message
construction from template to controller.

Make the selection of descendant or digital objects in the export page
mutually exclusive. Add several new styling rules (LESS) to the Dominium
theme in scaffolding that apply to the options panel (#export-options).

Change some help and alert messages to improve consistency.

Build additional export form parameter validation in the controller to
mitigate attempts to manually override parameters, and also to prevent
direct access to the export page URL without objects on the clipboard.

### DIFF
--- a/apps/qubit/modules/default/actions/alertsComponent.class.php
+++ b/apps/qubit/modules/default/actions/alertsComponent.class.php
@@ -43,22 +43,34 @@ class DefaultAlertsComponent extends sfComponent
         if (isset($firstJobAlertSkipped[$job->id]))
         {
           // Assemble job description
-          $message = $this->context->i18n->__('%1% (started: %2%, status: %3%).', array(
-            '%1%' => (string)$job,
-            '%2%' => $job->getCreationDateString(),
-            '%3%' => $job->getStatusString()));
+          $message = $this->context->i18n->__('%1% (started: %2%, status: %3%).', 
+            array(
+              '%1%' => (string)$job,
+              '%2%' => $job->getCreationDateString(),
+              '%3%' => $job->getStatusString()
+            )
+          );
 
           // Add download path. if applicable
           if (isset($job->downloadPath) && $job->statusId == QubitTerm::JOB_STATUS_COMPLETED_ID)
           {
-            $message .= $this->context->i18n->__(' %1%Download%2% (%3% b)', array(
-              '%1%' => sprintf('<a href="%s">', sfConfig::get('app_siteBaseUrl') .'/'. $job->downloadPath),
-              '%2%' => '</a>',
-              '%3%' => hr_filesize(filesize($job->downloadPath))));
+            $message .= $this->context->i18n->__(' %1%Download%2% (%3% b)', 
+              array(
+                '%1%' => sprintf('<a href="%s">', sfConfig::get('app_siteBaseUrl') .'/'. $job->downloadPath),
+                '%2%' => '</a>',
+                '%3%' => hr_filesize(filesize($job->downloadPath))
+              )
+            );
           }
-
-          // Add refresh link
-          $message .= ' &mdash; <a href="javascript:location.reload();">refresh the page</a> for updates.';
+          else
+          {
+            $message .= ' ' . $this->context->i18n->__('%1%Refresh the page%2% for progress updates.', 
+              array(
+                '%1%' => '<a href="javascript:location.reload();">',
+                '%2%' => '</a>'
+              )
+            );
+          }
 
           // Determine alert type to show
           $alertTypes = array(
--- a/apps/qubit/modules/object/actions/exportAction.class.php
+++ b/apps/qubit/modules/object/actions/exportAction.class.php
@@ -23,30 +23,113 @@ class ObjectExportAction extends DefaultEditAction
   public static
     $NAMES = array(
       'levels',
-      'objectType',
-      'format');
+      'type',
+      'format',
+      'includeDescendants',
+      'includeAllLevels',
+      'includeDigitalObjects',
+      'includeDrafts');
 
   private $choices = array();
 
   protected function earlyExecute()
   {
+    sfProjectConfiguration::getActive()->loadHelpers(array('I18N'));
+
+    // Initialize help array: messages added depending on visibility of fields
+    $this->helpMessages = array();
+
+    $this->typeChoices = array(
+      'informationObject' => sfConfig::get('app_ui_label_informationobject'),
+      'actor' => sfConfig::get('app_ui_label_actor'),
+      'repository' => sfConfig::get('app_ui_label_repository')
+    );
+
     $this->form->getValidatorSchema()->setOption('allow_extra_fields', true);
   }
 
   public function execute($request)
   {
-    $this->format = $request->getParameter('format', 'csv');
-    $this->objectType = $request->getParameter('objectType');
+
+    // Get object type and validate
+    // Currently 'switch' to process the inbound parameter, validate and set default - could be if/then if preferred
+    $this->objectType = trim(strtolower($request->getParameter('type')));
+    switch ($this->objectType)
+    {
+      case 'actor':
+        $className = 'QubitActor';
+
+        break;
+
+      case 'repository':
+        $className = 'QubitRepository';
+
+        break;
+
+      default:
+        $this->objectType = 'informationObject';
+        $className = 'QubitInformationObject';
+    }
+
+    // Stop direct access to export url or empty export attempts
+    // Check that the user has something on the clipboard for the current type - otherwise redirect back to clipboard
+    $items = $this->context->user->getClipboard()->countByType();
+    if(empty($items) || !array_key_exists($className, $items) || 1 > $items[$className])
+    {
+      $this->redirect($this->context->routing->generate(null, array('module' => 'user', 'action' => 'clipboard')) . '?type=' . $this->objectType);
+    }
+
+    // Get format and validate
+    // Currently 'switch' to process the inbound parameter, validate and set default - could be if/then if preferred
+    $this->formatType = trim(strtolower($request->getParameter('format')));
+    if(!('xml' == $this->formatType || 'csv' == $this->formatType) || 'repository' == $this->objectType)
+    {
+        $this->formatType = 'csv';
+    }
+
+    // Basic permission check to determine whether digital object export should be made available
+    $this->digitalObjectsAvailable = false;
+    if (sfConfig::get('app_clipboard_export_digitalobjects_enabled', false) && ('informationObject' == $this->objectType || ('actor' == $this->objectType && $this->context->user->isAuthenticated())))
+    {
+      $this->digitalObjectsAvailable = true;
+    }
+
+    // Show export options panel if:
+    // information object type
+    // or, if actor type and digital objects are on the clipboard
+    $this->showOptions = 'informationObject' == $this->objectType || ('actor' == $this->objectType && $this->digitalObjectsAvailable);
+
+    // Get field includeDescendants if:
+    // options enabled
+    // and, information object type
+    $this->descendantsIncluded = $this->showOptions && 'informationObject' == $this->objectType && 'on' == $request->getParameter('includeDescendants');
+
+    // get field includeAllLevels if:
+    // descendantsIncluded enabled
+    $this->descendantsAllLevels = $this->descendantsIncluded && 'on' == $request->getParameter('includeAllLevels');
+
+    // Get field includeDigitalObjects if:
+    // digital objects are on the clipboard
+    // and, information object type and descendants are not selected
+    // or, if actor type
+    $this->includeDigitalObjects = false;
+    if ($this->digitalObjectsAvailable)
+    {
+      $this->includeDigitalObjects = 'on' == $request->getParameter('includeDigitalObjects');
+    }
+
+    // Get field includeDrafts if:
+    // options enabled
+    // and, user is authenticated
+    $this->draftsIncluded = false;
+    if ($this->showOptions && $this->context->user->isAuthenticated())
+    {
+      $this->draftsIncluded = 'on' == $request->getParameter('includeDrafts');
+    }
 
     parent::execute($request);
 
     $this->response->addJavaScript('exportOptions', 'last');
-
-    $this->redirectUrl = array('module' => 'object', 'action' => 'export');
-    if (null !== $referrer = $request->getReferer())
-    {
-      $this->redirectUrl = $referrer;
-    }
 
     $this->title = $this->context->i18n->__('Export');
 
@@ -60,7 +143,7 @@ class ObjectExportAction extends DefaultEditAction
 
         $this->doBackgroundExport($request);
 
-        $this->redirect($this->redirectUrl);
+        $this->redirect($this->context->routing->generate(null, array('module' => 'user', 'action' => 'clipboard')) . '?type=' . $this->objectType);
       }
     }
   }
@@ -69,54 +152,122 @@ class ObjectExportAction extends DefaultEditAction
   {
     switch ($name)
     {
-      case 'levels':
-        $this->form->setValidator('levels', new sfValidatorPass);
+      case 'type':
+        $this->form->setValidator('type', new sfValidatorString(array('required' => true)));
+        $this->form->setWidget('type', new sfWidgetFormSelect(array('label' => __('Type'), 'choices' => $this->typeChoices)));
+        $this->form->setDefault('type', $this->objectType);
 
-        foreach (QubitTerm::getLevelsOfDescription() as $item)
+        break;
+
+      case 'format':
+        $this->form->setValidator('format', new sfValidatorString(array('required' => true)));
+        $choices = array();
+        $choices['csv'] = $this->context->i18n->__('CSV');
+        if('repository' != $this->objectType)
         {
-          $this->choices[$item->id] = $item->__toString();
+          $choices['xml'] = $this->context->i18n->__('XML');
+        }
+        $this->form->setWidget('format', new sfWidgetFormSelect(array('label' => __('Format'), 'choices' => $choices)));
+        $this->form->setDefault('format', 'actor' != $this->objectType ? 'xml' : 'csv');
+
+        break;
+
+      // Enable field includeDescendants if:
+      // options enabled
+      // and, information object type
+      case 'includeDescendants':
+        if ($this->showOptions && 'informationObject' == $this->objectType)
+        {
+          $this->form->setWidget('includeDescendants', new sfWidgetFormInputCheckbox(array('label' => __('Include descendants'))));
+          $this->form->setDefault('includeDescendants', false);
+          $this->helpMessages[] = __('Choosing "Include descendants" will include all lower-level records beneath those currently on the clipboard in the export.');
         }
 
-        $size = count($this->choices);
+        break;
+
+      // Enable field includeAllLevels if:
+      // options enabled
+      // and, information object type
+      case 'includeAllLevels':
+        if ($this->showOptions && 'informationObject' == $this->objectType)
+        {
+          $this->form->setWidget('includeAllLevels', new sfWidgetFormInputCheckbox(array('label' => __('Include all descendant levels of description'))));
+          $this->form->setDefault('includeAllLevels', true);
+        }
+
+        break;
+
+      // Enable field levels if:
+      // options enabled
+      // and, information object type
+      case 'levels':
+
+        $this->form->setValidator('levels', new sfValidatorPass);
+
+        $this->levelChoices = array();
+        foreach (QubitTerm::getLevelsOfDescription() as $item)
+        {
+          $this->levelChoices[$item->id] = $item->__toString();
+        }
+
+        $size = count($this->levelChoices);
         if ($size === 0)
         {
           $size = 4;
         }
 
-        $this->form->setWidget('levels', new sfWidgetFormSelect(array('choices' => $this->choices, 'multiple' => true), array('size' => $size)));
+        if ($this->showOptions && 'informationObject' == $this->objectType)
+        {
+          $this->form->setWidget('levels', new sfWidgetFormSelect(
+            array(
+              'label' => __('Select levels of descendant descriptions for inclusion'), 
+              'help' => __('If no levels are selected, the export will fail. You can use the control (Mac ⌘) and/or shift keys to multi-select values from the Levels of description menu. ' . 
+                           'It is necessary to include the level(s) above the desired export level, up to and including the level contained in the clipboard. ' . 
+                           'Otherwise, no records will be included in the export.'), 
+              'choices' => $this->levelChoices, 
+              'multiple' => true
+            ), 
+            array(
+              'size' => $size
+            )
+          ));
+        }
+        
+        break;
+
+      // Enable field includeDigitalObjects if:
+      // digital objects are available
+      case 'includeDigitalObjects':
+        if ($this->digitalObjectsAvailable)
+        {
+          switch ($this->objectType)
+          {
+            case 'informationObject':
+              $this->helpMessages[] = __('Note: Digital objects with restricted access or copyright will not be exported.') . ' ' . __('It is not possible to select both digital objects and descendants for export at the same time. Digital objects can only be exported for records that are on the clipboard.');
+
+              break;
+
+            case 'actor':
+              $this->helpMessages[] = __('Note: Digital objects with restricted access or copyright will not be exported.');
+
+              break;
+          }
+          $this->form->setWidget('includeDigitalObjects', new sfWidgetFormInputCheckbox(array('label' => __('Include digital objects'))));
+          $this->form->setDefault('includeDigitalObjects', true);
+        }
 
         break;
 
-      case 'objectType':
-        if (isset($this->objectType))
+      // Enable field includeDrafts if:
+      // options enabled
+      // and, user is authenticated
+      case 'includeDrafts':
+        if ($this->showOptions && $this->context->user->isAuthenticated())
         {
-          $choices = array(
-            $this->objectType => sfConfig::get('app_ui_label_'.strtolower($this->objectType))
-          );
+          $this->form->setWidget('includeDrafts', new sfWidgetFormInputCheckbox(array('label' => __('Include draft records'))));
+          $this->helpMessages[] = __('Choosing "Include draft records" will include those marked with a Draft publication status in the export. Note: if you do NOT choose this option, any descendants of a draft record will also be excluded, even if they are published.');
+          $this->form->setDefault('includeDrafts', true);
         }
-        else
-        {
-          $choices = array(
-            'informationObject' => sfConfig::get('app_ui_label_informationobject'),
-            'actor' => sfConfig::get('app_ui_label_actor'),
-            'repository' => sfConfig::get('app_ui_label_repository')
-          );
-        }
-
-        $this->form->setValidator('objectType', new sfValidatorString(array('required' => true)));
-        $this->form->setWidget('objectType', new sfWidgetFormSelect(array('choices' => $choices)));
-
-        break;
-
-      case 'format':
-        $choices = array(
-          'csv' => $this->context->i18n->__('CSV'),
-          'xml' => $this->context->i18n->__('XML')
-        );
-
-        $this->form->setDefault('format', $this->format);
-        $this->form->setValidator('format', new sfValidatorString(array('required' => true)));
-        $this->form->setWidget('format', new sfWidgetFormSelect(array('choices' => $choices)));
 
         break;
 
@@ -139,7 +290,7 @@ class ObjectExportAction extends DefaultEditAction
 
         break;
 
-      case 'objectType':
+      case 'type':
       case 'format':
         $this->$name = $this->form->getValue($name);
 
@@ -152,8 +303,9 @@ class ObjectExportAction extends DefaultEditAction
 
   protected function doBackgroundExport($request)
   {
-    // Create array of selections to pass to background job where Term ID will
-    // be key, and Term description is value.
+    // Create array of selections to pass to background job where 
+    // Term ID will be key, and Term description is value
+    $levelsOfDescription = array();
     foreach ($this->levels as $value)
     {
       $levelsOfDescription[$value] = $this->choices[$value];
@@ -161,20 +313,27 @@ class ObjectExportAction extends DefaultEditAction
 
     $options = array(
       'params' => array('fromClipboard' => true, 'slugs' => $this->context->user->getClipboard()->getAll()),
-      'current-level-only' => 'on' !== $request->getParameter('includeDescendants'),
-      'public' => 'on' !== $request->getParameter('includeDrafts'),
+      'current-level-only' => !$this->descendantsIncluded,
+      'public' => !$this->draftsIncluded,
       'objectType' => $this->objectType,
-      'levels' => $levelsOfDescription,
-      'name' => $this->context->i18n->__('CSV export')
+      'levels' => $levelsOfDescription
     );
 
-    if ('xml' === $this->format)
+    $options['name'] = $this->context->i18n->__('xml' === $this->formatType ? 'XML export' : 'CSV export' );
+
+    if($this->includeDigitalObjects)
     {
-      $options['name'] = $this->context->i18n->__('XML export');
+      $options['name'] = $this->context->i18n->__('%1% and %2%', 
+        array(
+          '%1%' => sfConfig::get('app_ui_label_digitalobject'),
+          '%2%' => $options['name']
+        )
+      );
+      $options['includeDigitalObjects'] = true;
     }
 
-    // When exporting actors, ensure aliases and relations are also exported.
-    if ('actor' === $this->objectType && 'csv' === $this->format)
+    // When exporting actors, ensure aliases and relations are also exported
+    if ('actor' === $this->objectType && 'csv' === $this->formatType)
     {
       $options['aliases'] = true;
       $options['relations'] = true;
@@ -182,16 +341,24 @@ class ObjectExportAction extends DefaultEditAction
 
     try
     {
-      // Preview export before attempting, if job supports this
-      $exportPreviewSupported = method_exists($this->getJobNameString(), 'exportOrCheckForResults');
+      $jobName = $this->getJobNameString();
 
-      if ($exportPreviewSupported && !call_user_func(array($this->getJobNameString(), 'exportOrCheckForResults'), $options))
+      // Preview export before attempting, if job supports this
+      $exportPreviewSupported = method_exists($jobName, 'exportOrCheckForResults');
+
+      if ($exportPreviewSupported && !call_user_func(array($jobName, 'exportOrCheckForResults'), $options))
       {
-        throw new sfException('Export complete - no records were exported.');
+        throw new sfException($this->context->i18n->__(
+          'No records were exported for your current selection. Please %1%refresh the page and choose different export options%2%.', 
+          array(
+            '%1%' => '<a href="javascript:location.reload();">',
+            '%2%' => '</a>',
+          )
+        ));
       }
       else
       {
-        $job = QubitJob::runJob($this->getJobNameString(), $options);
+        $job = QubitJob::runJob($jobName, $options);
 
         // If anonymous user, store job ID in session
         if (!$this->context->user->isAuthenticated())
@@ -200,21 +367,29 @@ class ObjectExportAction extends DefaultEditAction
           $manager->addJobAssociation($job);
         }
 
-        if ($this->context->user->isAuthenticated())
-        {
-          $message = $this->context->i18n->__('%1%Export initiated.%2% Check %3%job management%4% page to download the results when it has completed.', array(
-            '%1%' => '<strong>',
-            '%2%' => '</strong>',
-            '%3%' => sprintf('<a href="%s">', $this->context->routing->generate(null, array('module' => 'jobs', 'action' => 'browse'))),
-            '%4%' => '</a>'));
-        }
-        else
-        {
-          $message = $this->context->i18n->__('Export initiated. Progress will be reported, and a download link provided, in a subsequent notification &mdash; <a href="javascript:location.reload();">refresh the page</a> for updates.');
-        }
-      }
+        $message = array();
 
-      $this->context->user->setFlash('notice', $message);
+        $message[] = $this->context->i18n->__(
+          '%1%Your %2% export package is being built.%3% ' . ($this->context->user->isAuthenticated() ? 'The %4%job management page%5% will show progress and a download link when complete.' : 'Please %4%refresh the page%5% to see progress and a download link when complete.'), 
+          array(
+            '%1%' => '<strong>',
+            '%2%' => strtolower($this->typeChoices[$this->objectType]),
+            '%3%' => '</strong>',
+            '%4%' => ($this->context->user->isAuthenticated() ? sprintf('<a href="%s">', $this->context->routing->generate(null, array('module' => 'jobs', 'action' => 'browse'))) : '<a href="javascript:location.reload();">') . '<strong>',
+            '%5%' => '</strong></a>'
+          )
+        );
+
+        $message[] = $this->context->i18n->__(
+          '%1%Note:%2% AtoM may remove export packages after a period of time to free up storage space. When your export is ready you should download it as soon as possible.', 
+          array(
+            '%1%' => '<strong>',
+            '%2%' => '</strong>'
+          )
+        );
+
+      }
+      $this->context->user->setFlash('info', '<p>' . implode('</p><p>', $message) . '</p>');
     }
     catch (sfException $e)
     {
@@ -228,7 +403,7 @@ class ObjectExportAction extends DefaultEditAction
     switch ($this->objectType)
     {
       case 'informationObject':
-        if ('csv' == $this->format)
+        if ('csv' == $this->formatType)
         {
           return 'arInformationObjectCsvExportJob';
         }
@@ -238,7 +413,7 @@ class ObjectExportAction extends DefaultEditAction
         }
 
       case 'actor':
-        if ('csv' == $this->format)
+        if ('csv' == $this->formatType)
         {
           return 'arActorCsvExportJob';
         }
--- a/apps/qubit/modules/object/templates/exportSuccess.php
+++ b/apps/qubit/modules/object/templates/exportSuccess.php
@@ -18,62 +18,58 @@
     <?php echo $form->renderHiddenFields() ?>
 
     <section id="content">
+
       <div id="export-options" data-export-toggle="tooltip" data-export-title="<?php echo __('Export') ?>" data-export-alert-message="<?php echo __('Error: You must have at least one %1%Level of description%2% selected or choose %1%Include all descendant levels of description%2% to proceed.', array('%1%' => '<strong>', '%2%' => '</strong>')) ?>">
-      <fieldset class="collapsible">
+      <fieldset>
 
         <legend><?php echo __('Export options') ?></legend>
-        <div class="form-item">
-          <?php echo $form->objectType
-            ->label(__('Type'))
-            ->renderRow() ?>
-        </div>
 
-        <div class="form-item">
+        <div class="fieldset-wrapper">
+          <?php echo $form->type
+            ->renderRow() ?>
           <?php echo $form->format
-            ->label(__('Format'))
             ->renderRow() ?>
-        </div>
-
-        <div class="form-item">
+          <?php if ($showOptions): ?>
           <div class="panel panel-default" id="exportOptions">
             <div class="panel-body">
-              <div class="generic-help-box">
-                <label class="generic-help-type">
-                  <input name="includeDescendants" type="checkbox"/>
-                  <?php echo __('Include descendants') ?>
-                </label>
-                <a href="#" class="generic-help-icon" aria-expanded="false"><i class="fa fa-question-circle pull-right"></i></a>
-              </div>
-
-              <?php if ($sf_user->isAuthenticated()): ?>
-                <label>
-                  <input name="includeDrafts" type="checkbox"/>
-                  <?php echo __('Include draft records') ?>
-                </label>
+              <?php if (!empty($helpMessages)): ?>
+                <div class="generic-help-box">
+                  <a href="#" class="generic-help-icon" aria-expanded="false"><i class="fa fa-question-circle pull-right"></i></a>
+                </div>
               <?php endif; ?>
-              <label>
-                <input name="includeAllLevels" type="checkbox"/>
-                <?php echo __('Include all descendant levels of description') ?>
-              </label>
-              <div class="hidden" id="exportLevels">
-                <?php echo $form->levels
-                  ->renderLabel(__('Select levels of descendant descriptions for inclusion'), array('style' => 'font-weight: bold')) ?>
-                <?php echo $form->levels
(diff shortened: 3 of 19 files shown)


apps/qubit/modules/object/actions/exportAction.class.php
```diff
@@ -24,19 +24,47 @@ class ObjectExportAction extends DefaultEditAction
...
     $this->form->getValidatorSchema()->setOption('allow_extra_fields', true);
   }
 
   public function execute($request)
   {
     $this->format = $request->getParameter('format', 'csv');
     $this->objectType = $request->getParameter('objectType');
+    
+    $this->descendantsIncluded = ('on' == $request->getParameter('includeDescendants') ? true : false );
+    $this->descendantsAllLevels = ($this->descendantsIncluded && 'on' == $request->getParameter('includeAllLevels') ? true : false );
```
jraddaoui: Cosmetic: you could assign the comparison directly ...

```php
$this->descendantsIncluded = ('on' == $request->getParameter('includeDescendants'));
$this->descendantsAllLevels = ($this->descendantsIncluded && 'on' == $request->getParameter('includeAllLevels'));
```

The parenthesis may not be needed (specially in the first case), but I'm not completely sure.

apps/qubit/modules/object/actions/exportAction.class.php
```diff
@@ -24,19 +24,47 @@ class ObjectExportAction extends DefaultEditAction
...
+    
+    if ($this->context->user->isAuthenticated())
+    {
+      $this->draftsIncluded = ('on' == $request->getParameter('includeDrafts') ? true : false );
+    }
+    else
+    {
+      $this->draftsIncluded = false;
+    }
```
jraddaoui: Cosmetic:

```php
$this->draftsIncluded = false;

if ($this->context->user->isAuthenticated() && 'on' == $request->getParameter('includeDrafts'))
{
  $this->draftsIncluded = true;
}
```

apps/qubit/modules/object/actions/exportAction.class.php
```diff
@@ -24,19 +24,47 @@ class ObjectExportAction extends DefaultEditAction
...
+    
+    if ($this->getDigitalObjects())
+    {
+      $this->includeDigitalObjects = ('on' == $request->getParameter('includeDigitalObjects') ? true : false );
+    }
+    else
+    {
+      $this->includeDigitalObjects = false;
+    }
```
jraddaoui: Cosmetic: similar to the comment above for the `$this->includeDigitalObjects` assignation.

apps/qubit/modules/object/actions/exportAction.class.php
```diff
@@ -24,19 +24,47 @@ class ObjectExportAction extends DefaultEditAction
...
+    
+    $this->showOptions = ( 'actor' == $this->objectType && !$this->digitalObjectsAvailable ? false : true );
```
jraddaoui: I think you need to consider in here that `$this->objectType` could also be `'repository'`.

apps/qubit/modules/object/actions/exportAction.class.php
```diff
@@ -116,8 +126,64 @@ protected function addField($name)
...
         $this->form->setDefault('format', $this->format);
         $this->form->setValidator('format', new sfValidatorString(array('required' => true)));
-        $this->form->setWidget('format', new sfWidgetFormSelect(array('choices' => $choices)));
+        $this->form->setWidget('format', new sfWidgetFormSelect(array('label' => __('Format'), 'choices' => $choices)));
+
+        break;
+
+      case 'includeDescendants':
+          $this->form->setWidget('includeDescendants', new sfWidgetFormInputCheckbox(array('label' => 'Include descendants', 'help' => __('Choosing "Include descendants" will include all lower-level records beneath those currently on the clipboard in the export.'))));
+        break;
```
jraddaoui: Cosmetic:

- Fix indentation, reduce to 2 spaces on the first line.
- Split array items in different lines to reduce line length.
- Empty line before `break;`.

apps/qubit/modules/object/actions/exportAction.class.php
```diff
@@ -116,8 +126,64 @@ protected function addField($name)
...
+
+      case 'includeAllLevels':
+          $this->form->setWidget('includeAllLevels', new sfWidgetFormInputCheckbox(array('label' => 'Include all descendant levels of description')));
+        break;
```
jraddaoui: Cosmetic: same as above.

apps/qubit/modules/object/actions/exportAction.class.php
```diff
@@ -116,8 +126,64 @@ protected function addField($name)
 
         $this->form->setDefault('format', $this->format);
         $this->form->setValidator('format', new sfValidatorString(array('required' => true)));
-        $this->form->setWidget('format', new sfWidgetFormSelect(array('choices' => $choices)));
+        $this->form->setWidget('format', new sfWidgetFormSelect(array('label' => __('Format'), 'choices' => $choices)));
+
+        break;
+
+      case 'includeDescendants':
+          $this->form->setWidget('includeDescendants', new sfWidgetFormInputCheckbox(array('label' => 'Include descendants', 'help' => __('Choosing "Include descendants" will include all lower-level records beneath those currently on the clipboard in the export.'))));
```
jraddaoui: The label is missing the i18n helper.

apps/qubit/modules/object/actions/exportAction.class.php
```diff
@@ -116,8 +126,64 @@ protected function addField($name)
...
+        break;
+
+      case 'includeAllLevels':
+          $this->form->setWidget('includeAllLevels', new sfWidgetFormInputCheckbox(array('label' => 'Include all descendant levels of description')));
```
jraddaoui: The label is missing the i18n helper.

apps/qubit/modules/object/actions/exportAction.class.php
```diff
@@ -116,8 +126,64 @@ protected function addField($name)
...
+          $this->choices[$item->id] = $item->__toString();
+        }
 
+        $size = count($this->choices);
+        if ($size === 0)
+        {
+          $size = 4;
+        }
+
+        $this->form->setWidget('levels', new sfWidgetFormSelect(array('label' => __('Select levels of descendant descriptions for inclusion'), 'help' => __('If no levels are selected, the export will fail. You can use the control (Mac ⌘) and/or shift keys to multi-select values from the Levels of description menu. It is necessary to include the level(s) above the desired export level, up to and including the level contained in the clipboard. Otherwise, no records will be included in the export.'), 'choices' => $this->choices, 'multiple' => true), array('size' => $size)));
```
jraddaoui: Cosmetic: it would be nice to reduce this line length concatenating the text with `.` splitting the text in different lines.

apps/qubit/modules/object/actions/exportAction.class.php
```diff
@@ -116,8 +126,64 @@ protected function addField($name)
...
+        {
+          switch ($this->objectType)
+          {
+            case 'actor':
+              $helpMessage = '';
+              break;
+
+            default:
+              $helpMessage = __('Please note: digital objects will not be exported for any included descendant records, only for records that are on the clipboard.');
+          }
```
jraddaoui: Not considering 'repository' in this `switch`, which is probably cleaner as an `if`.

apps/qubit/modules/object/actions/exportAction.class.php
```diff
@@ -116,8 +126,64 @@ protected function addField($name)
...
+          
+          $this->form->setWidget('includeDigitalObjects', new sfWidgetFormInputCheckbox(array('label' => 'Include digital objects', 'help' => $helpMessage)));
+        } else {
```
jraddaoui: Cosmetic: new lines for the `else` and the `{`.

apps/qubit/modules/object/actions/exportAction.class.php
```diff
@@ -116,8 +126,64 @@ protected function addField($name)
...
+          $this->form->setWidget('includeDigitalObjects', new sfWidgetFormInputHidden(array('label' => false)));
```
jraddaoui: There is a JS file that takes care of hiding/showing fields based on another input values. I think we need to reconsider that JS (I'll follow-up about this later).

apps/qubit/modules/object/actions/exportAction.class.php
```diff
@@ -116,8 +126,64 @@ protected function addField($name)
...
+        }
+        $this->form->setDefault('includeDigitalObjects', false);
+        break;
```
jraddaoui, line 259: Cosmetic: empty line before `break`.

apps/qubit/modules/object/actions/exportAction.class.php
```diff
@@ -116,8 +126,64 @@ protected function addField($name)
...
+          {
+            case 'actor':
+              $helpMessage = '';
+              break;
+
+            default:
+              $helpMessage = __('Please note: digital objects will not be exported for any included descendant records, only for records that are on the clipboard.');
+          }
+          
+          $this->form->setWidget('includeDigitalObjects', new sfWidgetFormInputCheckbox(array('label' => 'Include digital objects', 'help' => $helpMessage)));
```
jraddaoui: The label is missing the i18n helper.

apps/qubit/modules/object/actions/exportAction.class.php
```diff
@@ -116,8 +126,64 @@ protected function addField($name)
...
+        } else {
+          $this->form->setWidget('includeDigitalObjects', new sfWidgetFormInputHidden(array('label' => false)));
+        }
+        $this->form->setDefault('includeDigitalObjects', false);
+        break;
+
+      case 'includeDrafts':
+        if ($this->context->user->isAuthenticated())
+        {
+          $this->form->setWidget('includeDrafts', new sfWidgetFormInputCheckbox(array('label' => 'Include draft records', 'help' => __('Choosing "Include draft records" will include those marked with a Draft publication status in the export. Note: if you do NOT choose this option, any descendants of a draft record will also be excluded, even if they are published.'))));
```
jraddaoui: The label is missing the i18n helper.

apps/qubit/modules/object/actions/exportAction.class.php
```diff
@@ -116,8 +126,64 @@ protected function addField($name)
...
+        } else {
+          $this->form->setWidget('includeDrafts', new sfWidgetFormInputHidden(array('label' => false)));
+        }
+        $this->form->setDefault('includeDrafts', false);
```
jraddaoui: Cosmetic: similar issues than above with `else` and `break`.

apps/qubit/modules/object/actions/exportAction.class.php
```diff
@@ -161,16 +227,17 @@ protected function doBackgroundExport($request)
 
     $options = array(
       'params' => array('fromClipboard' => true, 'slugs' => $this->context->user->getClipboard()->getAll()),
-      'current-level-only' => 'on' !== $request->getParameter('includeDescendants'),
-      'public' => 'on' !== $request->getParameter('includeDrafts'),
+      'current-level-only' => !$this->descendantsIncluded,
+      'public' => $this->draftsIncluded,
```
jraddaoui: Should this be `!$this->draftsIncluded`?

apps/qubit/modules/object/actions/exportAction.class.php
```diff
@@ -161,16 +227,17 @@ protected function doBackgroundExport($request)
...
       'objectType' => $this->objectType,
-      'levels' => $levelsOfDescription,
-      'name' => $this->context->i18n->__('CSV export')
+      'levels' => $levelsOfDescription
     );
 
-    if ('xml' === $this->format)
-    {
-      $options['name'] = $this->context->i18n->__('XML export');
+    $options['name'] = $this->context->i18n->__( ('xml' === $this->format ? 'XML export' : 'CSV export' ) );
```
jraddaoui: Cosmetic: no need for spaces between parenthesis.

apps/qubit/modules/object/actions/exportAction.class.php
```diff
@@ -161,16 +227,17 @@ protected function doBackgroundExport($request)
...
+
+    if($this->includeDigitalObjects){
```
jraddaoui: Cosmetic: space after `if` and new line for the `{`.

apps/qubit/modules/object/actions/exportAction.class.php
```diff
@@ -161,16 +227,17 @@ protected function doBackgroundExport($request)
...
+      $options['name'] = $this->context->i18n->__('%1% and', array('%1%' => sfConfig::get('app_ui_label_digitalobject'))) . ' ' . $options['name'];
```
jraddaoui: `'%1% and %2%'` and including the name in the array will probably be clearer for translators.

apps/qubit/modules/object/actions/exportAction.class.php
```diff
@@ -182,16 +249,18 @@ protected function doBackgroundExport($request)
...
 
-      if ($exportPreviewSupported && !call_user_func(array($this->getJobNameString(), 'exportOrCheckForResults'), $options))
+      if ($exportPreviewSupported && !call_user_func(array($jobName, 'exportOrCheckForResults'), $options))
       {
         throw new sfException('Export complete - no records were exported.');
       }
       else
       {
-        $job = QubitJob::runJob($this->getJobNameString(), $options);
+        $job = QubitJob::runJob($jobName, $options);
```
jraddaoui: Nice!

apps/qubit/modules/object/actions/exportAction.class.php
```diff
@@ -254,4 +323,125 @@ private function getJobNameString()
...
+   * Build an array of digital object ids if any are attached to clipboard items
+   * and current user has permission to view masters
+   *
+   * Sets class property:digitalObjects -> array of digital object ids
+   *
+   * Return a boolean class property if any objects are available for export
+   *
+   * @return boolean:digitalObjectsAvailable
+   */
+  protected function getDigitalObjects()
```
jraddaoui: Cosmetic: with the current implementation, I think it makes more sense to call it `loadDigitalObjects()`, do not return and check the global vars outside after it's called.

apps/qubit/modules/object/actions/exportAction.class.php
```diff
@@ -254,4 +323,125 @@ private function getJobNameString()
...
+  {
+    // set default value
+    $this->digitalObjectsAvailable = false;
+    
+    // prepare array for  digital object ids
+    $this->digitalObjects = array();
+    
+    // process if export option is set and this is a description or actor export
+    if (sfConfig::get('app_clipboard_export_digitalobjects_enabled') && ('informationObject' == $this->objectType || 'actor' == $this->objectType))
```
jraddaoui: It's better to use a default when possible (just in case):

```php
sfConfig::get('app_clipboard_export_digitalobjects_enabled', false);
```

apps/qubit/modules/object/actions/exportAction.class.php
```diff
@@ -254,4 +323,125 @@ private function getJobNameString()
...
+   *
+   * Sets class property:digitalObjects -> array of digital object ids
+   *
+   * Return a boolean class property if any objects are available for export
+   *
+   * @return boolean:digitalObjectsAvailable
+   */
+  protected function getDigitalObjects()
+  {
+    // set default value
```
jraddaoui: Cosmetic: uppercase first letter in new line comments (we're very inconsistent with this).

apps/qubit/modules/object/actions/exportAction.class.php
```diff
@@ -254,4 +323,125 @@ private function getJobNameString()
...
+    if (sfConfig::get('app_clipboard_export_digitalobjects_enabled') && ('informationObject' == $this->objectType || 'actor' == $this->objectType))
+    {
+      // get clipboard objects
+      $criteria = new Criteria;
+      // filter on clipboard slugs
+      $criteria->add(QubitSlug::SLUG, $this->context->user->getClipboard()->getAll(), Criteria::IN);
+      switch ($this->objectType)
+      {
+        case 'informationObject':
+          $criteria->addJoin(QubitInformationObject::ID, QubitDigitalObject::OBJECT_ID);
```
jraddaoui: This join is unnecessary.

apps/qubit/modules/object/actions/exportAction.class.php
```diff
@@ -254,4 +323,125 @@ private function getJobNameString()
...
+          // hide drafts if necessary
+          if(!$this->draftsIncluded)
+          {
+            $criteria = QubitAcl::addFilterDraftsCriteria($criteria);
+          }
+          $items = QubitInformationObject::get($criteria);
+          break;
+
+        case 'actor':
+          $criteria->addJoin(QubitActor::ID, QubitDigitalObject::OBJECT_ID);
```
jraddaoui: This join is unnecessary.

apps/qubit/modules/object/actions/exportAction.class.php
```diff
@@ -254,4 +323,125 @@ private function getJobNameString()
...
+          // get master object
+          $digitalObject = $item->digitalObjectsRelatedByobjectId[0];
+          
+          // permission check system modified from /apps/modules/digitalobject/actions/viewAction.class.php
+          // if we need to add in check for images only, then use:
+          // $digitalObject->isImage() or $digitalObject->isWebCompatibleImageFormat()
+          // ----------
+          // Do appropriate ACL check(s). Master copy of text objects are always allowed for reading
+          // QubitActor does not have a ACL check for readmaster - so only enable for authenticated users.
+          if ( QubitTerm::TEXT_ID == $obj->mediaTypeId
```
jraddaoui: Shouldn't it be `$digitalObject->mediaTypeId `?

apps/qubit/modules/object/actions/exportAction.class.php
```diff
@@ -254,4 +323,125 @@ private function getJobNameString()
...
+          
+          // permission check system modified from /apps/modules/digitalobject/actions/viewAction.class.php
+          // if we need to add in check for images only, then use:
+          // $digitalObject->isImage() or $digitalObject->isWebCompatibleImageFormat()
+          // ----------
+          // Do appropriate ACL check(s). Master copy of text objects are always allowed for reading
+          // QubitActor does not have a ACL check for readmaster - so only enable for authenticated users.
+          if ( QubitTerm::TEXT_ID == $obj->mediaTypeId 
+              || (
+                $digitalObject->masterAccessibleViaUrl()
```
jraddaoui: Should this check affect text DO and information obj DO too?

apps/qubit/modules/object/actions/exportAction.class.php
```diff
@@ -254,4 +323,125 @@ private function getJobNameString()
...
+          // if we need to add in check for images only, then use:
+          // $digitalObject->isImage() or $digitalObject->isWebCompatibleImageFormat()
+          // ----------
+          // Do appropriate ACL check(s). Master copy of text objects are always allowed for reading
+          // QubitActor does not have a ACL check for readmaster - so only enable for authenticated users.
+          if ( QubitTerm::TEXT_ID == $obj->mediaTypeId 
+              || (
+                $digitalObject->masterAccessibleViaUrl() 
+                && ('actor' == $this->objectType && $this->context->user->isAuthenticated() && QubitAcl::check($item, 'read')) 
+                || ('informationObject' == $this->objectType && (QubitAcl::check($item, 'readMaster') && QubitGrantedRight::checkPremis($item->id, 'readMaster') && !$this->isCopyrightedMaster($item->id)))
```
jraddaoui: Cosmetic: the way the parenthesis and indentation are used in here confuses a bit, making you believe the initial && and || above belong to the same group when the don't.

apps/qubit/modules/object/actions/exportAction.class.php
```diff
@@ -254,4 +323,125 @@ private function getJobNameString()
...
+              )
+          )
+          {
+            // add master image id to array
+            $this->digitalObjects[] = $digitalObject->id;
+          }
+        }
+      }
+    }
+    if (!empty($this->digitalObjects))
```
jraddaoui: Cosmetic: empty line before control statement.

apps/qubit/modules/object/actions/exportAction.class.php
```diff
@@ -254,4 +323,125 @@ private function getJobNameString()
...
+    {
+      $this->digitalObjectsAvailable = true;
+    }
+    return $this->digitalObjectsAvailable;
```
jraddaoui: Cosmetic: if you keep the `return`, empty line before it.

apps/qubit/modules/object/actions/exportAction.class.php
```diff
@@ -254,4 +323,125 @@ private function getJobNameString()
...
+  }
+
+  /**
+   * Check if digital object (master) has an active copyright statement
+   * modified from /apps/modules/digitalobject/actions/viewAction.class.php
+   *
+   * @return boolean
+   */
+  protected function isCopyrightedMaster($objId)
```
jraddaoui: I think this is a nice to have in the `QubitDigitalObject` model class as a public static function. And maybe reuse it from /apps/modules/digitalobject/actions/viewAction.class.php too.

apps/qubit/modules/object/actions/exportAction.class.php
```diff
@@ -254,4 +323,125 @@ private function getJobNameString()
...
+      $objId,
+      QubitTerm::RIGHT_ID,
+      QubitGrantedRight::CONDITIONAL_RIGHT,
+      QubitTerm::RIGHT_BASIS_COPYRIGHT_ID));
+
+    if (false === $r || !isset($r->has))
+    {
+      throw new sfException('Unexpected error');
+    }
+    return ( $r->has ? true : false );
```
jraddaoui: Cosmetic: empty line before return and you could return `$r->has` directly.

apps/qubit/modules/object/actions/exportAction.class.php
```diff
@@ -254,4 +323,125 @@ private function getJobNameString()
...
+      LIMIT 1) AS has';
+    $r = QubitPdo::fetchOne($sql, array(
+      $objId,
+      QubitTerm::RIGHT_ID,
+      QubitGrantedRight::CONDITIONAL_RIGHT,
+      QubitTerm::RIGHT_BASIS_COPYRIGHT_ID));
+
+    if (false === $r || !isset($r->has))
+    {
+      throw new sfException('Unexpected error');
```
jraddaoui: Apart from a connection error or a malformed database, what could cause this? Not sure if throwing an exception is a good idea in here, maybe just return `true`?

apps/qubit/modules/object/templates/exportSuccess.php
```diff
@@ -17,63 +17,90 @@
...
+        margin-bottom: 15px;
+      }
+      #exportLevels label {
+        font-weight: bold;
+        padding-bottom: 5px;
+      }
+      #exportLevels .export-options-row-l {
+        width: 33%;
+      }
+    </style>
```
jraddaoui: It would be nice to move this to https://github.com/artefactual/atom/blob/qa/2.6.x/plugins/arDominionPlugin/css/less/scaffolding.less.

And, if possible, use a bit of LESS structure, but it's not a problem if it's added as plain CSS. Ideally, we'd use as much of what's available in Bootstrap to reduce it and I think you could achieve most of what is done with the `row` and `span*` classes. Not a huge deal either.

apps/qubit/modules/settings/actions/clipboardAction.class.php
```diff
@@ -45,7 +46,8 @@ public function earlyExecute()
       'clipboard_send_enabled' => '0',
       'clipboard_send_button_text' => $this->i18n->__('Send'),
       'clipboard_send_message_html' => $this->i18n->__('Sending...'),
-      'clipboard_send_http_method' => 'POST'
+      'clipboard_send_http_method' => 'POST',
+      'clipboard_export_digitalobjects_enabled' => '1',
```
jraddaoui: I'd disable it by default but you may want to check with an analyst.

lib/job/arActorCsvExportJob.class.php
```diff
@@ -35,6 +35,9 @@ class arActorCsvExportJob extends arBaseJob
 
   public function runJob($parameters)
   {
+    // Look for any digital object ids to export and initialize this->digitalObjects
+    $this->digitalObjects = ( array_key_exists('digitalObjects', $parameters) && !empty($parameters['digitalObjects'] ) ? $parameters['digitalObjects'] : array());
```
jraddaoui: Cosmetic: avoid spaces inside and directly after/before the parenthesis.

lib/job/arBaseJob.class.php
```diff
@@ -284,34 +284,94 @@ private function createJobsDownloadsDirectory()
...
+                {
+                  // filename not used yet - add to tracker
+                  $fileNames[$fileName] = 0;
+                }
+                else
+                {
+                  // filename has been used - increment counter and add brackets
+                  $fileNames[$fileName]++;
+                  $doPathInfo = pathinfo( $doPath );
+                  $fileName = $doPathInfo['filename'] . ' (' . $fileNames[$fileName] . ').' . $doPathInfo['extension'];
```
jraddaoui: Nicely thought! I'd prefer `filename_n.ext` instead of spaces and parenthesis.

lib/job/arBaseJob.class.php
```diff
@@ -284,34 +284,94 @@ private function createJobsDownloadsDirectory()
...
+                try
+                {
+                  $zip->addFile($doPath, $fileName);
+                }
+                catch (Exception $e)
+                {
+                  if ($this->user->isAdministrator())
+                  {
+                    $errors[] = ('Exception: '.$e->getMessage());
+                  }
```
jraddaoui: Why the administrator check here and not in the previous two possible errors?

jraddaoui (changes requested):
This is looking really nice @PeteVox, great work!

I have added quite a few comments bellow, most of them cosmetic and expected in a first contribution. We're working on improving this with a syntax checker/fixer in the near future (https://projects.artefactual.com/issues/13317).

There are a couple of important points about the DO processing in the request cycle and the JS code that interacts with the export form, but I'll add this feedback now and address those later in a different comment.

Thanks!!

jraddaoui:
Hi @PeteVox,

Sorry for the delay in the follow-up, the other two points I wanted to consider:

**DO checks in the request cycle:**

It won't be the most usual case, but we mentioned the other day the case of having a big amount of resources in the clipboard and having to check the existence of DOs and their permissions in the request cycle only to determine if the check-box is needed. My suggestion would be move those checks to the background process and always show the check-box, we do the same for the other check-boxes and this way you'll only need to move the code to the job. Please, check if that's possible with the assigned analyst and if it's not, I'd suggest that you break the loop as soon as you find an available DO, to avoid looping over all resources. This could get tricky if you want to avoid duplicated code and that's one of the reasons of why I prefer to just not check in the request cycle.

**Default `objectType` and involved Javascript:**

I've noticed that the "object/export" URL is open and that you can access it without and `objectType` parameter. I'm not even sure if that was working properly before, AFAIK you can access the URL but there is no actual link in AtoM without the parameter. Nevertheless, it's not working properly with the new behavior and I understand that you are going to make changes in there to make some check-boxes enabled/disabled based in their value. I'd appreciate if you can take a look at this issue, it may be as easy as adding a default value when the parameter is obtained to avoid the multiple options in the `objectType` field:

https://github.com/artefactual/atom/blob/qa/2.6.x/apps/qubit/modules/object/actions/exportAction.class.php#L39

```php
$this->objectType = $request->getParameter('objectType', 'informationObject');
```

Bests,
Radda.

apps/qubit/modules/object/actions/exportAction.class.php
PeteVox: Have a couple of switches in the revised code (to be added to repo) that validate and set some default parameters. After you've been able to take a look at the revised code, no problem at all to convert to if/then if you prefer.

apps/qubit/modules/object/actions/exportAction.class.php
PeteVox: This now moved to _arBaseJob.class.php_ as part of background processing and does now return the valid digital object set. Will keep existing for the moment if ok - but no problem to change after you've been able to review the updated version.

apps/qubit/modules/object/actions/exportAction.class.php
PeteVox: Absolutely. Fixed.

apps/qubit/modules/object/actions/exportAction.class.php
PeteVox: Have simplified the logic in latest code. Still not sure the layout of parenthesis is ideal, but can edit after you've been able to see it.

apps/qubit/modules/object/actions/exportAction.class.php
PeteVox: Done.

apps/qubit/modules/object/actions/exportAction.class.php
PeteVox: Not sure about the need for this check at all - but copied from original version in `QubitDigitalObject`. Have updated in revised code.

apps/qubit/modules/object/templates/exportSuccess.php
PeteVox: This has been changed completely in revised code - but still using styles inline as they only apply to the Export page. Agree that ideally, they could be rolled in to the Dominion LESS files but perhaps we can assess the impact of doing that after QA? If we were to add to Dominion, then might have implications for users with custom themes?

apps/qubit/modules/settings/actions/clipboardAction.class.php
PeteVox: Agree. Have changed to disabled by default, but will review with analyst.

lib/job/arBaseJob.class.php
PeteVox: Have expanded this in revised code. Administrators should see full Exception (message), but any other user will see some more helpful text. To be reviewed.

apps/qubit/modules/object/templates/exportSuccess.php
jraddaoui: Custom themes should extend Dominion and get this changes by default. There is also the possibility of a custom theme overwriting this template so it's better to do this changes in the existing LESS files from Dominion. And, in general, it's better to have separation of concerns.

apps/qubit/modules/default/actions/alertsComponent.class.php
```diff
@@ -43,22 +43,34 @@ public function execute($request)
...
+              array(
+                '%1%' => sprintf('<a href="%s">', sfConfig::get('app_siteBaseUrl') .'/'. $job->downloadPath),
+                '%2%' => '</a>',
+                '%3%' => hr_filesize(filesize($job->downloadPath))
+              )
+            );
+          }
+          else
+          {
+            $message .= ' ' . $this->context->i18n->__('%1%Refresh the page%2% for progress updates.',
```
jraddaoui, line 67: Nice catch!

apps/qubit/modules/object/templates/exportSuccess.php
```diff
@@ -17,63 +17,105 @@
...
-                <label>
-                  <input name="includeDrafts" type="checkbox"/>
-                  <?php echo __('Include draft records') ?>
-                </label>
+                if (!$form->includeAllLevels->isHidden())
+                {
+                  echo $form->includeAllLevels->renderRow();
+                }
+
+              ?>
```
jraddaoui: Cosmetic: use the same templates format ...
```
<?php if (...): ?>
  <?php echo ...?>
<?php endif; ?>
```

apps/qubit/modules/object/actions/exportAction.class.php
```diff
@@ -69,54 +157,142 @@ protected function addField($name)
...
+        if ($this->showOptions && $this->context->user->isAuthenticated())
+        {
+          $this->form->setWidget('includeDrafts', new sfWidgetFormInputCheckbox(array('label' => __('Include draft records'))));
+          $this->helpMessage['30'] = __('Choosing "Include draft records" will include those marked with a Draft publication status in the export. Note: if you do NOT choose this option, any descendants of a draft record will also be excluded, even if they are published.');
+          $this->form->setDefault('includeDrafts', true);
+        }
+        else
+        {
+          $this->form->setWidget('includeDrafts', new sfWidgetFormInputHidden(array('label' => false)));
+        }
```
jraddaoui: Are this hidden fields really needed? Could we just use `isset()` in the templates to check the field existence?

apps/qubit/modules/object/actions/exportAction.class.php
```diff
@@ -23,33 +23,121 @@ class ObjectExportAction extends DefaultEditAction
...
+    $this->response->addJavaScript('exportOptions', 'last');
+
     $this->title = $this->context->i18n->__('Export');
 
+    if (0 != count($this->helpMessage))
+    {
+      // Convert sorted help message array to string
+      ksort($this->helpMessage);
+      $this->helpMessage = '<p>' . implode('</p><p>', $this->helpMessage) . '</p>';
+    }
```
jraddaoui: What's the reason for this help to this help messages manipulation? I think it's clearer to have them associated to a field as they were or just add them in the template.

apps/qubit/modules/object/templates/exportSuccess.php
```diff
@@ -17,63 +17,105 @@
...
-            <div class="alert alert-info generic-help animateNicely">
-              <p><?php echo __('Choosing "Include descendants" will include all lower-level records beneath those currently on the clipboard in the export.') ?></p>
-              <?php if ($sf_user->isAuthenticated()): ?>
-                <p><?php echo __('Choosing "Include draft records" will include those marked with a Draft publication status in the export. Note: if you do NOT choose this option, any descendants of a draft record will also be excluded, even if they are published.') ?></p>
+                if (!$form->includeDrafts->isHidden())
+                {
+                  echo $form->includeDrafts->renderRow();
+                }
+
+              ?>
```
jraddaoui: Same here.

apps/qubit/modules/object/templates/exportSuccess.php
```diff
@@ -17,63 +17,105 @@
...
+              <?php if (!empty($helpMessage)): ?>
+              <div class="alert alert-info generic-help animateNicely">
+                <?php echo(sfOutputEscaper::unescape($helpMessage)); ?>
+              </div>
               <?php endif; ?>
```
jraddaoui: If you refactor the messages manipulation, even if there shouldn't be an issue considering its content, it's better to avoid the unescaping.

js/exportOptions.js
```diff
@@ -4,26 +4,32 @@
...
   var ExportOptions = function (element)
     {
-
       this.$element = $(element);
 
+      this.$type = this.$element.find('select[name="type"]');
+      this.$formatSelect = this.$element.find('select[name="format"]');
+      this.formatDescriptionCsv = this.$formatSelect.find('option[value="csv"]').text();
+      this.formatDescriptionXml = this.$formatSelect.find('option[value="xml"]').text();
+      this.$includeAllLevels = this.$element.find('input[name="includeAllLevels"]:not([type="hidden"])');
```
jraddaoui: Similar in here, you could rely on the existence instead of the type.

js/exportOptions.js
```diff
@@ -33,46 +39,119 @@
...
-      this.setDefaults();
+      this.resetLevelsOptions();
     },
 
     listen: function()
     {
-      this.$includeDescendants.on('change', $.proxy(this.onIncludeDescendantsChange, this));
-      this.$includeAllLevels.on('change', $.proxy(this.onIncludeAllLevelsChange, this));
-      this.$objectType.on('change', $.proxy(this.onObjectTypeChange, this));
+      this.$type.on('change', $.proxy(this.onObjectTypeChange, this));
```
jraddaoui, line 47: I think this is no longer possible with the new changes. Otherwise, we may be duplicating the logic between the controller and the Javascript.

js/exportOptions.js
```diff
@@ -33,46 +39,119 @@
...
+        }
+        this.setDescendantsState(false);
+        this.setDigitalObjectsState(true);
       }
       else
       {
-        this.resetLevelsOptions();
+        this.setDescendantsState(true);
+      }
+    },
```
jraddaoui: Cosmetic: this two listeners could have been implemented in one, adding an extra argument to the function to know the direction of the switch.

js/exportOptions.js
```diff
@@ -33,46 +39,119 @@
...
+        {
+          o.removeClass('muted');
+        }
+        else
+        {
+          this.$includeDigitalObjects.attr('checked', false);
+          o.addClass('muted');
+        }
       }
     },
```
jraddaoui: Cosmetic: similar in here.

js/exportOptions.js
```diff
@@ -81,71 +160,28 @@
...
-
-    resetExportOptionsPanel: function ()
-    {
-      this.$includeDescendants.attr('checked', false);
-      this.$includeDrafts.attr('checked', false);
-      this.$includeAllLevels.parent().hide();
-      this.$levelDiv.addClass('hidden');
-      this.$includeAllLevels.attr('checked', true);
-      this.$levelSelect.val('');
+      window.location.href = url;
```
jraddaoui, line 186: Is this switch still possible?

lib/job/arBaseJob.class.php
```diff
@@ -284,34 +284,120 @@ private function createJobsDownloadsDirectory()
    * Create ZIP file from results
    *
    * @param string  Path of file to write CSV data to
+   * @param string  Optional array of digital object ids to include
```
jraddaoui: Cosmetic: no longer an array.

lib/job/arBaseJob.class.php
```diff
@@ -284,34 +284,120 @@ private function createJobsDownloadsDirectory()
...
 
-    if ($success == true)
-    {
-      foreach(scandir($path) as $file)
-      {
-        if (!is_dir($file))
+          // Get permitted digital object ids (if any) and iterate
+          foreach($this->getDigitalObjects() as $id)
+          {
+            if(!$error)
```
jraddaoui: Maybe stop the loop entirely:

```
if ($error)
{
  break;
}

// Otherwise continue
```

lib/job/arBaseJob.class.php
```diff
@@ -284,34 +284,120 @@ private function createJobsDownloadsDirectory()
...
+          // Set error flag
+          $error = false;
 
-    if ($success == true)
-    {
-      foreach(scandir($path) as $file)
-      {
-        if (!is_dir($file))
+          // Get permitted digital object ids (if any) and iterate
+          foreach($this->getDigitalObjects() as $id)
```
jraddaoui, line 317: Cosmetic: missing whitespace after `foreach`.

lib/job/arBaseJob.class.php
```diff
@@ -284,34 +284,120 @@ private function createJobsDownloadsDirectory()
...
+                    else {
+                      $errors[] = $this->i18n->__(
+                        'Sorry, but there was an error locating a digital object (#%1%). ' . 
+                        'This has prevented any further digital objects from being exported. ' .
+                        'Please contact an administrator.' , 
+                        array(
+                          '%1%' => $id
+                        )
+                      );
+                      $error = true;
```
jraddaoui: Should this be moved outside the `else` so it also acts on errors for admins?

lib/job/arBaseJob.class.php
```diff
@@ -401,4 +487,87 @@ protected function canBeFullyExecuted()
...
+
+      // Filter on clipboard slugs
+      $criteria->add(QubitSlug::SLUG, $this->params['params']['slugs'], Criteria::IN);
+
+      switch ($this->params['objectType'])
+      {
+        case 'informationObject':
+          $criteria->addJoin(QubitInformationObject::ID, QubitSlug::OBJECT_ID);
+          // Hide drafts if necessary
+          if(!$this->params['public'])
```
jraddaoui: Should this be the other way around? If it's public, then filter drafts?

lib/job/arBaseJob.class.php
```diff
@@ -401,4 +487,87 @@ protected function canBeFullyExecuted()
...
+                && QubitAcl::check($item, 'read')
+              )
+              || (
+                'informationObject' == $this->params['objectType']
+                && QubitAcl::check($item, 'readMaster')
+                && QubitGrantedRight::checkPremis($item->id, 'readMaster')
+                && !$digitalObject->hasConditionalCopyright($item->id)
+              )
+            )
+          )
```
jraddaoui: Really nice, quite and `if`!!

lib/job/arRepositoryCsvExportJob.class.php
```diff
@@ -55,7 +55,13 @@ public function runJob($parameters)
 
     // Compress CSV export files as a ZIP archive
     $this->info($this->i18n->__('Creating ZIP file %1', array('%1' => $this->getDownloadFilePath())));
-    $success = $this->createZipForDownload($tempPath);
+    $errors = $this->createZipForDownload($tempPath);
+
+    if (!empty($errors))
+    {
+      $this->error($this->i18n->__('Failed to create ZIP file.') . ' : ' . implode(' : ', $errors));
+      return false;
```
jraddaoui, line 64: Cosmetic: empty line before return.

lib/job/arInformationObjectXmlExportJob.class.php
```diff
@@ -38,6 +38,9 @@ public function runJob($parameters)
   {
     $this->params = $parameters;
 
+    // Do we include digital objects?
+    $this->includeDigitalObjects = array_key_exists('includeDigitalObjects', $parameters) && $parameters['includeDigitalObjects'];
```
jraddaoui: Cosmetic: could we move this to `arBaseJob` using `$this->params` inside `createZipForDownload`? That way it could be removed from all export jobs and not passed to that function.

jraddaoui (changes requested):
Thanks @PeteVox!

This is looking great and it's pretty close to merge. I've added a few comments below, please let me know your thoughts about them before spending a lot of time as most of them are not really important.

My biggest concerns are:

- Moving the styles to the LESS files.
- Reduce form logic (hidden fields, duplicated logic in JS/controller/templates) and avoid messages array if possible.
- Drafts condition on export query.
- All errors for admins and only one for public users?

apps/qubit/modules/object/templates/exportSuccess.php
PeteVox: Have updated this. Added to Dominion scaffolding.

apps/qubit/modules/object/actions/exportAction.class.php
PeteVox: Could do that Radda - but was trying to avoid conditional logic in the templates where possible. No problem to remove the hiddens and go with if/then in template if preferred?

apps/qubit/modules/object/actions/exportAction.class.php
PeteVox: Did initially add them to the relevant field (properties) in the first version, but Analyst team preferred the single pop-up help panel - for most fields. We need to build the help message conditionally depending on view/permissions/settings and I usually stack this type of thing in an array, rather than concatenating a string. No problem to change this to use string if preferred.

apps/qubit/modules/object/actions/exportAction.class.php
PeteVox: Didn't fully realise that isset() is ok to use here - thought that non-defined fields might impact any validators. Have updated in form and action class. Much cleaner.

apps/qubit/modules/object/templates/exportSuccess.php
PeteVox: Absolutely agree about the unescaping - but had to do something as the help message contains `<p></p>` tags. Did some more digging and found an alternative method using `$sf_data`.

js/exportOptions.js
PeteVox: Think we're ok here. The `.js` doesn't change visibility of fields in the options panel any longer - this all done by the controller on page refresh depending on permissions, but also on whether there are valid objects of the selected type on the clipboard that can be exported. This logic also fixes attempts to directly access (and manipulate parameters of) the export page.

apps/qubit/modules/object/actions/exportAction.class.php
jraddaoui: I think you could render the help messages in a different container, but I see your point. I'd prefer to build the array without keys (you can modify the order of the $names array to get the messages order needed) and the use a foreach in the template appending the paragraph tags, that way there's no need to access sf_data nor unescaping, and I think it's a bit more clear in the code.

apps/qubit/modules/object/templates/exportSuccess.php
jraddaoui: A lot nicer, thanks! But see my comment above.

js/exportOptions.js
jraddaoui: Sounds good, thanks @PeteVox. This is likely going to change after moving the clipboard content to the browser local storage, so it's more than okay as it is.

js/exportOptions.js
jraddaoui: @PeteVox, if you're not on this, please ignore. As mentioned, this is likely going to change after moving the clipboard content to the browser local storage.

js/exportOptions.js
jraddaoui: Same, ignore this part.

js/exportOptions.js
PeteVox: Have combined these in dev code.

js/exportOptions.js
PeteVox: This just refreshes the export page, but passes type so that controller can rebuild the options panel depending on permissions etc.

js/exportOptions.js
PeteVox: Would prefer to leave as they are for the moment if that's ok - until we've been able to fully discuss the UI with the Analyst team. Once we know for sure if we need to make any significant changes to the logic that's applied to the visibility of the digital objects check box, can streamline the functions.

lib/job/arBaseJob.class.php
PeteVox: Have actually removed the error flags and using breaks instead as per suggestion. MUCH cleaner.

lib/job/arBaseJob.class.php
PeteVox: Nice catch. Fixed.

lib/job/arInformationObjectXmlExportJob.class.php
PeteVox: Absolutely. Fixed. Much cleaner.

apps/qubit/modules/object/actions/exportAction.class.php
PeteVox: Done.

apps/qubit/modules/object/actions/exportAction.class.php
PeteVox: Have updated in repo. Removed keys and used foreach in template instead of constructing in controller. Had to still use `$sf_data->getRaw` as the array is `sfOutputEscaperArrayDecorator` object in template. Can update this if there's a better way to iterate.

jraddaoui: Awesome, thanks @PeteVox!

jraddaoui (approved):
Thanks @PeteVox!

I've copied the final commit message to the PR title and description. Commit body is not wrapped at 72 chars but not a big deal, I'll merge.

PR #1156, opened 2020-07-09
Clear statements in nested set builds, refs #13357

jraddaoui:
Clear the statements array ($this->rows) in between tables rebuild
to avoid mixing previous updates in different transactions.


sbreker (approved):
Looks great @jraddaoui !

PR #1157, opened 2020-07-09
Fix multi-row JS/date picker issue, refs #13315

mcantelon:
There were issues with the date picker being used as a field with the
multi-row JS. The date picker generates table HTML that was interfering
with the multi-row JS and when the multi-row JS cloned rows the date
picker in the new row wasn't being activated. Fixed issues.


djjuhasz (approved):
Looks good 👍

PR #1159, opened 2020-07-15
Reuse autocomplete iframes, refs #13366

djjuhasz:
- Don't create a new iframe every time an autocomplete "onchange" event is
triggered
- Remove obsolete related resource forms from submission queue


mcantelon (approved):
Code looks good! 👍

PR #1160, opened 2020-07-17
Setup Varnish in Docker environment, refs #11556

jraddaoui:
Extend Docker Compose environment with a Varnish container. Use a read
only volume for a basic configuration file to point to the Nginx
container and cache all (removing cookies and adding X-Cache headers).

PR #1161, opened 2020-07-19
Multicultural Varnish cache support, refs #13389

mcantelon:
Changes to allow, with the correct Varnish configuration, content to be
cached in non-default cultures.

### DIFF
--- a/apps/qubit/config/qubitConfiguration.class.php
+++ b/apps/qubit/config/qubitConfiguration.class.php
@@ -30,6 +30,11 @@ class qubitConfiguration extends sfApplicationConfiguration
     return str_ireplace('</head>', javascript_tag('jQuery.extend(Qubit, '.json_encode(array('relativeUrlRoot' => sfContext::getInstance()->request->getRelativeUrlRoot())).');').'</head>', $content);
   }
 
+  public function listenToChangeCultureEvent(sfEvent $event)
+  {
+    setcookie('atom_culture', $event['culture']);
+  }
+
   /**
    * @see sfApplicationConfiguration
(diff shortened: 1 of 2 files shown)


jraddaoui (commented):
Thanks @mcantelon,

I didn't know about that event and the code changes look really good. I wonder what is the configuration needed in Varnish to make this work. Would you need to allow the set-cookie header in the responses to requests using the `sf_culture` parameter? And then, would you use the `HTTP_X_ATOM_CULTURE` request header to generate the cache hash based on the culture? How is that header sent on each request?

mcantelon:
Hi @jraddaoui ... here are the VCL changes needed for it to work:

https://projects.artefactual.com/issues/13389#note-3

Basically the Varnish changes do the following:

1. Let requests with `sf_culture` specified as a GET parameter through (to AtoM can set the `atom_culture` cookie)
2. Don't remove `set-cookie` response headers for responses to requests with `sf_culture` set
3. Use the cookie value to set the X-Atom-Culture and incorporate its value into the hash key

jraddaoui (approved):
Nicely thought @mcantelon, great work!

jraddaoui:
@mcantelon, if it's not a low of work and you're planing to give it a try to https://github.com/artefactual/atom/pull/1160, it would be amazing if you implement those changes in the Varnish config from that PR.

jraddaoui:
Note to self, read ticket updates before review, sorry Mike :P

mcantelon:
@jraddaoui Thanks Radda! I'll give #1160 a try!

PR #1162, opened 2020-07-21
Fix actor relation ES update job, refs #13391

mcantelon:
Fixed issue with actor relation search index update job. Relationship
type data wasn't being updated.


jraddaoui (approved):
:+1:

PR #1166, opened 2020-07-30
Move clipboard to browser's storage, refs #11556

jraddaoui:
Move clipboard content and related exports association to the browser's
local storage. Use Ajax requests to interact with the back-end in those
processes where the clipboard content is needed. Avoid using flash
alerts and move them to the JS code. Use a mix of POST and GET requests
and HTML and JSON responses for convenience and given the current
implementation. Secure exports checks and job deletion by using tokens
instead of job ids to relate the jobs with the user in the front-end.
Cleanup and reorganize code. Minor style fixes.

See commit messages for more information.


jraddaoui:
Hi @djjuhasz, this is ready for code review.

I'd suggest to go commit by commit as there are some re-organization that will complicate jumping to the full diff, and you may find useful the comments from those commits messages (I'll reflect them in the squash).

As noted in the description, it's not the best implementation. One could say there are some anti-patterns here, mixing GET and POST requests and HTML and JSON responses. It's just convenient to do it like that to simplify the Varnish config and avoid edge cases, and also considering the current implementation and the time constraints for this changes.

One thing that I'd have liked to do is to move all the clipboard code to its own module or even a configurable plugin, now that it's not directly related with the user session, but I didn't want to mess with migrations (for the menus, etc.) and settings considering this has to be cherry-picked to a stable branch, so maybe something for the future.

jraddaoui:
I've added some extra notes to https://projects.artefactual.com/issues/11556#note-6.

js/clipboard.js
```diff
@@ -10,6 +10,15 @@
       this.$menuClearAll = this.$element.find('li#node_clearClipboard a');
       this.$toggleButtons = $('button.clipboard');
 
+      this.storage = localStorage;
```
djjuhasz: Wow, this is all it takes to use local storage, eh? Very cool. 😮

jraddaoui: Yes, that's it :)

And we could replace it in this line by `sessionStorage` if we prefer that one.

js/clipboard.js
```diff
@@ -28,20 +28,59 @@
```
djjuhasz: Is this a common javascript constructor pattern?  Steve used the same pattern in https://github.com/artefactual/atom/pull/1167/commits/f8011e73d89a791c56e5ac96242b4b4d28502e1b.  If it is a common pattern, do you have a link to a description of the pattern anywhere?

I don't understand the advantage of assigning the constructor function to a variable (i.e. `var Clipboard = function (element))` , then declaring a prototype on that variable (`Clipboard.prototype = {}`), then assigning the Clipboard variable as the prototype.constructor over the [Constructor with prototypes](https://www.oreilly.com/library/view/learning-javascript-design/9781449334840/ch09s01.html#constructors-with-prototypes) pattern that I've seen from multiple sources.

js/clipboard.js
```diff
@@ -165,6 +166,56 @@
...
+      {
+        data.repository_slugs = JSON.stringify(this.items['repository']);
+      }
+
+      // Show sending alert and assign it to a variable
+      var $sendingAlert = this.showAlert($sendButton.data('message'), 'alert-info');
+
+      $.ajax({
+        url: $sendButton.data('url'),
+        type: 'get',
```
jraddaoui: Fixed in a later commit to use the `data-method` from the button.

apps/qubit/modules/user/actions/clipboardSaveAction.class.php
```diff
@@ -21,37 +21,36 @@ class UserClipboardSaveAction extends sfAction
...
-      else
-      {
-        $this->saveClipboard($allSlugs, $password);
+      $this->saveClipboard($slugs, $password);
 
-        // Inform user of progress
-        $messageText = 'Clipboard saved. Clipoard ID is <b>%1%</b>. Please write this number down. ';
-        $messageText .= 'When you want to reload this clipboard in the future, open the Clipboard menu, ';
-        $messageText .= 'select <a href="%2%">Load clipboard</a>, and enter this number in the Clipboard ID field.';
+      $messageText = 'Clipboard saved. Clipoard ID is <b>%1%</b>. Please write this number down. ';
```
djjuhasz: Typo: "_Clipoard_ ID is... "

apps/qubit/modules/user/actions/clipboardSaveAction.class.php
```diff
@@ -21,37 +21,36 @@ class UserClipboardSaveAction extends sfAction
 {
   public function execute($request)
   {
-    $allSlugs = $this->context->user->getClipboard()->getAllByClassName();
+    $slugs = $request->getPostParameter('slugs', []);
```
djjuhasz: Does AtoM do any automatic sanitation of POST parameters?  I'm not sure if there is a potential for exploits here via a crafted POST.

apps/qubit/modules/user/actions/clipboardSaveAction.class.php
```diff
@@ -100,13 +99,13 @@ private function saveClipboard($allSlugs, $password)
...
-    foreach($allSlugs as $className => $slugs)
+    foreach($allSlugs as $type => $slugs)
     {
       foreach($slugs as $slug)
       {
         $item = new QubitClipboardSaveItem;
         $item->saveId = $save->id;
-        $item->itemClassName = $className;
+        $item->itemClassName = 'Qubit'.ucfirst($type);
         $item->slug = $slug;
```
djjuhasz: I think it would be good to validate that slug exists before saving it to the database, it looks to me like any arbitrary value passed to the method will be saved as-is.

djjuhasz: Actually, maybe the best solution is to add a fk constraint on `clipboard_save_item.slug` to `slug.slug` so if a resource is deleted it cascade deletes the clipboard item.

apps/qubit/modules/user/actions/clipboardExportAction.class.php
```diff
@@ -131,21 +123,133 @@ public function execute($request)
...
+      'current-level-only' => !$this->descendantsIncluded,
+      'public' => !$this->draftsIncluded,
+      'objectType' => $this->objectType,
+      'levels' => $levelsOfDescription
+    );
+
+    $options['name'] = $this->context->i18n->__('xml' === $this->formatType ? 'XML export' : 'CSV export' );
 
-    if ($request->isMethod('post'))
+    if($this->includeDigitalObjects)
```
djjuhasz: Nitpick: missing space after "if"

apps/qubit/modules/user/actions/clipboardExportAction.class.php
```diff
@@ -229,15 +236,19 @@ public function execute($request)
             '%5%' => '</strong></a>'
           )
         );
-        $message .= '</p><p>';
-        $message .= $this->context->i18n->__(
+        $responseData['success'] .= '</p><p>';
+        $responseData['success'] .= $this->context->i18n->__(
           '%1%Note:%2% AtoM may remove export packages after a period of time to free up storage space. When your export is ready you should download it as soon as possible.',
```
djjuhasz: I think it would be better to remove the &lt;strong&gt; emphasis on the "Note:" text to avoid having to add placeholders for the opening and closing &lt;strong&gt; tags.   The placeholders make localization harder and more error-prone.

apps/qubit/modules/user/actions/clipboardExportAction.class.php
```diff
@@ -210,16 +210,23 @@ public function execute($request)
...
+          // TODO: REMOVE WHEN DONE!!
           $manager = new QubitUnauthenticatedUserJobManager($this->context->user);
           $manager->addJobAssociation($job);
         }
 
-        $message = '<p>';
-        $message .= $this->context->i18n->__(
+        $responseData['success'] = '<p>';
+        $responseData['success'] .= $this->context->i18n->__(
           '%1%Your %2% export package is being built.%3% ' . ($this->context->user->isAuthenticated() ? 'The %4%job management page%5% will show progress and a download link when complete.' : 'Please %4%refresh the page%5% to see progress and a download link when complete.'),
```
djjuhasz: I think this should be broken into two translation strings and the conditional link should be removed to avoid having to add placeholder tags in the i18n text for &lt;strong&gt; tags and the link logic.  The placeholders make localization harder and more error-prone.

apps/qubit/modules/actor/templates/_searchResult.php
```diff
@@ -26,7 +26,7 @@
 
     <p class="title"><?php echo link_to(render_value_inline(get_search_i18n($doc, 'authorizedFormOfName', array('allowEmpty' => false, 'culture' => $culture))), array('module' => 'actor', 'slug' => $doc['slug'])) ?></p>
 
-    <?php echo get_component('object', 'clipboardButton', array('slug' => $doc['slug'], 'wide' => false)) ?>
+    <?php echo get_component('user', 'clipboardButton', array('slug' => $doc['slug'], 'wide' => false, 'type' => $clipboardType)) ?>
```
djjuhasz: I was expecting to see the clipboard functionality moved to it's own `clipboard` method, instead of being a subset of the `user` method, so all "http://example.com/clipboard/" URLs could be excluded from caching.  The way you've implemented it works well, so I debated whether to raise the issue at all, but in the end I'm a bit uncomfortable with the choice to make the cache exception rule specific to clipboard POST requests, and using POST requests when GET would be canonical.  I would prefer to use GET, if possible, for all requests getting data from an endpoint, and reserve POST for requests that are adding data.

apps/qubit/modules/user/actions/clipboardExportAction.class.php
```diff
@@ -131,21 +123,136 @@ public function execute($request)
...
+    try
     {
-      $this->form->bind($request->getPostParameters());
+      $jobName = $this->getJobNameString();
 
-      if ($this->form->isValid())
+      // Preview export before attempting, if job supports this
+      $exportPreviewSupported = method_exists($jobName, 'exportOrCheckForResults');
+
+      if ($exportPreviewSupported && !call_user_func(array($jobName, 'exportOrCheckForResults'), $options))
```
djjuhasz: From my reading of this, I don't think it will actually use the `exportOrCheckResults()` return value at all - it just throws an exception if the return is "falsey".  I don't see the point of running calling a separate method to check if there will be results from the "real" method, unless it's much faster.  In any case I think the name 'exportOrCheckResults()' method name is misleading if it's sole purpose is to check if the export will be empty or not.

apps/qubit/modules/user/actions/clipboardExportCheckAction.class.php
```diff
@@ -0,0 +1,103 @@
...
+      // Add to response data
+      $alerts[] = ['type' => $type, 'message' => $message, 'deleteUrl' => $deleteUrl];
+    }
+
+    $this->response->setHttpHeader('Content-Type', 'application/json; charset=utf-8');
+    $this->response->setStatusCode(200);
+
+    return $this->renderText(json_encode(['alerts' => $alerts, 'missingTokens' => $missingTokens]));
+  }
+}
```
djjuhasz: Nitpick: Missing newline at EOF

apps/qubit/templates/_alerts.php
```diff
@@ -0,0 +1,8 @@
+<?php foreach(array('notice', 'info', 'error', 'success') as $alertType): ?>
+  <?php if ($sf_user->hasFlash($alertType)): ?>
+    <div class="alert alert-<?php echo $alertType ?>">
+      <button type="button" class="close" data-dismiss="alert">&times;</button>
+      <?php echo $sf_user->getFlash($alertType, ESC_RAW) ?>
+    </div>
+  <?php endif; ?>
+<?php endforeach; ?>
```
djjuhasz: Nitpick: missing newline at EOF

js/clipboard.js
```diff
@@ -19,93 +34,335 @@
 
     init: function()
```
djjuhasz, line 33: Overall I love the way you've written the Clipboard script, it's easy to read and understand and the individual methods are nicely factored and it's easy to understand the expected functionality from the method names.  Well done! 👏

djjuhasz (changes requested):
This looks really good overall @jraddaoui.  I've added a bunch of comments but I'm really happy with the overall architecture and the feature is working great! I was surprised by how much functionality there was for the clipboard (I'd forgotten about save/load, and "send" features), good work on updating all of the various nooks and crannies to work with localStorage. :)

apps/qubit/modules/user/actions/clipboardExportAction.class.php
djjuhasz: I also don't know how the `__()` function will render the ternary condition here, but I don't think it will allow translation of the conditional strings.

js/clipboard.js
jraddaoui: I agree with you that it seems unnecessary to have all those functions/constructors with the same name and how everything is created at the bottom. Without knowing a lot about this, I believe some of that is needed to be able to share the Clipboard with other JS files, like [clipboardToggleAll.js](https://github.com/artefactual/atom/blob/qa/2.x/js/clipboardToggleAll.js#L5), but I'll take a look.

apps/qubit/modules/user/actions/clipboardSaveAction.class.php
jraddaoui: I don't think so ... I'll add some validation.

apps/qubit/modules/user/actions/clipboardSaveAction.class.php
jraddaoui: Yeah, great point about validating. I think that should be enough for now for a couple of reasons:

- If we go with a foreign key, I'd prefer to use the `slug.id` and that will require a few more changes.
- The clipboard could get easily out of sync in the previous and this implementation when resources are edited/deleted. Ideally, we'd have a process to reflect those changes in the saves, but also in the current clipboards. This can get complicated and after all, in the processes that actually load the resources (view and export), the missing slugs will be filtered in the query.

apps/qubit/modules/user/actions/clipboardExportAction.class.php
jraddaoui: Yes, this is all quite a mess, I didn't even want to look at it during the main work but I'll tweak it now. Thanks!

apps/qubit/modules/actor/templates/_searchResult.php
jraddaoui: Totally ... I'm not sure if you saw [this comment](https://github.com/artefactual/atom/pull/1166#issuecomment-669286043) above.

I think the convenience of doing it this way overcomes the anti-patterns. One of the edge cases (maybe the only one that we can't control) is the length of the URLs sending the clipboard slugs in get requests. [This](https://stackoverflow.com/a/417184) is a nice Stack Overflow answer about the matter. Also, not a huge deal for these pages, but it's nice to be able to cache those get requests.

About moving it to its own module/plugin, I'd like to do that but, knowing this is going to be cherry-picked to an stable branch, I'm a bit worried about the migrations. I know we have a task now to run individual migrations making them idempotent, but this could still cause issues in the next release upgrade. Although, moving it to it's own module may only require to modify the existing menus, and that's probably not a big deal ATM. I'll check the migrations currently added to that stable branch and the ones we have in qa/2.x to see what can be done.

apps/qubit/modules/user/actions/clipboardExportAction.class.php
jraddaoui: Not really part of this work, but I'll check it out.

js/clipboard.js
jraddaoui: Thank you, @djjuhasz!

js/clipboard.js
jraddaoui: I'm not sure if passing the constructor that way has some benefits [for inheritance](https://developer.mozilla.org/en-US/docs/Learn/JavaScript/Objects/Inheritance#Setting_Teachers_prototype_and_constructor_reference), but in any case we don't need it. Specially the extra prototype and constructor created with `$.fn` at the bottom. All this is not needed for clipboardToggleAll.js either, where the association is made using the element data.

I've removed the unnecessary bits, but It would be great to be able to use [ES6 classes](https://hacks.mozilla.org/2015/07/es6-in-depth-classes/) at some point. I've created [this gist](https://gist.github.com/jraddaoui/85c7b8583af51158e6fabffa773f0890) using them (including current diff) and I'll mention it in SAD, but I guess [IE support](https://caniuse.com/#search=es6) is still a thing.

apps/qubit/modules/user/actions/clipboardExportAction.class.php
jraddaoui: I agree that it's not the easiest to understand but I think it's okay ...

[The method](https://github.com/artefactual/atom/blob/qa/2.x/lib/job/arInformationObjectCsvExportJob.class.php#L133) is called from two places:
- From [here](https://github.com/artefactual/atom/blob/qa/2.x/apps/qubit/modules/object/actions/exportAction.class.php#L349) without path nor job parameters, which avoids the actual export and returns the count. It should be faster and when the count is "falsey" (no items "exported") it will avoid the actual job execution.
- From [here](https://github.com/artefactual/atom/blob/qa/2.x/lib/job/arInformationObjectCsvExportJob.class.php#L47) with the parameters, which performs the export and uses the returned count for logging info.

I can't find a good change to improve readability without code duplication, but I'm open to suggestions.

apps/qubit/modules/user/actions/clipboardExportAction.class.php
jraddaoui: Still not perfect but it should be easy to read now and give some context to translators.

apps/qubit/modules/user/actions/clipboardExportAction.class.php
jraddaoui: Used placeholders with context for translators to maintain the strong tags.

apps/qubit/modules/user/actions/clipboardSaveAction.class.php
jraddaoui: Added validation on clipboard save but the out of sync issue is a bigger problem. I'll add a note in the ticket about it to get feedback from the analysts too.

jraddaoui: https://projects.artefactual.com/issues/11556#note-8

apps/qubit/modules/actor/templates/_searchResult.php
jraddaoui: The migration should be pretty safe, so I've moved all to its own module. It's following other migrations where we look for the menu by its name, which is not the most bullet proof as I just noticed that some of this sub-menus' name can be changed from the GUI. Please, let me know if you want me to tweak that, looking for the old path could be more robust and also update custom menus.

jraddaoui:
Thanks for the feedback @djjuhasz, looking better now :)

djjuhasz (approved):
Looks great @jraddaoui! :shipit:

apps/qubit/modules/user/actions/clipboardExportAction.class.php
djjuhasz: Okay, thanks for checking it out @jraddaoui. :)

apps/qubit/modules/actor/templates/_searchResult.php
djjuhasz: Awesome, thanks!  I'm in favor of migrating the clipboard menus by searching on the old path, if you are willing to do that work, but I'm also okay with leaving it as is.

js/clipboard.js
djjuhasz: Great thanks @jraddaoui. :)  I read the article you linked on ES6 classes and the changes look great.  It is too bad that ES6 classes aren't IE compatible - hopefully we can ditch IE support some time in the near future 🤞

jraddaoui:
@djjuhasz, I've changed the migration to use paths. Also, noticed that clear doesn't have an action anymore, so it'll be changed by `#`, not a big deal since it's watched from the JS code, but better than an un-existing endpoint.

PR #1167, opened 2020-07-31
Dev/issue 13351 download aip

sbreker:
This commit adds Storage Service integration to AtoM for the purpose of downloading AIPs and objects
from within those AIPs. AIP and Object downloading is only available to users in specific user groups - 
by default this is the administrator group only.

Group access is handled by adding groups directly into the security.yml file in the arStorageServicePlugin's  arStorageService module.

### DIFF
--- a/plugins/arStorageServicePlugin/lib/arStorageServiceUtils.class.php
+++ b/plugins/arStorageServicePlugin/lib/arStorageServiceUtils.class.php
@@ -1,0 +1,104 @@
+<?php
+
+/*
+ * This file is part of the Access to Memory (AtoM) software.
+ *
+ * Access to Memory (AtoM) is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * Access to Memory (AtoM) is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with Access to Memory (AtoM).  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+class arStorageServiceUtils
+{
+  const STORAGE_SERVICE_PACKAGE_PATH = 'file';
+
+  /**
+   * Use phpcurl to request a URL and pass the header and stream back to
+   * the browser.
+   *
+   */
+  public static function getFileFromStorageService($url)
+  {
+    $ch = curl_init();
+
+    curl_setopt($ch, CURLOPT_CUSTOMREQUEST, 'GET');
+
+    curl_setopt($ch, CURLOPT_URL, $url);
+    curl_setopt($ch, CURLOPT_RETURNTRANSFER, true);
+    curl_setopt($ch, CURLOPT_CONNECTTIMEOUT, 500);
+    curl_setopt($ch, CURLOPT_WRITEFUNCTION, function($ch, $data) {
+        echo $data;
+
+        return strlen($data);
+    });
+
+    curl_setopt($ch, CURLOPT_HEADERFUNCTION, function($ch, $header) {
+        header($header);
+
+        return strlen($header);
+    });
+
+    curl_setopt($ch, CURLOPT_FOLLOWLOCATION, 1); // Storage service redirects
+    curl_setopt($ch, CURLOPT_FAILONERROR, true);
+    curl_setopt($ch, CURLOPT_HTTPHEADER, array(
+      sprintf(
+        'Authorization: ApiKey %s:%s',
+        (string) QubitSetting::getByName('storage_service_username'),
+        (string) QubitSetting::getByName('storage_service_api_key')
+      ),
+      'User-Agent: DRMC',
+    ));
+
+    curl_exec($ch);
+    $status = curl_getinfo($ch, CURLINFO_RESPONSE_CODE);
+
+    curl_close($ch);
+
+    return $status;
+  }
+
+  /**
+   * Return true if plugin is enabled and feature is activated.
+   *
+   */
+  public static function getAipDownloadEnabled()
+  {
+    $configuration = ProjectConfiguration::getActive();
+
+    if ($configuration->isPluginEnabled('arStorageServicePlugin')
+      && null !== $setting = QubitSetting::getByName('download_aip_enabled'))
+    {
+      return boolval($setting->getValue(array('sourceCulture' => true)));
+    }
+  }
+
+  public static function getStorageServiceException($status)
+  {
+    switch ($status)
+    {
+      case '400':
+        return new QubitApiBadRequestException("Storage service bad request");
+
+      case '404':
+        return new QubitApi404Exception("Storage service resource not found");
+
+      case '401':
+        return new QubitApiNotAuthorizedException("Storage service resource not authorized");
+
+      case '403':
+        return new QubitApiForbiddenException("Storage service resource forbidden");
+
+      default:
+        return new Exception(sprintf('Storage service error %s', (string)$status));
+    }
+  }
+}
--- a/plugins/arStorageServicePlugin/modules/arStorageService/actions/downloadAction.class.php
+++ b/plugins/arStorageServicePlugin/modules/arStorageService/actions/downloadAction.class.php
@@ -1,0 +1,105 @@
+<?php
+
+/*
+ * This file is part of the Access to Memory (AtoM) software.
+ *
+ * Access to Memory (AtoM) is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * Access to Memory (AtoM) is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with Access to Memory (AtoM).  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+class arStorageServiceDownloadAction extends sfAction
+{
+  public function execute($request)
+  {
+    $view = sfView::NONE;
+
+    $this->resource = $this->getRoute()->resource;
+
+    // Check that object exists and that it is not the root
+    if (!isset($this->resource))
+    {
+      $this->forward404();
+    }
+
+    try
+    {
+      $status = $this->downloadAip($request);
+    }
+    catch (QubitApi404Exception $e)
+    {
+      $this->response->setStatusCode(404, $e->getMessage());
+      throw $e;
+    }
+    catch (QubitApiNotAuthorizedException $e)
+    {
+      $this->response->setStatusCode(401, $e->getMessage());
+      throw $e;
+    }
+    catch (QubitApiForbiddenException $e)
+    {
+      $this->response->setStatusCode(403, $e->getMessage());
+      throw $e;
+    }
+    catch (QubitApiBadRequestException $e)
+    {
+      $this->response->setStatusCode(400, $e->getMessage());
+      throw $e;
+    }
+    catch (Exception $e)
+    {
+      $this->response->setStatusCode(500, $e->getMessage());
+      throw $e;
+    }
+
+    return $view;
+  }
+
+  /**
+   * Build the request URL for the Storage Service API's 'download' endpoint
+   * and make the request. Ensure if there is an error that the Storage Service
+   * return status is passed back to the browser.
+   *
+   */
+  protected function downloadAip($request)
+  {
+    if (!arStorageServiceUtils::getAipDownloadEnabled())
+    {
+      throw new QubitApiForbiddenException('AIP Download disabled');
+    }
+
+    if (null === $aipUUID = $this->resource->object->aipUUID)
+    {
+      throw new QubitApiBadRequestException('Missing parameter: aipuuid');
+    }
+
+    if (null === $baseUrl = QubitSetting::getByName('storage_service_api_url'))
+    {
+      throw new QubitApiBadRequestException('Missing setting: storage_service_api_url');
+    }
+
+    $url = sprintf('%s/%s/%s/download/',
+      trim($baseUrl, "/"),
+      arStorageServiceUtils::STORAGE_SERVICE_PACKAGE_PATH,
+      $aipUUID
+    );
+
+    // Check return status from Storage Service
+    if (200 !== $status = arStorageServiceUtils::getFileFromStorageService($url))
+    {
+      $ex = arStorageServiceUtils::getStorageServiceException($status);
+      throw $ex;
+    }
+
+    exit;
+  }
+}
--- a/plugins/arStorageServicePlugin/modules/arStorageService/actions/extractFileAction.class.php
+++ b/plugins/arStorageServicePlugin/modules/arStorageService/actions/extractFileAction.class.php
@@ -1,0 +1,118 @@
+<?php
+
+/*
+ * This file is part of the Access to Memory (AtoM) software.
+ *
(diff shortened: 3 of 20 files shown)


plugins/arStorageServicePlugin/js/storageServiceDownload.js
```diff
@@ -0,0 +1,71 @@
...
+    // Content-Disposition: attachment; filename="test01-36d4efb7-83ba-46bf-a171-8c7b965d9dcb.7z"
+    var filenameRegex = /filename[^;=\n]*=((['"]).*?\2|[^;\n]*)/;
+    var matches = filenameRegex.exec(contentDispositionText);
+    if (matches != null && matches[1]) {
+      return(matches[1].replace(/['"]/g, ''));
+    }
+    return defaultFilename;
+  }
+
+  StorageServiceDownload.prototype = {
```
djjuhasz: I'm don't know a lot about Javascript prototypes, so my suggestion below may just be a wrong take on how to write this script.  With that caveat, I find the way the StorageServiceDownload object is constructed and called to be confusing, and I'm not sure if there is any advantage to the current complexity.  Specifically I'm not clear why the object is defined with `var StorageServiceDownload = function ()` then the prototype adds the returned anonymous function as a prototype.constructor?

I would prefer to see a more conventional class definition and instantiation, if it provides the same functionality, e.g.

```
function StorageServiceDownload() {
  this.listen();
}

StorageServiceDownload.prototype = {
  listen: function ()
  {
    /* etc... */
  },

  onDownloadBtn: function(event)
  {
    /*  etc... */
  }
}

$(function ()
{
  new StorageServiceDownload();
});
```

plugins/arStorageServicePlugin/js/storageServiceDownload.js
```diff
@@ -0,0 +1,71 @@
+(function ($) {
+
+  'use strict';
+
+  var StorageServiceDownload = function ()
+    {
+      this.listen();
+    };
+
+  function parseContentDisposition (contentDispositionText, defaultFilename = "download.7z")
```
djjuhasz: If the `parseContentDisposition()` function isn't used in any other context, I think it would be better to make it a method of the `StorageServiceDownload` class.

plugins/arStorageServicePlugin/js/storageServiceDownload.js
```diff
@@ -0,0 +1,71 @@
...
+    constructor: StorageServiceDownload,
+
+    listen: function()
+    {
+      $(document).on('click', 'button.storageServiceDownloadButton', $.proxy(this.onDownloadBtn, this));
+    },
+
+    onDownloadBtn: function (event)
+    {
+      if (typeof event.preventDefault === 'function')
```
djjuhasz: I don't understand why this is necessary?  Maybe you could add a comment to describe why `preventDefault` would be a function and why that's undesirable?

plugins/arStorageServicePlugin/js/storageServiceDownload.js
```diff
@@ -0,0 +1,71 @@
...
+    return defaultFilename;
+  }
+
+  StorageServiceDownload.prototype = {
+
+    constructor: StorageServiceDownload,
+
+    listen: function()
+    {
+      $(document).on('click', 'button.storageServiceDownloadButton', $.proxy(this.onDownloadBtn, this));
```
djjuhasz: TIL about $.proxy() :grin:

plugins/arStorageServicePlugin/js/storageServiceDownload.js
```diff
@@ -0,0 +1,71 @@
...
+
+      req.responseType = "blob";
+
+      req.onload = function (event) {
+        if (req.status != 200) {
+          alert(`Error ${req.status}: ${req.statusText}`);
+        }
+        else {
+          var blob = req.response;
+          var link = document.createElement('a');
```
djjuhasz: I looked up this XMLHttpRequest download pattern on Stackoverflow, and I'm a bit worried about the following comment, because the feature will be used to download large files: 

> This is a good solution for small files, but notice that this puts the whole file in memory before sending it to disk, so take that into account if downloading very large files. Also the user won't be able to see the progress of the actual download (like in Chrome's statusbar), what they'll see is the progress of the "download" from memory to disk instead (once the actual download has finished).

-- https://stackoverflow.com/questions/22724070/prompt-file-download-with-xmlhttprequest#comment88913700_44435573

I wonder if using a regular HTML <a> link to a PHP page that then streams the file contents (via fpassthru() or maybe curl()) will avoid the need to load the whole file into memory before download, and let the browser show the download progress bar, without blocking the user from continuing to navigate the site?

plugins/arStorageServicePlugin/modules/arStorageServicePlugin/actions/baseAction.class.php
```diff
@@ -0,0 +1,152 @@
...
+    return $this->renderText(json_encode($data, $options));
+  }
+
+  public static function getFileFromStorageService($url)
+  {
+    $ch = curl_init();
+
+    curl_setopt($ch, CURLOPT_CUSTOMREQUEST, 'GET');
+
+    curl_setopt($ch, CURLOPT_URL,$url);
```
djjuhasz: Nitpick: missing space after comma.

plugins/arStorageServicePlugin/modules/arStorageServicePlugin/actions/baseAction.class.php
```diff
@@ -0,0 +1,152 @@
...
+    curl_setopt($ch, CURLOPT_RETURNTRANSFER, true);
+    curl_setopt($ch, CURLOPT_CONNECTTIMEOUT, 500);
+    curl_setopt($ch, CURLOPT_WRITEFUNCTION, function($ch, $data) {
+        echo $data;
+        return strlen($data);
```
djjuhasz: Nitpick: missing blank line before return.

plugins/arStorageServicePlugin/modules/arStorageServicePlugin/actions/baseAction.class.php
```diff
@@ -0,0 +1,152 @@
...
+    });
+
+    curl_setopt($ch, CURLOPT_HEADERFUNCTION, function($ch, $header) {
+        header($header);
+        return strlen($header);
```
djjuhasz: Nitpick: missing blank line before return.

plugins/arStorageServicePlugin/modules/arStorageServicePlugin/actions/baseAction.class.php
```diff
@@ -0,0 +1,152 @@
...
+    });
+
+    curl_setopt($ch, CURLOPT_FOLLOWLOCATION, 1); // Storage service redirects
+    curl_setopt($ch, CURLOPT_FAILONERROR, true);
+    curl_setopt($ch, CURLOPT_HTTPHEADER, array(
+      sprintf('Authorization: ApiKey %s:%s', (string)QubitSetting::getByName('storage_service_username'), (string)QubitSetting::getByName('storage_service_api_key')),
```
djjuhasz: Nitpick: missing space after `(string)`

plugins/arStorageServicePlugin/modules/arStorageServicePlugin/actions/baseAction.class.php
```diff
@@ -0,0 +1,152 @@
...
+    curl_close($ch);
+
+    return $status;
+  }
+
+  public static function getAipDownloadEnabled()
+  {
+    $setting = QubitSetting::getByName('download_aip_enabled');
+
+    return (null === $setting || boolval($setting->getValue(array('sourceCulture' => true))));
```
djjuhasz: I think this will return true if `null === $setting`, which I don't think is the intention?

plugins/arStorageServicePlugin/modules/arStorageServicePlugin/actions/baseAction.class.php
```diff
@@ -0,0 +1,152 @@
...
+      throw $e;
+    }
+
+    return $view;
+  }
+
+  public function process($request)
+  {
+    $method = strtoupper($request->getMethod());
+    if (!method_exists($this, $method))
```
djjuhasz: Opinion: I think it would be better to explicitly list the methods that can be called via a request parameter, to prevent weird exploits by passing unexpected method names (e.g. passing "method=process" may cause an infinite loop?), and to make it clear in the code what methods are available via the public API.

plugins/arStorageServicePlugin/modules/arStorageServicePlugin/actions/baseAction.class.php
```diff
@@ -0,0 +1,152 @@
...
+
+    // Define function callable
+    $fnCallable = array($this, $method);
+    $fnParamaters = array($request);
+
+    // Modern frameworks support application/json, Symfony1 is too old :)
+    // AngularJS doesn't use application/x-www-form-urlencoded
+    if (('PUT' == $method || 'POST' == $method) && 'application/json' == $request->getContentType())
+    {
+      $fnParamaters[] = json_decode($request->getContent());
```
djjuhasz: I think passing unsanitized parameters to `call_user_func_array()` is probably insecure.

djjuhasz (commented):
This is looking good overall Steve!  I've added a handful of comments, but I think the only one that could require fundamental changes is the question about whether to calling the download endpoint in a XMLHttpRequest will scale for large files.

sbreker:
See commit: d0f9fce501609c9053f1a1d39d57f71b6c1c48a2 - I refactored the code heavily in this commit and removed the js completely.

plugins/arStorageServicePlugin/modules/arStorageServicePlugin/actions/baseAction.class.php
sbreker: This is no longer performed.

plugins/arStorageServicePlugin/modules/arStorageServicePlugin/actions/baseAction.class.php
sbreker: Fixed!

plugins/arStorageServicePlugin/modules/arStorageServicePlugin/actions/baseAction.class.php
sbreker: fixed

plugins/arStorageServicePlugin/modules/arStorageServicePlugin/actions/baseAction.class.php
sbreker: Fixed

plugins/arStorageServicePlugin/js/storageServiceDownload.js
sbreker: This js code has been removed completely

plugins/arStorageServicePlugin/js/storageServiceDownload.js
sbreker: This js has been removed

plugins/arStorageServicePlugin/lib/arStorageServiceUtils.class.php
```diff
@@ -0,0 +1,100 @@
...
+    curl_setopt($ch, CURLOPT_HEADERFUNCTION, function($ch, $header) {
+        header($header);
+
+        return strlen($header);
+    });
+
+    curl_setopt($ch, CURLOPT_FOLLOWLOCATION, 1); // Storage service redirects
+    curl_setopt($ch, CURLOPT_FAILONERROR, true);
+    curl_setopt($ch, CURLOPT_HTTPHEADER, array(
+      sprintf('Authorization: ApiKey %s:%s', (string) QubitSetting::getByName('storage_service_username'), (string)QubitSetting::getByName('storage_service_api_key')),
```
djjuhasz: Nitpick: Long line is too long :P

plugins/arStorageServicePlugin/modules/arStorageService/actions/downloadAction.class.php
```diff
@@ -0,0 +1,107 @@
...
+
+    // Check return status from Storage Service
+    if (200 !== $status = arStorageServiceUtils::getFileFromStorageService($url))
+    {
+      $ex = arStorageServiceUtils::getStorageServiceException($status);
+
+      throw $ex;
+    }
+
+    exit;
```
djjuhasz, line 103: Huh, "exit" is unexpected here.  Is this debugging code that didn't get removed, or is it intentional?

plugins/arStorageServicePlugin/modules/arStorageService/actions/extractFileAction.class.php
```diff
@@ -0,0 +1,120 @@
...
+    }
+    catch (QubitApiBadRequestException $e)
+    {
+      $this->response->setStatusCode(400, $e->getMessage());
+      throw $e;
+    }
+    catch (Exception $e)
+    {
+      $this->response->setStatusCode(500, $e->getMessage());
+
```
djjuhasz: Nitpick: either this blank line should be removed, or a blank line should be added before the preceding "throws" for consistency.  I don't know if our style guide suggests a blank line before a throw, or not?

djjuhasz (changes requested):
Looking good @sbreker! I added a few nitpicks and a question about an unexpected exit. :)

plugins/arStorageServicePlugin/modules/arStorageService/actions/downloadAction.class.php
sbreker: This is intentional and modelled after this: https://github.com/artefactual/binder/blob/qa/0.9.x/plugins/arRestApiPlugin/modules/api/actions/aipsDownloadAction.class.php#L61

If the return code from the Storage Service is 200, then absolutely nothing further should be forwarded to the client wrt this response after the response from the Storage Service has been forwarded. The exit ensures that is the case.

plugins/arStorageServicePlugin/modules/arStorageService/actions/extractFileAction.class.php
sbreker: I think you are correct. I have removed any blank lines before a 'throw' statement.

plugins/arStorageServicePlugin/modules/arStorageServicePlugin/actions/baseAction.class.php
sbreker: I have removed this, but this same logic exists in the rest API plugin in both AtoM and Binder and should be reviewed if this is the case.

plugins/arStorageServicePlugin/lib/arStorageServiceUtils.class.php
sbreker: Fixed

plugins/arStorageServicePlugin/lib/arStorageServiceUtils.class.php
```diff
@@ -0,0 +1,102 @@
...
+
+        return strlen($header);
+    });
+
+    curl_setopt($ch, CURLOPT_FOLLOWLOCATION, 1); // Storage service redirects
+    curl_setopt($ch, CURLOPT_FAILONERROR, true);
+    curl_setopt($ch, CURLOPT_HTTPHEADER, array(
+      sprintf('Authorization: ApiKey %s:%s',
+      (string) QubitSetting::getByName('storage_service_username'),
+      (string)QubitSetting::getByName('storage_service_api_key')),
```
djjuhasz: Nitpick: needs a space after `(string)`

djjuhasz (changes requested):
One last nitpick @sbreker :)

plugins/arStorageServicePlugin/lib/arStorageServiceUtils.class.php
sbreker: fixed

djjuhasz: oh, one more style nitpick: I think it would be more readable to indent the sprintf arguments, and drop the closing parenthesis to the next line, e.g. 

```
    curl_setopt($ch, CURLOPT_HTTPHEADER, array(
      sprintf(
        'Authorization: ApiKey %s:%s',
        (string) QubitSetting::getByName('storage_service_username'),
        (string) QubitSetting::getByName('storage_service_api_key')
      ),
      'User-Agent: DRMC',
    ));
```

sbreker: I agree - fixed.

djjuhasz (approved):
Looks good @sbreker! :shipit:

PR #1168, opened 2020-08-11
Fix carousel's thumbnail retrieval, refs #13402

replaceafill:
This adds a missing argument to the
`QubitDigitalObject::getGenericRepresentation` call in the carousel
when the user doesn't have a group or PREMIS permission to access a
digital object's thumbnail.


djjuhasz (approved):
Looks good @replaceafill 👍

replaceafill:
Thank you @djjuhasz!

PR #1169, opened 2020-08-11
Allow disabling copyright pop-up, refs #13399

djjuhasz:
Fix a regression introduced in commit 676305b that displayed the conditional
copyright pop-up whenever a "master" digital object is requested.

The error was introduced by inverting the `hasConditionalCopyright()` check to
return true on success, instead of returning false (thus avoiding the following
valid token check) when `hasConditionalCopyring()` returned false.

### DIFF
--- a/apps/qubit/modules/digitalobject/actions/viewAction.class.php
+++ b/apps/qubit/modules/digitalobject/actions/viewAction.class.php
@@ -75,17 +75,14 @@ class DigitalObjectViewAction extends sfAction
 
   protected function needsPopup($action)
   {
-    // Only if the user is reading the master digital object
-    if ($action !== 'readMaster')
+    // Only if the user is reading the master digital object, and the resource
+    // has a PREMIS conditional copyright restriction
+    if ($action != 'readMaster' || !$this->resource->hasConditionalCopyright())
     {
       return false;
     }
 
-    if ($this->resource->hasConditionalCopyright())
-    {
-      return true;
-    }
-
+    // Show the pop-up if a valid access token was not submitted
     return false === $this->isAccessTokenValid();
   }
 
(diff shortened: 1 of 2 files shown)


apps/qubit/modules/digitalobject/actions/viewAction.class.php
```diff
@@ -75,17 +75,14 @@ public function execute($request)
 
   protected function needsPopup($action)
   {
-    // Only if the user is reading the master digital object
-    if ($action !== 'readMaster')
-    {
+    // Only if the user is reading the master digital object, and the resource
+    // has a PREMIS conditional copyright restriction
+    if ($action != 'readMaster' || !$this->resource->hasConditionalCopyright()
+    ) {
```
jraddaoui: Misplaced parenthesis.

jraddaoui (approved):

jraddaoui:
Thanks @djjuhasz! I think the following condition (introduced in that commit too) doesn't need changes, but just in case I'm missing something ...

https://github.com/artefactual/atom/blob/qa/2.x/lib/job/arBaseJob.class.php#L539-L555

apps/qubit/modules/digitalobject/actions/viewAction.class.php
djjuhasz: Fixed

djjuhasz:
> Thanks @djjuhasz! I think the following condition (introduced in that commit too) doesn't need changes, but just in case I'm missing something ...
> 
> https://github.com/artefactual/atom/blob/qa/2.x/lib/job/arBaseJob.class.php#L539-L555

I looked at that too @jraddaoui, and I agree that it's fine as is.

jraddaoui:
Thanks @djjuhasz!

PR #1170, opened 2020-08-12
Hide drafts checkbox in actors export, refs #13395

jraddaoui:
Do not show include drafts checkbox in clipboard actors export.


mcantelon (approved):
Looks good to me!

jraddaoui:
Thanks @mcantelon!

PR #1171, opened 2020-08-12
DIP Upload METS parsing issue, refs #13351

sbreker:
This commit fixes an issue where the dmdSec value was not being
properly assigned resulting in an error during DIP upload. This
issue affected transfers that include metadata.

The $dmdSec value is dependent on the $dmdId being assigned in
the preceeding conditions - parentheses added to ensure the
clauses are evaluated in the correct order.


djjuhasz (approved):
Good job figuring out the problem @sbreker  👍

PR #1172, opened 2020-08-13
Fix some PHP warnings & notices, refs #13348

djjuhasz:
Fix several non-critical notices and warnings on:
* archival description view and edit pages
* browse/search pages for archival descriptions and actors

### DIFF
--- a/apps/qubit/modules/actor/actions/browseAction.class.php
+++ b/apps/qubit/modules/actor/actions/browseAction.class.php
@@ -578,6 +578,12 @@ class ActorBrowseAction extends DefaultBrowseAction
 
     $this->populateAggs($resultSet);
 
+    // Default to hide the advanced search panel
+    $this->showAdvanced = filter_var(
+      $request->showAdvanced,
+      FILTER_VALIDATE_BOOLEAN
+    );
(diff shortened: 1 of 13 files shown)


apps/qubit/modules/actor/actions/browseAction.class.php
```diff
@@ -578,6 +578,13 @@ public function execute($request)
 
     $this->populateAggs($resultSet);
 
+    // Default to hide the advanced search panel
+    $this->showAdvanced = false;
+    if (filter_var($request->showAdvanced, FILTER_VALIDATE_BOOLEAN))
```
mcantelon: Could you just do `$this->showAdvanced = filter_var($request->showAdvanced, FILTER_VALIDATE_BOOLEAN)`?

mcantelon (approved):
One question but otherwise looks 💯

apps/qubit/modules/actor/actions/browseAction.class.php
djjuhasz: Why yes, you can! Very elegant @mcantelon. 👍

PR #1173, opened 2020-08-13
Refactor common export code, refs #13395

djjuhasz:
Create a separate 'arExportJob' subclass for methods specific to export jobs,
and refactor for readability.


lib/job/arExportJob.class.php
```diff
@@ -0,0 +1,349 @@
...
+      {
+        // Filename not used yet - add to tracker
+        $fileNames[$fileName] = 0;
+      }
+      else
+      {
+        // Filename has been used - increment counter and add to filename
+        $fileNames[$fileName]++;
+        $doPathInfo = pathinfo($doPath);
+        $fileName = "{$doPathInfo['filename']}_{$fileNames[$fileName]}.{$doPathInfo['extension']}";
```
mcantelon: Using sprintf might be cleaner.

mcantelon (approved):
One quibble but otherwise 👍 !

lib/job/arExportJob.class.php
djjuhasz: Yeah, good call. 👍

PR #1176, opened 2020-08-18
Add CAS authentication, refs #13383

tw4l:
This commit adds support for CAS (Central Authentication Service) user authentication in AtoM, implemented in a new `arCasPlugin`.

The plugin also provides an optional mechanism for dynamically setting `AclUserGroup` membership on each login, based on the presence or absence of expected values in the attributes returned from the CAS server during the `p3/serviceValidate` stage of CAS authentication (CAS version 3.0 only).

Configuration of the plugin is handled manually by editing configuration files rather than through the AtoM user interface to avoid a "chicken-egg" problem of needing to log into AtoM as an administrator in order to set up the authentication mechanism. The steps required to enable the plugin are as follows, with all paths relative to the AtoM root directory:

* Activate the plugin by adding it to the `$plugins` array in `config/ProjectConfiguration.class.php`
* Configure the CAS settings (CAS version, server name, server port, server path, server SSL certificate, and settings for parsing CAS attributes into `AclUserGroup` membership) in `plugins/arCasPlugin/config/app.yml`
* Change the default login module from `user` to `cas` in `apps/qubit/config/settings.yml`
* Change the user class to `casUser` in `apps/qubit/config/factories.yml`

Any users needed for tasks such as DIP upload should be created prior to enabling CAS authentication, or will need to be associated with a user account on the CAS server.

### DIFF
--- a/plugins/arCasPlugin/lib/arCAS.class.php
+++ b/plugins/arCasPlugin/lib/arCAS.class.php
@@ -1,0 +1,83 @@
+<?php
+
+/*
+ * This file is part of the Access to Memory (AtoM) software.
+ *
+ * Access to Memory (AtoM) is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * Access to Memory (AtoM) is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with Access to Memory (AtoM).  If not, see <http://www.gnu.org/licenses/>.
+ *
+ * This file is based heavily on the sfCASPlugin developed by D.Jeanmonod and
+ * maintained by H.Lepesant, MIT License, https://github.com/jeanmonod/sfCASPlugin.
+ */
+
+class arCAS
+{
+  protected static $phpCASIsInitialized = false;
+
+  /**
+   * Initialize the phpCAS library
+   */
+  public static function initializePhpCAS()
+  {  
+    // Return if phpCAS is already initialized
+    if (self::$phpCASIsInitialized) {
+      return;
+    }
+
+    require_once sfConfig::get('sf_root_dir').'/vendor/composer/jasig/phpcas/CAS.php';
+
+    if (true == sfConfig::get('sf_debug', false))
+    {
+      $debugLogPath = sfConfig::get('sf_log_dir').'/phpcas.log';
+      phpCAS::setDebug($debugLogPath);
+      phpCAS::setVerbose(true);
+    }
+
+    $casVersion = sfConfig::get('app_cas_cas_version', '3.0');
+    $validCasVersions = array('1.0', '2.0', '3.0');
+    $casVersion = in_array($casVersion, $validCasVersions) ? $casVersion : CAS_VERSION_3_0;
+
+    phpCAS::client(
+      $casVersion,
+      sfConfig::get('app_cas_server_name'),
+      sfConfig::get('app_cas_server_port'),
+      sfConfig::get('app_cas_server_path'),
+      false  // Let Symfony handle the session
+    );
+
+    // This setting prevents a redirection loop that is otherwise caused by the
+    // interaction of phpCAS with Symfony's session management.
+    phpCAS::setNoClearTicketsFromUrl();
+
+    // Validate the server SSL certificate according to configuration.
+    $certPath = sfConfig::get('app_cas_server_cert', false);
+    if (!strpos($certPath, '/') === 0)
+    {
+      $certPath = sfConfig::get('sf_root_dir') . DIRECTORY_SEPARATOR . $certPath;
+    }
+    if (file_exists($certPath))
+    {
+      phpCAS::setCasServerCACert($certPath);
+    }
+    else if ($certPath === false)
+    {
+      phpCAS::setNoCasServerValidation();
+    }
+    else
+    {
+      throw new Exception("Invalid SSL certificate settings. Please review the app_cas_server_cert parameter in plugin app.yml.");
+    }
+
+    self::$phpCASIsInitialized = true;   
+  }
+}
--- a/plugins/arCasPlugin/lib/casUser.class.php
+++ b/plugins/arCasPlugin/lib/casUser.class.php
@@ -1,0 +1,132 @@
+<?php
+
+/*
+ * This file is part of the Access to Memory (AtoM) software.
+ *
+ * Access to Memory (AtoM) is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * Access to Memory (AtoM) is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with Access to Memory (AtoM).  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+class casUser extends myUser implements Zend_Acl_Role_Interface
+{
+  public function initialize(sfEventDispatcher $dispatcher, sfStorage $storage, $options = array())
+  {
+    // initialize parent
+    parent::initialize($dispatcher, $storage, $options);
+  }
+
+  /**
+   * Try to login with the CAS server.
+   */
+  public function authenticate($username = null, $password = null)
+  {
+    $authenticated = false;
+
+    arCAS::initializePhpCAS();
+    phpCAS::forceAuthentication();
+    $username = phpCAS::getUser();
+
+    // Load user using username or, if one doesn't exist, create it.
+    $criteria = new Criteria;
+    $criteria->add(QubitUser::USERNAME, $username);
+    if (null === $user = QubitUser::getOne($criteria))
+    {
+      $user = new QubitUser();
+      $user->username = $username;
+      $user->save();
+    }
+
+    // Parse CAS attributes into group memberships. If enabled, we perform this
+    // check each time a user authenticates so that changes made on the CAS
+    // server are applied in AtoM on the next login.
+    if (true == sfConfig::get('app_cas_set_groups_from_attributes', false))
+    {
+      $attributes = phpCAS::getAttributes();
+      $this->setGroupsFromCasAttributes($user, $attributes);
+    }
+    
+    $authenticated = true;
+    $this->signIn($user);
+
+    return $authenticated;
+  }
+
+  /**
+   * Set group membership based on user attributes returned by CAS server.
+   */
+  protected function setGroupsFromCasAttributes($user, $attributes)
+  {
+    // Get the CAS attribute we're checking for AtoM group membership. If the
+    // attribute doesn't exist or is null, log the error and return.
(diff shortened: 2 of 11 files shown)


plugins/arCasPlugin/lib/arCAS.class.php
```diff
@@ -0,0 +1,83 @@
...
+  public static function initializePhpCAS()
+  {  
+    // Return if phpCAS is already initialized
+    if (self::$phpCASIsInitialized) {
+      return;
+    }
+
+    require_once sfConfig::get('sf_root_dir').'/vendor/composer/jasig/phpcas/CAS.php';
+
+    if (sfConfig::get('sf_debug', false) == true)
```
sbreker: This comparison should be reversed. Symfony's coding standards specify "Yoda conditions" (https://en.wikipedia.org/wiki/Yoda_conditions) which reverses the typical order of the comparison: e.g. "true == sfConfig::get('sf_debug', false)".

plugins/arCasPlugin/lib/arCAS.class.php
```diff
@@ -0,0 +1,83 @@
...
+    {
+      $debugLogPath = sfConfig::get('sf_log_dir').'/phpcas.log';
+      phpCAS::setDebug($debugLogPath);
+      phpCAS::setVerbose(true);
+    }
+
+    $casVersion = sfConfig::get('app_cas_cas_version', '3.0');
+    $validCasVersions = array('1.0', '2.0', '3.0', 'S1');
```
sbreker: Wondering if versions other than 3.0 have been tested? Is it valid/safe to allow selection of 1.0, 2.0 etc?

plugins/arCasPlugin/lib/casUser.class.php
```diff
@@ -0,0 +1,164 @@
...
+    {
+      $user = new QubitUser();
+      $user->username = $username;
+      $user->save();
+    }
+
+    // Parse CAS attributes into group memberships. If enabled, we perform this
+    // check each time a user authenticates so that changes made on the CAS
+    // server are applied in AtoM on the next login.
+    if (sfConfig::get('app_cas_set_groups_from_attributes', false) == true)
```
sbreker: 'Yoda order' should be used in the comparison as above

plugins/arCasPlugin/lib/arCAS.class.php
```diff
@@ -0,0 +1,83 @@
...
+      return;
+    }
+
+    require_once sfConfig::get('sf_root_dir').'/vendor/composer/jasig/phpcas/CAS.php';
+
+    if (sfConfig::get('sf_debug', false) == true)
+    {
+      $debugLogPath = sfConfig::get('sf_log_dir').'/phpcas.log';
+      phpCAS::setDebug($debugLogPath);
+      phpCAS::setVerbose(true);
```
sbreker, line 43: This debugging log will be a helpful feature!

plugins/arCasPlugin/lib/casUser.class.php
```diff
@@ -0,0 +1,164 @@
...
+
+    if (!array_key_exists($attributeKey, $attributes))
+    {
+      sfContext::getInstance()->getLogger()->err('Key not found in CAS attributes');
+      return;
+    }
+    
+    $attributeToCheck = $attributes[$attributeKey];
+    
+    if ($attributeToCheck === null)
```
sbreker: Yoda order

plugins/arCasPlugin/lib/casUser.class.php
```diff
@@ -0,0 +1,164 @@
...
+    {
+      sfContext::getInstance()->getLogger()->err('CAS attribute used for setting AtoM group membership is null');
+      return;
+    }
+
+    // The value for a given CAS attribute can be an array or a string. If it's
+    // a string, we convert into an array to simplify the checking routine.
+    if (!is_array($attributeToCheck))
+    {
+      $attributeToCheck = array($attributeToCheck);
```
sbreker, line 91: Nitpick: It looks like $attributeToCheck could contain more than one attribute from CAS?  If so perhaps $attributesToCheck (plural) might be clearer

plugins/arCasPlugin/lib/casUser.class.php
```diff
@@ -0,0 +1,164 @@
...
+      ),
+      array(
+        'expectedCasValue' => sfConfig::get('app_cas_contributor_value', 'atom-contributor'),
+        'groupID' => QubitAclGroup::CONTRIBUTOR_ID
+      ),
+      array(
+        'expectedCasValue' => sfConfig::get('app_cas_translator_value', 'atom-translator'),
+        'groupID' => QubitAclGroup::TRANSLATOR_ID
+      )
+    );
```
sbreker: This has a limitation in that only these four groups will be checked regardless of how the app.yml file is configured. It will require a coding change to make use of custom groups defined in AtoM, and will always check these values even if one of the groups has been removed from the plugin's yml config file. 

I am wondering if these values could be defined in the plugin app.yml file as an associative array? Is it possible to build this array dynamically from the app.yml file entries?

plugins/arCasPlugin/lib/casUser.class.php
```diff
@@ -0,0 +1,164 @@
...
+  /**
+  * Set membership in QubitAclUserGroup based on presence of expected CAS attribute.
+  */
+  private function setGroupMembership($user, $attributeToCheck, $expectedValue, $groupID)
+  {
+    if (in_array($expectedValue, $attributeToCheck))
+    {
+      // CAS attributes say the user belongs to the group, so add them to the
+      // group if they're not already a member.
+      if (!$user->hasGroup($groupID))
```
sbreker: Is it possible to grab all the groups at once for a user, reducing the number of trips to the DB? In the best case scenario, this will always be 4 accesses to the database minimum - does it make sense to grab all the groups for a user in one request?

plugins/arCasPlugin/modules/cas/actions/loginAction.class.php
```diff
@@ -0,0 +1,39 @@
...
+      $this->redirect('@homepage');
+    }
+
+    $this->getUser()->authenticate();
+
+    // Redirect to module/action the user was trying to reach before being redirected
+    // to CAS for authentication. We prefer a redirect to a forward so that the ticket
+    // parameter is not accidentally exposed in the user's browser.
+    $redirectUrl = $request->getParameter('module') . '/' . $request->getParameter('action');
+    $this->redirect($redirectUrl);
```
sbreker, line 37: This is great!  👍

sbreker (changes requested):
Hi Tessa! Looks great so far! I've added a couple nitpicky comments, and a couple of questions - one regarding the various CAS versions and some about the user group setting/lookups. Great work implementing this in Symfony's plugin structure, and also with integrating with AtoM's redirect behaviour!

plugins/arCasPlugin/lib/arCAS.class.php
tw4l: Interesting, what a name!

plugins/arCasPlugin/lib/arCAS.class.php
tw4l: Thanks, it really helped in development!

plugins/arCasPlugin/modules/cas/actions/loginAction.class.php
tw4l: Thanks for your help with it!

plugins/arCasPlugin/lib/casUser.class.php
tw4l: The way I've been thinking about it is that we're checking a single attribute from CAS that might have one or more values (i.e. a string or an array of strings). But I'd be happy to rename it if you think pluralizing will help with readability.

plugins/arCasPlugin/lib/arCAS.class.php
tw4l: My understanding of the CAS standard and the phpcas library is that there's good support for all four versions. I've tested with `1.0`, `2.0` and `3.0` during development and confirmed that login/logout works as expected across all versions. They're quite similar, with the main difference in relation to AtoM being that CAS user attributes are only available in `3.0`. It [looks like `S1`adds some additional methods we're not making use of](https://apereo.github.io/phpCAS/api/group__internalSAML.html), so it might be safe to drop that one.

plugins/arCasPlugin/lib/casUser.class.php
tw4l: Thanks, good point! I've made the change to build the array dynamically from the plugin's app.yml file in https://github.com/artefactual/atom/pull/1176/commits/0ac2b518c8ec35981c74cc6692e06a78b901103c.

plugins/arCasPlugin/lib/casUser.class.php
tw4l: Absolutely, great suggestion! Added in commit https://github.com/artefactual/atom/pull/1176/commits/bd631bfc8e995ce4d7a10293b3cc126145ff6948.

plugins/arCasPlugin/lib/arCAS.class.php
tw4l: I removed `S1` as one of the valid CAS version options in https://github.com/artefactual/atom/pull/1176/commits/8c00e4c493746aa9fca7bbea458d402e3e87eb97, since in theory it should work but we don't have the ability to test it.

tw4l:
Thanks @sbreker, for the review and for your guidance along the way! This should be ready for re-review. I've left responses in line to your questions, and indicated where I've made changes.

plugins/arCasPlugin/lib/casUser.class.php
```diff
@@ -122,7 +132,7 @@ private function setGroupMembership($user, $attributeToCheck, $expectedValue, $g
     {
       // CAS attributes say the user does not belong to the group, so remove
       // them if they are currently a member.
-      if ($user->hasGroup($groupID))
+      if (in_array($groupID, $currentAclUserGroupIds))
```
sbreker: In the edge case where:
- a user assigned to a group is removed from a group on the CAS side, 
- and the CAS group is valid inside app_cas_user_groups setting 
- and the group is still linked to the user in AtoM, 
...will the group be successfully removed during CAS authentication? 

A second edge case:
- a user is currently assigned to a group
- they authenticate via CAS which also has them in that group
- the group has been removed from the 'app_cas_user_groups' setting
Will the invalid group be successfully removed from the user?

sbreker (commented):
Looks great so far! I've added a comment with a few comments about edge cases during the group removal process

plugins/arCasPlugin/lib/casUser.class.php
tw4l: > In the edge case where:
> 
>     * a user assigned to a group is removed from a group on the CAS side,
> 
>     * and the CAS group is valid inside app_cas_user_groups setting
> 
>     * and the group is still linked to the user in AtoM,
>       ...will the group be successfully removed during CAS authentication?

Yes, exactly. If a user is downgraded on the CAS server and no configuration change has been made in AtoM, AtoM will reflect the change and update the group membership on the next login.

> A second edge case:
> 
>     * a user is currently assigned to a group
> 
>     * they authenticate via CAS which also has them in that group
> 
>     * the group has been removed from the 'app_cas_user_groups' setting
>       Will the invalid group be successfully removed from the user?

Interesting.This edge case is not currently handled by the code. As-is, it would currently require an administrator to manually remove the user group that's no longer configured in `app_cas_user_groups` from users who belong to it. But I think we could accomplish this by implementing logic that's a bit closer to the `user` module's `editAction`, whereby a user's group membership is reset on each login and only appropriately configured groups are included.

tw4l:
Thanks @sbreker! Flagging this for re-review - I think the design's much nicer for your last round of feedback and help on Slack. I've tested against the client's CAS server and verified that the changes are working as expected.

sbreker (approved):
Looks great @tw4l 👍 - Thanks for making these changes - it will ensure that the AtoM user's ACL groups are always in sync with CAS, while respecting what groups are available via the plugin config file & AtoM Settings

tw4l:
Merged as https://github.com/artefactual/atom/commit/f937ceda7ddc2870ac5d63b908e5c666b6318a2b, closing manually.

PR #1177, opened 2020-08-21
Refactor and fix clipboard export, refs #13395

djjuhasz:
- Refactor clipboard export code to reduce duplication of code
- Fix coding  standard issues such as long lines, and trailing whitespace
- Add conditional to avoid trying to export external digital objects, and add notice to log
- Fix bug that always set the job name to "XML export" even for CSV exports
- Add more comments

### DIFF
--- a/apps/qubit/modules/clipboard/actions/exportAction.class.php
+++ b/apps/qubit/modules/clipboard/actions/exportAction.class.php
@@ -50,9 +50,9 @@ class ClipboardExportAction extends DefaultEditAction
 
   public function execute($request)
   {
-
     // Get object type and validate
-    // Currently 'switch' to process the inbound parameter, validate and set default - could be if/then if preferred
+    // Currently 'switch' to process the inbound parameter, validate and set
+    // default - could be if/then if preferred
     $this->objectType = trim(strtolower($request->getParameter('type')));
     switch ($this->objectType)
     {
@@ -72,16 +72,28 @@ class ClipboardExportAction extends DefaultEditAction
     }
 
     // Get format and validate
-    // Currently 'switch' to process the inbound parameter, validate and set default - could be if/then if preferred
+    // Currently 'switch' to process the inbound parameter, validate and set
+    // default - could be if/then if preferred
     $this->formatType = trim(strtolower($request->getParameter('format')));
-    if(!('xml' == $this->formatType || 'csv' == $this->formatType) || 'repository' == $this->objectType)
+    if ($this->formatType != 'xml' || $this->objectType == 'repository')
     {
-        $this->formatType = 'csv';
+      $this->formatType = 'csv';
     }
 
-    // Basic permission check to determine whether digital object export should be made available
+    // Basic permission check to determine whether digital object export should
+    // be made available
     $this->digitalObjectsAvailable = false;
-    if (sfConfig::get('app_clipboard_export_digitalobjects_enabled', false) && ('informationObject' == $this->objectType || ('actor' == $this->objectType && $this->context->user->isAuthenticated())))
+
+    if (
+      sfConfig::get('app_clipboard_export_digitalobjects_enabled', false)
+      && (
+        'informationObject' == $this->objectType
+        || (
+          'actor' == $this->objectType
+          && $this->context->user->isAuthenticated()
+        )
+      )
+    )
     {
       $this->digitalObjectsAvailable = true;
     }
@@ -89,35 +101,32 @@ class ClipboardExportAction extends DefaultEditAction
     // Show export options panel if:
     // information object type
     // or, if actor type and digital objects are on the clipboard
-    $this->showOptions = 'informationObject' == $this->objectType || ('actor' == $this->objectType && $this->digitalObjectsAvailable);
+    $this->showOptions = 'informationObject' == $this->objectType
+      || ('actor' == $this->objectType && $this->digitalObjectsAvailable);
 
     // Get field includeDescendants if:
     // options enabled
     // and, information object type
-    $this->descendantsIncluded = $this->showOptions && 'informationObject' == $this->objectType && 'on' == $request->getParameter('includeDescendants');
+    $this->descendantsIncluded = $this->showOptions
+      && 'informationObject' == $this->objectType
+      && 'on' == $request->getParameter('includeDescendants');
 
     // get field includeAllLevels if:
     // descendantsIncluded enabled
-    $this->descendantsAllLevels = $this->descendantsIncluded && 'on' == $request->getParameter('includeAllLevels');
+    $this->descendantsAllLevels = $this->descendantsIncluded
+      && 'on' == $request->getParameter('includeAllLevels');
 
     // Get field includeDigitalObjects if:
-    // digital objects are on the clipboard
-    // and, information object type and descendants are not selected
-    // or, if actor type
-    $this->includeDigitalObjects = false;
-    if ($this->digitalObjectsAvailable)
-    {
-      $this->includeDigitalObjects = 'on' == $request->getParameter('includeDigitalObjects');
-    }
+    // digital object export option is available
+    $this->includeDigitalObjects = $this->digitalObjectsAvailable
+      && 'on' == $request->getParameter('includeDigitalObjects');
 
     // Get field includeDrafts if:
     // options enabled
     // and, user is authenticated
-    $this->draftsIncluded = false;
-    if ($this->showOptions && $this->context->user->isAuthenticated())
-    {
-      $this->draftsIncluded = 'on' == $request->getParameter('includeDrafts');
-    }
+    $this->draftsIncluded = $this->showOptions
+      && $this->context->user->isAuthenticated()
+      && 'on' == $request->getParameter('includeDrafts');
 
     parent::execute($request);
 
@@ -130,7 +139,10 @@ class ClipboardExportAction extends DefaultEditAction
       return;
     }
 
-    $this->response->setHttpHeader('Content-Type', 'application/json; charset=utf-8');
+    $this->response->setHttpHeader(
+      'Content-Type',
+      'application/json; charset=utf-8'
+    );
 
     $this->form->bind($request->getPostParameters());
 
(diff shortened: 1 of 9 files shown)


apps/qubit/modules/clipboard/actions/exportAction.class.php
```diff
@@ -3,18 +3,18 @@
...
  *
  * Access to Memory (AtoM) is distributed in the hope that it will be useful,
  * but WITHOUT ANY WARRANTY; without even the implied warranty of
  * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
  * GNU General Public License for more details.
  *
- * You should have received a copy of the GNU General Public License
- * along with Access to Memory (AtoM).  If not, see <http://www.gnu.org/licenses/>.
+ * You should have received a copy of the GNU General Public License along with
+ * Access to Memory (AtoM).  If not, see <http://www.gnu.org/licenses/>.
```
jraddaoui: This is going to be needed in **a lot** of files. We still need to decide the standard we want to follow (re: #13317) but both, PSR-2 and PSR-12, state that _"There MUST NOT be a hard limit on line length; the soft limit MUST be 120 characters; lines SHOULD be 80 characters or less"_.

https://www.php-fig.org/psr/psr-12/#23-lines

So maybe we should not worry that much about it, specially in here.

djjuhasz: Yes, you have a good point.  The thing that galls me about the current statement is that it takes just as many lines when it's wrapped at 80 characters. 😜 

I find that PSR statement on line lengths to be so confusing and ambiguous - but I agree that it allows for the current statement as is. 😕

apps/qubit/modules/clipboard/actions/exportAction.class.php
```diff
@@ -432,30 +433,20 @@ protected function addField($name)
...
+              'It is not possible to select both digital objects and'
+              . ' descendants for export at the same time. Digital objects can'
+              . ' only be exported for records that are on the clipboard.'
+            );
           }
+
+          $this->helpMessages[] = __(
+            'Digital objects with restricted access or copyright will not'
+            . ' be exported.'
+          );
```
jraddaoui, line 445: This should not appear for repositories.

lib/job/arActorExportJob.class.php
```diff
@@ -0,0 +1,91 @@
...
+ * (at your option) any later version.
+ *
+ * Access to Memory (AtoM) is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with Access to Memory (AtoM).  If not, see <http://www.gnu.org/licenses/>.
+ */
```
jraddaoui, line 18: For example ... :stuck_out_tongue:

If you really want to change this, maybe do a search and replace over the entire AtoM folder in a different commit.

lib/job/arActorExportJob.class.php
```diff
@@ -0,0 +1,91 @@
...
+   */
+  protected function doExport($path)
+  {
+    $search = self::findExportRecords($this->params);
+
+    // Scroll through results then iterate through resulting IDs
+    foreach (
+      arElasticSearchPluginUtil::getScrolledSearchResultIdentifiers($search)
+      as $id
+    ) {
```
jraddaoui: New line for the bracket. This is one case I'd prefer in one line even if it's higher than 80 chars.

lib/job/arExportJob.class.php
```diff
@@ -229,126 +241,104 @@ protected function addDigitalObjects(&$zip, &$errors)
...
-  protected function getDigitalObjects()
+  protected function addDigitalObject($resource, $tempDir)
   {
-    // Prepare array for digital object ids
-    $digitalObjects = array();
+    // Skip if the includeDigitalObjects option is not set, or it is disabled
+    if (
+      !isset($this->params['includeDigitalObjects'])
+      || !$this->params['includeDigitalObjects']
+    ) {
```
jraddaoui: New line for bracket.

lib/job/arExportJob.class.php
```diff
@@ -229,126 +241,104 @@ protected function addDigitalObjects(&$zip, &$errors)
...
+          && $this->user->isAuthenticated()
+          && QubitAcl::check($resource, 'read')
+        ) || (
+          'informationObject' == $this->params['objectType']
+          && QubitAcl::check($resource, 'readMaster')
+          && QubitGrantedRight::checkPremis($resource->id, 'readMaster')
+          && !$digitalObject->hasConditionalCopyright()
+        )
+      )
+    ) {
```
jraddaoui: New line for bracket.

lib/job/arExportJob.class.php
```diff
@@ -229,126 +241,104 @@ protected function addDigitalObjects(&$zip, &$errors)
...
+      // export is allowed
```
jraddaoui: Uppercase first.

lib/job/arInformationObjectCsvExportJob.class.php
```diff
@@ -23,232 +23,82 @@
...
-          $job->info($job->i18n->__(
-            '%1 items exported.', array('%1' => $itemsExported))
-          );
-        }
-
-        $itemsExported++;
+    // Export descendants if option was selected
+    if (!$this->params['current-level-only'])
+    {
+      foreach ($resource->getDescendantsForExport($options) as $item) {
```
jraddaoui: New line for bracket.

lib/job/arInformationObjectExportJob.class.php
```diff
@@ -0,0 +1,172 @@
...
+   * @return arElasticSearchPluginQuery  AtoM Elasticsearch query
+   */
+  static public function getCurrentArchivalStandard()
+  {
+    // If not using RAD, default to ISAD CSV export format
+    $archivalStandard = 'isad';
+    if (
+      QubitSetting::getByNameAndScope('informationobject', 'default_template')
+      == 'rad'
+    ) {
```
jraddaoui: New line for bracket. And another one I'd prefer in a single line.

lib/job/arInformationObjectExportJob.class.php
```diff
@@ -0,0 +1,172 @@
...
+    if (0 == $search->count())
+    {
+      return;
+    }
+
+    // Scroll through results then iterate through resulting IDs
+    foreach (
+      arElasticSearchPluginUtil::getScrolledSearchResultIdentifiers($search)
+      as $id
+    ) {
```
jraddaoui: New line for bracket. And the same.

lib/job/arExportJob.class.php
```diff
@@ -255,18 +258,34 @@ protected function addDigitalObject($resource, $tempDir)
 
     $digitalObject = $resource->getDigitalObject();
 
+    // Skip this digital object if it doesn't exist in the database, is a URI
+    // for an external object, or export is not authorized
     if (
       empty($digitalObject)
       || !$this->allowDigitalObjectExport($resource, $digitalObject)
     ) {
```
jraddaoui: Not sure if I missed this one.

lib/job/arExportJob.class.php
```diff
@@ -255,18 +258,34 @@ protected function addDigitalObject($resource, $tempDir)
...
       return false;
     }
 
+    // Don't try to export an external digital object
+    if (QubitTerm::EXTERNAL_URI_ID == $digitalObject->usageId)
```
jraddaoui: I think we may want to check OFFLINE_ID too (I think it comes only from the metadata DIP upload from AM).

jraddaoui (changes requested):
Looks great @djjuhasz, thanks!

Only a couple of important comments, all the other cosmetics.

lib/job/arActorExportJob.class.php
djjuhasz: Fixed

lib/job/arExportJob.class.php
djjuhasz: I like the `) {` convention (obviously 😜), but I think you are right that putting open bracket on a new line is canonical. 😢

djjuhasz: Fixed

apps/qubit/modules/clipboard/actions/exportAction.class.php
djjuhasz: Reverted

apps/qubit/modules/clipboard/actions/exportAction.class.php
djjuhasz: It doesn't, the `$this->digitalObjectsAvailable` flag above is false for the repository type.

lib/job/arInformationObjectCsvExportJob.class.php
djjuhasz: Fixed

lib/job/arInformationObjectExportJob.class.php
djjuhasz: Done. 😄

lib/job/arInformationObjectExportJob.class.php
djjuhasz: Done.

djjuhasz (commented):
Feedback addressed :)

lib/job/arExportJob.class.php
djjuhasz: Oops, missed this one in my commit :(

lib/job/arExportJob.class.php
djjuhasz: Do you mean it doesn't check if it's an external URI like the comment indicates, or the `) {`?

jraddaoui: Yeah, the bracket.

jraddaoui (approved):
Thanks @djjuhasz!

Just a few not addressed comments, the only one that may be important is to check the OFFLINE_ID usage type.

lib/job/arExportJob.class.php
djjuhasz: Addressed in commit 9a5ba5e3a

lib/job/arExportJob.class.php
djjuhasz: Okay, got this in a *third* commit 🤦

lib/job/arExportJob.class.php
djjuhasz: Fixed

djjuhasz:
> Just a few not addressed comments, the only one that may be important is to check the OFFLINE_ID usage type.

Sorry, missed this the first time I went through your comments. :(  I've addressed it now.

jraddaoui (approved):
:+1:

PR #1178, opened 2020-08-21
Remove redundant QubitObject method

djjuhasz:
Remove QubitObject::getDigitalObjectRelatedByobjectId() method which duplicates
QubitObject::getDigitalObject() functionality

### DIFF
--- a/lib/model/QubitObject.php
+++ b/lib/model/QubitObject.php
@@ -390,24 +390,14 @@ class QubitObject extends BaseObject implements Zend_Acl_Resource_Interface
    */
   public function getDigitalObject()
   {
-    $digitalObjects = $this->getDigitalObjectsRelatedByobjectId();
+    $digitalObjects = $this->digitalObjectsRelatedByobjectId;
+
     if (count($digitalObjects) > 0)
     {
       return $digitalObjects[0];
     }
-    else
-    {
-      return null;
-    }
-  }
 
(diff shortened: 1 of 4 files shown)


lib/model/QubitObject.php
```diff
@@ -390,24 +390,14 @@ public function getNotesByTaxonomy(array $options = array())
...
-  }
 
-  public function getDigitalObjectRelatedByobjectId()
-  {
-    $digitalObjects = $this->getDigitalObjectsRelatedByobjectId();
-    if (0 < count($digitalObjects))
-    {
-      return $digitalObjects[0];
-    }
+    return null;
```
jraddaoui, line 400: No need to return null.

jraddaoui (approved):

lib/model/QubitObject.php
djjuhasz: Yeah, I debated that.  I know that `return null;`, `return;` and omitting a final `return` are all functionally equivalent, but I think there is an argument for explicitly using `return null;` when a function usually returns a value.  

This StackOverflow question debates the point, but the most popular answer is to use `return null;` when a function usually returns a value: https://stackoverflow.com/questions/8749091/should-a-function-use-return-null

The symfony coding standard says "Use return null; when a function explicitly returns null values and use return; when the function returns void values;" -- https://symfony.com/doc/current/contributing/code/standards.html

PR #1179, opened 2020-08-23
Fix slug transliteration in Docker, refs #13406

jraddaoui:
Install gnu-libiconv and set LD_PRELOAD env. var. to make `iconv()` work
as expected in Alpine/Musl.


djjuhasz (approved):
Looks good to me 👍

PR #1180, opened 2020-08-27
Update digital object metadata display, refs #13388

replaceafill:
This updates the `Digital object metadata` section of the index views with two collapsible fieldsets that look like this:

Screenshot_2020-08-27 Beihai, Guanxi, China, 1988 - Demo site

The rest of the changes include:

* New properties have been added to information objects to store METS information about the original and preservation files.
* New visibility settings have been added to control what fields are displayed in the new areas.
* A new preservation system access statement setting has been added (similar to the copyright statement) to show a message to unauthenticated users on how to get access to the preservation copies

### DIFF
--- a/apps/qubit/modules/digitalobject/actions/metadataComponent.class.php
+++ b/apps/qubit/modules/digitalobject/actions/metadataComponent.class.php
@@ -28,34 +28,320 @@ class DigitalObjectMetadataComponent extends sfComponent
 {
   public function execute($request)
   {
-    $id = $this->object->id;
-    $this->denyFileNameByPremis = false;
+    // Check related object type to display IO properties in the template
+    $this->relatedToIo = $this->resource->object instanceOf QubitInformationObject;
+    $this->relatedToActor = $this->resource->object instanceOf QubitActor;
 
-    // Special case: If all digital object representations are
-    // denied by premis, we will hide the filename when displaying
-    // the digital object metadata for security reasons. Only check
-    // rights for info obj linked digital objects.
-    if (!QubitGrantedRight::checkPremis($id, 'readThumb') &&
-        !QubitGrantedRight::checkPremis($id, 'readReference') &&
-        !QubitGrantedRight::checkPremis($id, 'readMaster') &&
-        $this->object instanceOf QubitInformationObject)
+    $this->user = $this->context->getUser();
+    $this->storageServicePluginEnabled = (
+      $this->context->getConfiguration()->isPluginEnabled('arStorageServicePlugin')
+      && arStorageServiceUtils::getAipDownloadEnabled()
+    );
+
+    $this->referenceCopy = $this->resource->getRepresentationByUsage(QubitTerm::REFERENCE_ID);
+    $this->thumbnailCopy = $this->resource->getRepresentationByUsage(QubitTerm::THUMBNAIL_ID);
+
+    // An authenticated user can download the original preservation file only if the Storage Service plugin is enabled
+    $this->canAccessOriginalFile = (
+      $this->user->isAuthenticated()
+      && $this->relatedToIo
+      && $this->storageServicePluginEnabled
+    );
+
+    // The preservation copy cannot be downloaded from AtoM
+    $this->canAccessPreservationCopy = false;
+
+    // AtoM's representations cannot be downloaded by default
+    $this->canAccessMasterFile = false;
+    $this->canAccessReferenceCopy = false;
+    $this->canAccessThumbnailCopy = false;
+
+    // Reasons for denying access to a AtoM's representations based on PREMIS rights
+    $this->masterFileDenyReason = false;
+    $this->referenceCopyDenyReason = false;
+    $this->thumbnailCopyDenyReason = false;
+
+    // Group permissions always control access to AtoM's representations if the digital object is related to an actor
+    if ($this->relatedToActor && QubitAcl::check($this->resource->object, 'read'))
     {
-      $this->denyFileNameByPremis = true;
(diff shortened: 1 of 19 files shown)


apps/qubit/modules/digitalobject/actions/metadataComponent.class.php
```diff
@@ -28,34 +28,333 @@ class DigitalObjectMetadataComponent extends sfComponent
...
+    $this->showOriginalFileMetadata = $this->setOriginalFileShowProperties();
+    $this->showPreservationCopyMetadata = $this->setPreservationCopyShowProperties();
+    $this->showMasterFileMetadata = $this->setMasterFileShowProperties();
+    $this->showReferenceCopyMetadata = $this->setReferenceCopyShowProperties();
+    $this->showThumbnailCopyMetadata = $this->setThumbnailCopyShowProperties();
+  }
+
+  protected function isPreservationSystemAccessStatementEnabled()
+  {
+    if (null !== $setting = QubitSetting::getByName('digitalobject_preservation_system_access_statement_enabled'))
```
mcantelon: Would it work to use `sfConfig::get` to get the property instead?

apps/qubit/modules/digitalobject/actions/metadataComponent.class.php
```diff
@@ -28,34 +28,333 @@ class DigitalObjectMetadataComponent extends sfComponent
...
+        return (bool)$value;
+      }
+    }
+  }
+
+  protected function getPreservationSystemAccessStatement()
+  {
+    if (
+      $this->isPreservationSystemAccessStatementEnabled()
+      && (null !== $setting = QubitSetting::getByName('digitalobject_preservation_system_access_statement'))
```
mcantelon: `sfConfig::get` instead?

apps/qubit/modules/digitalobject/actions/metadataComponent.class.php
```diff
@@ -28,34 +28,333 @@ class DigitalObjectMetadataComponent extends sfComponent
...
+    ) {
+        $value = $setting->getValue(array('sourceCulture' => true));
+        if (!empty($value))
+        {
+          return (string)$value;
+        }
+    }
+  }
+
+  protected function isEmpty($value)
```
mcantelon, line 122: Is this being used instead of `empty()` because `empty()` would report `0` as empty?

apps/qubit/modules/digitalobject/actions/metadataComponent.class.php
```diff
@@ -28,34 +28,333 @@ class DigitalObjectMetadataComponent extends sfComponent
...
     // Provide longitude to template
-    $longitudeProperty = $this->object->digitalObjectsRelatedByobjectId[0]->getPropertyByName('longitude');
+    $longitudeProperty = $this->resource->object->digitalObjectsRelatedByobjectId[0]->getPropertyByName('longitude');
     $this->longitude = $longitudeProperty->value;
 
-    // Check related object type to display IO properties in the template
-    $this->relatedToIo = $this->resource->object instanceOf QubitInformationObject;
-    $this->relatedToActor = $this->resource->object instanceOf QubitActor;
+    $this->showMasterFileGoogleMap = (
+      (bool)sfConfig::get('app_toggleDigitalObjectMap')
```
mcantelon: Is the casting of the value as bool to deal with the setting being null? Seems like we normally use this way of dealing with null values: `sfConfig::get('app_toggleDigitalObjectMap', false)`.

apps/qubit/modules/digitalobject/actions/metadataComponent.class.php
```diff
@@ -28,34 +28,333 @@ class DigitalObjectMetadataComponent extends sfComponent
...
+      && is_numeric($this->latitude)
+      && is_numeric($this->longitude)
+      && (bool)$this->googleMapsApiKey
+    );
+
+    $this->showMasterFileGeolocation = (
+      (bool)check_field_visibility('app_element_visibility_digital_object_geolocation')
```
mcantelon: `check_field_visibility` should, I think, always return a bool so not sure if we need to cast it as bool.

lib/QubitMetsParser.class.php
```diff
@@ -1166,4 +1166,217 @@ public function getOriginalPathInAip($fileId)
...
+   *
+   * @param string $fileId METS FILEID
+   *
+   * @return SimpleXMLElement the <mets:file> element
+   */
+  protected function getOriginalFile($fileId)
+  {
+    if (
+      (false !== $file = $this->document->xpath(
+        '//m:fileSec/m:fileGrp[@USE="original"]/m:file[@ID="'.$fileId.'"]'
```
mcantelon: Nitpick: using sprintf instead of string concatenation's a bit cleaner, but not a big deal either way.

mcantelon (commented):
Nicely done! A few questions/concerns, but nothing major.

apps/qubit/modules/digitalobject/actions/metadataComponent.class.php
replaceafill: Yes. Is there a better approach?

replaceafill:
Thanks for the code review and your recommendations @mcantelon! :bow: I think I addressed all your observations.

apps/qubit/modules/digitalobject/actions/metadataComponent.class.php
mcantelon: No, your approach seems correct given that!

mcantelon (approved):
Great work... looks good! 👍

PR #1181, opened 2020-08-27
Avoid slow descendant dates query, refs #13407

djjuhasz:
- Remove descendantEventTypes check when determining whether to show "Calculate
  dates" link as the query can be very slow when the resource has 100,000+
  descendants
- Do authentication and children checks before running the descendant event
  types query, to avoid running it when it's not necessary
- Show an error page with a user-friendly warning if information object has no
  descendants with event dates
- Show an error page with a user-friendly warning when trying to calculate
  accumulated dates on an information object with no children, instead of a
  generic 403 error

### DIFF
--- a/apps/qubit/modules/informationobject/actions/calculateDatesAction.class.php
+++ b/apps/qubit/modules/informationobject/actions/calculateDatesAction.class.php
@@ -17,12 +17,10 @@
  * along with Access to Memory (AtoM).  If not, see <http://www.gnu.org/licenses/>.
  */
 
-class InformationObjectCalculateDatesAction extends DefaultEditAction
+class InformationObjectCalculateDatesAction extends sfAction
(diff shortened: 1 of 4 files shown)


mcantelon (approved):
Other than some lines in the commit message being longer than 72 chars (I think we're going by this standard, but if not then disregard: https://tbaggery.com/2008/04/19/a-note-about-git-commit-messages.html), everything looks good! 👍

djjuhasz:
> Other than some lines in the commit message being longer than 72 chars (I think we're going by this standard, but if not then disregard: https://tbaggery.com/2008/04/19/a-note-about-git-commit-messages.html), everything looks good! 👍

Thanks @mcantelon. I reformatted the commit message to wrap lines to 72 chars or less. :)

PR #1182, opened 2020-08-31
Fix storage service message, refs #13351

sbreker:
Fixed exception message to correctly display the aip UUID.


djjuhasz (approved):
Looks good  👍

PR #1184, opened 2020-09-01
Stop index update on uploadLimit save, refs #13412

djjuhasz:
- In `QubitRepository::save()` only update search index if `$indexOnSave` is true
- In `RepositoryEditUploadLimitAction` set `$indexOnSave` to false before saving
  `QubitRepository` changes


mcantelon (approved):
Looks good! 👍

PR #1185, opened 2020-09-02
Add repo upload limits setting, refs #13411

djjuhasz:
- Create a new "Uploads" setting page, and move three existing settings
  from the "Global" page
- Add a setting to allow disabling per repository "Upload limits"
- Don't load the "Upload limits" component when it is disabled in the
  settings
- Move the QubitAcl check for displaying the `uploadLimit` component
  from the `_contextMenu.php` template to the `uploadLimit` controller
- Cosmetic: Sort settings menu array alphabetically by label
- Add additional comments


mcantelon (approved):
Looks good!

PR #1186, opened 2020-09-03
Fix for running single migrations, refs #13368

mcantelon:


lib/task/migrate/arUpgradeSqlTask.class.php
```diff
@@ -336,7 +336,7 @@ protected function updateDatabaseVersion($version)
    * @param int $previousMilestone  Previous milestone
    * @param int $currentMilestone  Current milestone
    */
-  private function runMigrationsInDirectory($migrationsDirectory, $previousMilestone, $currentMilestone)
+  private function runMigrationsInDirectory($version, $migrationsDirectory, $previousMilestone, $currentMilestone)
```
replaceafill, line 342: Add `$version` to the `@param`s list?

replaceafill (commented):
I tested your branch and I can now upgrade to the latest version. However the task output seems to be printing the current version (186 in my case) and not the migrated one (it should be 188 for me).

replaceafill (approved):
Tested the last commit and everything works as expected! :shipit:

PR #1187, opened 2020-09-10
Get digital object count from ES, refs #13415

djjuhasz:
- Query the Elasticsearch index to get a count of descendant digital
  objects instead of a slow database query
- Wrap long lines


mcantelon (approved):
Good solution! Good looks clean as a whistle!

PR #1188, opened 2020-09-15
Remove user admin options for CAS, refs #13413

tw4l:
This commit removes AtoM user management UI options that are made redundant by CAS authentication when the `arCasPlugin` is enabled. This includes the Reset password link, the option to create new users from within the AtoM GUI, and the option to the username, email, and password fields from the Edit user screen. Other Edit profile options such as marking users as active or inactive, API key generation, translation languages, and group management are retained.

When the `arCasPlugin`'s `set_groups_from_attributes` is enabled, user groups are reset on each login based on values in CAS attributes and so any changes made to group membership in the AtoM GUI will be nullified the next time the user logs in. This should be noted in the AtoM docs. See: https://github.com/artefactual/atom-docs/pull/162.

Connected to https://projects.artefactual.com/issues/13413

### DIFF
--- a/plugins/arCasPlugin/modules/user/actions/editAction.class.php
+++ b/plugins/arCasPlugin/modules/user/actions/editAction.class.php
@@ -1,0 +1,315 @@
+<?php
+
+/*
+ * This file is part of the Access to Memory (AtoM) software.
+ *
+ * Access to Memory (AtoM) is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as published by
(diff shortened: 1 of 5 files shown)


sbreker:
Looks great Tessa! I think this strategy makes sense (to override these controllers and templates to create CAS-ified user maintenance screens). This keeps the CAS specific logic self contained in the plugin!

sbreker (approved):
👍

tw4l:
Merged as https://github.com/artefactual/atom/commit/2a65e665ee01560f577737af603c1b8687e57a23, closing manually.

PR #1189, opened 2020-09-17
Use basic auth for DIP Upload regardless of authentication scheme

tw4l:
This commit adds a new `authenticateWithBasicAuth` method to AtoM's base `myUser` user class and uses the new method to allow Archivematica to authenticate for DIP Upload using basic auth even when a single sign-on method such as CAS is enabled in AtoM.

This implementation allows user classes that subclass `myUser` to retain or void the basic auth method as needed by keeping or overriding `authenticateWithBasicAuth`.

Connected to https://projects.artefactual.com/issues/13416


mcantelon (approved):
Looks good to me! 👍

PR #1190, opened 2020-09-17
Fix repeated rows in CSV export, refs #13395

djjuhasz:
- getDescendantsForExport() was being called recursively for each
  descendant, exponentially increasing the number of rows exported
  for information object CSV exports with descendants
- Normalize export progress log code
- Log number of items found (without descendants) at start of the job
  to help debugging

### DIFF
--- a/lib/job/arExportJob.class.php
+++ b/lib/job/arExportJob.class.php
@@ -27,6 +27,9 @@
 
 class arExportJob extends arBaseJob
 {
+  // Log progress every n rows
+  const LOG_INTERVAL = 100;
+
   // Child class should set this if creating user downloads
   protected $downloadFileExtension = null;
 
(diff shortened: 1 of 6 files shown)


lib/job/arExportJob.class.php
```diff
@@ -364,4 +367,19 @@ protected function getUniqueFilename($filepath)
       $pathinfo['extension']
     );
   }
+
+  /**
+   * Log export progress every LOG_INTERVAL rows
```
mcantelon: Nitpick: maybe add the cache clearing to the description of the function.

mcantelon (approved):
Nice PR! Other than one comment nitpick, seems good to go!

lib/job/arExportJob.class.php
djjuhasz: Done!

PR #1191, opened 2020-09-18
Fix error viewing actor with DO, refs #13388

djjuhasz:
Fix 500 error when authenticated and viewing an actor with a digital
object.  Actors will never have Archivematica AIP digital object
metadata (i.e. original file and preservation copy) so avoid checking
if these metadata exist.

### DIFF
--- a/apps/qubit/modules/digitalobject/actions/metadataComponent.class.php
+++ b/apps/qubit/modules/digitalobject/actions/metadataComponent.class.php
@@ -98,9 +98,16 @@ class DigitalObjectMetadataComponent extends sfComponent
     // Statement shown as the Permissions field for preservation copies
     $this->accessStatement = $this->getPreservationSystemAccessStatement();
 
-    // What metadata sections should be shown
-    $this->showOriginalFileMetadata = $this->setOriginalFileShowProperties();
-    $this->showPreservationCopyMetadata = $this->setPreservationCopyShowProperties();
+    // Determine which metadata sections should be shown
+    if ($this->relatedToIo)
+    {
+      // Only an information object can have Archivematica "original file" and
+      // "preservation copy" metadata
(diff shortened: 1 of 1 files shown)


apps/qubit/modules/digitalobject/actions/metadataComponent.class.php
```diff
@@ -61,10 +61,15 @@ public function execute($request)
     $this->referenceCopyDenyReason = false;
     $this->thumbnailCopyDenyReason = false;
 
-    // Group permissions always control access to AtoM's representations if the digital object is related to an actor
-    if ($this->relatedToActor && QubitAcl::check($this->resource->object, 'read'))
```
replaceafill: Maybe I misunderstood how permissions work, but if we remove the `QubitAcl::check` doesn't that mean we can't limit access anymore? I think I picked up that logic for actors from the `Must have read privileges to view` comment (and related changes) in https://github.com/artefactual/atom/commit/77f91b2909f276950ae295b55cbb8fdce589dbdc

replaceafill (commented):
Thanks for catching this @djjuhasz. The change looks good to me, but I left a question regarding permissions for actors.

apps/qubit/modules/digitalobject/actions/metadataComponent.class.php
replaceafill: Oh, following the permissions checks a bit in the code. Maybe the check is irrelevant because if you don't have `read` permissions on an actor you're not even supposed to be accessing its index view?

djjuhasz: After carefully considering and reviewing the digital object permissions, I have come to the conclusion I don't want to touch them with a 10 foot pole. 😆   

I'll remove this change from the PR, and resubmit. 😅

replaceafill (approved):
:shipit:

PR #1192, opened 2020-09-18
Fix generate identifier counter, refs #13418

jraddaoui:
Add "usingMask" input even when the identifier mask is not enabled as
this field is modified when the "Generate identifier" link is used to
indicate that the counter needs to be incremented.


djjuhasz (approved):
Looks good. :)

PR #1194, opened 2020-09-21
Fix EAC descriptive note display, refs #13421

jraddaoui:
Mixing a tag and the content in the `render_value` call breaks the XML
output and it's not necessary as `render_value` already adds the
paragraph tags.


djjuhasz (approved):
Looks good to me! 👍

jraddaoui:
Thanks @djjuhasz!

PR #1196, opened 2020-09-22
Hide AIP Download links, refs #13351

sbreker:
This commit fixes an issue where the aip and file download links
were still being displayed for authenticated users even when the
arStorageService security.yml credentials should prevent this.

A check has been added to prevent display of these download links
when a user's group is not in the credentials section of the
arStorageService plugin security.yml file.

### DIFF
--- a/apps/qubit/modules/digitalobject/templates/_metadata.php
+++ b/apps/qubit/modules/digitalobject/templates/_metadata.php
@@ -66,7 +66,7 @@
               <?php echo render_show(
                 __('AIP UUID'),
                 render_value($resource->object->aipUUID),
-                 array('fieldLabel' => 'aipUUID')
+                array('fieldLabel' => 'aipUUID')
               ) ?>
             <?php endif; // arStorageService is disabled ?>
           <?php endif; ?>
--- a/plugins/arStorageServicePlugin/modules/arStorageService/templates/_aipDownload.php
+++ b/plugins/arStorageServicePlugin/modules/arStorageService/templates/_aipDownload.php
@@ -2,10 +2,12 @@
     <h3><?php echo __('File UUID') ?></h3>
     <div class="aip-download">
         <?php echo render_value_inline($resource->object->objectUUID) ?>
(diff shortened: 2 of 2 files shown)


apps/qubit/modules/digitalobject/actions/metadataComponent.class.php
```diff
@@ -36,6 +36,8 @@ public function execute($request)
     $this->storageServicePluginEnabled = (
       $this->context->getConfiguration()->isPluginEnabled('arStorageServicePlugin')
       && arStorageServiceUtils::getAipDownloadEnabled()
+      && $this->user->checkModuleActionAccess('arStorageService', 'download')
+      && $this->user->checkModuleActionAccess('arStorageService', 'extractFile')
```
djjuhasz: Is it necessary to check for both "download" and "extractFile" permissions? I would expect checking for 'download' authorization to be sufficient.

djjuhasz (commented):
One question @sbreker :)

apps/qubit/modules/digitalobject/actions/metadataComponent.class.php
sbreker: This is because there are two plugin actions to protect and they _could_ be configured independently in the security.yml file (This code is using the security.yml file to determine user access): 

https://github.com/artefactual/atom/blob/qa/2.x/plugins/arStorageServicePlugin/modules/arStorageService/config/security.yml

I am checking both because the component being protected will display both download links.

djjuhasz: @sbreker I think it would be better to make the secuirty.yml checks specific to the relevant link then, rather then disabling both links when only one is authorized.

sbreker: @djjuhasz Good idea! See new commit - I have moved the checks into the Storage Service plugin WebUI component. This allows individual config of the two download links (see commit message).

djjuhasz (approved):
Looks good @sbreker 👍

PR #1197, opened 2020-09-29
Create plugin i18n folders, refs #13426

sbreker:
This commit ensures that plugin folders contain an i18n folder
before extracting strings from AtoM. It fixes an error where AtoM's
i18n extract task was failing if a plugin did not contain an i18n
folder.


jraddaoui (approved):
Looks good!

`i18n-create-missing-plugin-i18n-folders` may be an easier name, but not a big deal of course.

PR #1198, opened 2020-09-29
i18n rectify plugin fix, refs #13427

sbreker:
Update AtoM's i18n rectify task to allow an existing translation to be
overwritten when updated translations are imported from Weblate.

### DIFF
--- a/lib/task/i18n/i18nRectifyTask.class.php
+++ b/lib/task/i18n/i18nRectifyTask.class.php
@@ -99,7 +99,7 @@ EOF;
       {
         foreach ($translations as $key => &$value)
         {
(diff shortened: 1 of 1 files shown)


lib/task/i18n/i18nRectifyTask.class.php
```diff
@@ -99,7 +99,7 @@ public function execute($arguments = array(), $options = array())
       {
         foreach ($translations as $key => &$value)
         {
-          if (0 == strlen(trim($value[0])) && isset($currentMessages[$key]))
+          if (isset($currentMessages[$key]))
```
jraddaoui, line 102: (╯°□°）╯︵ ┻━┻

jraddaoui (approved):
Thanks!

PR #1202, opened 2020-09-30
Refactor treeview display code

djjuhasz:
Refactor the code that builds the node text for the full width treeview


djjuhasz:
I did this work as a precursor to a fix for https://projects.artefactual.com/issues/13423, but it didn't seem directly related to the ticket, so I didn't "ref" it.  I'm not sure if that was the right choice.

djjuhasz:
I've got some more refactoring to do

PR #1204, opened 2020-10-07
Remove extraneous ')' from array, refs #13429

djjuhasz:
With thanks to https://github.com/xojc for finding the error and
suggesting the fix.

Cosmetic: reformat code to make it easier to distinguish between array
elements and the arguments of the `getI18nFieldNames()` method.


mcantelon (approved):
Looks good to me!

PR #1205, opened 2020-10-16
Get treeview data from Elasticsearch, refs #13423

djjuhasz:
- Refactor full width treeview code
- Get information object data for the treeview from Elasticsearch 
   because it is already normalized, runs faster, and uses less memory
- Add information_object.lft column data to Elasticsearch index to
  allow sorting search results on the column
- Wrap long lines

### DIFF
--- a/apps/qubit/modules/default/actions/fullTreeViewAction.class.php
+++ b/apps/qubit/modules/default/actions/fullTreeViewAction.class.php
@@ -37,48 +37,43 @@ class DefaultFullTreeViewAction extends sfAction
     $this->getResponse()->setContentType('application/json');
   }
 
-  protected function getNodeOrChildrenNodes($id, $baseReferenceCode, $children = false, $options = array())
+  /**
+   * Get treeview node data for a description or it's children
+   *
+   * @param int $id information_object.id value
+   * @param string $baseReferenceCode reference code of parent node
+   * @param bool $children return children of target node when true
+   * @param array $options optional parameters
+   *
+   * @return array full width treeview data
+   */
+  protected function getNodeOrChildrenNodes(
+    $id,
+    $baseReferenceCode,
+    $children = false,
+    $options = array()
+  )
   {
-    $i18n = sfContext::getInstance()->i18n;
-    $culture = $this->getUser()->getCulture();
+    // Hide drafts from unauthenticated users
+    $draftsSql = $this->getDraftsCriteria();
 
-    if (sfConfig::get('app_markdown_enabled', true))
-    {
-      $untitled = '_'.$i18n->__('Untitled').'_';
-    }
-    else
-    {
-      $untitled = '<em>'.$i18n->__('Untitled').'</em>';
-    }
+    // Get current node data (for the collection root on the first load) or
+    // children data
+    $condition = $children ? 'io.parent_id = :id' : 'io.id = :id';
 
-    // Remove drafts for unauthenticated users
-    $draftsSql = !$this->getUser()->user ? 'AND status.status_id <> ' . QubitTerm::PUBLICATION_STATUS_DRAFT_ID : '' ;
+    // Sort column
+    $orderColumn = (isset($options['orderColumn'])) ?
+      $options['orderColumn'] : 'io.lft';
 
-    // Get current node data (for the collection root on the first load) or children data
-    $condition = $children ? 'io.parent_id = :id' : 'io.id = :id' ;
-
-    // Determine ordering
-    $orderColumn = (isset($options['orderColumn'])) ? $options['orderColumn'] : 'io.lft';
-
-    // If we're currently fetching children and paging options are set,
-    // use paging options to assemble LIMIT clause
-    $limitClause = "";
-    $skip = (isset($options['skip'])) ? $options['skip'] : null;
-    $limit = (isset($options['limit'])) ? $options['limit'] : null;
-
-    if ($children && (ctype_digit($skip) || ctype_digit($limit)))
-    {
-      $limitClause = "LIMIT ";
-      $limitClause .= (ctype_digit($skip)) ? $skip : "0";
-      $limitClause .= (ctype_digit($limit)) ? ", ". $limit : "";
-    }
+    // Limit results based on paging $options[skip, limit]
+    $limitClause = $this->getLimitClause($children, $options);
 
     $sql = "SELECT SQL_CALC_FOUND_ROWS
       io.id, io.lft, io.rgt,
       io.source_culture,
       io.identifier,
       current_i18n.title AS current_title,
-      IFNULL(source_i18n.title, '$untitled') as text,
+      source_i18n.title as source_title,
       io.parent_id AS parent,
       slug.slug,
       IFNULL(lod.name, '') AS lod,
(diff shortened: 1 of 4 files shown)


jraddaoui:
Hi @djjuhasz,

I only read the PR description so I'm not sure if you took in consideration the recent changes from the CSV import, where the nested set is not updated until the end of the process (without updating those values in ES):

https://projects.artefactual.com/issues/13356
https://projects.artefactual.com/issues/13354

The idea was to use similar changes in other import process to improve their performance, with a bigger goal to remove the nested set altogether:

https://projects.artefactual.com/issues/13385
https://projects.artefactual.com/issues/13240

I can see the benefits of using ES for this matter and we still need to evaluate the pros of fully removing the nested set (better hierarchy updates) versus its cons (worst hierarchy reads), but I think this change goes in a different direction and that we should think a bit more about how we manage hierarchical data and how it's passed to ES so we can use it in places like the tree-view for reads without major impact in process like imports for updates.

Please, let me know your thoughts and sorry for not being involved in this sooner, I didn't know it was coming.

djjuhasz:
@jraddaoui I didn't switch to Elasticsearch because of the nested set queries - I did it because getting the data from MySQL required six joins and a one-to-many query on the `event` table for each row returned.  The query also caused PHP to eat up a ton of memory (over 4 GB).    I started trying to optimize the query by using a recursive CTE to traverse the hierarchy, but pretty quickly came to the conclusion that it wasn't going to significantly improve the query performance.

All of the data required for a full width treeview node is in a single ES information object document, so it's naturally much faster and more memory efficient to get the data from ES.

jraddaoui:
@djjuhasz, I see that, I think we are using a recursive process with lazy load, not nested set queries. But I can see the benefits of using ES, as I said.

My only concern is that we can't trust the `lft` values to sort siblings after the latest changes from the CSV import. In fact, we can't even trust them in the DB (see [#13414](https://projects.artefactual.com/issues/13414)), but in ES they will be empty unless the index is rebuilt after the import.

I'm not saying this changes are not good or that they won't benefit the tree-view performance, just that we may move in a different direction to manage hierarchical data, which will require to modify this implementation again (as it would with the current one). I'd love to see all happening together but I guess that won't be the case, since removing the nested set entirely is not a minor task and you already have these changes almost ready. And after all, we may only need to change the sort field for the tree-view as it's already a recursive implementation. Nevertheless, until them, I think we should consider the issues mentioned with the `lft` values.

djjuhasz:
Ah, thanks for explaining further @jraddaoui.

Is there an alternative column I can use to sort siblings other than `lft` that will match the previous treeview sorting (which was on the `lft` value)?  I understand that the ES `lft` value might be out of date with the MySQL `lft` value, but I worry changing the treeview sort will annoy AtoM users that have sorted their descriptions in a meaningful way.

jraddaoui:
Yeah, there is no alternative so far, not until we move to that "order" column if we decide to do so at some point. And it's hard to see how can we trust the `lft` value without going back to the performance/deadlock issues.

As a workaround, maybe we could use a placeholder left value in the import process, auto-incrementing for siblings without updating the nested set in other rows from the IO table. That value will go to ES until the nested set update changes those values in the DB at the end of the import, and it could act as an order column because the process is already recursive. Just an initial idea as there may be a lot of things to consider, for example existing siblings to start the placeholder counter or how rebuilding the index will probably change those values after the nested set update; although, I think it could help with #13414 too, so maybe it's worth a try.

djjuhasz:
@jraddaoui hmm, I've been wrestling with how to proceed on this Pull Request all morning.  🙁 

To summarize my understanding of the problem:

1. [Issue 13414](https://projects.artefactual.com/issues/13414): CSV import no longer updates the `lft` column in the database during the import (to fix [issue 13354](https://projects.artefactual.com/issues/13354)), so the `lft` values for siblings created by an import do not match the original ordering in the CSV import file .
2. Because the `lft` values in the database are not populated until the end of the import process, if the `--index` import option is used the Elasticsearch documents created during the import will have an empty `lft` value
3. Running a full `search:populate` after an import will populate the Elasticsearch `lft` values, but because of (1) the sibling order will still be random
4. Because of (1) and (2), the database `lft` values for newly imported records will be random, but the Elasticsearch `lft` values will be random *and* may be empty

Is that your understanding as well?

I think fixing issue 13414 (random sibling ordering) is important for AtoM, but it's too big of a change to include in this Pull Request.   The problem of the `lft` value in the Elasticsearch index being empty in some cases is not great, but there is a work around (running `search:populate`).  Since we can't guarantee the sibling order for the treeview in any case without fixing issue 13414 I think using the ES `lft` value for the treeview sort is an acceptable trade-off for better memory performance and treeview load speed.  I also don't think sorting by ES `lft` now makes it significantly harder to fix Issue 13414 in the future.

My proposal is to ignore the ES `lft` sort problem for this Pull Request and address it with the fix for Issue 13414.  I know it's not an ideal solution, but I think it's the best compromise for now.

jraddaoui:
Hi @djjuhasz,

> Is that your understanding as well?

Yes, exactly. It's worth noting that CSV GUI imports behave that way by default.

> I think fixing issue 13414 (random sibling ordering) is important for AtoM, but it's too big of a change to include in this Pull Request. The problem of the lft value in the Elasticsearch index being empty in some cases is not great, but there is a work around (running search:populate). Since we can't guarantee the sibling order for the treeview in any case without fixing issue 13414 I think using the ES lft value for the treeview sort is an acceptable trade-off for better memory performance and treeview load speed. I also don't think sorting by ES lft now makes it significantly harder to fix Issue 13414 in the future.

I agree.

> My proposal is to ignore the ES lft sort problem for this Pull Request and address it with the fix for Issue 13414. I know it's not an ideal solution, but I think it's the best compromise for now.

Sounds good.

I see @mcantelon is assigned for review and I'm feeling a bit off today, so I'll let him deal with the actual code.

Thanks David.

apps/qubit/modules/default/actions/fullTreeViewAction.class.php
```diff
@@ -23,257 +23,542 @@
...
+
+    $data = $result->getData();
+
+    return $data['ancestors'];
+  }
+
+  /**
+   * Get an array of ids for the ancestors of the given information object
+   *
+   * @param int $informationObjectId information object id
```
mcantelon: The function doesn't have this parameter.

mcantelon (changes requested):
Great improvement/refctoring... cleans things up a lot!

The file `apps/qubit/modules/browse/actions/hierarchyDataAction.class.php` seems like it needs to be updated as enabling "Show browse hierarchy page" in treeview settings and visiting  browse/hierarchy results in a 500 error as the function "getNodeOrChildrenNodes" has been removed.

Other thn that I only found one issue with a comment. Nicely done!

apps/qubit/modules/default/actions/fullTreeViewAction.class.php
djjuhasz: Oops, yeah that comment is totally unrelated to that method 🤦

djjuhasz: Copy & paste error.... 😛

djjuhasz:
Thanks @mcantelon! Good catch on the browse hierarchy page using the old `getNodeOrChildrenNodes()` method. 😅 

I've added new commits with the fixes so you can review them easily - I'll squash them into the main commit before I merge.

mcantelon (approved):
Looks good to me! Really nice refactoring! 👍

PR #1206, opened 2020-10-29
User Active setting fix, Refs #13393

sbreker:
This commit fixes an issue where editing your own user account sets the
account to inactive on save. The 'active' checkbox is not displayed when
editing one's own user resulting in the field being assigned 'false' on
save.


jraddaoui (approved):
Thanks @sbreker! Out of curiosity, why the capitalized Refs? I've seen it in various commits and I wonder if it's your editor or something like that doing it automatically.

PR #1207, opened 2020-10-29
Protect reference copy link, refs #13388

replaceafill:
This updates the DO metadata template to show the link to the
reference copy only to authenticated users.

For more context on the change see https://projects.artefactual.com/issues/13388#note-16


djjuhasz (approved):
Looks good. 👍

PR #1208, opened 2020-10-30
Save clipboard message i18n, refs #13382

sbreker:
Allow the Save dialog user message to be picked up in the i18n string
export.


djjuhasz (approved):
Looks good @sbreker! 👍

PR #1209, opened 2020-10-30
Update clipboard routes, refs #11556

sbreker:
Update Clipboard menu routes to point to new clipboard module.


djjuhasz (approved):
Looks good @sbreker. 👍

PR #1210, opened 2020-10-30
PHP 7.4 array ref curly braces, refs #13433

sbreker:
In PHP 7.4 array references using curly braces have been deprecated.
This commit updates all instances in Symfony where this occurs. The
fix is a direct replacement with square brackets.

e.g. $somearray{0}
Becomes:  $somearray[0]

https://wiki.php.net/rfc/deprecate_curly_braces_array_access


mcantelon (approved):
Looks good to me!

Issue #1211, opened 2020-11-02
The index functions associated with bulk updating publication status do not get initialized with a language code to be able to read title into the index


ross-spencer:
The index functions associated with bulk updating publication status do not get initialized with a language code to be able to read title into the index. NB. for the casual reader, this makes the search results look pretty bad - but it doesn't affect what is retrieved for the archival description.

For example, initial import (doesn't matter if it is Draft or Published) looks as follows:

![image](https://user-images.githubusercontent.com/1880412/97824359-ad635e80-1c89-11eb-905a-a2c69564baa5.png)

After running the update publication status task we see the following:

![image](https://user-images.githubusercontent.com/1880412/97824411-d4ba2b80-1c89-11eb-9ee3-f3f30b83e30d.png)

Various fields are not correctly read and output in the ES index here. 

![image](https://user-images.githubusercontent.com/1880412/97826010-6e83d780-1c8e-11eb-9fd3-e90231191029.png)

The correct output should normally look something like: 

Vs. 

![image](https://user-images.githubusercontent.com/1880412/97826153-d1756e80-1c8e-11eb-9f22-334630d55161.png)

_Docker commands used:_

* `docker-compose exec atom php symfony tools:purge --demo`
* `docker-compose exec atom php symfony csv:import --source-name=un /home/www-data/error-demo-draft.csv --index`
* `docker-compose exec atom php symfony tools:update-publication-status published a-simple-title`

_Test CSV:_

[error-demo-draft.zip](https://github.com/artefactual/atom/files/5472570/error-demo-draft.zip)

**What are we seeing in the code**

* [updatePublicationStatusTask](https://github.com/artefactual/atom/blob/487c9ddf6144fb2a4caad05665c9e60e25e9cc77/lib/task/tools/updatePublicationStatusTask.class.php#L47-L51) is the primary runner for this work.
* [arElasticSearchModelBase](https://github.com/artefactual/atom/blob/487c9ddf6144fb2a4caad05665c9e60e25e9cc77/plugins/arElasticSearchPlugin/lib/model/arElasticSearchModelBase.class.php#L76-L79) is the code run by both the initial CSV import and the update task to add data to the ES index. We never progress far enough into this [loop](https://github.com/artefactual/atom/blob/487c9ddf6144fb2a4caad05665c9e60e25e9cc77/plugins/arElasticSearchPlugin/lib/model/arElasticSearchModelBase.class.php#L106-L109) because `$allowedLanguages` is never set. 
[...]

PR #1212, opened 2020-11-02
Initialize settings array, refs #13434

ross-spencer:
Ensure that the classes required for the update-publication-status
task are properly initialized. This commit makes sure that the
search index can be properly updated when the task is run, i.e. by
populating the allowed languages array to be properly used when
updating the index.

**NB.** _This is just an initial commit to fix this issue, but also looking for guidance on what else will be needed for it to be a complete PR._ 

Connected to https://github.com/artefactual/atom/issues/1211

### DIFF
--- a/lib/task/tools/updatePublicationStatusTask.class.php
+++ b/lib/task/tools/updatePublicationStatusTask.class.php
@@ -17,7 +17,7 @@
  * along with Access to Memory (AtoM).  If not, see <http://www.gnu.org/licenses/>.
  */
 
-class updatePublicationStatusTask extends sfBaseTask
+class updatePublicationStatusTask extends arBaseTask
 {
   protected function configure()
   {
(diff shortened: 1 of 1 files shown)


jraddaoui (approved):
Thanks a lot @ross-spencer!

It's amazing that this still comes up ... long story short ... this is done in arBaseTask and the idea was to extend all the tasks that need those settings from it to have it. However, some of the tasks (tools:puge specially) may have an empty database when they start causing that config load to fail, so that was never done. Lately, I added the need to have that config loaded in the indexing processes for performance, which most likely introduced this issue.

So, the fix looks great, but I think this task could extend arBaseTask and remove the bits that are already on its execute method.

ross-spencer:
@jraddaoui thanks for the review and feedback. I wanted to see if I could write the change based on inheriting from `arBaseTask`. I have tried that here: https://github.com/artefactual/atom/commit/340f737c8328400cd544514138443bfdd26e9891 it works, but based on your comment, is it all I need to consider or is it missing something? If it looks okay, or needs some small changes, I can convert it to a PR to supersede this one.

jraddaoui:
That looks great @ross-spencer, thanks! Feel free to create a new PR or to overwrite this one (rename the new branch as this one and force push).

ross-spencer:
Nice! Done @jraddaoui! So that's okay to merge? Also, because this came up in a 2.6.x site during a support task, do you have a recommendation how we can make use of it there if it's needed?

jraddaoui:
Sure, okay to merge! It looks like a 2.6.2 candidate too so I'd suggest to cherry-pick to stable/2.6.x and then rebase that site branch against it.

Thanks @ross-spencer!

PR #1213, opened 2020-11-03
Fix permission denied adding child, refs #13435

djjuhasz:
- Use the parent resource slug (e.g. 'my-slug') not the relative path
  (e.g. '/my-slug') for the "parent" parameter when creating a new
  information object
- Check POST as well as GET parameters for a "parent" param when
  creating a new informtion object, to allow authenticating the
  "create" action against the parent IO on form submit.


jraddaoui (approved):
Thanks @djjuhasz!

djjuhasz:
Thanks @jraddaoui! :)

PR #1214, opened 2020-11-07
Consolidate DO auth logic, refs #13437

djjuhasz:
- Move information object specific ACL logic to the QubitInformationObjectAcl
  subclass
- Consolidate information object digital object authorization logic
- Consolidate actor digital object authorization logic
- Update all digital object authorization checks to use the new consolidated
  auth code
- Merge duplicate code for getting a digital object URL from QubitActor and 
  QubitInformationObject classes to the QubitObject class

### DIFF
--- a/apps/qubit/modules/digitalobject/actions/showGenericIconComponent.class.php
+++ b/apps/qubit/modules/digitalobject/actions/showGenericIconComponent.class.php
@@ -18,7 +18,7 @@
  */
 
 /**
- * Display an 'image' digital asset
+ * Display a generic representation of a digital object
  *
  * @package    AccesstoMemory
  * @subpackage digital object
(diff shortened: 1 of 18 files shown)


plugins/qbAclPlugin/lib/QubitAcl.class.php
```diff
@@ -263,6 +260,37 @@ public static function isAllowed($role, $resource, $action, $options = array())
...
+   * the parent of the resource
+   *
+   * @param myUser $role actor's role to authorize
+   * @param mixed $resource target of the requested action
+   * @param string $action requested for authorization (e.g. 'read')
+   * @param array|null $options optional parameters
+   *
+   * @return bool true if the access request is authorized
+   */
+  protected function authorizeAgainstParent($role, $resource, $action, $options = array())
```
mcantelon: The `$options` arg doesn't seem to be used.

plugins/qbAclPlugin/lib/QubitActorAcl.class.php
```diff
@@ -0,0 +1,99 @@
...
+    'publish' => 'Publish',
+    'readMaster' => 'Access master',
+    'readReference' => 'Access reference',
+    'readThumbnail' => 'Access thumbnail'
+  );
+
+  protected static
+    $_digitalObjectActions = ['readMaster', 'readReference', 'readThumbnail'];
+
+   /**
```
mcantelon: Looks like this line is indented one space too much.

apps/qubit/modules/digitalobject/actions/showGenericIconComponent.class.php
```diff
@@ -18,7 +18,7 @@
  */
 
 /**
- * Display an 'image' digital asset
+ * Display a generic representation of a digital object
```
mcantelon, line 21: Nitpick: no period at the end of the PHPDoc-style comment I think.

mcantelon (approved):
Really nice reorg/cleanup and well commented!

Looks good aside from a few nitpicks.

plugins/qbAclPlugin/lib/QubitAcl.class.php
djjuhasz: Good catch! The `$options` array should be passed to the `isAllowed()` method on line 291:

```
return self::getInstance()->acl->isAllowed($role, $parent, $action, $options);
```

I'll update the PR.

djjuhasz: Huh, I was wrong. Just removing the `$options` parameter from the `authorizeAgainstParent()` method works fine and is simpler, so I'll do that.

PR #1217, opened 2020-11-13
Authorize export of DO metadata, refs #13395

djjuhasz:
- If user is authorized to view master or reference digital objects,
  include the URL and checksum for the highest quality digital object
  allowed
- Pass current user from export job (UI) to CSV writer
- Pass current user from export task (CLI) to CSV writer


mcantelon (approved):
Nicely done... looks good to me! 👍

PR #1219, opened 2020-11-19
Adjust Dockerfile to take advantage of docker caching

jasonhildebrand:
I've upgraded archives.mhsc.ca to AtoM 2.6 - it's now hosted using docker.  Kudos to you for your work on AtoM and your work to dockerize it!

I've worked with your Dockerfile a bit in updating our customizations and theming. One pain point I hit was that lessc (from make -C /atom/src/plugins/arDominionPlugin) failed due to a theming customization on our end.  Fixing and re-running meant re-building the entire container.  That meant several very slow iterations while I fixed that up.

I am submitting this PR which re-organizes your Dockerfile a bit.  The local source tree almost always has changes from the previous build, so docker can't usually use any cached steps below `COPY . /atom/src`.  The main idea in the PR is to move the COPY lower down in the file to allow docker's cache to work.  Building PHP doesn't depend on the codebase, so it's not necessary to have COPY above the PHP build.  

I also added the fcgi package - this allows a docker healthcheck to be defined in the atom container which tests that PHP-FPM is running and producing expected result (I can share the healthcheck code if you are interested).

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -6,8 +6,6 @@ ENV FOP_HOME=/usr/share/fop-2.1 \
 
 COPY --from=composer /usr/bin/composer /usr/bin/composer
 
-COPY . /atom/src
-
 RUN set -xe \
 		&& apk add --no-cache --virtual .phpext-builddeps \
 			gettext-dev \
@@ -48,12 +46,18 @@ RUN set -xe \
 			make \
 			bash \
 			gnu-libiconv \
+			fcgi \
 		&& npm install -g "less@<2.0.0" \
 		&& curl -Ls https://archive.apache.org/dist/xmlgraphics/fop/binaries/fop-2.1-bin.tar.gz | tar xz -C /usr/share \
-		&& ln -sf /usr/share/fop-2.1/fop /usr/local/bin/fop \
+		&& ln -sf /usr/share/fop-2.1/fop /usr/local/bin/fop
+
+COPY . /atom/src
+
+RUN set -xe \
 		&& make -C /atom/src/plugins/arDominionPlugin \
 		&& make -C /atom/src/plugins/arArchivesCanadaPlugin \
 		&& composer install -d /atom/src
+
 
 WORKDIR /atom/src
 


Dockerfile
```diff
@@ -48,13 +46,19 @@ RUN set -xe \
...
+		&& ln -sf /usr/share/fop-2.1/fop /usr/local/bin/fop
+
+COPY . /atom/src
+
+RUN set -xe \
 		&& make -C /atom/src/plugins/arDominionPlugin \
 		&& make -C /atom/src/plugins/arArchivesCanadaPlugin \
 		&& composer install -d /atom/src
 
+
```
jraddaoui, line 61: Extra empty line.

Dockerfile
```diff
@@ -6,8 +6,6 @@ ENV FOP_HOME=/usr/share/fop-2.1 \
 
 COPY --from=composer /usr/bin/composer /usr/bin/composer
```
jraddaoui, line 7: I think it would be nice to move this layer too, just in case there are changes in the composer executable.

jraddaoui (commented):
This is great @jasonhildebrand, thanks a lot!

We had [a similar version](https://github.com/jraddaoui/atom/pull/1/commits/234b5856645ead79de99c57efe32664c0659823a) of this changes that we're using for some CI tests in a personal fork, but it never made it to vanilla, so we appreciate you bring this up.

I've added a couple of minor suggestions below and I'd appreciate if you can change the commit message, to reference [the Redmine ticket I just created](https://projects.artefactual.com/issues/13441) for this matter and to wrap the subject to 50 chars. Something like "Adjust Dockerfile, refs #13441", then an empty line and the body wrapped to 72 chars lines, if you want to explaining the changes a bit more.

Regards.

jraddaoui:
Thanks again @jasonhildebrand!

I've moved it to https://github.com/artefactual/atom/pull/1247 with the suggestions I mentioned.

jasonhildebrand:
OK, great.  Thanks for going ahead and fixing this up - I had already moved on to a different project and wasn't able to get back ot this.

PR #1222, opened 2020-11-23
Add job data support to expiry task, refs #13440

mcantelon:
* Fixes bug where clipboard maximum age setting wasn't getting read
* Adds the ability ro expire job data and related temporary user
  download files
* Adds the ability to expire multiple types of data in one run
* Code refactoring/cleanup

### DIFF
--- a/lib/task/tools/expireDataTask.class.php
+++ b/lib/task/tools/expireDataTask.class.php
@@ -25,13 +25,21 @@ class expireDataTask extends arBaseTask
       'clipboard' => array(
         'name' => 'saved clipboard',
         'plural_name' => 'saved clipboards',
-        'age_setting_name' => 'clipboard_save_max_age'
+        'age_setting_name' => 'app_clipboard_save_max_age'
+      ),
+      'job' => array(
+        'name' => 'job (and any related file)',
+        'plural_name' => 'jobs (and any related files)'
       )
     );
 
   protected function configure()
   {
-    $dataTypeArgDescription = sprintf('Data type (supported types: %s)', $this->supportedTypesDescription());
+    $dataTypeArgDescription = sprintf(
+      'Data type(s), comma-separated (supported types: %s)',
+      $this->supportedTypesDescription()
+    );
+
     $this->addArguments(array(
       new sfCommandArgument('data-type', sfCommandArgument::REQUIRED, $dataTypeArgDescription)
     ));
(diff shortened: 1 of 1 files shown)


lib/task/tools/expireDataTask.class.php
```diff
@@ -70,50 +78,91 @@ protected function execute($arguments = array(), $options = array())
...
-      // Throw error if setting value isn't an integer
-      if (!ctype_digit($maxAge))
+      $maxAge = sfConfig::get($typeSpec['age_setting_name']);
+
+      if ($checkAgeSetting && $maxAge)
       {
-        throw new sfException(
-          sprintf('Error: setting %s value "%s" is non-numeric.', $typeSpec['age_setting_name'], $maxAge)
+        // Throw error if setting value isn't an integer
+        if (!ctype_digit($maxAge))
```
djjuhasz: I think `ctype_digit()` will return false here if `$maxAge` is an integer between -128 and 255 (See https://www.php.net/manual/en/function.ctype-digit.php#example-5540).  It's not a problem here if `sfConfig::get($typeSpec['age_setting_name'])` always returns a string value, but I'm not sure if that is the case. 🤔

lib/task/tools/expireDataTask.class.php
```diff
@@ -70,50 +78,91 @@ protected function execute($arguments = array(), $options = array())
...
+      $checkAgeSetting = !isset($options['older-than'])
+                         && isset($typeSpec['age_setting_name']);
 
-    // Set older than option if not set and a non-zero maximum age is set for data type
-    $checkAgeSetting = !isset($options['older-than']) && isset($typeSpec['age_setting_name']);
-    if ($checkAgeSetting && $maxAge = sfConfig::get('app_clipboard_save_max_age'))
-    {
-      // Throw error if setting value isn't an integer
-      if (!ctype_digit($maxAge))
+      $maxAge = sfConfig::get($typeSpec['age_setting_name']);
```
djjuhasz: I think this will emit a warning if the `$typeSpec['age_setting_name']` key is not set for `$typeSpec`.  Instead of putting `isset($typeSpec['age_setting_name'])` in the previous line, I think it would be better to do something like:

`$maxAge = isset($typeSpec['age_setting_name']) ? sfConfig::get($typeSpec['age_setting_name']) : null;`

The following conditional should still evaluate to false if $maxAge is null.

lib/task/tools/expireDataTask.class.php
```diff
@@ -70,50 +78,91 @@ protected function execute($arguments = array(), $options = array())
...
+    $this->validateDataTypes($dataTypes);
+
+    foreach ($dataTypes as $dataType)
     {
-      throw new sfException('Aborted: unsupported data type.');
-    }
+      $typeSpec = self::$TYPE_SPECIFICATONS[$dataType];
 
-    $typeSpec = self::$TYPE_SPECIFICATONS[($arguments['data-type'])];
+      // Set older than option if not set and a non-zero maximum age is set for data type
```
djjuhasz: Opinion: I think it would be better to move the logic for determining `$options['older-than']` when it's not explicitly set to it's own method to make it easier to track the main execution path.

lib/task/tools/expireDataTask.class.php
```diff
@@ -70,50 +78,91 @@ protected function execute($arguments = array(), $options = array())
...
+          );
+        }
+
+        // Use date type's maximum age setting to calculate older than option
+        $options['older-than'] = $this->calculateOlderThanDate(
+          sfConfig::get($typeSpec['age_setting_name'])
+        );
+
+        // Let user know that date type's maximum age setting was used
+        $this->logSection(
```
djjuhasz: Nice 👍

lib/task/tools/expireDataTask.class.php
```diff
@@ -153,4 +202,32 @@ private function clipboardExpireData($options)
...
+    {
+      $criteria->add(QubitJob::CREATED_AT, $options['older-than'], Criteria::LESS_THAN);
+    }
+
+    // Delete jobs and save items
+    $deletedCount = 0;
+
+    foreach(QubitJob::get($criteria) as $job)
+    {
+      // Finding aids won't be deleted because the job doesn't store the path to them
```
djjuhasz: _Not_ deleting finding aids is the desired behavior, right?  I think it would be helpful to make this explicit in the comment. :)

djjuhasz (changes requested):
Looks good overall @mcantelon, I just had a few questions and comments.

lib/task/tools/expireDataTask.class.php
mcantelon: Yikes! Fixed.

mcantelon:
Thanks @djjuhasz... I've committed an update for the PR!

lib/task/tools/expireDataTask.class.php
mcantelon: Good idea.

lib/task/tools/expireDataTask.class.php
```diff
@@ -70,50 +78,100 @@ protected function execute($arguments = array(), $options = array())
...
+    }
+  }
+
+  private function determineOlderThanDate($options, $typeSpec)
+  {
     // Set older than option if not set and a non-zero maximum age is set for data type
-    $checkAgeSetting = !isset($options['older-than']) && isset($typeSpec['age_setting_name']);
-    if ($checkAgeSetting && $maxAge = sfConfig::get('app_clipboard_save_max_age'))
+    $checkAgeSetting = !isset($options['older-than'])
+                         && isset($typeSpec['age_setting_name']);
```
djjuhasz: You can drop the `&& isset($typeSpec['age_setting_name']);` here, as it is tested below. :)

lib/task/tools/expireDataTask.class.php
```diff
@@ -70,50 +78,100 @@ protected function execute($arguments = array(), $options = array())
...
         );
       }
 
       // Use date type's maximum age setting to calculate older than option
-      $date = new DateTime();
-      $interval = new DateInterval('P'. sfConfig::get('app_clipboard_save_max_age') .'D');
-      $date->sub($interval);
-      $options['older-than'] = $date->format('Y-m-d');
+      $options['older-than'] = $this->calculateOlderThanDate(
+        sfConfig::get($typeSpec['age_setting_name'])
```
djjuhasz: I'd use `$maxAge` instead of `sfConfig::get($typeSpec['age_setting_name'])` here, as it's shorter and already assigned.

lib/task/tools/expireDataTask.class.php
```diff
@@ -153,4 +211,34 @@ private function clipboardExpireData($options)
...
+    {
+      $criteria->add(QubitJob::CREATED_AT, $options['older-than'], Criteria::LESS_THAN);
+    }
+
+    // Delete jobs and save items
+    $deletedCount = 0;
+
+    foreach(QubitJob::get($criteria) as $job)
+    {
+      // Jobs generate finding aids, which we *don't* want to delete... finding
```
djjuhasz, line 225: Nice! 👍

djjuhasz (changes requested):
Looking good @mcantelon, I just suggested a couple of minor tweaks.

mcantelon:
Thanks again @djjuhasz... the max age check is now considerably saner... I've updated the PR! ;)

djjuhasz (approved):

PR #1224, opened 2020-11-27
Fix syntax in actor relationship area, refs #13431

replaceafill:
This removes the word `the` from the labels in the Relationship area
of actors.


tw4l (approved):
LGTM! :sunglasses:

PR #1225, opened 2020-11-27
Fix storage location display, refs #13314

mcantelon:
A function that was generating data for an HTML select form field was unnessarily modifying QubitTerm instances. The modifed instances were getting cached and this caching was leading to incorrect display of the names of the cached QubitTerm instances in logic following the generation of the HTML select form field data.

This was fixed by not modifying the QubitTerm instances.


jraddaoui (approved):
Thanks @mcantelon!

PR #1226, opened 2020-12-02
Get rid of global import functions, refs #13308

mcantelon:
Changed import-related global function declarations into static class methods. Renamed one function to conform to method naming convention.


djjuhasz (approved):

PR #1229, opened 2020-12-03
Don't show deleted quick link pages, refs #13297

mcantelon:
Added logic to the quick links menu so links won't be shown to any static pages that have been deleted.

### DIFF
--- a/apps/qubit/modules/menu/templates/_quickLinksMenu.php
+++ b/apps/qubit/modules/menu/templates/_quickLinksMenu.php
@@ -14,7 +14,7 @@
 
     <div class="top-dropdown-body">
       <ul>
-        <?php foreach ($quickLinksMenu->getChildren() as $child): ?>
+        <?php foreach ($quickLinks as $child): ?>
           <?php if ('login' != $child->getName() && 'logout' != $child->getName() && 'myProfile' != $child->getName()): ?>
             <li<?php if ($child->isSelected()): ?> class="active"<?php endif; ?>><?php echo link_to($child->getLabel(array('cultureFallback' => true)), $child->getPath(array('getUrl' => true, 'resolveAlias' => true))) ?></li>
           <?php endif; ?>
--- a/data/fixtures/settings.yml
+++ b/data/fixtures/settings.yml
@@ -3,7 +3,7 @@ QubitSetting:
     name: version
     editable: 0
     deleteable: 0
-    value: 188
+    value: 189
   milestone:
     name: milestone
     editable: 0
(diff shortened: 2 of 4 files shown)


apps/qubit/modules/menu/actions/quickLinksMenuComponent.class.php
```diff
@@ -28,12 +28,26 @@ class menuQuickLinksMenuComponent extends sfComponent
...
   {
-    // Get menu objects
-    $this->quickLinksMenu = QubitMenu::getById(QubitMenu::QUICK_LINKS_ID);
+    // Get menu
+    $quickLinksMenu = QubitMenu::getById(QubitMenu::QUICK_LINKS_ID);
 
-    if (!$this->quickLinksMenu instanceof QubitMenu)
+    if (!$quickLinksMenu instanceof QubitMenu)
     {
       return sfView::NONE;
```
djjuhasz: Nitpick: I believe `return sfView::NONE` is only relevant for sfAction objects, and a bare `return;` is sufficient to skip component output.

lib/Qubit.class.php
```diff
@@ -452,4 +452,13 @@ public static function generateIdentifierFromCounterAndMask($counter, $mask)
       }
     }, $mask);
   }
+
+  public static function actionExistsForUrl($url)
```
djjuhasz: This method looks like a duplicate of the `QubitObject::actionExistsForUrl()` method, and it doesn't look like it's called anywhere.

apps/qubit/modules/menu/actions/quickLinksMenuComponent.class.php
```diff
@@ -28,12 +28,26 @@ class menuQuickLinksMenuComponent extends sfComponent
...
+
+    // Get menu items that correspond to an external URL or an internal path
+    $this->quickLinks = [];
+
+    foreach ($quickLinksMenu->getChildren() as $child)
+    {
+      $url = $child->getPath(['getUrl' => true, 'resolveAlias' => true]);
+      $urlParsed = parse_url($url);
+
+      if (isset($urlParsed['scheme']) || QubitObject::actionExistsForUrl($url))
```
djjuhasz, line 47: There's a bug in this logic somewhere - `$this->quickLinks` is always empty in my test AtoM instance, even though I have quick link menu items.

empty_quick_links_menu

djjuhasz (changes requested):
@mcantelon I think there is a bug in the logic here somewhere - my "Quick links" menu is always empty when I check out this dev branch.

lib/Qubit.class.php
mcantelon: Oops... accidentally committed that.

apps/qubit/modules/menu/actions/quickLinksMenuComponent.class.php
mcantelon: Fixed!

mcantelon:
@djjuhasz I tried this branch in another dev VM and it works for me. Maybe restart your VM in case there's some stubborn caching going on?

apps/qubit/modules/menu/actions/quickLinksMenuComponent.class.php
djjuhasz: I retested, and the quick links are working for me now.

djjuhasz (approved):
@mcantelon you are right, I restarted my test VM and the quick links are working now. 🤷 

The code looks good to me.  👍 

I am curious about the added check to make sure the menu path is a valid URL or internal route.   I doesn't mention anything in the ticket about checking the paths. Is the path checking logic something you thought would be good to do, or was there another reason for it?

mcantelon:
@djjuhasz The path checking is a means of checking to make sure a menu item doesn't point to a static page, or any other AtoM resource/action, that has been deleted.

If the path is an absolute URL then it's deemed external and there's no check done to see if the external resource exists. If the path isn't an absolute URL then it's treated as a path to a Symfony resource/action. If a static page has been deleted then the menu item's path (a static page slug) won't resolve to a Symfony action and the menu item ends up getting skipped and not displayed.

djjuhasz:
@mcantelon oh, so instead of deleting the menu item to stop e.g. the privacy policy page from showing in the quick links menus one would delete the static page?  🤔   From the ticket description I was just expecting a "delete" button on the menu item edit page.

mcantelon:
@djjuhasz Ah, you're right I misread the issue! I'll fix the inability to delete the menu item itself and add that to this PR.

mcantelon:
@djjuhasz I've added the ability to delete the privacy menu item. Let me know if you see anything that might need to be tweaked (I'll update the commit message, before merging. to reflect the DB tweak to unprotect the privacy menu item).Thanks!

djjuhasz (approved):
Looks good to me @mcantelon. 😄

mcantelon:
Woot! Thanks @djjuhasz !

Jimadine:
@mcantelon I upgraded from `2.6.4` to `2.7.1` and noticed that our custom quick links are missing from the quick links menu. These are links with the path set to `staticpage/index?slug=<slug>` (external `http`/`https` links unaffected).  I tried a VM restart as suggested earlier in this PR but no luck. So I edited `apps/qubit/modules/menu/actions/quickLinksMenuComponent.class.php`, commenting out the conditional as follows:
```
            //if (isset($urlParsed['scheme']) || QubitObject::actionExistsForUrl($url)) {
                $this->quickLinks[] = $child;
            //}
```
The static pages definitely exist and have the slug field set.
With the above conditional commented out, after running `php symfony cc` & `systemctl restart php7.4-fpm`, the quick links appeared.
I've also checked this in a clean VM with `stable/2.7.x` to rule out any local customisations, and the problem is still apparent. Are you able to take a look, please? An example missing link is to the privacy page.

PR #1230, opened 2020-12-06
Make sure password hash algo. exists, refs #13448

mcantelon:
Check, when generating a password hash, to make sure the password hashing algorithm actually exists in the PHP installation.

### DIFF
--- a/lib/model/QubitUser.php
+++ b/lib/model/QubitUser.php
@@ -74,10 +74,41 @@ class QubitUser extends BaseUser
 
   public static function generatePasswordHash($password)
   {
-    $hashAlgo = constant(sfConfig::get('app_password_hash_algorithm', 'PASSWORD_ARGON2I'));
-    $hashAlgoOptions = json_decode(sfConfig::get('app_password_hash_algorithm_options', '{}'), true);
+    $hashAlgoConstant = sfConfig::get('app_password_hash_algorithm', 'PASSWORD_ARGON2I');
 
-    return password_hash($password, $hashAlgo, $hashAlgoOptions);
+    // Check to make sure hashing constant specified in settings is defined
+    if (defined($hashAlgoConstant))
+    {
+      // Hashing constant is defined so use it and options specified in settings
+      $hashAlgo = constant($hashAlgoConstant);
+
+      $hashAlgoOptions = json_decode(
+        sfConfig::get('app_password_hash_algorithm_options', '{}'),
+        true
+      );
+    }
+    else
+    {
+      // Hashing constant specified in settings isn't defined: use default
+      // alghorithm (PASSWORD_DEFAULT, usually PASSWORD_BCRYPT) and options
+      $hashAlgo = PASSWORD_DEFAULT;
+      $hashAlgoOptions = [];
+    }
+
+    $passwordHash = password_hash($password, $hashAlgo, $hashAlgoOptions);
+
+    // Make sure hashing completed successfully
+    if (empty($passwordHash))
+    {
+      $errorMessage = sprintf(
+        "Password hashing using the %s algorithm was unsuccessful.",
+        $hashAlgoConstant
+      );
+
+      throw new ErrorException($errorMessage);
+    }
+
+    return $passwordHash;
   }
 
   public function getAclGroups()


mcantelon:
How I've tested this PR so far:

1) I generated a DB dump of AtoM, with a 16.04-based VM, using the commit before commit with the migration that updates the password hashes (migration 175)

2) I loaded this DB dump into a 18.04-based VM and tested using tools:upgrade-sql with this PR to make sure it would properly update if Argon2 is available

3) I loaded this DB dump into a 18.04-based VM and tested using tools:upgrade-sql with this PR to make sure it would properly update if Argon2 isn't available (simulating this by adding a line to the generatePasswordHash method, in QubitUser, to set $hashAlgoConstant to a non-existing constant's name)

scollazo:
Other way to test this would be to deploy on centos, where ARGON cipher is not available for now

DavidHume-ArtefactualSystems:
@mcantelon - I concur with @scollazo 's test suggestion

(also been musing with @mamedin per his suggestion a PR on the deployment side)

@mcantelon - also, (where) is PASSWORD_DEFAULT in fact defined?

mcantelon:
@DavidHume-ArtefactualSystems PASSWORD_DEFAULT is a PHP predefined constant.

mcantelon:
@scollazo Yeah, trying this in Centos would be a good way to test. I didn't see AtoM/Centos in our public deployment repo so haven't tried it yet.

djjuhasz:
@mcantelon @scollazo @DavidHume-ArtefactualSystems I will test this works with CentOS7 when I do code review

djjuhasz:
I did a fresh install on CentOS7 and created a user with the `php symfony tools:add-superuser` CLI script, and the password was hashed correctly using the bcrypt hashing algorithm. 👍   I'll test a database upgrade next.

djjuhasz:
Upgrading an AtoM 2.5 data set (with `symfony tools:upgrade-sql`) on CentOS7 worked like a charm, all passwords are hashed with bcrypt and I can log in fine. 😁

Now to code review the actual code. :)

lib/model/QubitUser.php
```diff
@@ -74,8 +74,26 @@ public function setPassword($password)
...
+        true
+      );
+    }
+    else
+    {
+      // Hashing constant specified in settings isn't defined: use default
+      // alghorithm (PASSWORD_DEFAULT, usually PASSWORD_BCRYPT) and options
+      $hashAlgo = PASSWORD_DEFAULT;
+      $hashAlgoOptions = [];
+    }
```
djjuhasz, line 97: Maybe I am being extra paranoid, but I would like to see this function throw an exception if `$hashAlgo` is empty, before calling `password_hash()`, to ensure existing passwords aren't destroyed.

djjuhasz (changes requested):
Hi @mcantelon, just one requested change - and I may be being overly cautious.

lib/model/QubitUser.php
mcantelon: Sounds good... I'll add that!

mcantelon:
I added some logic to check to make sure $hashAlgo isn't empty and tested it by:

1. Defining a constant in `lib/model/QubitUser.php` as `""` then setting the `password_hash_algorithm` in `config/app.yml` to the name of it

2. Triggering the exception by running `php symfony tools:purge --demo`

Another possible sanity check I could do is check the resulting hash value itself... causing an exception if the password hash generated is empty.

djjuhasz:
> Another possible sanity check I could do is check the resulting hash value itself... causing an exception if the password hash generated is empty.

I love this @mcantelon!  My concern is writing an empty string to the database and deleting an existing password hash, and this is the best way to prevent that scenario I think. 👍

mcantelon:
Thanks @djjuhasz ! I've updated the PR adding this check. I've removed the previous check I added as this check with also catch the use of an algo constant that is defined but empty.

lib/model/QubitUser.php
djjuhasz: Awesome!

djjuhasz (approved):
Looks good @mcantelon :shipit: !

PR #1232, opened 2020-12-17
Hide missing DO master file, refs #13388

replaceafill:
Archivematica's metadata-only DIP upload generates a DO with no master
file. This updates the DO metadata component to hide the master file
section when the master file is missing.

The existing conditions for the reference and thumbnail copy areas
have been moved to the component class for uniformity.


mcantelon (approved):
Looks good to me!

replaceafill:
Thank you Mike!

PR #1234, opened 2021-01-06
D.O. download API draft access check, ref #13436

mcantelon:
Added check, to API endpoint for downloading master digital objects, to make sure user has draft access if the digital object's information object isn't published.

### DIFF
--- a/plugins/arRestApiPlugin/modules/api/actions/informationobjectsDownloadDigitalObjectAction.class.php
+++ b/plugins/arRestApiPlugin/modules/api/actions/informationobjectsDownloadDigitalObjectAction.class.php
@@ -39,6 +39,17 @@ class ApiInformationObjectsDownloadDigitalObjectAction extends QubitApiAction
       throw new QubitApiNotAuthorizedException;
     }
 
+    // Check, if I.O. isn't published, that user is authorized to access drafts
+    $publicationStatusId = $this->resource->getPublicationStatus()->statusId;
+
+    if (
+      $publicationStatusId != QubitTerm::PUBLICATION_STATUS_PUBLISHED_ID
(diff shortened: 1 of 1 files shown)


tw4l:
Hi @mcantelon, thanks for submitting this! The code looks good to my eyes but the permissions check doesn't seem to be working as expected when I test it in my development environment.

I created two draft IOs as an admin user with attached DOs (one externally linked, one uploaded), and then attempted to download the DOs from the admin user (success) as well as from a new user with permissions that shouldn't allow it (screenshots below, this also worked though it shouldn't):

Screenshot of non-admin user

Screenshot of non-admin user permissions

I was seeing something similar in my first steps at attempting this development before the holidays. Maybe something about the `QubitAcl::check($this->resource, 'viewDraft')` check isn't working quite as expected? Can you replicate this on your end?

mcantelon:
Weird... when I tried to replicate your test things seems to work for me (doing a request to the endpoint via the web browser rather than curl)... the user with "view draft" deny wasn't able to download the digital object. I create the non-admin user I added it to the "translator" group then clicked on "Archival description permissions" then "edit" then set "view draft" to "deny". Was that how you created your test user (just asking in case there's some edge case at work with a user being created a certain way)?

tw4l:
Super weird! That's exactly how I created my test user. I think it might have had something to do with my local development environment.

I tried out this branch on the PDT Mills AtoM test site and it's working as expected there. Given that it's working on a proper deploy and the code looks good, I'm going to say this LGTM. Thanks again!

tw4l (approved):

mcantelon:
Awesome... Thanks @tw4l !

PR #1235, opened 2021-01-06
Respect 'no_script_name' for EAD URLs, refs #13456

djjuhasz:
- Use symfony web routing to generate resource URL when in web request
  context
- In cli and worker contexts, build resource URL using siteBaseUrl
  setting
- Include "index.php" script name when generating the <eadid @url> only
  when production environment "no_script_name" value is not true
- Create qubitConfiguration::getConfigForEnvironment() method to get
  qubit configuration values outside current context
- Use new method to get "no_script_name" setting for prod environment
  when generating a resource URL in cli or worker context

### DIFF
--- a/apps/qubit/config/qubitConfiguration.class.php
+++ b/apps/qubit/config/qubitConfiguration.class.php
@@ -118,4 +118,45 @@ class qubitConfiguration extends sfApplicationConfiguration
 
     $this->setWebDir($path);
   }
+
+  /**
+   * Get a config variable from an application config file (YAML) for a specific
+   * environment (e.g. "prod", "dev", "cli")
+   *
+   * N.B. to get a config variable for the current context/environment, use
+   * sfConfing::get() instead!
+   *
+   * @param string $varname config variable name
+   * @param string $env Environment name (e.g. 'prod', 'cli')
+   * @param string $configFile config file to check (e.g. 'config/settings.yml')
+   *
+   * @return string|null config value or null if variable is not set
+   */
+  public static function getConfigForEnvironment($varname, $env, $configFile)
+  {
+    // Parse the YAML data to an array
+    $config = sfSimpleYamlConfigHandler::getConfiguration(
+      sfContext::getInstance()
+        ->getConfiguration()
+        ->getConfigPaths($configFile)
+    );
+
+    // The settings.yml file requires an intermediate '.settings' key for
+    // some reason :-/
+    if (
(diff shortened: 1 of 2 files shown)


plugins/sfEadPlugin/lib/sfEadPlugin.class.php
```diff
@@ -197,22 +197,59 @@ public function renderEadId()
...
+      $url = rtrim($this->siteBaseUrl, " \n\r\t\v\0/");
+
+      // Get the "no_script_name" config for the "prod" environment
+      $noScriptName = qubitConfiguration::getConfigForEnvironment(
+        'no_script_name', 'prod', 'config/settings.yml'
+      );
+
+      // Add 'index.php' to the URL when "no_script_name" is not true in
+      // production
+      if (null === $noScriptName || !$noScriptName)
```
mcantelon: Nitpick: could use `empty()` to check.

apps/qubit/config/qubitConfiguration.class.php
```diff
@@ -118,4 +118,45 @@ public function setRootDir($path)
...
+  public static function getConfigForEnvironment($varname, $env, $configFile)
+  {
+    // Parse the YAML data to an array
+    $config = sfSimpleYamlConfigHandler::getConfiguration(
+      sfContext::getInstance()
+        ->getConfiguration()
+        ->getConfigPaths($configFile)
+    );
+
+    // The settings.yml file requires an intermediate '.settings' key for
```
mcantelon, line 144: Good ol' Symfony magic. ;)

mcantelon (approved):
Nicely done! One nitpick but looks good. 👍

mcantelon:
(One more nitpick: commit header line doesn't include Redmine issue.)

plugins/sfEadPlugin/lib/sfEadPlugin.class.php
djjuhasz: Nice! I didn't realize empty() would handle an unset variable without complaint. 😎

apps/qubit/config/qubitConfiguration.class.php
djjuhasz: hehe, yep 🧙

djjuhasz:
@mcantelon I've fixed the commit summary in my second commit.  I'll squash both commits and use the 2nd commit summary when I merge the branch.

PR #1236, opened 2021-01-08
Add import:delete task, refs #13457 

djjuhasz:
Finds all records in the keymap table that match the given import name
(keymap.source_name) and deletes the matching "target" objects and
keymap rows from the database.

Because AtoM imports only create a keymap row for the "primary" import
object (e.g. information_objects for an archival decription import)
the task will only delete these "primary" objects, so other data
created by the import (e.g. new term objects) will *not* be deleted.

- Include "-f|--force" option to skip confirmation before deleting data
- Include "-v|--verbose" option to output additional debugging info
- Include "-l|--logfile" option to write output to the designated file

### DIFF
--- a/lib/task/import/importDeleteTask.class.php
+++ b/lib/task/import/importDeleteTask.class.php
@@ -1,0 +1,284 @@
+<?php
+
+/*
+ * This file is part of the Access to Memory (AtoM) software.
+ *
+ * Access to Memory (AtoM) is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * Access to Memory (AtoM) is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with Access to Memory (AtoM).  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+/**
+ * Delete import data
+ *
+ * @package    AccessToMemory
+ * @subpackage task
+ * @author     David Juhasz <djjuhasz@gmail.com>
+ */
+class importDeleteTask extends arBaseTask
+{
+  protected $namespace        = 'import';
+  protected $name             = 'delete';
+  protected $briefDescription = 'Delete data created by an import';
+
+  protected $detailedDescription = <<<EOF
+Delete data created by the named import from the AtoM database
+EOF;
+
+  protected $count = 0;
+  protected $objectIds = [];
+  protected $verbose = false;
+  protected $timer;
+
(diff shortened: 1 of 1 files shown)


lib/task/import/importDeleteTask.class.php
```diff
@@ -0,0 +1,283 @@
...
+        '[+%ss] %s ',
+        str_pad($this->timer->elapsed(), 7, '0', STR_PAD_LEFT),
+        $message
+      )
+    );
+  }
+
+  private function registerFileLogger($filename)
+  {
+    $flogger = new sfFileLogger($this->dispatcher, ['file' => $filename]);
```
mcantelon, line 174: Nice!

lib/task/import/importDeleteTask.class.php
```diff
@@ -0,0 +1,283 @@
...
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with Access to Memory (AtoM).  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+/**
+ * Run ad-hoc PHP logic contained in a file
```
mcantelon: Leftover comment.

lib/task/import/importDeleteTask.class.php
```diff
@@ -0,0 +1,283 @@
...
+  private function confirmDeletion($isForced)
+  {
+    if ($isForced)
+    {
+      return;
+    }
+
+    $confirmed = $this->askConfirmation(
+      [
+"Continuing will delete $this->count records and related data.",
```
mcantelon: Nitpick: indenting (but maybe indenting would make lines too long in which case ignore this comment ;) ).

lib/task/import/importDeleteTask.class.php
```diff
@@ -0,0 +1,283 @@
...
+    $this->logMsg(
+      sprintf(
+        'Found %s resources created by import "%s"', $this->count, $name
+      )
+    );
+  }
+
+  /**
+   * Allow the user to bail out if they aren't sure they want to delete targeted
+   * descriptions.
```
mcantelon: Nitpick: maybe "AtomM resources" given the import might have created other types of resources.

mcantelon (approved):
Good to have this task! Some nitpicks, but otherwise looks 👍 ,

lib/task/import/importDeleteTask.class.php
djjuhasz: Thanks! Using the Symfony file logger and event dispatcher together is new to me, and I was pleased with the results. 😁

lib/task/import/importDeleteTask.class.php
djjuhasz: Fixed.

lib/task/import/importDeleteTask.class.php
djjuhasz: I was all over the place with "resources", "objects" and "records" in the comments, so I standardized on "database records" in all cases.

PR #1237, opened 2021-01-08
AIP Download metadata only DIPs, refs 13454

sbreker:
This commit fixes an issue where the 'Download file' link on the digital
object page was not working if the source of the digital object was a
metadata only DIP upload.

Now when a metadata only DIP upload is performed, the
relative_path_within_aip and aip_name will be saved as properties of the
object in AtoM's create digital object REST API endpoint. This
information is necessary when calling the AM Storage Service
'extract_file' API endpoint.

### DIFF
--- a/lib/model/QubitInformationObject.php
+++ b/lib/model/QubitInformationObject.php
@@ -74,6 +74,7 @@ class QubitInformationObject extends BaseInformationObject
       case 'formatRegistryName':
       case 'objectUUID':
       case 'aipUUID':
+      case 'aipName':
       case 'relativePathWithinAip':
       case 'originalFileName':
       case 'originalFileSize':
--- a/plugins/arRestApiPlugin/modules/api/actions/digitalobjectsCreateAction.class.php
+++ b/plugins/arRestApiPlugin/modules/api/actions/digitalobjectsCreateAction.class.php
@@ -54,7 +54,9 @@ class ApiDigitalObjectsCreateAction extends QubitApiAction
         'format_name' => 'formatName',
         'format_version' => 'formatVersion',
         'format_registry_key' => 'formatRegistryKey',
-        'format_registry_name' => 'formatRegistryName'
+        'format_registry_name' => 'formatRegistryName',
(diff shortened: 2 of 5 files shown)


plugins/arRestApiPlugin/modules/api/actions/digitalobjectsCreateAction.class.php
```diff
@@ -54,7 +54,9 @@ protected function post($request, $payload)
         'format_name' => 'formatName',
         'format_version' => 'formatVersion',
         'format_registry_key' => 'formatRegistryKey',
-        'format_registry_name' => 'formatRegistryName'
+        'format_registry_name' => 'formatRegistryName',
+        'relative_path_within_aip' => 'relativePathWithinAip',
+        'aip_name' => 'aipName',
```
replaceafill: Just curious if this trailing comma may be code guideline issue since I don't see one in the deleted line?

lib/model/QubitInformationObject.php
```diff
@@ -74,6 +74,7 @@ public function __get($name)
       case 'formatRegistryName':
       case 'objectUUID':
       case 'aipUUID':
+      case 'aipName':
```
replaceafill, line 77: This property won't be created in the regular DIP upload (METS based), correct? I'm wondering if it's worth creating it even if it duplicates what's contained in `$aip->filename` just for consistency.

replaceafill (approved):
LGTM :+1: @sbreker. I just left a question about the new property and the regular DIP upload workflow.

plugins/arRestApiPlugin/modules/api/actions/digitalobjectsCreateAction.class.php
sbreker: I can remove this comma. 👍

lib/model/QubitInformationObject.php
sbreker: Great idea - adding....

PR #1240, opened 2021-01-16
Fix missing treeview children, refs #13423

djjuhasz:
In a number of scenarios an Elasticsearch information object document
will not have a value for the "children" field, which was preventing
the documents children from being displayed in the full width treeview.

To work around the problem I am doing an Elasticsearch query to count
each the children of each node in the tree.  This workaround is
significantly slower for large trees, but loads all of the tree nodes
correctly.

### DIFF
--- a/apps/qubit/modules/default/actions/fullTreeViewAction.class.php
+++ b/apps/qubit/modules/default/actions/fullTreeViewAction.class.php
@@ -132,14 +132,14 @@ class DefaultFullTreeViewAction extends sfAction
   }
 
   /**
-   * Do Elasticsearch query for given $term
+   * Get Elasticsearch query object for the given $term
    *
    * @param \Elastica\Query\Term $term query term
    * @param array $options optional arguments
    *
-   * @return \Elastica\ResultSet search result set
+   * @return \Elastica\Query query object
    */
-  protected function doElasticsearchQuery($term, $options = [])
+  protected function getElasticSearchQuery($term, $options = [])
   {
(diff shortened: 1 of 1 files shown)


mcantelon:
The code look's good! Would making sure "children" is populated in our ES data, instead, take a long time?

djjuhasz:
> The code look's good! Would making sure "children" is populated in our ES data, instead, take a long time?

To be honest, I don't know how much time it would take to make sure "children" is populated in ES, but I am assuming it would require updating code in multiple places where new archival descriptions are created.  I'm also not convinced that it's a good idea to have the "children" field in it's current form, which is a list of ids for *all* children of an archival description.   Because descriptions can have tens or hundreds of thousands of children, I'm am worried that storing all their ids will cause problems with ES search speeds or storage requirements in large collections.  For the full-width treeview, a list of children isn't required, the script just needs to know if the current information object has one or more children - and a simple boolean "hasChildren" or a count of the number of children would be sufficient, and would require storing a lot less data.

Right now I just don't have a lot of time to explore the different options for addressing this bug.  @mcantelon if you have time and want to look into the other options I would love to have a better solution.

mcantelon:
Good point about the problem with storing child IDs. I'd say :shipit: !

mcantelon (approved):
👍

djjuhasz:
Merged 40766011ccb64cc21c2591b15b7ef995d6a213e7 to qa/2.x

PR #1241, opened 2021-01-19
Reduce nested set build memory usage, refs #13384

jraddaoui:
- Do not save all statements and execute them while they are generated.
- Avoid keeping the entire tree in memory during the process.
- Remove already processed children from array.


sbreker (approved):
👍  Great work @jraddaoui!

jraddaoui:
Thanks @sbreker!

PR #1242, opened 2021-01-19
Add unit tests GA workflow, refs #13460

jraddaoui:
Adds a Github Actions workflow to run the existing unit tests over
Ubuntu 18.04 and a matrix of PHP versions. These tests require access
to MySQL/Percona and Elasticsearch, which are setup using the Docker
Compose environment (only those two services). The workflow will be
triggered on all PRs and new commits on qa/* and stable/* branches.

### DIFF
--- a/.github/workflows/unit-tests.yml
+++ b/.github/workflows/unit-tests.yml
@@ -1,0 +1,48 @@
+name: Unit tests
+on:
+  pull_request:
+  push:
+    branches:
+    - qa/**
+    - stable/**
+jobs:
+  unit-tests:
+    runs-on: ubuntu-18.04
+    strategy:
+      fail-fast: false
+      matrix:
+        php: [7.2, 7.3, 7.4]
+    name: PHP ${{ matrix.php }}
+    env:
+      COMPOSE_FILE: ${{ github.workspace }}/docker/docker-compose.dev.yml
+    steps:
+    - name: Setup PHP ${{ matrix.php }}
+      uses: shivammathur/setup-php@v2
+      with:
+        php-version: ${{ matrix.php }}
+        coverage: none
+    - name: Checkout code
+      uses: actions/checkout@v2
+    - name: Cache dependencies
+      uses: actions/cache@v2
+      with:
+        path: ~/.composer/cache/files
+        key: php-${{ matrix.php }}-composer-${{ hashFiles('composer.json') }}
+    - name: Install dependencies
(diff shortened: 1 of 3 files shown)


.github/workflows/unit-tests.yml
```diff
@@ -0,0 +1,48 @@
...
+    branches:
+    - qa/**
+    - stable/**
+jobs:
+  unit-tests:
+    runs-on: ubuntu-18.04
+    strategy:
+      fail-fast: false
+      matrix:
+        php: [7.2, 7.3, 7.4, 8.0]
```
djjuhasz: Let's remove the PHP 8.0 test until we have time to update the code so the tests pass.

test/bootstrap/config.php
```diff
@@ -0,0 +1,70 @@
...
+    array (
+      'class' => 'sfPropelDatabase',
+      'param' =>
+      array (
+        'encoding' => 'utf8mb4',
+        'persistent' => true,
+        'pooling' => true,
+        'dsn' => 'mysql:host=127.0.0.1;port=63003;dbname=atom;charset=utf8mb4',
+        'username' => 'atom',
+        'password' => 'atom_12345',
```
djjuhasz, line 16: Is having the password in plain text a security risk?

djjuhasz (changes requested):
Looks good to me @jraddaoui.  :)  I did request that we drop the PHP 8.0 test until we can get it to pass.

test/bootstrap/config.php
jraddaoui: I don't think it is ...

It's just an initial database for the tests and we already have that password in the code as default for the development env. Also the VM is up only for a few minutes and I guess it's not accessible from the outside and even if it's, the Percona port is only accessible internally:

https://github.com/artefactual/atom/blob/qa/2.x/docker/docker-compose.dev.yml#L57

jraddaoui:
It's worth mention that, apart from the unit tests, this workflow also tests the database and Elasticsearch index initialization using the purge task ...

https://github.com/artefactual/atom/blob/dde852f601ff620e47120a7608fb2e7f50a3b9f9/.github/workflows/unit-tests.yml#L45-L46

djjuhasz (approved):
Looks great! Thanks @jraddaoui :shipit:

jraddaoui:
c89065a in stable/2.6.x.

PR #1244, opened 2021-01-20
Replace Flash Uploader Uppy, refs #13319

sbreker:
This commit removes the YAHOO.widget.Uploader SFW Flash library. The
mulitupload window now uses Uppy.io to achieve equivalent functionality.

This commit also removes view.yml settings for classes that were
previously removed from the informationobject module.

### DIFF
--- a/apps/qubit/modules/informationobject/actions/multiFileUpdateAction.class.php
+++ b/apps/qubit/modules/informationobject/actions/multiFileUpdateAction.class.php
@@ -1,0 +1,122 @@
+<?php
+
+/*
+ * This file is part of the Access to Memory (AtoM) software.
+ *
+ * Access to Memory (AtoM) is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * Access to Memory (AtoM) is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with Access to Memory (AtoM).  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+class InformationObjectMultiFileUpdateAction extends sfAction
+{
+  protected $informationObjectList = array();
+  protected $informationObjectOrignalTitles = array();
+
+  public function execute($request)
+  { 
+    $this->resource = $this->getRoute()->resource;
+
+    // Check user authorization
+    if (!QubitAcl::check($this->resource, 'update'))
+    {
+      QubitAcl::forwardUnauthorized();
+    }
+
+    // Check that object exists and that it is not the root
+    if (!isset($this->resource) || !isset($this->resource->parent))
+    {
+      $this->forward404();
+    }
+
+    if (isset($request->items))
+    {
+      if (false === $this->items = explode(",", $request->items))
+      {
+        $this->forward404();
+      }
+
+      foreach ($this->items as $slug)
+      {
+        if (null !== $io = QubitInformationObject::getBySlug($slug))
+        {
+          if (!QubitAcl::check($io, 'update'))
+          {
+            continue;
+          }
+
+          // Child IOs should not be root and should be direct descendants of resource.
+          if (!isset($io->parent) || $io->parentId !== $this->resource->id)
+          {
+            continue;
+          }
+
+          $this->informationObjectList[$slug] = $io;
+          $this->informationObjectOrignalTitles[$slug] = $io->title;
+        }
+        else
+        {
+          continue;
+        }
+      }
+    }
+
+    $this->digitalObjectTitleForm = new DigitalObjectTitleUpdateForm(array(), 
+      array('informationObjects' => $this->informationObjectList));
+
+    // Handle POST data (form submit)
+    if ($request->isMethod('post'))
+    {
+      $this->digitalObjectTitleForm->bind($request->titles);
+
+        if ($this->digitalObjectTitleForm->isValid())
+        {
+          $this->updateDigitalObjectTitles();
+          $this->redirect(array($this->resource, 'module' => 'informationobject'));
+        }
+    }
+
+    $this->populateDigitalObjectTitleForm();
+  }
+
+  /**
+   * Populate the ui_label form with database values (localized)
+   */
+  protected function populateDigitalObjectTitleForm()
+  {
+    foreach ($this->digitalObjectTitleForm->getInformationObjects() as $io)
+    {
+      $this->digitalObjectTitleForm->setDefault($io->id, $io->getTitle(array('cultureFallback' => true)));
+    }
+  }
+
+  /**
+   * Update ui_label db values with form values (localized)
+   *
+   * @return $this
+   */
+  protected function updateDigitalObjectTitles()
+  {
+    foreach ($this->digitalObjectTitleForm->getInformationObjects() as $informationObject)
+    {
+      if (null !== $title = $this->digitalObjectTitleForm->getValue($informationObject->id))
+      {
+        // Test if title changed.
+        if ($this->informationObjectOrignalTitles[$informationObject->id] !== $title)
+        {
+          $informationObject->title = $title;
+          $informationObject->save();
+        }
+      }
+    }
+  }
+}
--- a/apps/qubit/modules/informationobject/actions/multiFileUploadAction.class.php
+++ b/apps/qubit/modules/informationobject/actions/multiFileUploadAction.class.php
@@ -43,16 +43,11 @@ class InformationObjectMultiFileUploadAction extends sfAction
       QubitAcl::forwardToSecureAction();
     }
 
-    // Add javascript libraries
-    $this->response->addJavaScript('/vendor/yui/logger/logger', 'last');
-    $this->response->addJavaScript('/vendor/yui/uploader/uploader-min', 'last');
-    $this->response->addJavaScript('multiFileUpload', 'last');
-
     // Get max upload size limits
-    $this->maxUploadSize = QubitDigitalObject::getMaxUploadSize();
+    $this->maxFileSize = QubitDigitalObject::getMaxUploadSize();
+    $this->maxPostSize = QubitDigitalObject::getMaxPostSize();
 
     // Paths for uploader javascript
-    $this->uploadSwfPath = "{$this->request->getRelativeUrlRoot()}/vendor/yui/uploader/assets/uploader.swf";
     $this->uploadResponsePath = "{$this->context->routing->generate(null, array('module' => 'digitalobject', 'action' => 'upload'))}?".http_build_query(array(session_name() => session_id()));
     $this->uploadTmpDir = "{$this->request->getRelativeUrlRoot()}/uploads/tmp";
 
@@ -90,6 +85,7 @@ class InformationObjectMultiFileUploadAction extends sfAction
 
     // Upload files
     $i = 0;
+    $informationObjectSlugList = array();
 
     foreach ($this->form->getValue('files') as $file)
     {
@@ -131,19 +127,15 @@ class InformationObjectMultiFileUploadAction extends sfAction
         $digitalObject->save();
       }
 
-      $thumbnailIsGeneric = (bool) strstr($file['thumb'], 'generic-icons');
+      $informationObjectSlugList[] = $informationObject->slug;
 
       // Clean up temp files
       if (file_exists("$tmpPath/$file[tmpName]"))
       {
         unlink("$tmpPath/$file[tmpName]");
       }
-      if (!$thumbnailIsGeneric && file_exists("$tmpPath/$file[thumb]"))
-      {
-        unlink("$tmpPath/$file[thumb]");
-      }
     }
 
-    $this->redirect(array($this->resource, 'module' => 'informationobject'));
+    $this->redirect(array($this->resource, 'module' => 'informationobject', 'action' => 'multiFileUpdate', 'items' => implode(",", $informationObjectSlugList)));
   }
 }
--- a/apps/qubit/modules/informationobject/templates/multiFileUpdateSuccess.php
+++ b/apps/qubit/modules/informationobject/templates/multiFileUpdateSuccess.php
@@ -1,0 +1,65 @@
+<?php decorate_with('layout_1col.php') ?>
+
+<?php slot('title') ?>
+  <h1 class="multiline">
+    <?php echo __('Update digital object titles') ?>
+    <span class="sub"><?php echo render_title(new sfIsadPlugin($resource)) ?> </span>
+  </h1>
+<?php end_slot() ?>
+
+<?php slot('content') ?>
+
+  <?php echo $digitalObjectTitleForm->renderGlobalErrors() ?>
+  <?php echo $digitalObjectTitleForm->renderFormTag(url_for(array($resource, 'module' => 'informationobject', 'action' => 'multiFileUpdate', 'items' => $sf_request->items)), array('method' => 'post', 'id' => 'bulk-title-update-form')) ?>
+    <?php echo $digitalObjectTitleForm->renderHiddenFields() ?>
+
+    <div id="content">
+
+      <table class="table sticky-enabled">
+        <thead>
+          <tr>
+            <th><?php echo __('Object')?></th>
+            <th><?php echo __('Title')?></th>
+          </tr>
+        </thead>
+        <tbody>
+          <?php foreach ($digitalObjectTitleForm->getInformationObjects() as $io): ?>
+            <tr>
+              <td class="thumbnail-container">
+                <?php foreach ($io->digitalObjectsRelatedByobjectId as $do): ?>
+                  <?php if (
+                    (null !== $thumbnail = $do->getRepresentationByUsage(QubitTerm::THUMBNAIL_ID))
+                    && QubitAcl::check($io, 'readThumbnail')
+                  ): ?>
+                    <?php echo image_tag($thumbnail->getFullPath(), array('alt' => __($do->getDigitalObjectAltText() ?: 'Original %1% not accessible', array('%1%' => sfConfig::get('app_ui_label_digitalobject'))))) ?>
+                  <?php else: ?>
+                    <?php echo image_tag(QubitDigitalObject::getGenericIconPathByMediaTypeId($do->mediaTypeId), array('alt' => __($do->getDigitalObjectAltText() ?: 'Original %1% not accessible', array('%1%' => sfConfig::get('app_ui_label_digitalobject'))))) ?>
+                  <?php endif; ?>
+                <?php endforeach; ?>
+              </td>
+              <td>
+                <?php if ($sf_user->getCulture() != $io->getSourceCulture() && !strlen($io->title)): ?>
+                  <div class="default-translation">
+                    <?php echo render_value($digitalObjectTitleForm[$io->id]->getValue(), $io) ?>
+                  </div>
+                <?php endif; ?>
+
+                <?php echo $digitalObjectTitleForm[$io->id]
+                    ->label(__('Title')) ?>
+                <?php echo __($io->digitalObjectsRelatedByobjectId[0]->name) ?>
+                <?php echo render_show(__('Level of description'), render_value_inline($io->levelOfDescription)) ?>
+              </td>
+            </tr>
+          <?php endforeach; ?>
+        </tbody>
+      </table>
+    </div>
+
+    <section class="actions">
+      <ul>
+        <li><input class="c-btn c-btn-submit" id="rename-form-submit" type="submit" value="<?php echo __('Save') ?>"/></li>
+      </ul>
+    </section>
+  </form>
+
+<?php end_slot() ?>
--- a/apps/qubit/modules/informationobject/templates/multiFileUploadSuccess.php
+++ b/apps/qubit/modules/informationobject/templates/multiFileUploadSuccess.php
@@ -22,18 +22,6 @@
     </section>
   </noscript>
 
-  <div id="no-flash" style="display: none;">
-    <div class="messages warning">
-      <?php echo __('Your browser does not have Flash player installed. To import digital objects, please <a href="https://get.adobe.com/flashplayer/">Download Adobe Flash Player</a> (requires version 9.0.45 or higher)') ?>
-    </div>
-
-    <section class="actions">
-      <ul>
-        <li><?php echo link_to(__('Cancel'), array($resource, 'module' => 'informationobject'), array('class' => 'c-btn')) ?></li>
-      </ul>
-    </section>
-  </div>
-
   <?php if (QubitDigitalObject::reachedAppUploadLimit()): ?>
 
     <div id="upload_limit_reached">
@@ -45,14 +33,14 @@
         <ul>
           <li><?php echo link_to(__('Cancel'), array($resource, 'module' => 'informationobject')) ?></li>
         </ul>
-      </sectin>
+      </section>
     </div>
 
   <?php else: ?>
 
     <?php echo $form->renderGlobalErrors() ?>
 
-    <?php echo $form->renderFormTag(url_for(array($resource, 'module' => 'informationobject', 'action' => 'multiFileUpload')), array('id' => 'multiFileUploadForm', 'style' => 'display: none')) ?>
+    <?php echo $form->renderFormTag(url_for(array($resource, 'module' => 'informationobject', 'action' => 'multiFileUpload')), array('id' => 'multiFileUploadForm', 'style' => 'inline')) ?>
 
       <?php echo $form->renderHiddenFields() ?>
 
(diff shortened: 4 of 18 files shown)


apps/qubit/modules/informationobject/actions/multiFileUpdateAction.class.php
```diff
@@ -0,0 +1,141 @@
...
+{
+  protected $informationObjectList = array();
+  protected $informationObjectOrignalTitles = array();
+
+  protected function earlyExecute()
+  {
+    $this->resource = $this->getRoute()->resource;
+
+    // Check user authorization
+    if (!QubitAcl::check($this->resource, 'update') && !$this->getUser()->hasGroup(QubitAclGroup::EDITOR_ID))
```
jraddaoui: Why are we forcing editor group? Shouldn't we let QubitAcl handle that?

apps/qubit/modules/informationobject/actions/multiFileUpdateAction.class.php
```diff
@@ -0,0 +1,141 @@
...
+    {
+      QubitAcl::forwardUnauthorized();
+    }
+  }
+
+  public function execute($request)
+  { 
+    $this->resource = $this->getRoute()->resource;
```
jraddaoui: Already populated in `earlyExecute`.

apps/qubit/modules/informationobject/actions/multiFileUpdateAction.class.php
```diff
@@ -0,0 +1,141 @@
...
+
+    // Check that object exists and that it is not the root
+    if (!isset($this->resource) || !isset($this->resource->parent))
+    {
+      $this->forward404();
+    }
```
jraddaoui, line 39: Maybe this should be part of the `earlyExecute` before the ACL check, or just remove the `earlyExecute` method, I don't think it has a lot of value in here.

apps/qubit/modules/informationobject/actions/multiFileUpdateAction.class.php
```diff
@@ -0,0 +1,141 @@
...
+
+    $this->requestIds = $request->ids;
```
jraddaoui: Not a big deal, but `$this->requestIds` is only used in the template, where you could access the request directly and avoid this assignment.

apps/qubit/modules/informationobject/actions/multiFileUpdateAction.class.php
```diff
@@ -0,0 +1,141 @@
...
+      if (false === $this->ids = explode(",", $request->ids))
+      {
+        $this->forward404();
+      }
+
+      foreach ($this->ids as $id)
+      {
+        if (!ctype_digit($id))
+        {
+          $this->forward404();
```
jraddaoui: I know we "have control" over those ids and that a non digit value is not expected, but this feels like a `continue`.

apps/qubit/modules/informationobject/actions/multiFileUpdateAction.class.php
```diff
@@ -0,0 +1,141 @@
...
+        }
+
+        if (null !== $io = QubitInformationObject::getById($id))
+        {
+          if (!QubitAcl::check($io, 'update') && !$this->getUser()->hasGroup(QubitAclGroup::EDITOR_ID))
```
jraddaoui: Same here about the editor group.

apps/qubit/modules/informationobject/actions/multiFileUpdateAction.class.php
```diff
@@ -0,0 +1,141 @@
...
+          {
+            QubitAcl::forwardUnauthorized();
```
jraddaoui: Same here about the continue.

apps/qubit/modules/informationobject/actions/multiFileUpdateAction.class.php
```diff
@@ -0,0 +1,141 @@
...
+          }
+
+          // Child IOs should not be root and should be direct descendants of resource.
+          if (!isset($io->parent) || $io->parentId !== $this->resource->id)
+          {
+            $this->forward404();
```
jraddaoui: Continue?

apps/qubit/modules/informationobject/actions/multiFileUpdateAction.class.php
```diff
@@ -0,0 +1,141 @@
...
+          }
+
+          $this->informationObjectList[$id] = $io;
+          $this->informationObjectOrignalTitles[$id] = $io->title;
+        }
+        else
+        {
+          $this->forward404();
```
jraddaoui: Continue?

apps/qubit/modules/informationobject/actions/multiFileUpdateAction.class.php
```diff
@@ -0,0 +1,141 @@
...
+      array('informationObjects' => $this->informationObjectList));
+
+    // Handle POST data (form submit)
+    if ($request->isMethod('post'))
+    {
+      $this->digitalObjectTitleForm->bind($request->titles);
+
+        if ($this->digitalObjectTitleForm->isValid())
+        {
+          $this->updateDigitalObjectTitles($this->digitalObjectTitleForm);
```
jraddaoui: No need to pass the form, being a global var.

apps/qubit/modules/informationobject/actions/multiFileUpdateAction.class.php
```diff
@@ -0,0 +1,141 @@
...
+          $this->redirect(array($this->resource, 'module' => 'informationobject'));
+        }
+    }
+
+    $this->populateDigitalObjectTitleForm($this->digitalObjectTitleForm);
```
jraddaoui: Same here, no need to pass the form.

apps/qubit/modules/informationobject/actions/multiFileUpdateAction.class.php
```diff
@@ -0,0 +1,141 @@
...
+        // Test if title changed.
+        if ($this->informationObjectOrignalTitles[$informationObject->id] !== $title)
+        {
+          $informationObject->title = $title;
+          $informationObject->save();
+        }
+      }
+    }
+
+    return $this;
```
jraddaoui: Not needed.

apps/qubit/modules/informationobject/actions/multiFileUpdateAction.class.php
```diff
@@ -0,0 +1,141 @@
...
+  }
+
+  protected function processForm()
+  {
+    $this->redirect(array($this->resource, 'module' => 'informationobject'));
+  }
```
jraddaoui: I think this function is not used in the end.

apps/qubit/modules/informationobject/actions/multiFileUploadAction.class.php
```diff
@@ -90,6 +90,7 @@ public function processForm()
 
     // Upload files
     $i = 0;
+    $idList = array();
```
jraddaoui: Called `$informationObjectIdList` later on ...

apps/qubit/modules/informationobject/actions/multiFileUploadAction.class.php
```diff
@@ -44,15 +44,15 @@ public function execute($request)
     }
 
     // Add javascript libraries
-    $this->response->addJavaScript('/vendor/yui/logger/logger', 'last');
-    $this->response->addJavaScript('/vendor/yui/uploader/uploader-min', 'last');
+    $this->response->addJavaScript('/vendor/uppy/uppy-bundle.js', 'last');
+    $this->response->addStylesheet('/vendor/uppy/uppy-bundle.css', 'last');
     $this->response->addJavaScript('multiFileUpload', 'last');
```
jraddaoui: Could we remove these from here and add them to the view.yml file for consistency? I could see a point if they were optional, like the fullwidth treeview.

apps/qubit/modules/informationobject/actions/multiFileUpdateAction.class.php
jraddaoui: Also, exposing database ids is not ideal, I'd suggest to change the entire process to use slugs, specially if you can avoid passing them in the request (see notes below, in the template).

apps/qubit/modules/informationobject/templates/multiFileUpdateSuccess.php
```diff
@@ -0,0 +1,65 @@
...
+  <h1 class="multiline">
+    <?php echo __('Update digital object titles') ?>
+    <span class="sub"><?php echo render_title(new sfIsadPlugin($resource)) ?> </span>
+  </h1>
+<?php end_slot() ?>
+
+<?php slot('content') ?>
+
+  <?php echo $digitalObjectTitleForm->renderGlobalErrors() ?>
+  <?php echo $digitalObjectTitleForm->renderFormTag(url_for(array($resource, 'module' => 'informationobject', 'action' => 'multiFileUpdate', 'ids' => $requestIds)), array('method' => 'post', 'id' => 'bulk-title-update-form')) ?>
```
jraddaoui: Why do we need to add the ids here? Could we handle them entirely in the controller? Could we use slugs instead?

apps/qubit/modules/informationobject/templates/multiFileUploadSuccess.php
```diff
@@ -22,18 +22,6 @@
...
-      <?php echo __('Your browser does not have Flash player installed. To import digital objects, please <a href="https://get.adobe.com/flashplayer/">Download Adobe Flash Player</a> (requires version 9.0.45 or higher)') ?>
-    </div>
-
-    <section class="actions">
-      <ul>
-        <li><?php echo link_to(__('Cancel'), array($resource, 'module' => 'informationobject'), array('class' => 'c-btn')) ?></li>
-      </ul>
-    </section>
-  </div>
-
```
jraddaoui, line 36: (╯°□°）╯︵ ┻━┻

apps/qubit/modules/informationobject/templates/multiFileUploadSuccess.php
```diff
@@ -45,14 +33,14 @@
         <ul>
           <li><?php echo link_to(__('Cancel'), array($resource, 'module' => 'informationobject')) ?></li>
         </ul>
-      </sectin>
+      </section>
```
jraddaoui, line 36: lol

apps/qubit/modules/informationobject/templates/multiFileUploadSuccess.php
```diff
@@ -62,6 +50,12 @@
 
           <legend><?php echo __('Import multiple digital objects') ?></legend>
 
+          <div class="alert alert-info">
+            <p><?php echo __('Add your digital objects by dragging and dropping local files into the pane below, or by clicking the browse link to open your local file explorer.') ?></p>
+            <p><?php echo __('The Title and Level of description values entered on this page will be applied to each child description created for the associated digital objects - %dd% represents an incrementing 2-value number, so by default descriptions created via this uploader will be named image 01, image 02, etc.') ?></p>
```
jraddaoui: Is `%dd%` going to be an issue with the translations? I think the %whatever% pattern is used to replace content.

lib/form/DigitalObjectTitleUpdateForm.class.php
```diff
@@ -0,0 +1,73 @@
...
+ * @package    AccesstoMemory
+ * @subpackage Information Object
+ * @author     Steve Breker <sbreker@artefactual.com>
+ */
+class DigitalObjectTitleUpdateForm extends sfForm
+{
+  public function configure()
+  {
+    // Build widgets and validators
+    foreach($this->getInformationObjects() as $informationObject) {
```
jraddaoui: At the moment, line break before brackets. We'll see when we implement the syntax checker.

lib/form/DigitalObjectTitleUpdateForm.class.php
```diff
@@ -0,0 +1,73 @@
...
+  {
+    $this->setOption('informationObjects', $objects);
+
+    return $this;
+  }
+
+  public function getInformationObjects()
+  {
+    return $this->getOption('informationObjects');
+  }
```
jraddaoui, line 73: Could you pass the IOs ids/slugs on the form creation? That way I think we could remove this functions, assign the default values directly here (instead of in `populateDigitalObjectTitleForm` at the controller) and work directly with the form fields instead of with the option.

js/multiFileUpload.js
```diff
@@ -1,351 +1,390 @@
-(function ($)
+(function ($) {
+
+  "use strict";
+
+  Qubit.multiFileUpload = Qubit.multiFileUpload || {};
```
jraddaoui, line 5: I doubt this will work if `Qubit.multiFileUpload` doesn't exist and it's populated with and empty object, so maybe we can remove this line.

js/multiFileUpload.js
```diff
@@ -1,351 +1,390 @@
...
+
+  var MultiFileUpload = function (element)
```
jraddaoui: Something that I learnt recently and will make this a bit cleaner. A var is not actually needed, you can name the function directly ... `function MultiFileUpload(element)`. Then, the prototype will extend this function and the function content is assigned directly as the constructor, so it doesn't need to be passed again. Check the latest version of `clipboard.js`.

js/multiFileUpload.js
```diff
@@ -1,351 +1,390 @@
...
+      .attr('type','hidden')
+      .text(Qubit.multiFileUpload.i18nRetry)
+      .appendTo('.actions');
+
+    this.init();
+    this.listen();
+  };
+
+  MultiFileUpload.prototype = {
+    constructor: MultiFileUpload,
```
jraddaoui: We won't need this if we follow the comment above, as it's the default behavior when a function is extended with prototype.

js/multiFileUpload.js
```diff
@@ -1,351 +1,390 @@
...
-          function parseHtmlId(id)
-          {
-            var yuiId = id.match(/([0-9]+)$/).shift();
-
-            if (null != yuiId && !isNaN(parseInt(yuiId)))
-            {
-              return yuiId;
+    this.uppy = new Uppy.Core({ 
+      debug: false,
+      id: 'Uppy-atom',
```
jraddaoui: Just cosmetic, is the uppercase needed?

js/multiFileUpload.js
```diff
@@ -1,351 +1,390 @@
...
+            { id: 'title', name: Qubit.multiFileUpload.i18nInfoObjectTitle },
+          ],
+        })
+        .use(Uppy.XHRUpload, {
+          endpoint: Qubit.multiFileUpload.uploadResponsePath,
+          formData: true,
+          method: 'post',
+          limit: 10,
+          fieldName: 'Filedata',
+          objectId: Qubit.multiFileUpload.objectId,
```
jraddaoui: I'd prefer to use the slug if possible.

apps/qubit/modules/informationobject/templates/multiFileUploadSuccess.php
```diff
@@ -104,42 +95,48 @@
...
-YAHOO.widget.Uploader.SWFURL = '$uploadSwfPath';
-
-Qubit.multiFileUpload.maxUploadSize = '$maxUploadSize';
-Qubit.multiFileUpload.uploadTmpDir = '$uploadTmpDir';
+
+<?php echo javascript_tag(<<<content
+Qubit.multiFileUpload.maxFileSize = '$maxFileSize';
+Qubit.multiFileUpload.maxPostSize = '$maxPostSize';
 Qubit.multiFileUpload.uploadResponsePath = '$uploadResponsePath';
 Qubit.multiFileUpload.objectId = '$resource->id';
```
jraddaoui: I'd prefer to use the slug, if possible.

js/multiFileUpload.js
```diff
@@ -1,351 +1,390 @@
...
+    {
+      // Intercept AtoM's Submit button.
+      this.$submitButton.on('click', $.proxy(this.onSubmitButton, this));
+      this.$retryButton.on('click', $.proxy(this.onRetryButton, this));
+    },
+
+    // Retry is available if some/all DO's do not successfully upload.
+    onRetryButton: function ()
+    {
+      this.uppy.retryAll().then((result) => {
```
jraddaoui, line 141: I see some ES6 arrow functions over here, something that I think we have not used before, but I think it's okay to have them nowadays, considering that only (the close to death) IE doesn't support them.

js/multiFileUpload.js
```diff
@@ -1,351 +1,390 @@
...
+          this.$retryButton.hide();
+          this.showAlert(Qubit.multiFileUpload.i18nRetrySuccess, 'alert-info');
+        }
+      })
+    },
+
+    // Checks if ANY uploads were successful.
+    checkUploadSuccessful: function ()
+    {
+      const uploaded = (element) => element.progress.uploadComplete === true;
```
jraddaoui, line 152: Love them!

js/multiFileUpload.js
```diff
@@ -1,351 +1,390 @@
...
+              '<input type="hidden" class="md5sum" value="' + fileResponse.body.md5sum + '"/>' +
+              '<input type="hidden" name="files[' + file.id + '][name]" value="' + fileResponse.body.name + '"/>' +
+              '<input type="hidden" name="files[' + file.id + '][md5sum]" value="' + fileResponse.body.md5sum + '"/>' +
+              '<input type="hidden" name="files[' + file.id + '][tmpName]" value="' + fileResponse.body.tmpName + '"/>' +
+              '<input type="hidden" class="title" name="files[' + file.id + '][infoObjectTitle]" value="' + file.meta.title + '"/>' +
+            '</div>' +
+          '</div>' +
+        '</div>')
+        .appendTo("#uploads");
+      });
```
jraddaoui, line 249: I know it's not the best at this stage, but if it's possible, I'd rather see the slugs being used all over the place instead of DB ids.

js/multiFileUpload.js
```diff
@@ -1,351 +1,390 @@
...
-          // uploader.addListener('mouseUp', function () { });
-        }
+      return fileName;
     }
-  })(jQuery);
+  };
+
+  $(function ()
+  {
+    var $node = $('body');
```
jraddaoui: Could we look for a more restrictive element, even if it's added only in this page, starting it after the body doesn't feel right.

jraddaoui (changes requested):
A lot less disruptive than I was expecting @sbreker, fantastic job!

I've added a bunch of comments below, but nothing really important.

lib/form/DigitalObjectTitleUpdateForm.class.php
sbreker: Fixed.

apps/qubit/modules/informationobject/actions/multiFileUpdateAction.class.php
sbreker: Good point. This was historical I seem to recall but your point is valid. I will remove these extra checks.

apps/qubit/modules/informationobject/actions/multiFileUpdateAction.class.php
sbreker: Fixed. I have removed earlyExecute() entirely.

apps/qubit/modules/informationobject/actions/multiFileUpdateAction.class.php
sbreker: See above. earlyExecute() has been removed.

apps/qubit/modules/informationobject/actions/multiFileUpdateAction.class.php
sbreker: Removed `$this->requestIds` in favour of referring to sf_request in the template.

apps/qubit/modules/informationobject/actions/multiFileUpdateAction.class.php
sbreker: Updated to `continue`.

apps/qubit/modules/informationobject/actions/multiFileUpdateAction.class.php
sbreker: Fixed as above.

apps/qubit/modules/informationobject/actions/multiFileUpdateAction.class.php
sbreker: Updated to `continue`.

apps/qubit/modules/informationobject/actions/multiFileUpdateAction.class.php
sbreker: Updated to `continue`.

apps/qubit/modules/informationobject/actions/multiFileUpdateAction.class.php
sbreker: Updated to `continue`.

apps/qubit/modules/informationobject/actions/multiFileUpdateAction.class.php
sbreker: Fixed in both methods.

apps/qubit/modules/informationobject/actions/multiFileUpdateAction.class.php
sbreker: Fixed in both methods.

apps/qubit/modules/informationobject/actions/multiFileUpdateAction.class.php
sbreker: Removed.

apps/qubit/modules/informationobject/actions/multiFileUpdateAction.class.php
sbreker: Removed.

apps/qubit/modules/informationobject/actions/multiFileUploadAction.class.php
sbreker: Good catch.

apps/qubit/modules/informationobject/actions/multiFileUploadAction.class.php
sbreker: Good point. These were historically in this file... I have moved these to the `view.yml` file.

apps/qubit/modules/informationobject/templates/multiFileUploadSuccess.php
sbreker: 👍

js/multiFileUpload.js
sbreker: 👍

js/multiFileUpload.js
sbreker: Using slug instead of object id now for image upload from Uppy to digitalobject/uploadAction.php. See latest commits.

apps/qubit/modules/informationobject/templates/multiFileUploadSuccess.php
sbreker: Using slug instead of object id now for image upload from Uppy to digitalobject/uploadAction.php. See latest commits.

js/multiFileUpload.js
sbreker: My thinking as well.

js/multiFileUpload.js
sbreker: That was the Uppy way. I have updated it to be all lower case as it should not affect anything.

js/multiFileUpload.js
sbreker: Removed.

js/multiFileUpload.js
sbreker: Interesting - I've changed the syntax as you suggested. See latest commits.

js/multiFileUpload.js
sbreker: Updated as above.

js/multiFileUpload.js
sbreker: I have moved this to the .multiFileUpload div.

js/multiFileUpload.js
sbreker: Leaving in as it seems necessary. If removed then Qubit.multiFileUpload is unknown to the js function.

apps/qubit/modules/informationobject/templates/multiFileUpdateSuccess.php
sbreker: Updated to use slugs.

apps/qubit/modules/informationobject/templates/multiFileUploadSuccess.php
sbreker: I have exported the XLIFF with this '%dd%' string included and it does not seem to cause any issues with extracting the string. I have added single quotes around '%dd%' for clarity when reading.

```
<trans-unit id="906116b1dd3e40cf3c978815d31c78ba03e1f034">
        <source>The Title and Level of description values entered on this page will be applied to each child description created for the associated digital objects - '%dd%' represents an incrementing 2-value number, so by default descriptions created via this uploader will be named image 01, image 02, etc.</source>
        <target/>
        <note>https://github.com/artefactual/atom/blob/master/apps/qubit/modules/informationobject/templates/multiFileUploadSuccess.php</note>
      </trans-unit>
```

js/multiFileUpload.js
sbreker: These are not object.id values but rather Uppy's unique id's for each image. At this point in the import, these digital objects do not yet have a slug or id.

lib/form/DigitalObjectTitleUpdateForm.class.php
jraddaoui: Not a big deal if you want to skip this suggestion.

js/multiFileUpload.js
jraddaoui: Oh, I see!

jraddaoui (approved):
Thanks a lot Steve! As discussed, we can leave the modification to avoid passing multiple slugs in the URL for a following iteration, hopefully for 2.7.

PR #1245, opened 2021-01-21
Clarify menu label/description for reference_image_maxwidth setting to better reflect behaviour

mInnes-archives:
The labels in the 'Digital object derivatives' menu for the `reference_image_maxwidth` setting don't accurately reflect its behaviour. The `reference_image_maxwidth` value is actually being used to determine the maximum number of pixels on the longest edge for the derived image. This can be either its height or width depending on whichever is larger.

While this wording may be a bit awkward, thinking in terms of the number of pixels on the longest edge is pretty standard for those working with image assets. For example, this is commonly used to determine the DPI when digitising photographs.

This pull request does introduce some inaccuracies in the translations but the values in the`<source>` tags have been updated to still facilitate the switch to the existing translations.

This is a pretty minor pull request, but since the original description caused me a bit of confusion, I thought it might help clarify things for others (or at least those who have AtoM set to English).

### DIFF
--- a/apps/qubit/i18n/ar/messages.xml
+++ b/apps/qubit/i18n/ar/messages.xml
@@ -2821,7 +2821,7 @@
         <note>https://github.com/artefactual/atom/blob/master/apps/qubit/modules/digitalobject/actions/editAction.class.php</note>
       </trans-unit>
       <trans-unit id="9917379c2d4f57eb0424e346a3a86a01f46d0d49">
-        <source>Maximum image width (pixels)</source>
+        <source>Maximum length on longest edge (pixels)</source>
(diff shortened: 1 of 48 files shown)


jraddaoui:
Thanks @mInnes-archives, 

That's a great point and I just confirmed this behavior. The messages files are updated with a CLI task that is used in the process to upload the new strings to the translation platform, so we'd only need to merge the first commit and the messages files will be updated later before the translations announcement and release.

@fiver-watson, are you okay with the changes? Do we need a ticket and a reference?

fiver-watson:
Hi @mInnes-archives and @jraddaoui, 

Matt, thanks for catching this and submitting another PR! I've created a corresponding ticket in our issue tracker here: 

* https://projects.artefactual.com/issues/13471

I can't think of a better way to word this, so I'm happy with the changes. Since we'll need to pick just the first commit, Radda can handle ensuring that the commit message refs the issue ticket number. 

Thanks again!

mInnes-archives:
>The messages files are updated with a CLI task that is used in the process to upload the new strings to the translation platform, so we'd only need to merge the first commit and the messages files will be updated later before the translations announcement and release.

Great, good to know!

jraddaoui:
Thanks again @mInnes-archives! Merged in 6629c1f.

PR #1246, opened 2021-01-26
Add logic to paginate I.O. indexing, refs #13464

mcantelon:
Ran into a situation with a client where memory ran out during indexing. Paging indexing, via a custom script, worked around this.

Modified classes, related to Elastsearch indexing of information objects, to allow indexing to be paginated using custom scripts.

### DIFF
--- a/plugins/arElasticSearchPlugin/lib/model/arElasticSearchInformationObject.class.php
+++ b/plugins/arElasticSearchPlugin/lib/model/arElasticSearchInformationObject.class.php
@@ -41,15 +41,7 @@ class arElasticSearchInformationObject extends arElasticSearchModelBase
   {
     $this->load();
 
-    // Pass root data to top-levels to avoid ancestors query
-    $ancestors = array(array(
-      'id' => QubitInformationObject::ROOT_ID,
-      'identifier' => null,
-      'repository_id' => null
-    ));
-
-    // Recursively descend down hierarchy
-    $this->addInformationObjects(['ancestors' => $ancestors]);
+    $this->addInformationObjects();
 
     return $this->errors;
   }
@@ -59,10 +51,19 @@ class arElasticSearchInformationObject extends arElasticSearchModelBase
     $skip = isset($options['skip']) ? $options['skip'] : 0;
     $limit = isset($options['limit']) ? $options['limit'] : $this->count;
 
-    // Loop through children and add to search index
-    $sql = "SELECT id, parent_id FROM information_object";
-    $sql .= " WHERE id <> ? ";
-    $sql .= " ORDER BY lft, parent_id, id ";
+    // Loop through hierarchy and add to search index
+    $sql = "WITH RECURSIVE cte (id, parent_id) AS (
+              SELECT id, parent_id
+              FROM information_object
+              WHERE parent_id = ?
+              UNION ALL
+              SELECT i.id, i.parent_id
+              FROM information_object i
+              INNER JOIN cte
+              ON i.parent_id = cte.id
+            )
+            SELECT * FROM cte";
+
     $sql .= sprintf(" LIMIT %d, %d", $skip, $limit);
 
     $rows = QubitPdo::fetchAll(
@@ -71,22 +72,20 @@ class arElasticSearchInformationObject extends arElasticSearchModelBase
       ['fetchMode' => PDO::FETCH_ASSOC]
     );
 
+    $lastParentId = null;
+
     foreach ($rows as $row)
     {
-      $ancestors = $inheritedCreators = array();
-      $repository = null;
       self::$counter++;
 
 print 'ID:'. $row['id'] ."\n";
       try
       {
-        if ($row['parent_id'] == 1)
+        if ($lastParentId != $row['parent_id'])
         {
-          $options['ancestors'] = array(array(
-            'id' => QubitInformationObject::ROOT_ID,
-            'identifier' => null,
-            'repository_id' => null
-          ));
+          unset($options['ancestors']);
+          unset($options['repository']);
+          unset($option['inheritedCreators']);
         }
 
         $node = new arElasticSearchInformationObjectPdo($row['id'], $options);
@@ -96,17 +95,15 @@ print 'ID:'. $row['id'] ."\n";
 
         $this->logEntry($data['i18n'][$data['sourceCulture']]['title'], self::$counter);
 
-        $options['ancestors'] = array_merge($node->getAncestors(), array(array(
-          'id' => $node->id,
-          'identifier' => $node->identifier,
-          'repository_id' => $node->repository_id
-        )));
-        $options['repository'] = $node->getRepository();
-        $option['inheritedCreators'] = array_merge($node->inheritedCreators, $node->creators);
+        // If under a different parent then cache related data
+        if ($lastParentId != $row['parent_id'])
+        {
+          $options['ancestors'] = $node->getAncestors();
+          $options['repository'] = $node->getRepository();
+          $option['inheritedCreators'] = $node->inheritedCreators;          
+        }
 
-#$options['ancestors'] = $ancestors;
-#$options['inheritedCreators'] = $inheritedCreators;
-#$options['repository'] = $repository;
+        $lastParentId = $row['parent_id'];
       }
       catch (sfException $e)
       {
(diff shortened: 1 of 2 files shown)


djjuhasz (changes requested):
@mcantelon I don't understand how this is going to work with hierarchical data. 

I.e. Archival description Alpha has 150 children.  You set a "skip" value of 100.   When you run the script, Alpha is indexed and so are the first 99 children of Alpha (Alpha + 99 = 100).  How do you then index the other 51 children of Alpha?  

Why not break out of the foreach loop after $counter > $skip? After $counter > $skip nothing else in the loop is going to get indexed.

Why "$skip" for a limit on the number of docs to index? "Skip" usually means "skip the first X records" so I find it confusing that it is meant to "stop indexing after X records".

plugins/arElasticSearchPlugin/lib/model/arElasticSearchInformationObject.class.php
```diff
@@ -198,4 +206,9 @@ public static function setBoostValues(&$fields, $cultures)
 
     self::addBoostValuesToFields($fields, $i18nBoostFields, $nonI18nBoostFields);
   }
+
+  public function getErrors()
```
djjuhasz, line 204: This doesn't seem related to the $skip code.

plugins/arElasticSearchPlugin/lib/model/arElasticSearchModelBase.class.php
```diff
@@ -45,6 +45,11 @@ public function getCount()
     return $this->count;
   }
 
+  public function setCount($count)
```
djjuhasz: I don't see how this related to the $skip code either.

djjuhasz:
> 
> 
> @mcantelon I don't understand how this is going to work with hierarchical data.
> 
> I.e. Archival description Alpha has 150 children. You set a "skip" value of 100. When you run the script, Alpha is indexed and so are the first 99 children of Alpha (Alpha + 99 = 100). How do you then index the other 51 children of Alpha?
> 
> Why not break out of the foreach loop after $counter > $skip? After $counter > $skip nothing else in the loop is going to get indexed.
> 
> Why "$skip" for a limit on the number of docs to index? "Skip" usually means "skip the first X records" so I find it confusing that it is meant to "stop indexing after X records".

Oh, I see now, the $skip is actually skipping the first X records, I was reading the conditional wrong. :(

So how do you limit the number of descriptions indexed on each "page"?  Is there code in AtoM already to stop after indexing Y records?

plugins/arElasticSearchPlugin/lib/model/arElasticSearchInformationObject.class.php
```diff
@@ -60,6 +60,8 @@ public function populate()
 
   public function recursivelyAddInformationObjects($parentId, $totalRows, $options = array())
   {
+    $skip = isset($options['skip']) ? $options['skip'] : null;
```
djjuhasz: If you set `$skip` to zero by default, it makes the conditional check later easier.

plugins/arElasticSearchPlugin/lib/model/arElasticSearchInformationObject.class.php
```diff
@@ -69,12 +71,17 @@ public function recursivelyAddInformationObjects($parentId, $totalRows, $options
 
       try
       {
+        // Index document if within paging limits
         $node = new arElasticSearchInformationObjectPdo($item->id, $options);
-        $data = $node->serialize();
 
-        QubitSearch::getInstance()->addDocument($data, 'QubitInformationObject');
+        if (!is_numeric($skip) || (self::$counter > $skip && self::$counter <= ($skip + $totalRows)))
```
djjuhasz: If `$skip` defaults to zero, then you don't need the `!is_numeric($skip)` check.

plugins/arElasticSearchPlugin/lib/model/arElasticSearchInformationObject.class.php
mcantelon: This just allows search errors to be fetched by an external script. I forgot that I didn't include a link to an example script in this PR. My bad.

plugins/arElasticSearchPlugin/lib/model/arElasticSearchModelBase.class.php
mcantelon: This sets the "limit" of the indexing according to the logic.

plugins/arElasticSearchPlugin/lib/model/arElasticSearchInformationObject.class.php
mcantelon: Good call. I'll update the PR.

mcantelon:
Here's the external script that I was using to do paged indexing (for a client that was running out of memory indexing millions of information objects)... hopefully this clears up the usage (I should have included a link to this in the PR):

https://gist.githubusercontent.com/mcantelon/5260291a778568f755cc2a82e9e5127e/raw/7bd4455b8edae43055a5e4767a33950b80f11e70/io-index.php

djjuhasz:
> Here's the external script that I was using to do paged indexing (for a client that was running out of memory indexing millions of information objects)... hopefully this clears up the usage (I should have included a link to this in the PR):
> 
> https://gist.githubusercontent.com/mcantelon/5260291a778568f755cc2a82e9e5127e/raw/7bd4455b8edae43055a5e4767a33950b80f11e70/io-index.php

Thanks for that extra context @mcantelon :)

djjuhasz:
@mcantelon I have two major complaints with the $skip functionality as written:

1. It requires loading and scrolling through all the information objects before the `$skip` value
2. The script keep scrolling through the hierarchy after `$counter > $skip + $totalCount`, even though no furhter ES records will be added to the database

I think both problems will waste a lot of time and resources when indexing a large dataset.  

The second problem is trivial to solve, the script could just stop when `$counter > $skip + $totalCount`.  

The first problem is a lot harder to solve with the current method of recursively adding children to the index.  I'm think we *could* craft a SQL query to skip to the right place in the hierarchy by using the nested set, but (a) we'd need to change how the script traverses the hierarchy to allow indexing from an arbitrary information object, and (b) our long term goal is to remove nested set from AtoM.

*But*, with MySQL8 and @jraddaoui's CTE ancestor queries it's now a lot faster and more efficient to get an information object's ancestors than it used to be in AtoM.  I think with the CTE queries it's no longer necessary to use recursive queries to cache ancestor data and minimize ancestor queries.  And if we don't need to do recursive SQL queries it's a lot easier to page SQL data with the standard MySQL offset and limit functionality.

For instance, I think the following simple query to get for records 1001 to 2000 could be used in place of the current recursive child queries:

`SELECT * FROM information_object WHERE id <> 1 ORDER BY parent_id, id LIMIT 1000, 1000;`

By ordering primarily on parent_id we can get the ancestors of the first sibling then cache the results for the subsequent siblings (i.e. until the parent_id changes).  Doing the secondary sort on id just gives us a unique value to ensure the information_objects data always sorts in the same order (assuming no data is added of course).

Radda's changes in PR #997 allow doing a CTE query for an information object's ancestors with a simple `$informationObject->ancestors` call.

I know the change I'm proposing requires more work than your current solution, but I think it has two key benefits:

1. It avoids loading unnecessary data from the database for the sole purpose of figuring out where to resume ES indexing
2. It moves us in the direction of getting rid of nested set in AtoM

Being able to specify $skip and $limit parameters during search indexing also opens up the possibility of running multiple `search:populate` tasks (with different offsets) in parallel to speed up indexing of large data sets.

mcantelon:
@djjuhasz Sounds like we can make this PR a lot simpler given the CTE stuff. I'm going to close this PR for the time being until project work here dies down and I can re-visit.

mcantelon:
Reopening the PR as WiP.

mcantelon:
Hi @djjuhasz ... I've updated the PR. Good thinking on the simplified query.

jraddaoui:
This is nice @mcantelon (and @djjuhasz), I got to a similar solution while experimenting with PHP threads and, while this pagination allows to run the process without hitting memory limits and could be used to run the process in parallel (which may improve performance too in a powerful enough env.), we should test the performance of this changes indexing a big database. Even if the ancestor queries are improved, this may introduce a lot of them and other ancestor look-ups like the creators and repository.

mcantelon:
Thanks @jraddaoui... I'll see if I can cherry pick this to a client repo (with a few million records) to do some quick testing.

jraddaoui:
@mcantelon, I think @sbreker is already on it ;)

mcantelon:
Closing as I've reworked this in another PR.

PR #1247, opened 2021-01-26
Re-organize Dockerfile layers, refs #13441

jraddaoui:
- Install OS and composer dependencies before copying the source code to
  avoid the invalidation of the those cached layers when the container
  is built after any code change.
- Add fcgi package to allow a Docker healthcheck to be defined.
- Fix indentation.


sbreker (approved):
Looks good @jraddaoui !

jraddaoui:
Thanks @sbreker!

PR #1248, opened 2021-01-27
Fix ES' donors mapping, refs #13466

jraddaoui:
Fix i18n fields mapping generation and make it strict.


sbreker (approved):
Looks good @jraddaoui!  Thanks!

jraddaoui:
Thanks @sbreker!

PR #1250, opened 2021-01-27
Fix composer install on Docker, refs #13441

jraddaoui:
- Make sure an existing vendor/composer folder is not added
  to the container.
- Do not copy entire composer folder within vendor/composer.
- Remove build directory after moving composer contents.


sbreker (approved):
👍 Looks good @jraddaoui !

PR #1252, opened 2021-02-08
Maintain siblings order on CSV import, refs #13414

jraddaoui:
Maintain siblings order when the nested set update is disabled.
Add a placeholder lft value incrementing the max lft value from
the resource's siblings. This placeholder will overlap with the
parent's rgt value until the nested set is rebuilt.

### DIFF
--- a/lib/model/QubitInformationObject.php
+++ b/lib/model/QubitInformationObject.php
@@ -312,6 +312,50 @@ class QubitInformationObject extends BaseInformationObject
       $status->save($connection);
     }
 
+    // Maintain siblings order when the nested set update is disabled.
+    // Add a placeholder lft value incrementing the max lft value from
+    // the resource's siblings. This placeholder will overlap with the
+    // parent's rgt value until the nested set is rebuilt.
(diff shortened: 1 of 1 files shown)


lib/model/QubitInformationObject.php
```diff
@@ -225,6 +225,50 @@ public function save($connection = null)
...
+          
+        $connection->exec($sql);
+
+        $connection->commit();
+      }
+      catch (Exception $e)
+      {
+        $connection->rollBack();
+
+        throw $e;
```
sbreker: What state will io be in if this rollback and exception are thrown? parent::save() is called before, but all child records etc are called after. Will the io be saved, but all child recs not be created?

sbreker (commented):
Added one question regarding save() behaviour if the exception is thrown

lib/model/QubitInformationObject.php
jraddaoui: @sbreker, that's a great point, thanks!

I'll move down in the method, just before indexing.

sbreker (approved):
Thanks for making that change @jraddaoui !  Looks good

PR #1254, opened 2021-02-12
Fix ES' query string queries, refs #12186

jraddaoui:
Splits multi-field string queries into boolean queries with a string query
clause per field. This allows ES to use the field's analyzer on each clause
instead of the default analyzer and therefore it generates better results,
specially working with stop-words in a multi-culture index.

Additionally:

- Simplify retrieval of all fields by not processing the enabled
  languages until the query generation.
- Modify and simplify boosting process, now required as part of the
  boolean query clauses and not appended to the fields.
- Remove unused parse function from arElasticSearchPlugin class.

More info in https://projects.artefactual.com/issues/12186.

### DIFF
--- a/apps/qubit/modules/default/actions/moveAction.class.php
+++ b/apps/qubit/modules/default/actions/moveAction.class.php
@@ -139,13 +139,13 @@ class DefaultMoveAction extends sfAction
 
     if (isset($request->query))
     {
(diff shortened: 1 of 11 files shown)


sbreker (approved):
Looks good @jraddaoui - great to have this change!

PR #1256, opened 2021-02-18
Add parent ID checking to CSV check, refs #13474

mcantelon:
Added, to the csv:check CLI task, logic to check parent IDs to make sure they correspond to legacy IDs in the CSV file(s) being checked.

### DIFF
--- a/lib/task/import/csvCheckImportTask.class.php
+++ b/lib/task/import/csvCheckImportTask.class.php
@@ -131,14 +131,15 @@ EOF;
             }
             else
             {
-              // Note that the legacy ID is a duplicate
+              // Note that the legacy ID is used more than once
               $self->status['duplicateLegacyIds'][$self->columnValue('legacyId')] = true;
             }
           }
 
-          // Count parent ID uses
+          // Count uses of parent IDs and note premature use of an ID
           if ($self->columnExists('parentId') && !empty($self->columnValue('parentId')))
           {
+            // Count parent ID uses
             $parentId = $self->columnValue('parentId');
 
             if (empty($self->status['parentIds'][$parentId]))
--- a/lib/task/import/csvCheckImportTask.class.php
+++ b/lib/task/import/csvCheckImportTask.class.php
@@ -131,7 +131,7 @@ EOF;
             }
             else
             {
-              // Note that the legacy ID is used more than once
+              // Note that the legacy ID has been used more than once
               $self->status['duplicateLegacyIds'][$self->columnValue('legacyId')] = true;
             }
           }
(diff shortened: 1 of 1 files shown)


mcantelon:
(This'll check across multiple CSV files too... if you do `php symfony csv:check a.csv,b.csv` and a parent row is in `a.csv`, or vice versa, and the child is in `b.csv` it won't deem the parent to be missing.)

lib/task/import/csvCheckImportTask.class.php
```diff
@@ -189,6 +215,43 @@ public function execute($arguments = array(), $options = array())
...
+      print sprintf("Number of unique missing parent IDs: %d\n", count($missingParentIds));
+
+      // Display details about each missing parent and total number of rows affected
+      if (count($missingParentIds))
+      {
+        $orphans = 0;
+
+        foreach ($missingParentIds as $parentId => $numberOfChildren)
+        {
+          print sprintf("* %d (%d children)\n", $parentId, $numberOfChildren);
```
djjuhasz: The output from this is:

```
* 10000 (1 children)
```

I think it would be helpful to indicate the first number is the parentId value, e.g.

```
* parentId: 10000 (1 children)
```

lib/task/import/csvCheckImportTask.class.php
```diff
@@ -189,6 +215,43 @@ public function execute($arguments = array(), $options = array())
       }
     }
 
+    // Check that parent IDs actually exist if legacy and parent IDs were found
+    if (count($legacyIds) && count($parentIds))
```
djjuhasz: I think the checker should report **all** `parentIds` as missing if `0 == count(legacyIds)`.

djjuhasz (changes requested):
@mcantelon when using the `csv:import --source` option the import script will also look up parentIds in the `keymap.source_id` table if the --source value is the same as the `keymap.source_name` value.   In this case it's valid to use a `parentId` value that is not a `legacyId` in the current CSV file.  It doesn't look like `csv:check-import` supports a `--source` option right now, so it's not possible to check parentIds against previous imports, but it would be a good feature to consider in the future.

I requested a couple of small changes to the code, but other than that it looks good to me. :)

mcantelon:
Thanks for the review! I've updated the PR with these changes... good idea re: a future `--source` option.

Here's what I made the output if no legacy IDs are present yet there are parent IDs.... let me know if this seems clear enough.

    Missing parent IDs:
    -------------------
    
    No legacyId values were found.

djjuhasz (commented):
@mcantelon thanks, those changes look good!  :)

I just realized though that the order of the legacyId and parentId values is important - if a parentId comes before the legacyId in the import, the parent/child relationship won't be created.  It would be nice if the check-import tool flagged this problem too.

mcantelon:
I've updated the PR to report on prematurely used parents and also to report on any legacy IDs that have been used more than once.

mcantelon:
Steve's been working on the same thing, as part of a larger effort, so I'll just keep this PR open for now as a reference.

PR #1258, opened 2021-02-19
Avoid empty language setting, refs #13465

jraddaoui:
Make `languageCode` required but keep the allow empty option for
display. Restructure validation to show the enabled languages on error.


sbreker (approved):
Nice work! Thanks @jraddaoui!

PR #1260, opened 2021-02-26
Restore access info to inventory list, refs #13480

djjuhasz:
Add the access restriction information back to the PDF Finding Aid
"inventory summary" table in the file and item level table row "body".

- Move non DID elements out of "ead:did" template
- Add ead:accessrestrict to "body" of file/item table row


jraddaoui (approved):
:ok_hand:

PR #1249, opened 2021-01-27
Setup php-cs-fixer, refs #13317

jraddaoui:
Format PHP code using PHP CS Fixer and the @PhpCsFixer ruleset with
one exception to avoid inconsitencies in templates (check `.php_cs`).
Setup Github Actions workflow to run syntax checks in the CI process.

PR #1261, opened 2021-02-27
Add video/audio caption/subtitle/chapter file support, refs #13490

mInnes-archives:
### Administrative

~~PR needs an issue id.~~ -> 13490

### Description

PR adds the ability to add caption/subtitle/chapter files in SRT or VTT format for use with video reference files. This takes advantage of HTML5 `<track>` tags in `<video>` elements: 590a164, e87a47e. As many caption/subtitle files may be uploaded and included as needed b1c7f45, and a single chapter file. Only files with extensions .srt or .vtt are permitted. This restriction is handled by a new validator that works off of the `$qubitMimeTypes` array in `QubitDigitalObject.php`: ff6a95b, ce3649e. File uploading and deletion uses the existing 'Edit digital object' module: 081418f, e8e384b. These sections will only show when `media_type_id` corresponds to video.

This feature is largely achieved by adding two additional usage types [`usage_id`] for use with the `digital_object` table, one for captions/subtitles and one for chapters: 5bc3001. These new types are thus treated very similarly to other representation types. The only change in the database structure is the addition of a `language` column in the `digital_object` table: 787b0d7. Data type for this field has been set to varchar(50), despite currently only storing ISO-639 codes. The extra room is added to facilitate custom language values, should this be needed in the future.

I18n has been implemented for this feature. Symfony handles the translation of language names in all situations. I've added the new usage terms to the taxonomy for the section labeling and added French translations so you can see it working: 2c84abb.

Two minor digital object related changes are also included:
- ~~Restrict the maximum width of the video frame while stretching. 2ba823e~~ [removed]
  - ~~This was briefly discussed in PR #1238, but I threw it in at the end so it didn't get implemented.~~
  - ~~This is intended to make the size of the video frame in the 'Edit digital object' module less overbearing.~~
- Typos and mislabeling in comments. 8ac160a

### Limitations

- Can upload as many captions/subs as needed, but only 1 at a time
- Can't change language once uploaded
- Can't specify a custom language
  - Limited to symfony list, but the vast majority of ISO 639 languages are there.
  - Two captions/subs of the same language can't be labeled differently, which may cause some confusion between CC and standard subs in the selection menu [eg. English, English CC]. I have some ideas about how to add a little '[CC]' label when it's needed without requiring a whole section dedicated to captions, but I think this PR covers enough ground for now.
- Subtitle style hasn't been changed.
  - The background transparency of for subtitles may not be appropriate for accessibility reasons.
  - The text position in fullscreen seems quite high. This might be a personal preference, though.

### Review

Feel free to ask me any questions about any of the code here.

### Images

subs_sample1
*Caption/subtitle display on record.*

subs_sample2
*Restricted video width in 'Edit digital object' module.  2ba823e*

subs_sample3
*Caption/subtitle/chapter upload area in 'Edit digital object' module.*

### DIFF
--- a/apps/qubit/modules/digitalobject/actions/editAction.class.php
+++ b/apps/qubit/modules/digitalobject/actions/editAction.class.php
@@ -94,19 +94,19 @@ class DigitalObjectEditAction extends sfAction
     {
       if (null === $representation)
       {
-        $repName = "repFile_$usageId";
-        $derName = "generateDerivative_$usageId";
+        $repName = "repFile_{$usageId}";
+        $derName = "generateDerivative_{$usageId}";
 
         $this->form->setValidator($repName, new sfValidatorFile);
         $this->form->setWidget($repName, new sfWidgetFormInputFile);
 
         if (-1 < $maxUploadSize)
         {
-          $this->form->getWidgetSchema()->$repName->setHelp($this->context->i18n->__('Max. size ~%1%', array('%1%' => hr_filesize($maxUploadSize))));
+          $this->form->getWidgetSchema()->{$repName}->setHelp($this->context->i18n->__('Max. size ~%1%', array('%1%' => hr_filesize($maxUploadSize))));
         }
         else
         {
-          $this->form->getWidgetSchema()->$repName->setHelp('');
+          $this->form->getWidgetSchema()->{$repName}->setHelp('');
         }
 
         // Add "auto-generate" checkbox
@@ -114,6 +114,58 @@ class DigitalObjectEditAction extends sfAction
         $this->form->setWidget($derName, new sfWidgetFormInputCheckbox(array(), array('value' => 1)));
       }
     }
+
+    // If video track doesn't exist, include upload widget
+    // But always include subtitle upload widget
+    foreach ($this->videoTracks as $usageId => $videoTrack)
+    {
+      if ($usageId != QubitTerm::SUBTITLES_ID)
+      {
+        if (null === $videoTrack)
+        {
+          $trackName = "trackFile_{$usageId}";
+  
+          $this->form->setValidator($trackName, new sfValidatorAnd(array(
+            new QubitValidatorMimeType(array('mime_types' => array('text/vtt','application/x-subrip'))),
+            new sfValidatorFile()
+          )));
+          $this->form->setWidget($trackName, new sfWidgetFormInputFile);
+    
+          if (-1 < $maxUploadSize)
+          {
+            $this->form->getWidgetSchema()->{$trackName}->setHelp($this->context->i18n->__('Max. size ~%1%', array('%1%' => hr_filesize($maxUploadSize))));
+          }
+          else
+          {
+            $this->form->getWidgetSchema()->{$trackName}->setHelp('');
+          }
+        }
+      }
+      else
+      {
+        $trackName = "trackFile_$usageId";
+        $langName = "lang_$usageId";
+
+        $this->form->setValidator($trackName, new sfValidatorAnd([
+          new QubitValidatorMimeType(['mime_types' => ['text/vtt','application/x-subrip']]),
+          new sfValidatorFile()
+        ]));
+        $this->form->setWidget($trackName, new sfWidgetFormInputFile);
+
+        $this->form->setValidator($langName, new sfValidatorI18nChoiceLanguage);
+        $this->form->setWidget($langName, new sfWidgetFormI18nChoiceLanguage);
+
+        if (-1 < $maxUploadSize)
+        {
+          $this->form->getWidgetSchema()->$trackName->setHelp($this->context->i18n->__('Max. size ~%1%', array('%1%' => hr_filesize($maxUploadSize))));
+        }
+        else
+        {
+          $this->form->getWidgetSchema()->$trackName->setHelp('');
+        }
+      }
+    }
+  
 
     // Add latitude and longitude fields
     foreach (array('latitude', 'longitude') as $geoPropertyField)
@@ -162,6 +214,11 @@ class DigitalObjectEditAction extends sfAction
       QubitTerm::REFERENCE_ID => $this->resource->getChildByUsageId(QubitTerm::REFERENCE_ID),
       QubitTerm::THUMBNAIL_ID => $this->resource->getChildByUsageId(QubitTerm::THUMBNAIL_ID));
 
+    // Get video track files
+    $this->videoTracks = array(
+      QubitTerm::CHAPTERS_ID => $this->resource->getChildByUsageId(QubitTerm::CHAPTERS_ID),
+      QubitTerm::SUBTITLES_ID => $this->resource->getChildByUsageId(QubitTerm::SUBTITLES_ID));
+
     $this->addFormFields();
 
     // Process forms
@@ -203,9 +260,9 @@ class DigitalObjectEditAction extends sfAction
     $uploadedFiles = array();
     foreach ($this->representations as $usageId => $representation)
     {
-      if (null !== $uf = $this->form->getValue("repFile_$usageId"))
+      if (null !== $uploadedFile = $this->form->getValue("repFile_$usageId"))
       {
-        $uploadedFiles[$usageId] = $uf;
+        $uploadedFiles[$usageId] = $uploadedFile;
       }
     }
 
@@ -239,6 +296,30 @@ class DigitalObjectEditAction extends sfAction
       $representation->createDerivatives = false;
 
       $representation->save();
+    }
+
+    // Upload new video track files
+    foreach ($this->videoTracks as $usageId => $videoTrack)
+    {
+      if (null !== $uploadedTrack = $this->form->getValue("trackFile_$usageId"))
+      {
+        $lang = $this->form->getValue("lang_$usageId");
+        $uploadedTracks[$usageId] = ["track" => $uploadedTrack, "language" => $lang];
+      }
+    }
+
+    foreach ($uploadedTracks as $usageId => $uploadTrack)
+    {
+      $content = file_get_contents($uploadTrack["track"]->getTempName());
+
+      $track = new QubitDigitalObject;
+      $track->usageId = $usageId;
+      $track->assets[] = new QubitAsset($uploadTrack["track"]->getOriginalName(), $content);
+      $track->parentId = $this->resource->id;
+      $track->createDerivatives = false;
+      $track->language = $uploadTrack["language"];
+
+      $track->save();
     }
 
     // Generate new reference
--- a/apps/qubit/modules/digitalobject/templates/_editRepresentation.php
+++ b/apps/qubit/modules/digitalobject/templates/_editRepresentation.php
@@ -1,13 +1,18 @@
 <div>
 
-  <div style="float: right;">
+  <div class="digital-object-preview">
 
-    <?php echo get_component('digitalobject', 'show', array(
-      'iconOnly' => true,
-      'link' => public_path($representation->getFullPath()),
-      'resource' => $representation,
-      'usageType' => QubitTerm::THUMBNAIL_ID)) ?>
+    <?php if ($representation->usageId == QubitTerm::CHAPTERS_ID): ?>
+      <a href="<?php echo $representation->getFullPath() ?>"><?php echo __('View file') ?></a>
 
+    <?php else: ?>
+      <?php echo get_component('digitalobject', 'show', array(
+        'iconOnly' => true,
+        'link' => public_path($representation->getFullPath()),
+        'resource' => $representation,
+        'usageType' => QubitTerm::THUMBNAIL_ID)) ?>
+
+    <?php endif; ?>
   </div>
 
   <div>
--- a/apps/qubit/modules/digitalobject/templates/_editSubtitles.php
+++ b/apps/qubit/modules/digitalobject/templates/_editSubtitles.php
@@ -1,0 +1,42 @@
+<fieldset class="collapsible">
+<legend><?php echo __('%1%', array('%1%' => QubitTerm::getById($usageId))) ?></legend> 
+
+<?php foreach ($subtitles as $subtitle): ?>
+  <table class="table table-bordered">
+    <thead>
+      <tr><th><?php echo format_language($subtitle->language) ?></th></tr>
+    </thead>
+    <tbody>
+      <tr>
+        <td>
+        
+          <div class="digital-object-preview">
+            <a href="<?php echo $subtitle->getFullPath() ?>"><?php echo __('View file') ?></a>
+          </div>
+        
+          <div>
+            <?php echo render_show(__('Filename'), render_value($subtitle->name)) ?>
+            <?php echo render_show(__('Filesize'), hr_filesize($subtitle->byteSize)) ?>
+            <?php echo link_to(__('Delete'), array($subtitle, 'module' => 'digitalobject', 'action' => 'delete'), array('class' => 'delete')) ?>
+          </div>
+      
+        </td>
+      </tr>
+    </tbody>
+  </table>
+<?php endforeach ?>
+
+<table class="table table-bordered">
+  <tr>
+    <td>
+      <?php echo $form["trackFile_$usageId"]
+        ->label(__('Select a file to upload (.vtt|.srt)'))
+        ->renderRow() ?>
+              
+      <?php echo $form["lang_$usageId"]
+        ->label(__('Language'))
+        ->renderRow() ?>
+    </td>
+  </tr>
+</table>
+</fieldset>
--- a/apps/qubit/modules/digitalobject/templates/_show.php
+++ b/apps/qubit/modules/digitalobject/templates/_show.php
@@ -1,9 +1,20 @@
-<div class="digital-object-reference">
+<?php if ($usageType == QubitTerm::CHAPTERS_ID || $usageType == QubitTerm::SUBTITLES_ID): ?>
   <?php if (!empty($accessWarning)): ?>
-    <div class="access-warning">
-      <?php echo $accessWarning ?>
-    </div>
-  <?php else: ?>
-    <?php echo get_component('digitalobject', $showComponent, array('iconOnly' => $iconOnly, 'link' => $link, 'resource' => $resource, 'usageType' => $usageType)) ?>
+      <div class="access-warning">
+        <?php echo $accessWarning ?>
+      </div>
+    <?php else: ?>
+      <?php echo get_component('digitalobject', $showComponent, array('iconOnly' => $iconOnly, 'link' => $link, 'resource' => $resource, 'usageType' => $usageType)) ?>
   <?php endif; ?>
-</div>
+    
+<?php else: ?>
+  <div class="digital-object-reference">
+    <?php if (!empty($accessWarning)): ?>
+      <div class="access-warning">
+        <?php echo $accessWarning ?>
+      </div>
+    <?php else: ?>
+      <?php echo get_component('digitalobject', $showComponent, array('iconOnly' => $iconOnly, 'link' => $link, 'resource' => $resource, 'usageType' => $usageType)) ?>
+    <?php endif; ?>
+  </div>
+<?php endif; ?>
--- a/apps/qubit/modules/digitalobject/templates/_showVideo.php
+++ b/apps/qubit/modules/digitalobject/templates/_showVideo.php
@@ -8,10 +8,27 @@
     <?php echo link_to(image_tag($representation->getFullPath(), array('alt' => __($resource->getDigitalObjectAltText() ?: 'Open original %1%', array('%1%' => sfConfig::get('app_ui_label_digitalobject'))))), $link) ?>
   <?php endif; ?>
 
+<?php elseif ($usageType == QubitTerm::CHAPTERS_ID): ?>
+
+  <?php if ($showMediaPlayer): ?>
+    <track kind="chapters" src="<?php echo public_path($representation->getFullPath()) ?>" srclang="">
+  <?php endif; ?>
+
+<?php elseif ($usageType == QubitTerm::SUBTITLES_ID): ?>
+
+  <?php if ($showMediaPlayer): ?>
+    <?php foreach ($representation as $subtitle): ?>
+      <track kind="subtitles" src="<?php echo public_path($subtitle->getFullPath()) ?>" srclang="<?php echo $subtitle->language ?>" label="<?php echo format_language($subtitle->language) ?>">
+    <?php endforeach; ?>
+  <?php endif; ?>
+
 <?php elseif ($usageType == QubitTerm::REFERENCE_ID): ?>
 
   <?php if ($showMediaPlayer): ?>
-    <video preload="metadata" class="mediaelement-player" src="<?php echo public_path($representation->getFullPath()) ?>"></video>
+    <video preload="metadata" class="mediaelement-player" src="<?php echo public_path($representation->getFullPath()) ?>">
+      <?php echo get_component('digitalobject', 'show', array('resource' => $resource, 'usageType' => QubitTerm::CHAPTERS_ID)) ?>
+      <?php echo get_component('digitalobject', 'show', array('resource' => $resource, 'usageType' => QubitTerm::SUBTITLES_ID)) ?>
+    </video>
   <?php else: ?>
     <div style="text-align: center">
       <?php echo image_tag($representation->getFullPath(), array('style' => 'border: #999 1px solid', 'alt' => __($resource->getDigitalObjectAltText() ?: 'Original %1% not accessible', array('%1%' => sfConfig::get('app_ui_label_digitalobject'))))) ?>
--- a/apps/qubit/modules/digitalobject/templates/editSuccess.php
+++ b/apps/qubit/modules/digitalobject/templates/editSuccess.php
@@ -80,6 +80,36 @@
 
       <?php endforeach; ?>
 
+      <?php if ($resource->mediaTypeId == QubitTerm::VIDEO_ID || $resource->mediaTypeId == QubitTerm::AUDIO_ID): ?>
+        
+        <?php foreach ($videoTracks as $usageId => $videoTrack): ?>
+
+          <?php if ($resource->mediaTypeId == QubitTerm::VIDEO_ID && $usageId == QubitTerm::SUBTITLES_ID): ?>
+            
+            <?php echo include_partial('editSubtitles', array('resource' => $resource, 'subtitles' => $videoTrack, 'form' => $form, 'usageId' => $usageId)) ?>
+
+          <?php elseif ($usageId != QubitTerm::SUBTITLES_ID): ?>                 
+          
+            <fieldset class="collapsible">
+
+              <legend><?php echo __('%1%', array('%1%' => QubitTerm::getById($usageId))) ?></legend>
+
+              <?php if (isset($videoTrack)): ?>
+              
+                <?php echo get_component('digitalobject', 'editRepresentation', array('resource' => $resource, 'representation' => $videoTrack)) ?>
+
+              <?php else: ?>
+
+                <?php echo $form["trackFile_$usageId"]
+                  ->label(__('Select a file to upload (.vtt|.srt)'))
+                  ->renderRow() ?>
+            
+              <?php endif; ?>
+            </fieldset> 
+          <?php endif; ?>
+        <?php endforeach; ?>
+      <?php endif; ?>
+
     </section>
 
     <section class="actions">
--- a/config/schema.yml
+++ b/config/schema.yml
@@ -98,6 +98,7 @@ propel:
     id: { type: integer, required: true, primaryKey: true, foreignTable: object, foreignReference: id, onDelete: cascade, inheritanceKey: true }
     object_id: { type: integer, foreignTable: object, foreignReference: id, onDelete: cascade }
     usage_id: { type: integer, foreignTable: term, foreignReference: id, onDelete: setnull }
+    language: varchar(50)
     mime_type: varchar(255)
     media_type_id: { type: integer, foreignTable: term, foreignReference: id, onDelete: setnull }
     name: { type: varchar(1024), required: true }
--- a/data/fixtures/taxonomyTerms.yml
+++ b/data/fixtures/taxonomyTerms.yml
@@ -2678,6 +2678,22 @@ QubitTerm:
       hr: 'Vanjska datoteka'
       pt: 'Ficheiro externo'
       pt_BR: 'Arquivo externo'
+  QubitTerm_chapters_id:
+    taxonomy_id: QubitTaxonomy_17
+    parent_id: QubitTerm_110
+    id: 195
+    source_culture: en
+    name:
+      en: 'Chapters'
+      fr: 'Chapitres'
+  QubitTerm_subtitles_id:
+    taxonomy_id: QubitTaxonomy_17
+    parent_id: QubitTerm_110
+    id: 196
+    source_culture: en
+    name:
+      en: 'Captions/Subtitles'
+      fr: 'Légendes/Sous-titres'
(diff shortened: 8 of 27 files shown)


fiver-watson:
Hi @mInnes-archives  - thank you so much for submitting this pull request! As a bigger PR, I suspect it may take our team a bit longer to find the time to fully review it, and there will likely be a few change requests before we can merge it into a public release, just to ensure that it matches our coding standards and development patterns, and our team feels confident we can maintain it going forward. This is a great feature for accessibility and I would love to see it incorporated into our next major release, so I hope you'll be a bit patient as we catch up, and also willing to do a bit more work on it to incorporate some of the feedback once we've reviewed! 

Thanks again for sharing your work back with the public project! More soon...

config/schema.yml
```diff
@@ -98,6 +98,7 @@ propel:
     id: { type: integer, required: true, primaryKey: true, foreignTable: object, foreignReference: id, onDelete: cascade, inheritanceKey: true }
     object_id: { type: integer, foreignTable: object, foreignReference: id, onDelete: cascade }
     usage_id: { type: integer, foreignTable: term, foreignReference: id, onDelete: setnull }
+    language: varchar(50)
```
jraddaoui, line 101: This change requires some important extra steps:

- Build SQL:

For new installations, this column needs to be added to the initial SQL, use the following task to do so:

```
php symfony propel:build-sql
```

- Build models:

Needed to reflect the new column in the base model classes, similar task:

```
php symfony propel:build-model
```

- Add a migration:

For old installations, we need a migration to add this column to the existing database. Create a subsequent migration class in the migrations folder, this is a good example to add a new column ...

https://github.com/artefactual/atom/blob/qa/2.x/lib/task/migrate/migrations/arMigration0130.class.php

Use the next version number for the file/class name and the constant, add a proper comment on top (unlike this one) and, use the third parameter of `QubitMigrate::addColumn` to indicate the new column position. Something like:

```php
QubitMigrate::addColumn(
  QubitDigitalObject::TABLE_NAME,
  'language VARCHAR(50)',
  ['after' => 'usage_id']
);
```

To test this migration, use the upgrade task:

```
php symfony tools:upgrade-sql
```

data/fixtures/taxonomyTerms.yml
```diff
@@ -2678,6 +2678,22 @@ QubitTerm:
...
+      en: 'Chapters'
+      fr: 'Chapitres'
+  QubitTerm_subtitles_id:
+    taxonomy_id: QubitTaxonomy_17
+    parent_id: QubitTerm_110
+    id: 196
+    source_culture: en
+    name:
+      en: 'Captions/Subtitles'
+      fr: 'Légendes/Sous-titres'
```
jraddaoui, line 2696: Similar to the column, these terms are now a requirement in all installations. However, these fixtures are only used in the database initialization and not in the upgrade process. Also, in existing instances, these ids may be already in use so we need to bump the terms using them to free them for the new ones. Here an example of how we usually do that:

https://github.com/artefactual/atom/blob/qa/2.x/lib/task/migrate/migrations/arMigration0155.class.php#L47-L54

Use the same migration where you added the new column to add both terms.

lib/model/QubitDigitalObject.php
```diff
@@ -1768,7 +1769,30 @@ public function getChildByUsageId($usageId)
...
+        {
+          $result[] = $subtitle;
+          ++$i;
+        }
+        else
+        {
+          $fileExists = false;
+        }
+      }
+    }
```
jraddaoui: `QubitDigitalObject::get($criteria)` instead of `getOne` should be enough to get all the related subtitles. I'm not sure if "no results" is returned as `null`, empty `array` or `false`, but you could check that instead of using the while loop.

lib/model/QubitTerm.php
```diff
@@ -179,12 +179,16 @@ class QubitTerm extends BaseTerm
 
     // Accession alternative identifier taxonomy
     ACCESSION_ALTERNATIVE_IDENTIFIER_DEFAULT_TYPE_ID = 192,
-
+
```
jraddaoui: Nitpick: unexpected whitespace.

lib/model/QubitTerm.php
```diff
@@ -179,12 +179,16 @@ class QubitTerm extends BaseTerm
...
-
+    
     // Accession event type: physical transfer
     ACCESSION_EVENT_PHYSICAL_TRANSFER_ID = 193,
 
     // Accession event note
-    ACCESSION_EVENT_NOTE_ID = 194;
+    ACCESSION_EVENT_NOTE_ID = 194,
+    
+    // Video track file taxonomy
```
jraddaoui: Nitpick: they are part of the "Digital Object Usages" taxonomy.

lib/validator/QubitValidatorMimeType.class.php
```diff
@@ -0,0 +1,62 @@
...
+
+  protected function doClean($value)
+  {        
+    $mimeType = QubitDigitalObject::deriveMimeType($value['name']);
+    $mimeTypes = $this->getOption('mime_types');
+    
+    if (!in_array($mimeType, $mimeTypes))
+      {
+        throw new sfValidatorError($this, 'mime_types', array('value' => $value['name'], 'mime_type' => $mimeType));
+      }
```
jraddaoui: Nitpick: broken indentation.

lib/validator/QubitValidatorMimeType.class.php
```diff
@@ -0,0 +1,62 @@
...
+    return $value;
```
jraddaoui, line 48: Nitpick: blank line before `return`.

lib/validator/QubitValidatorMimeType.class.php
```diff
@@ -0,0 +1,62 @@
...
+  
+  /**
+   * @see sfValidatorBase
+   */
+  protected function isEmpty($value)
+  {
+    // empty if the value is not an array
+    // or if the value comes from PHP with an error of UPLOAD_ERR_NO_FILE
+    return
+      (!is_array($value))
```
jraddaoui: Nitpick: parenthesis not needed.

lib/validator/QubitValidatorMimeType.class.php
```diff
@@ -0,0 +1,62 @@
...
+        ||
+      (is_array($value) && isset($value['error']) && UPLOAD_ERR_NO_FILE === $value['error']);
```
jraddaoui: Nitpick: Should be in one line, without extra indentation ... `|| (is_array...`.

lib/validator/QubitValidatorMimeType.class.php
```diff
@@ -0,0 +1,62 @@
...
+    return $value;
+  }
+  
+  /**
+   * @see sfValidatorBase
+   */
+  protected function isEmpty($value)
+  {
+    // empty if the value is not an array
+    // or if the value comes from PHP with an error of UPLOAD_ERR_NO_FILE
```
jraddaoui: Nitpick: uppercase to start comment, dot to finish it if it's multiline.

lib/model/QubitTerm.php
```diff
@@ -245,6 +250,7 @@ public static function isProtected($id)
       QubitTerm::SOURCE_NOTE_ID,
       QubitTerm::STANDARDIZED_FORM_OF_NAME_ID,
       QubitTerm::STATUS_TYPE_PUBLICATION_ID,
+      QubitTerm::SUBTITLES_ID,
```
jraddaoui, line 253: Nice.

lib/model/QubitDigitalObject.php
```diff
@@ -2772,7 +2796,7 @@ public static function isAudioFile($filename)
 
   /*
    * -----------------------------------------------------------------------
-   * VIDEO
+   * Audio
```
jraddaoui, line 2783: Thanks!

apps/qubit/modules/digitalobject/templates/_showVideo.php
```diff
@@ -8,10 +8,29 @@
...
+    <?php foreach ($representation as $subtitle): ?>
+      <track kind="subtitles" src="<?php echo public_path($subtitle->getFullPath()) ?>" srclang="<?php echo $subtitle->language ?>" label="<?php echo format_language($subtitle->language) ?>">
+    <?php endforeach; ?>
+  <?php endif; ?>
+
 <?php elseif ($usageType == QubitTerm::REFERENCE_ID): ?>
 
   <?php if ($showMediaPlayer): ?>
-    <video preload="metadata" class="mediaelement-player" src="<?php echo public_path($representation->getFullPath()) ?>"></video>
+    <div style="max-width:640px;margin:auto;">
```
jraddaoui: I'd prefer to not limit its width but, if this is really needed, can we move it to the theme styles?

https://github.com/artefactual/atom/blob/qa/2.x/plugins/arDominionPlugin/css/less/scaffolding.less#L1241-L1243

apps/qubit/modules/digitalobject/templates/_showVideo.php
```diff
@@ -8,10 +8,29 @@
     <?php echo link_to(image_tag($representation->getFullPath(), array('alt' => __($resource->getDigitalObjectAltText() ?: 'Open original %1%', array('%1%' => sfConfig::get('app_ui_label_digitalobject'))))), $link) ?>
   <?php endif; ?>
 
+<?php elseif ($usageType == QubitTerm::CHAPTERS_ID): ?>
+
+  <?php if ($showMediaPlayer): ?>
+    <track kind="chapters" src="<?php echo public_path($representation->getFullPath()) ?>" srclang="">
```
jraddaoui, line 14: Is the empty `srclang` attribute needed?

apps/qubit/modules/digitalobject/templates/_show.php
```diff
@@ -1,9 +1,20 @@
...
+  <div class="digital-object-reference">
+    <?php if (!empty($accessWarning)): ?>
+      <div class="access-warning">
+        <?php echo $accessWarning ?>
+      </div>
+    <?php else: ?>
+      <?php echo get_component('digitalobject', $showComponent, array('iconOnly' => $iconOnly, 'link' => $link, 'resource' => $resource, 'usageType' => $usageType)) ?>
+    <?php endif; ?>
+  </div>
+<?php endif; ?>
```
jraddaoui: Except the `div` and the `class` both blocks look the same. The `div` shouldn't affect the layout and the class could be determined with a ternary operator to avoid the duplication:

```php
<div <?php echo $usageType == QubitTerm::CHAPTERS_ID || $usageType == QubitTerm::SUBTITLES_ID ? '' : 'class="digital-object-reference"' ?>>
```

apps/qubit/modules/digitalobject/templates/_editSubtitles.php
```diff
@@ -0,0 +1,38 @@
...
+      <?php echo $form["trackFile_$usageId"]
+        ->label(__('Select a file to upload (.vtt|.srt)'))
+        ->renderRow() ?>
+              
+      <?php echo $form["lang_$usageId"]
+        ->label(__('Language'))
+        ->renderRow() ?>
+    </td>
+  </tr>
+</table>
```
jraddaoui: Nitpick: missing blank line at the end of file.

apps/qubit/modules/digitalobject/templates/_editSubtitles.php
```diff
@@ -0,0 +1,38 @@
...
+        <div>
+          <?php echo render_show(__('Filename'), render_value($subtitle->name)) ?>
+          <?php echo render_show(__('Filesize'), hr_filesize($subtitle->byteSize)) ?>
+          <?php echo link_to(__('Delete'), array($subtitle, 'module' => 'digitalobject', 'action' => 'delete'), array('class' => 'delete')) ?>
+        </div>
+      
+      </td>
+    </tr>
+  </tbody>
+  </table>
```
jraddaoui, line 26: Nitpick: table content missing indentation.

apps/qubit/modules/digitalobject/templates/_editSubtitles.php
```diff
@@ -0,0 +1,38 @@
+<?php foreach ($subtitles as $subtitle): ?>
+  <table class="table table-bordered">
+  <thead>
+    <tr><th><?php echo format_language($subtitle->language) ?></th></tr>
+  </thead>    
+  <tbody>
+    <tr>
+      <td>
+        
+        <div style="float: right;">
```
jraddaoui: I think it's better to use a class in this div and have separation of concerns, moving the style to the theme:

https://github.com/artefactual/atom/tree/qa/2.x/plugins/arDominionPlugin/css/less

apps/qubit/modules/digitalobject/templates/_editRepresentation.php
```diff
@@ -20,4 +25,4 @@
 
   </div>
 
-</div>
+</div>
```
jraddaoui: Nitpick: missing blank line at the end of file.

apps/qubit/modules/digitalobject/actions/editAction.class.php
```diff
@@ -115,6 +115,58 @@ protected function addFormFields()
...
 
+    // If video track doesn't exist, include upload widget
+    // But always include subtitle upload widget
+    foreach ($this->videoTracks as $usageId => $videoTrack)
+    {
+      if ($usageId != QubitTerm::SUBTITLES_ID)
+      {
+        if (null === $videoTrack)
+        {
+          $trackName = "trackFile_$usageId";
```
jraddaoui: Nitpick: use curly braces in string variables: `"trackFile_{$usageId}"`

apps/qubit/modules/digitalobject/actions/editAction.class.php
```diff
@@ -115,6 +115,58 @@ protected function addFormFields()
...
+  
+          $this->form->setValidator($trackName, new sfValidatorAnd(array(
```
jraddaoui, line 128: Nitpick: we are moving to array short syntax `[]`.

apps/qubit/modules/digitalobject/actions/editAction.class.php
```diff
@@ -115,6 +115,58 @@ protected function addFormFields()
...
+          )));
+          $this->form->setWidget($trackName, new sfWidgetFormInputFile);
+    
+          if (-1 < $maxUploadSize)
+          {
+            $this->form->getWidgetSchema()->$trackName->setHelp($this->context->i18n->__('Max. size ~%1%', array('%1%' => hr_filesize($maxUploadSize))));
+          }
+          else
+          {
+            $this->form->getWidgetSchema()->$trackName->setHelp('');
```
jraddaoui: Is this needed?

apps/qubit/modules/digitalobject/actions/editAction.class.php
```diff
@@ -115,6 +115,58 @@ protected function addFormFields()
...
+  
+          $this->form->setValidator($trackName, new sfValidatorAnd(array(
+            new QubitValidatorMimeType(array('mime_types' => array('text/vtt','application/x-subrip'))),
+            new sfValidatorFile()
+          )));
+          $this->form->setWidget($trackName, new sfWidgetFormInputFile);
+    
+          if (-1 < $maxUploadSize)
+          {
+            $this->form->getWidgetSchema()->$trackName->setHelp($this->context->i18n->__('Max. size ~%1%', array('%1%' => hr_filesize($maxUploadSize))));
```
jraddaoui: Nitpick: use curly braces in indirect variables: `->{$trackName}->`

apps/qubit/modules/digitalobject/actions/editAction.class.php
```diff
@@ -241,6 +298,30 @@ public function processForm()
       $representation->save();
     }
 
+    // Upload new video track files
+    foreach ($this->videoTracks as $usageId => $videoTrack)
+    {
+      if (null !== $uf = $this->form->getValue("trackFile_$usageId"))
```
jraddaoui: Nitpick: use a more specific variable name, `$uploadedFile` for example.

apps/qubit/modules/digitalobject/actions/editAction.class.php
```diff
@@ -241,6 +298,30 @@ public function processForm()
...
+      {
+        $lang = $this->form->getValue("lang_$usageId");
+        $uploadedTracks[$usageId] = array($uf, $lang);
+      }
+    }
+
+    foreach ($uploadedTracks as $usageId => $uploadTrack)
+    {
+      $content = file_get_contents($uploadTrack[0]->getTempName());
```
jraddaoui: Even if this files are small and shouldn't have a big memory footprint, `QubitAsset` allows to pass directly a path to avoid saving the file contents in memory. For example ...

https://github.com/artefactual/atom/blob/qa/2.x/lib/task/digitalobject/digitalObjectLoadTask.class.php#L371

I'm not completely sure this will work with the uploaded files in the tmp folder, but I think we should check if it does.

apps/qubit/modules/digitalobject/actions/editAction.class.php
```diff
@@ -241,6 +298,30 @@ public function processForm()
       $representation->save();
     }
 
+    // Upload new video track files
+    foreach ($this->videoTracks as $usageId => $videoTrack)
+    {
+      if (null !== $uf = $this->form->getValue("trackFile_$usageId"))
+      {
+        $lang = $this->form->getValue("lang_$usageId");
+        $uploadedTracks[$usageId] = array($uf, $lang);
```
jraddaoui: Nitpick: it would be easier to understand with a key/value array.

apps/qubit/modules/digitalobject/actions/editSubtitlesComponent.class.php
```diff
@@ -0,0 +1,33 @@
...
+ * @subpackage digitalObject
+ * 
+ */
+class DigitalObjectEditSubtitlesComponent extends sfComponent
+{
+  public function execute($request)
+  {
+  }
+}
+
```
jraddaoui: An empty component class usually indicates that this could be directly a partial.

jraddaoui (changes requested):
Hi @mInnes-archives,

Thanks a lot for this contribution, nice feature and great work, it's really neat.

There are a couple of important comments bellow that need to be addressed before merging this PR, specially the ones about migrations. There are also a few minor suggestions that would be nice to address but are not really a huge deal.

Finally, there are some "nitpick" comments related to code syntax that you could ignore, but I thought it would be nice to mention. They can be ignored ATM because we're about to implement a syntax checker that does it automatically, and we can pass it over this changes when they get merged (see https://github.com/artefactual/atom/pull/1249 if you're interested on it).

Thanks again!

lib/model/QubitDigitalObject.php
mInnes-archives: You're right. This works fine, and it's much simpler.

```suggestion
    if ($usageId == QubitTerm::SUBTITLES_ID)
    {
      $result = QubitDigitalObject::get($criteria);
    }
```

apps/qubit/modules/digitalobject/actions/editAction.class.php
mInnes-archives: I mostly just copied and modified the code used for uploading other representations.
I'll use `$uploadedTrack` to distinguish a bit. I'll change the variable on line 307 as well.

```suggestion
      if (null !== $uploadedTracks = $this->form->getValue("trackFile_$usageId"))
```

apps/qubit/modules/digitalobject/templates/_showVideo.php
mInnes-archives: Yes, weirdly enough.

apps/qubit/modules/digitalobject/actions/editAction.class.php
mInnes-archives: I'm essentially using the same method that is currently being used for representations. I think that this is touching on an issue above my little PR, so I'll defer to you folks to explore this.

apps/qubit/modules/digitalobject/templates/_show.php
mInnes-archives: My main goal was to avoid having a random `div` in the `video` element. Mind you, everything works and looks fine whether it's there or not, but it technically shouldn't be there. I can't say for certain whether this will cause an issue in the future, so I went with the safe, albeit inelegant, approach.

apps/qubit/modules/digitalobject/actions/editAction.class.php
mInnes-archives: Honestly, I have no idea. This was copied from the existing code that handles representations.

apps/qubit/modules/digitalobject/actions/editSubtitlesComponent.class.php
mInnes-archives: Yep.
Partial call:

```php
<?php echo include_partial('editSubtitles', array('resource' => $resource, 'subtitles' => $videoTrack, 'form' => $form, 'usageId' => $usageId)) ?>
```

apps/qubit/modules/digitalobject/templates/_editSubtitles.php
mInnes-archives: Something like this?
```
+++ b/apps/qubit/modules/digitalobject/templates/_editRepresentation.php
@@ -1,6 +1,6 @@
 <div>
 
-  <div style="float: right;">
+  <div class="digital-object-preview">
 
     <?php if ($representation->usageId == QubitTerm::CHAPTERS_ID): ?>
       <a href="<?php echo $representation->getFullPath() ?>"><?php echo __('View file') ?></a>
```
```
+++ b/plugins/arDominionPlugin/css/less/forms.less
@@ -160,6 +160,11 @@ fieldset {
     }
   }
 
+  // Thumbnails or 'View file' links for digital objects
+  .digital-object-preview {
+  float: right;
+  }
+
 }
```

apps/qubit/modules/digitalobject/templates/_show.php
jraddaoui: I don't think it's a big deal but I see your point. We could encapsulate the duplication in a extra partial but, considering how complex is the current components/partials setup in the digital objects module, we could leave it as it is.

apps/qubit/modules/digitalobject/actions/editAction.class.php
jraddaoui: Sounds good! And not a little PR ;)

apps/qubit/modules/digitalobject/actions/editSubtitlesComponent.class.php
mInnes-archives: Just to flag it, same issue here:

https://github.com/artefactual/atom/blob/qa/2.x/apps/qubit/modules/digitalobject/actions/editRepresentationComponent.class.php

apps/qubit/modules/digitalobject/templates/_show.php
mInnes-archives: This also works. Less lines of code, but you're throwing in an extra conditional to handle each part of the `div`. It's also a bit more obvious what the conditionals are actually doing this way.
```php
<?php if ($usageType != QubitTerm::CHAPTERS_ID && $usageType != QubitTerm::SUBTITLES_ID): ?>
  <div class="digital-object-reference">
<?php endif; ?>
    <?php if (!empty($accessWarning)): ?>
      <div class="access-warning">
        <?php echo $accessWarning ?>
      </div>
    <?php else: ?>
      <?php echo get_component('digitalobject', $showComponent, array('iconOnly' => $iconOnly, 'link' => $link, 'resource' => $resource, 'usageType' => $usageType)) ?>
    <?php endif; ?>
<?php if ($usageType != QubitTerm::CHAPTERS_ID && $usageType != QubitTerm::SUBTITLES_ID): ?>
  </div>
<?php endif; ?>

```

jraddaoui: I personally don't like this solution as you have to be aware of both for any change and it "breaks" indentation. Not a big deal of course, so I'm good with all the options.

mInnes-archives: Fair. The indentation is weird. I'll stick with my original solution in this case.

config/schema.yml
mInnes-archives: I've tested this and it looks like it works:

```php
<?php

/*
 * This file is part of the Access to Memory (AtoM) software.
 *
 * Access to Memory (AtoM) is free software: you can redistribute it and/or modify
 * it under the terms of the GNU Affero General Public License as published by
 * the Free Software Foundation, either version 3 of the License, or
 * (at your option) any later version.
 *
 * Access to Memory (AtoM) is distributed in the hope that it will be useful,
 * but WITHOUT ANY WARRANTY; without even the implied warranty of
 * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
 * GNU General Public License for more details.
 *
 * You should have received a copy of the GNU General Public License
 */

/*
 * Modify database for caption/subtitle/chapter support.
 *
 * @package    AccesstoMemory
 * @subpackage migration
 */
class arMigration0190
{
  const
    VERSION = 190, // The new database version
    MIN_MILESTONE = 2; // The minimum milestone required

  public function up($configuration)
  {
    // Add extra column, language
    QubitMigrate::addColumn(
      QubitDigitalObject::TABLE_NAME,
      'language VARCHAR(50)',
      ['after' => 'usage_id']
    );
  
    // Add chapters term
    QubitMigrate::bumpTerm(QubitTerm::CHAPTERS_ID, $configuration);
    $term = new QubitTerm;
    $term->id = QubitTerm::CHAPTERS_ID;
    $term->parentId = QubitTerm::ROOT_ID;
    $term->taxonomyId = QubitTaxonomy::DIGITAL_OBJECT_USAGE_ID;
    $term->sourceCulture = 'en';
    $term->setName('Chapters', ['culture' => 'en']);
    $term->setName('Chapitres', ['culture' => 'fr']);
    $term->save();

    // Add subtitles term
    QubitMigrate::bumpTerm(QubitTerm::SUBTITLES_ID, $configuration);
    $term = new QubitTerm;
    $term->id = QubitTerm::SUBTITLES_ID;
    $term->parentId = QubitTerm::ROOT_ID;
    $term->taxonomyId = QubitTaxonomy::DIGITAL_OBJECT_USAGE_ID;
    $term->sourceCulture = 'en';
    $term->setName('Captions/Subtitles', ['culture' => 'en']);
    $term->setName('Légendes/Sous-titres', ['culture' => 'fr']);
    $term->save();

    return true;
  }
}

```

Just to clarify, do you want the results of `propel:build-sql` and `propel:build-model` included in the PR? These include modifications to these files:

`/usr/share/nginx/atom/lib/model/map/DigitalObjectTableMap.php`
`lib/model/om/BaseDigitalObject.php`
`data/sql/lib.model.schema.sql`

I originally excluded these since I wasn't sure if these tasks were run as part of initial setup.

As an aside, I'm also seeing modifications to `actor`, `object`, `term`, and `user` tables and classes when these tasks are run.

apps/qubit/modules/digitalobject/templates/_showVideo.php
mInnes-archives: I looked at this again. It's not as bad as a first thought, so I'm cool with excluding it. Saves any issues it might cause in the future as well.

config/schema.yml
jraddaoui: Thanks for checking @mInnes-archives,

> Just to clarify, do you want the results of propel:build-sql and propel:build-model included in the PR?

Yes, please. These tasks are not part of the installation/upgrade process.

> As an aside, I'm also seeing modifications to actor, object, term, and user tables and classes when these tasks are run.

Good catch! Please, commit them too. Long story short, the builder has never been consistent ordering some stuff and it may reintroduce those changes the next time the models are built; but I'd like to check if it keeps happening and I think this is a good time.

mInnes-archives:
I've committed all of the changes that were reviewed. After final approval, I'll clean up the commits by submitting a new PR. I think I also need an issue ID to append to the end of the final commits.

I've also added chapter support for audio 22b8cc9. It follows the same logic as with video. The way mediaElement displays captions for audio isn't appropriate for a production environment, so I've excluded it. This required a bit of refactoring to handle which upload widgets show for video vs. audio, so this needs a quick review.

plugins/arDominionPlugin/css/less/forms.less
```diff
@@ -160,6 +160,11 @@ fieldset {
     }
   }
 
+  // Thumbnails or 'View file' links for digital objects
+  .digital-object-preview {
+  float: right;
```
jraddaoui: Nitpick: missing indentation.

apps/qubit/modules/digitalobject/actions/editSubtitlesComponent.class.php
```diff
@@ -1,33 +0,0 @@
...
- * @subpackage digitalObject
- * 
- */
-class DigitalObjectEditSubtitlesComponent extends sfComponent
-{
-  public function execute($request)
-  {
-  }
-}
-
```
jraddaoui: Shouldn't this include the component/partial call modification too?

apps/qubit/modules/digitalobject/templates/editSuccess.php
```diff
@@ -80,20 +80,19 @@
...
-            <legend><?php echo __('%1%', array('%1%' => QubitTerm::getById($usageId))) ?></legend> 
+          <?php if ($resource->mediaTypeId == QubitTerm::VIDEO_ID && $usageId == QubitTerm::SUBTITLES_ID): ?>
+            
+            <?php echo include_partial('editSubtitles', array('resource' => $resource, 'subtitles' => $videoTrack, 'form' => $form, 'usageId' => $usageId)) ?>
 
-            <?php if ($usageId == QubitTerm::SUBTITLES_ID): ?>
+          <?php elseif ($usageId != QubitTerm::SUBTITLES_ID): ?>                 
+          
+            <fieldset class="collapsible">
+            <legend><?php echo __('%1%', array('%1%' => QubitTerm::getById($usageId))) ?></legend>
```
jraddaoui: Nitpick: missing indentation.

apps/qubit/modules/digitalobject/actions/editSubtitlesComponent.class.php
jraddaoui: I see, it's in a later commit ;)

jraddaoui (commented):
Thanks for addressing all feedback, @mInnes-archives!

Nice to have support for audio too, I wonder how do the chapters look in there, but we'll figure out during testing. I've created a placeholder ticket, linking to this pull request ...

https://projects.artefactual.com/issues/13490

It would be nice to keep using this PR, and you shouldn't need to close it for the refactoring. If I may suggest, I would ...

- Address the couple of indentation issues mentioned above.
- Add this remote (if not done already) and rebase this branch with the latest changes from qa/2.x.
- Move this branch to a temporary one (`git branch -m tmp`).
- Checkout qa/2.x (update if needed to get the latest changes).
- Create a new branch based on qa/2.x with the same name as this one.
- Merge with squash the temporary branch into the new one and commit (`git merge --squash tmp && git commit`).
- Give a nice subject (50 chars with ref) and body (blank line to separate it from the subject and lines wrapped to 72 chars) to that commit, it should be the only one over the latest one from qa/2.x, and the commit that will get merged.
- Push (with force) the new branch to your remote, which should update this PR leaving a single commit (`git push -f ...`).

Thanks again for this contribution.

mInnes-archives:
I think that does it.

Thanks for your help.

For a quick preview, the audio chapters looks like this:

chapters_sample

You select the chapter in the little popup menu and it jumps to the time code specified in the srt/vtt file. This just gives an alternative to segmenting audio without breaking it into different files or whatnot.

jraddaoui (approved):

jraddaoui:
Amazing @mInnes-archives, thank you! It has been my pleasure.

PR #1264, opened 2021-03-03
Add RAD notes fields to CSV export, refs #7130

mcantelon:
Added fields for RAD other notes types - cast, credit, and signatures - to CSV export.

Added these fields to the example RAD CSV as they're alreaady supported by CSV import.


sbreker (approved):
Look good @mcantelon!

PR #1268, opened 2021-03-22
Initiate config/propel.ini on Docker, refs #13487

jraddaoui:
Copy the contents from config/propel.ini.tmpl on the bootstrap process
of the Docker Compose environment to avoid errors in the Propel tasks.


djjuhasz (approved):
Looks good! 👍

PR #1269, opened 2021-03-22
Update unit tests workflow, refs #13485 and #13486

jraddaoui:
- Test only over Ubuntu 20.04 and PHP 7.4 for now.
- Use `composer install` to install deps using the `composer.lock` file.
- Start services earlier in the process and reduce sleep time.


djjuhasz (approved):
Looks good 👍

PR #1270, opened 2021-03-22
Use PHP 7.4 and Composer 2 in Docker, refs #13488

jraddaoui:


Dockerfile
```diff
@@ -11,6 +11,7 @@ RUN set -xe \
       zlib-dev \
       libmemcached-dev \
       libzip-dev \
+      oniguruma-dev \
```
jraddaoui, line 14: Needed by `mbstring` in this version.

djjuhasz (approved):
Looks good 👍

PR #1271, opened 2021-03-23
Simplify debug mode in Docker, refs #13489

jraddaoui:
If ATOM_DEVELOPMENT_MODE is enabled, get the container's network gateway
and add it to the AtoM debug IPs automatically in the bootstrap process.


djjuhasz (approved):
Nice! 👍

PR #1273, opened 2021-03-26
CSV Check task enhancements, refs #13475

sbreker:


lib/CsvImportValidator.class.php
```diff
@@ -0,0 +1,320 @@
...
+  public function getResults()
+  {
+    $results = array();
+
+    foreach ($this->getCsvTests() as $test)
+    {
+      $results[$test->getFileName()][] = $test->getTestResult();
+    }
+    
+    //var_dump($results);
```
mcantelon: Commented out code.

lib/CsvImportValidator.class.php
```diff
@@ -0,0 +1,320 @@
...
+
+  public function setImportType(string $value)
+  {
+    if (in_array($value, $this->defaultCsvTypeMap))
+    {
+      $this->validatorOptions['importType'] = $value;
+    }
+    else
+    {
+      throw new UnexpectedValueException(sprintf('Invalid option "%s".', $name));
```
mcantelon: Should it be "Invalid name"? Where is $name set?

lib/task/import/csvCheckImportTask.class.php
```diff
@@ -21,172 +21,152 @@
...
+  /**
+   * @see sfTask
+   */
+  public function execute($arguments = array(), $options = array())
+  {
+    parent::execute($arguments, $options);
+    
+    $validatorOptions = $this->setOptions($options);
+
+    if (isset($options['verbose']) && $options['verbose'])
```
mcantelon: Nitpick: could check for `!empty($options['verbose'])`.

lib/task/import/csvCheckImportTask.class.php
```diff
@@ -21,172 +21,152 @@
...
+    return $filenames;
+  }
+
+  protected function setOptions($options = [])
+  {
+    $this->validateOptions($options);
+
+    $opts = array();
+    
+    //$validatorOptions = [];
```
mcantelon: Commented-out code.

lib/CsvImportValidator.class.php
```diff
@@ -0,0 +1,466 @@
...
+    const UTF32_LITTLE_ENDIAN_BOM = "\xFF\xFE\x00\x00";
+    const UTF32_BIG_ENDIAN_BOM = "\x00\x00\xFE\xFF";
+
+    public static $bomTypeMap = [
+        'utf8Bom' => self::UTF8_BOM,
+        'utf16LittleEndianBom' => self::UTF16_LITTLE_ENDIAN_BOM,
+        'utf16BigEndianBom' => self::UTF16_BIG_ENDIAN_BOM,
+        'utf32LittleEndianBom' => self::UTF32_LITTLE_ENDIAN_BOM,
+        'utf32BigEndianBom' => self::UTF32_BIG_ENDIAN_BOM,
+    ];
```
djjuhasz, line 49: Cosmetic: linebreak after multi-line array closing bracket.

lib/CsvImportValidator.class.php
```diff
@@ -0,0 +1,466 @@
...
+    // Tests which pertain only to this class type.
+    protected $qubitInformationObjectTestList = [
+        'CsvParentValidator' => CsvParentValidator::class,
+        'CsvLegacyIdValidator' => CsvLegacyIdValidator::class,
+        'CsvEventValuesValidator' => CsvEventValuesValidator::class,
+        'CsvDigitalObjectPathValidator' => CsvDigitalObjectPathValidator::class,
+        'CsvDigitalObjectUriValidator' => CsvDigitalObjectUriValidator::class,
+    ];
+
+    public function __construct(sfContext $context = null,
```
djjuhasz: Cosmetic: sfContext argument should be on next line (https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/3.0/doc/rules/function_notation/method_argument_space.rst)

lib/CsvImportValidator.class.php
```diff
@@ -0,0 +1,466 @@
...
+    $dbcon = null,
```
djjuhasz: Cosmetic: indenting

lib/CsvImportValidator.class.php
```diff
@@ -0,0 +1,466 @@
...
+    $options = [])
```
djjuhasz: Cosmetic: Closing parenthesis should be on next line

lib/CsvImportValidator.class.php
```diff
@@ -0,0 +1,466 @@
...
+                'QubitObject' => QubitObject::class,
+            ]
+        );
+    }
+
+    public function getTestsByClassType()
+    {
+        $selectedTests = $this->generalTestList;
+
+        switch ($this->validatorOptions['className']) {
```
djjuhasz: Opinion: I think we should avoid switch/case constructs when possible in AtoM.  Switch/case is bad for cyclomatic complexity and not well supported by php-cs-fixer.  I think it would be better to use  separate classes to encapsulate differences between the test object types, and the factory pattern to choose the right object is better OOP architecture based on the principle of replacing conditionals with small objects (see: [Sandi Metz, all the Little Things](https://www.youtube.com/watch?v=8bZh5LMaSmE) and the Gilded Rose Kata).

lib/CsvImportValidator.class.php
```diff
@@ -0,0 +1,466 @@
...
+
+        $this->rows = [];
+        while ($item = fgetcsv($fh, 60000, $this->getOption('separator'), $this->getOption('enclosure'))) {
+            $this->rows[] = $item;
+        }
+
+        // Remove trailing blank lines. Iterate over array from bottom up,
+        // removing empty rows.
+        for ($i = count($this->rows) - 1; $i >= 0; --$i) {
+            if (!(1 == count($this->rows[$i]) && empty(trim(implode('', $this->rows[$i]))))) {
```
djjuhasz: I find this conditional really confusing.  

I think it would be clearer without the initial "not":

```
if (1 !== count($this->rows[$i]) || empty(trim(implode('', $this->rows[$i])))) {
    return;
}
```

I also don't think the `implode()` is necessary, because the the first condition ensures that there is one and only one element in the array.

lib/CsvImportValidator.class.php
```diff
@@ -0,0 +1,466 @@
...
+        }
+    }
+
+    public function validate()
+    {
+        if (empty($this->getCsvTests())) {
+            $this->setCsvTests($this->getTestsByClassType());
+        }
+
+        unset($this->resultCollection);
```
djjuhasz: I don't think `unset($this->resultCollection)` is necessary before assigning a new `CsvValidatorResultCollection` object to `$this->resultCollection`?

lib/CsvImportValidator.class.php
```diff
@@ -0,0 +1,466 @@
...
+            return "\nAnalysis complete.\n";
+        }
+
+        return $output;
+    }
+
+    protected function getLongestRow(): int
+    {
+        $rowsMaxCount = count(max($this->rows));
+        $headerCount = count($this->header);
```
djjuhasz: Instead could do:

```
return max($rowsMaxCount, count($this->header));
```

lib/task/import/csvCheckImportTask.class.php
```diff
@@ -40,153 +35,126 @@ class csvCheckImportTask extends csvImportBaseTask
...
-            // Add values of both arrays if possible
-            foreach ($a as $key => $value) {
-                if (isset($b[$key])) {
-                    $c[$key] = $a[$key] + $b[$key];
-                } else {
-                    $c[$key] = $a[$key];
-                }
-            }
+        $validator = new CsvImportValidator(
+            $this->context, $this->getDbConnection(), $validatorOptions);
```
djjuhasz: Cosmetic: closing parenthesis on new line

lib/task/import/csvCheckImportTask.class.php
```diff
@@ -40,153 +35,126 @@ class csvCheckImportTask extends csvImportBaseTask
...
             }
+
+            $opts[$newkey] = $options[$oldkey];
         }
 
-        echo "\n";
+        return $opts;
+    }
+
+    protected function validateOptions($options = [])
```
djjuhasz: I think this empty method should be removed or validation logic should be added.

lib/task/import/validate/CsvEventValuesValidator.class.php
```diff
@@ -0,0 +1,116 @@
...
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with Access to Memory (AtoM).  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+/**
+ * CSV Multi Value test. Test if specified mulitvalue columns are all of the same length.
```
djjuhasz: Typo: "mulitvalue", and standardize on "multi value" or "multivalue"

lib/task/import/validate/CsvEventValuesValidator.class.php
```diff
@@ -0,0 +1,116 @@
...
+ * e.g.: eventType, eventActor, etc...
+ *
+ * Issue warning if eventTypes has nonzero count of x eventTypes, and any another column
+ * has a different number of event values than eventTypes. Do not warn for empty fields.
+ *
+ * @author     Steve Breker <sbreker@artefactual.com>
+ *
+ * @internal
+ * @coversNothing
```
djjuhasz: Is this run as PHPUnit test? If not I don't think @coversNothing is appropriate.

lib/task/import/validate/CsvEventValuesValidator.class.php
```diff
@@ -0,0 +1,116 @@
...
+ */
+class CsvEventValuesValidator extends CsvBaseValidator
+{
+    const TITLE = 'Event Value Count Test';
+    protected $matchList = [
```
djjuhasz, line 34: Cosmetic: blank line between `const` and `protected` definitions.

lib/task/import/validate/CsvEventValuesValidator.class.php
```diff
@@ -0,0 +1,116 @@
...
+        $row = $this->combineRow($header, $row);
+
+        // Check which columns are present on first row read.
+        if (!$this->columnsChecked) {
+            // Loop over each configured column check and see if they are present in the import CSV.
+            foreach ($this->matchList as $columnName) {
+                if (array_key_exists($columnName, $row)) {
+                    $this->columnsFound[] = $columnName;
+                }
+            }
```
djjuhasz, line 78: Cosmetic: blank line after multi-line statement.

lib/task/import/validate/csvBaseValidator.class.php
```diff
@@ -0,0 +1,134 @@
...
+
+    public function reset()
+    {
+        $this->filename = '';
+        $this->testData = new CsvValidatorResult($this->title, $this->filename, $this->getClassName(), $this->getVerbose());
+    }
+
+    public function getVerbose()
+    {
+        return (isset($this->options['verbose']) && $this->options['verbose']) ? true : false;
```
djjuhasz: This can be simplified to:

```
return isset($this->options['verbose']) && $this->options['verbose'];
```

lib/task/import/validate/csvBaseValidator.class.php
```diff
@@ -0,0 +1,134 @@
...
+        return get_class($this);
+    }
+
+    protected function combineRow(array $header, array $row)
+    {
+        // Enforce header has $columnCount elements. Add elements if necessary.
+        for ($i = count($header); $i < $this->columnCount; ++$i) {
+            $header[] = sprintf('%s-%d', self::HEADER_PLACEHOLDER, $i);
+        }
+        // Enforce row has $columnCount elements.
```
djjuhasz, line 123: Cosmetic: blank line after multi-line statement

lib/task/import/validate/csvColumnCountValidator.class.php
```diff
@@ -0,0 +1,94 @@
...
+            $this->updateRowCountSummary($this->headerCount);
+        }
+
+        $this->updateRowCountSummary(count($row));
+    }
+
+    public function getTestResult()
+    {
+        // When rows are all same length then rowCountSummary will have 1 row.
+        if (1 == count($this->rowCountSummary)) {
```
djjuhasz: Cosmetic: "Always use identical comparison unless you need type juggling;" -- https://symfony.com/doc/current/contributing/code/standards.html

lib/task/import/validate/csvColumnNameValidator.class.php
```diff
@@ -0,0 +1,146 @@
...
+
+        parent::reset();
+    }
+
+    public function testRow(array $header, array $row)
+    {
+        parent::testRow($header, $row);
+
+        // Only do this check once per file.
+        if (!$this->complete) {
```
djjuhasz: Opinion: I think it's easier to read if you just return when there is nothing to do, e.g.

```
if ($this->complete) {
    return;
}
```

This makes it clear that no further logic happens in this method, and avoids indenting all of the following code unnecessarily.

lib/task/import/validate/csvColumnNameValidator.class.php
```diff
@@ -0,0 +1,146 @@
...
+            foreach ($header as $columnName) {
+                // If $columnName is not in validColumnNames.
+                if (!in_array($columnName, $this->validColumnNames)) {
```
djjuhasz: Opinion: I think this would be easier to read if the valid column check is refactored into a separate method.

Also, again I would prefer:

```
if (in_array($columnName, $this->validColumnNames)) {
     continue;
}
```

To make it clear that no further processing should be done in the branch, and avoid unnecessary indenting.

lib/task/import/validate/csvColumnNameValidator.class.php
```diff
@@ -0,0 +1,146 @@
...
+        return parent::getTestResult();
+    }
+
+    // Load the column name config from lib/flatfile/config.
+    protected function loadObjectColumnNames($resourceClass)
+    {
+        $resourceTypeBaseConfigFile = $resourceClass.'.yml';
+        $config = QubitFlatfileExport::loadResourceConfigFile($resourceTypeBaseConfigFile, 'base');
+
+        $this->validColumnNames = $config['columnNames'];
```
djjuhasz, line 133: I think this will trigger a PHP warning if 'columnNames' isn't defined in the config file?

lib/task/import/validate/csvCultureValidator.class.php
```diff
@@ -0,0 +1,146 @@
...
+        parent::testRow($header, $row);
+        $row = $this->combineRow($header, $row);
+
+        // Set if culture column is present.
+        if (!isset($this->cultureColumnPresent)) {
+            $this->cultureColumnPresent = array_key_exists('culture', $row);
+        }
+
+        // If present check contents.
+        if ($this->cultureColumnPresent) {
```
djjuhasz: Opinion: Return if culture column isn't present:

```
if (!$this->cultureColumnPresent) {
     return;
}
```

lib/task/import/validate/csvCultureValidator.class.php
```diff
@@ -0,0 +1,146 @@
...
+            if (empty($row['culture'])) {
+                ++$this->rowsWithBlankCulture;
```
djjuhasz: Opinion: I'd add a return here too, to avoid the nested if/else, i.e.

```
if (empty($row['culture'])) {
    ++$this->rowsWithBlankCulture;

    return;
}
```

lib/task/import/validate/csvCultureValidator.class.php
```diff
@@ -0,0 +1,146 @@
...
+            } else {
+                // Check if contains pipe.
+                if (0 < strpos($row['culture'], '|')) {
+                    ++$this->rowsWithPipeFoundInCulture;
+                    $this->testData->addDetail(implode(',', $row));
+
+                    // Keep a list of invalid culture values.
```
djjuhasz: I would restructure this if/elseif to avoid repeating this code block twice. e.g.

```
if ($this->isCultureValid($row['culture'])) {
    return;
}

// Keep a list of invalid culture values.
if (!in_array($row['culture'], $this->invalidCultures)) {
    $this->invalidCultures[] = $row['culture'];
}

... check for pipe or invalid culture string
```

lib/task/import/validate/csvCultureValidator.class.php
```diff
@@ -0,0 +1,146 @@
...
+                        $this->invalidCultures[] = $row['culture'];
+                    }
+                }
+            }
+        }
+    }
+
+    public function getTestResult()
+    {
+        if (false == $this->cultureColumnPresent) {
```
djjuhasz: This can be shortened to `if (!$this->cultureColumnPresent) {`

lib/task/import/validate/csvCultureValidator.class.php
```diff
@@ -0,0 +1,146 @@
...
+            // culture column not present in file.
+            $this->testData->setStatusWarn();
+            $this->testData->addResult(sprintf("'culture' column not present in file."));
+            $this->testData->addResult(sprintf("Rows without a valid culture value will be imported using AtoM's default source culture."));
+        } else {
```
djjuhasz: Opinion: Again, I'd return here to close the branch and avoid the long `else`:

```
   return parent::getTestResult();
}
```

lib/task/import/validate/csvColumnNameValidator.class.php
djjuhasz: Or could the column name validation be moved into the constructor, to avoid the need for the `$this->complete` flag to run the checks only on the header row?

lib/task/import/validate/CsvEventValuesValidator.class.php
```diff
@@ -0,0 +1,116 @@
...
+/**
+ * CSV Multi Value test. Test if specified mulitvalue columns are all of the same length.
+ * e.g.: eventType, eventActor, etc...
+ *
+ * Issue warning if eventTypes has nonzero count of x eventTypes, and any another column
+ * has a different number of event values than eventTypes. Do not warn for empty fields.
+ *
+ * @author     Steve Breker <sbreker@artefactual.com>
+ *
+ * @internal
```
djjuhasz: I don't understand why `@internal` is used here.

lib/task/import/validate/csvDigitalObjectPathValidator.class.php
```diff
@@ -0,0 +1,232 @@
...
+    // Do not reset between CSVs.
+    protected $fileList = [];
+    protected $pathToDigitalObjects = '';
+
+    protected $digitalObjectPathColumnPresent;
+    protected $digitalObjectUriColumnPresent;
+    protected $digitalObjectUses = [];
+    protected $overriddenByUriCount = 0;
+
+    public function __construct(array $options = null)
```
djjuhasz, line 44: Nitpick: the type declaration for the `$options` argument should be explicitly set as nullable, e.g.

```
public function __construct(?array $options = null)
```

This will require updating the definition for `__construct()` for all of the subclasses of this class.

lib/task/import/validate/csvBaseValidator.class.php
```diff
@@ -0,0 +1,134 @@
...
+
+    protected $filename = '';
+    protected $columnCount = 0;
+    protected $rowNumber = 1;
+    protected $title = '';
+    protected $options = [];
+    protected $ormClasses = [];
+    protected $testData;
+
+    public function __construct(array $options = null)
```
djjuhasz: Nitpick: the type declaration for the `$options` argument should be explicitly set as nullable, e.g.

```
public function __construct(?array $options = null)
```

This will require updating the definition for `__construct()` for all of the subclasses of this class.

lib/task/import/validate/csvDigitalObjectPathValidator.class.php
```diff
@@ -0,0 +1,232 @@
...
+                }
+            }
+        }
+    }
+
+    public function getTestResult()
+    {
+        if (false === $this->digitalObjectPathColumnPresent) {
+            $this->testData->addResult(sprintf("Column 'digitalObjectPath' not present in CSV. Nothing to verify."));
+        } else {
```
djjuhasz: Opinion: return instead of else

lib/task/import/validate/csvDigitalObjectPathValidator.class.php
```diff
@@ -0,0 +1,232 @@
...
+                if (!empty($row['digitalObjectPath']) && !empty($row['digitalObjectUri'])) {
+                    ++$this->overriddenByUriCount;
+                }
+            }
+        }
+    }
+
+    public function getTestResult()
+    {
+        if (false === $this->digitalObjectPathColumnPresent) {
```
djjuhasz, line 91: Is it not valid to have CSV file with a `digitalObjectUri` column, but no `digitalObjectPath` column?

lib/task/import/validate/csvDigitalObjectPathValidator.class.php
```diff
@@ -0,0 +1,232 @@
...
+            if (empty($this->pathToDigitalObjects)) {
+                // Option was not supplied.
+                if (empty($this->options['pathToDigitalObjects'])) {
+                    $this->testData->addResult(sprintf('Digital object folder location not specified.'));
+                }
+                // Path could not be found.
+                else {
+                    $this->testData->addResult(sprintf('Unable to open digital object folder path: %s', $this->options['pathToDigitalObjects']));
+                }
+                // If digitalObjectPath column is populated in CSV, this is an error.
```
djjuhasz, line 106: Cosmetic: blank line before comment

lib/task/import/validate/csvDigitalObjectPathValidator.class.php
```diff
@@ -0,0 +1,232 @@
...
+            // Digital object folder option not passed/is invalid.
+            if (empty($this->pathToDigitalObjects)) {
+                // Option was not supplied.
+                if (empty($this->options['pathToDigitalObjects'])) {
+                    $this->testData->addResult(sprintf('Digital object folder location not specified.'));
+                }
+                // Path could not be found.
+                else {
+                    $this->testData->addResult(sprintf('Unable to open digital object folder path: %s', $this->options['pathToDigitalObjects']));
+                }
```
djjuhasz: Cosmetic: Blank like after multi-line statement closing brace

lib/task/import/validate/csvDigitalObjectPathValidator.class.php
```diff
@@ -0,0 +1,232 @@
...
+
+    public function getTestResult()
+    {
+        if (false === $this->digitalObjectPathColumnPresent) {
+            $this->testData->addResult(sprintf("Column 'digitalObjectPath' not present in CSV. Nothing to verify."));
+        } else {
+            $this->testData->addResult(sprintf("Column 'digitalObjectPath' found."));
+
+            // Digital object folder option not passed/is invalid.
+            if (empty($this->pathToDigitalObjects)) {
```
djjuhasz: Opinion: It would be great to reduce the complexity of this highly nested if/else tree.  I'd recommend moving sub-branches of the tree into separate methods and using return statements to short circuit long logic trees where possible.

lib/task/import/validate/csvDigitalObjectPathValidator.class.php
```diff
@@ -0,0 +1,232 @@
...
+
+        return parent::getTestResult();
+    }
+
+    public function setPathToDigitalObjects(string $path)
+    {
+        $path = (!empty($path) && is_dir($path)) ? $path : '';
+
+        if (is_dir($path)) {
+            $this->pathToDigitalObjects = realpath($path) ? realpath($path) : $path;
```
djjuhasz, line 171: Cool, I didn't know about `realpath()` :)

djjuhasz: I think if `realpath($path)` returns false it should be considered an error instead of continuing with `$path`

lib/task/import/validate/csvDigitalObjectPathValidator.class.php
```diff
@@ -0,0 +1,232 @@
...
+        }
+
+        return $missingDigitalObjects;
+    }
+
+    private function getDigitalObjectFiles()
+    {
+        $this->fileList = [];
+
+        if (!empty($this->pathToDigitalObjects)) {
```
djjuhasz: Opinion:

```
if (empty($this->pathToDigitalObjects)) {
     return;
}
```

lib/task/import/validate/csvDigitalObjectUriValidator.class.php
```diff
@@ -0,0 +1,144 @@
...
+    {
+        $invalidUris = [];
+
+        foreach ($this->digitalObjectUses as $uri => $uses) {
+            if (false === filter_var($uri, FILTER_VALIDATE_URL)) {
+                array_push($invalidUris, $uri);
+            } else {
+                $parsedUri = parse_url($uri);
+
+                if (!('https' == $parsedUri['scheme'] || 'http' == $parsedUri['scheme'])) {
```
djjuhasz: Opinion: I find it harder to read conditionals when a multiple term conditional is negated with "!".

E.g.  This is easier for me to read:

```
if ('https' != $parsedUri['scheme'] && 'http' != $parsedUri['scheme']) {
```

lib/task/import/validate/csvDuplicateColumnNameValidator.class.php
```diff
@@ -0,0 +1,79 @@
...
+
+        parent::reset();
+    }
+
+    public function testRow(array $header, array $row)
+    {
+        parent::testRow($header, $row);
+        $header = array_map('trim', $header);
+
+        // Only empty on first iteration.
```
djjuhasz: Again, I'm curious if this could be checked in the constructor instead?

lib/task/import/validate/csvDuplicateColumnNameValidator.class.php
```diff
@@ -0,0 +1,79 @@
...
+        if (empty($this->columnFrequency)) {
+            foreach ($header as $columnName) {
+                if (!array_key_exists($columnName, $this->columnFrequency)) {
```
djjuhasz: Opinion: I prefer

```
if (!isset($this->columnFrequency[$columnName])) {
```

And it sounds like it might be faster than `array_key_exists()` (though the behaviour is slightly different) based on the comments on https://www.php.net/manual/en/function.array-key-exists.php

lib/task/import/validate/csvFieldLengthValidator.php
```diff
@@ -0,0 +1,126 @@
...
+ * @internal
+ * @coversNothing
+ */
+class CsvFieldLengthValidator extends CsvBaseValidator
+{
+    const TITLE = 'Field Length Check';
+    // Add fields to check legth of here with max len. Anything larger than this value will
+    // trigger the default action.
+    protected $fieldMaxSizes = [
+        'culture' => 6,
```
djjuhasz: This might be a problem for Valencian users.  In our data fixtures we allow 'ca@valencia' as a valid culture value (e.g.: https://github.com/artefactual/atom/blob/qa/2.x/data/fixtures/menus.yml#L21).  The AtoM `culture` column definition is `varchar(16)` ([ref](https://github.com/artefactual/atom/blob/qa/2.x/data/sql/lib.model.schema.sql#L95)) so I would use a value of 11 or 16.

lib/task/import/validate/csvFieldLengthValidator.php
```diff
@@ -0,0 +1,126 @@
...
+    public function testRow(array $header, array $row)
+    {
+        parent::testRow($header, $row);
+        $row = $this->combineRow($header, $row);
+
+        // Check which columns are present on first row read.
+        if (!$this->columnsChecked) {
+            // Loop over each configured column check and see if they are present in the import CSV.
+            foreach ($this->fieldMaxSizes as $columnName => $fieldSize) {
+                if (array_key_exists($columnName, $row)) {
```
djjuhasz: Again, `isset($row[$columnName])` sounds like it is faster than `array_key_exists($columnName, $row)`

lib/task/import/validate/csvFieldLengthValidator.php
```diff
@@ -0,0 +1,126 @@
...
+            $this->testData->addResult('No columns to check.');
+        } else {
+            $this->testData->addResult(sprintf('Checking columns: %s', implode(',', array_keys($this->columnsFound))));
+        }
+
+        foreach ($this->columnsFound as $columnName => $errorCount) {
+            if (0 < $errorCount) {
+                $this->testData->setStatusWarn();
+                $this->testData->addResult(sprintf("'%s' column may have invalid values.", $columnName));
+            }
```
djjuhasz, line 118: Blank line after closing brace

lib/task/import/validate/csvLanguageValidator.class.php
```diff
@@ -0,0 +1,120 @@
...
+        parent::testRow($header, $row);
+        $row = $this->combineRow($header, $row);
+
+        // Set if language column is present.
+        if (!isset($this->languageColumnPresent)) {
+            $this->languageColumnPresent = array_key_exists('language', $row);
+        }
+
+        // If present check contents.
+        if ($this->languageColumnPresent) {
```
djjuhasz: Opinion: again, I think it's easier to understand the conditional as:

```
if (!$this->languageColumnPresent || empty($row['language'])) {
    return;
}
```

lib/task/import/validate/csvLanguageValidator.class.php
```diff
@@ -0,0 +1,120 @@
...
+                        }
+                    }
+                }
+            }
+        }
+    }
+
+    public function getTestResult()
+    {
+        if (false == $this->languageColumnPresent) {
```
djjuhasz: I think this should either be a strict comparison (e.g. `if (false === $this->languageColumnPresent) {`) or use the shorter `if (!$this->languageColumnPresent) {`

lib/task/import/validate/csvLanguageValidator.class.php
```diff
@@ -0,0 +1,120 @@
...
+
+            if (0 == $this->rowsWithInvalidLanguage) {
+                $this->testData->addResult(sprintf("'language' column values are all valid."));
+            }
+        }
+
+        return parent::getTestResult();
+    }
+
+    // TODO: Remove these DB accesses to a wrapper class so it's not performed in the
```
djjuhasz: It looks like this TODO has been TODONE 😛

lib/task/import/validate/csvLanguageValidator.class.php
```diff
@@ -0,0 +1,120 @@
...
+            $this->languageColumnPresent = array_key_exists('language', $row);
+        }
+
+        // If present check contents.
+        if ($this->languageColumnPresent) {
+            if (!empty($row['language'])) {
+                // Validate language value against AtoM.
+                $errorDetailAdded = false;
+                foreach (explode('|', $row['language']) as $value) {
+                    if (!$this->isLanguageValid(trim($value))) {
```
djjuhasz: I think it's slightly better to call `trim($value)` in the `isLanguageValid()` method so it's applied consistently when the method is called.

lib/task/import/validate/csvLanguageValidator.class.php
```diff
@@ -0,0 +1,120 @@
...
+        }
+
+        return parent::getTestResult();
+    }
+
+    // TODO: Remove these DB accesses to a wrapper class so it's not performed in the
+    // test class itself.
+    protected function isLanguageValid(string $language)
+    {
+        if (!empty($language)) {
```
djjuhasz: If `$language` is an empty string, this method will return null, which evaluate to `false` if cast as a boolean.  I think it's better to explicitly return `false` if `$language === ''` make it clear that an empty string is considered invalid.

lib/task/import/validate/csvLegacyIdValidator.class.php
```diff
@@ -0,0 +1,109 @@
...
+        $this->rowsWithoutLegacyId = 0;
+        $this->nonUniqueLegacyIdValues = [];
+
+        parent::reset();
+    }
+
+    public function testRow(array $header, array $row)
+    {
+        parent::testRow($header, $row);
+        $row = $this->combineRow($header, $row);
```
djjuhasz, line 57: Is `combineRow()` called every time from `testRow()`? If so, maybe it should be moved into the parent class?

lib/task/import/validate/csvLegacyIdValidator.class.php
```diff
@@ -0,0 +1,109 @@
...
+
+        // Is legacyId column present?
+        if (!isset($this->legacyIdColumnPresent)) {
+            $this->legacyIdColumnPresent = array_key_exists('legacyId', $row);
```
djjuhasz: `array_key_exists` vs. `isset` again :)

lib/task/import/validate/csvLegacyIdValidator.class.php
```diff
@@ -0,0 +1,109 @@
...
+        }
+
+        if ($this->legacyIdColumnPresent) {
+            if (empty($row['legacyId'])) {
+                ++$this->rowsWithoutLegacyId;
+                $this->testData->addDetail(implode(',', $row));
+            } else {
+                if (in_array($row['legacyId'], $this->legacyIdValues)) {
+                    // This is a duplicate legacyId. Add to list of non-unique legacyIds.
+                    $this->nonUniqueLegacyIdValues[] = $row['legacyId'];
```
djjuhasz: This will add multiple instances of `$row['legacyId']` to `$this->nonUniqueLegacyIdValues[]` if the same legacy id is used more than twice.  I think that is unintentional?

lib/task/import/validate/csvBaseValidator.class.php
mcantelon: And further to `return !empty($this->options['verbose'])`.

lib/task/import/validate/csvParentValidator.class.php
```diff
@@ -0,0 +1,214 @@
...
+            if (!$this->canFindByParentId($row['parentId'], $this->options['source'])) {
+                $this->orphanRowsFound = true;
+                ++$this->unmatchedCount;
+
+                // Add row that triggered this to the output.
+                $this->testData->addDetail(implode(',', $row));
+            }
+        }
+
+        if ($this->parentIdColumnPresent && empty($row['parentId'])
```
djjuhasz: I don't know what the "canonical" formatting is for multi-line conditionals like this, but my preference is:

```
if (
     $this->parentIdColumnPresent 
     && empty($row['parentId'])
     && $this->qubitParentSlugColumnPresent
     && empty($row['qubitParentSlug'])
) {
    ++$this->rowsWithoutParentIdQubitParentSlug;
}
```

djjuhasz: Confirmed that my suggestion is canonical: https://wiki.accesstomemory.org/wiki/Development/Coding_standard#Multi-line_control_structures

lib/task/import/validate/csvParentValidator.class.php
```diff
@@ -0,0 +1,214 @@
...
+        }
+
+        if ($this->legacyIdColumnPresent) {
+            $this->legacyIdList[] = $row['legacyId'];
+        }
+    }
+
+    public function getTestResult()
+    {
+        if (false == $this->parentIdColumnPresent && false == $this->qubitParentSlugColumnPresent) {
```
djjuhasz: Again, I'd use the shorter `if (!$this->parentIdColumnPresent && ...` unless strict type comparison is appropriate.

lib/task/import/validate/csvParentValidator.class.php
```diff
@@ -0,0 +1,214 @@
...
+            $this->testData->setStatusWarn();
+            $this->testData->addResult(sprintf("'parentId' and 'qubitParentSlugColumnPresent' columns not present. CSV contents will be imported as top level records."));
+        } else {
+            if ($this->parentIdColumnPresent) {
+                $this->testData->addResult(sprintf('Rows with parentId populated: %s', $this->rowsWithParentId));
+            }
+            if ($this->qubitParentSlugColumnPresent) {
```
djjuhasz, line 142: Blank line before control statement

lib/task/import/validate/csvParentValidator.class.php
```diff
@@ -0,0 +1,214 @@
...
+
+        if ($this->orphanRowsFound) {
+            $this->testData->setStatusError();
+            $this->testData->addResult(sprintf('Number of rows for which parents could not be found (will import as top level records): %s', $this->unmatchedCount));
+        }
+
+        return parent::getTestResult();
+    }
+
+    // TODO: Remove this DB access to a wrapper class so it's not performed in the
```
djjuhasz, line 167: This comment doesn't seem related to the method

lib/task/import/validate/csvValidatorResult.class.php
```diff
@@ -0,0 +1,160 @@
...
+    }
+
+    public function setTitle(string $title)
+    {
+        $this->testData[self::TEST_TITLE] = $title;
+    }
+
+    public function setStatus(int $status)
+    {
+        if (in_array($status, [self::RESULT_INFO, self::RESULT_WARN, self::RESULT_ERROR])
```
djjuhasz, line 62: First expression of a multi-line if statement should be on the next line.

djjuhasz: Should be:

```
if (
    in_array($status, [self::RESULT_INFO, self::RESULT_WARN, self::RESULT_ERROR])
    && ($status > $this->testData[self::TEST_STATUS])
) {
    $this->testData[self::TEST_STATUS] = intval($status);
}
```

lib/task/import/validate/csvValidatorResult.class.php
```diff
@@ -0,0 +1,160 @@
...
+{
+    // Integer type to allow comparison of severity values.
+    const RESULT_INFO = 0;
+    const RESULT_WARN = 1;
+    const RESULT_ERROR = 2;
+
+    const TEST_TITLE = 'title';
+    const TEST_STATUS = 'status';
+    const TEST_RESULTS = 'results';
+    const TEST_DETAIL = 'details';
```
djjuhasz: I think this should be `TEST_DETAILS` (plural) to match `TEST_RESULTS`.

lib/task/import/validate/csvValidatorResultCollection.php
```diff
@@ -0,0 +1,167 @@
...
+/**
+ * CSV validation result collection class.
+ *
+ * @author     Steve Breker <sbreker@artefactual.com>
+ */
+class CsvValidatorResultCollection
+{
+    protected $results = [];
+
+    public function __construct()
```
djjuhasz: Is there a reason to have an empty constructor?

lib/task/import/validate/csvValidatorResultCollection.php
```diff
@@ -0,0 +1,167 @@
...
+    }
+
+    public function appendResult(CsvValidatorResult $result)
+    {
+        $this->results[] = $result;
+
+        $this->sortByFilenameStatusDescending();
+    }
+
+    public function toArray(bool $verbose = false)
```
djjuhasz, line 40: I think it's better to make the argument type declaration explicitly nullable, e.g.

```
public function toArray(?bool $verbose = false)
```

djjuhasz, line 36: It doesn't look like `$verbose` does anything?

lib/task/import/validate/csvValidatorResultCollection.php
```diff
@@ -0,0 +1,167 @@
...
+
+    public function getByFilenameTestname(string $filename, string $testname)
+    {
+        foreach ($this->results as $result) {
+            if ($filename == $result->getFilename() && $testname == $result->getClassname()) {
+                return $result->toArray();
+            }
+        }
+
+        return $this->results[$filename][$testname]->toArray();
```
djjuhasz: I think this should be removed?

test/phpunit/csvImportValidator/CsvBaseValidatorTest.php
```diff
@@ -0,0 +1,63 @@
...
+    {
+        $csvValidator = new CsvMockValidator([]);
+        $csvValidator->setTitle('Title');
+        $this->assertSame('Title', $csvValidator->getTitle());
+    }
+
+    public function testSetFilename()
+    {
+        $csvValidator = new CsvMockValidator([]);
+        $csvValidator->setTitle('Title');
```
djjuhasz: I think this `setTitle()` is unnecessary

test/phpunit/csvImportValidator/CsvBaseValidatorTest.php
```diff
@@ -0,0 +1,63 @@
...
+        $this->assertInstanceOf(CsvValidatorResult::class, $csvValidator->getTestResult());
+    }
+
+    public function testGetColumnCount()
+    {
+        $csvValidator = new CsvMockValidator([]);
+        $csvValidator->setColumnCount(12);
+        $this->assertInstanceOf(CsvValidatorResult::class, $csvValidator->getTestResult());
```
djjuhasz: This looks like a copy & paste error.

test/phpunit/csvImportValidator/CsvColumnCountTest.php
```diff
@@ -0,0 +1,279 @@
...
+    }
+
+    /**
+     * @dataProvider csvValidatorTestProvider
+     *
+     * Generic test - options and expected results from csvValidatorTestProvider()
+     *
+     * @param mixed $options
+     */
+    public function testCsvValidator($options)
```
djjuhasz, line 81: I think this class should go in it's own file

lib/task/import/validate/csvValidatorResultCollection.php
sbreker: I realized this after you'd begun reviewing. There was a time when the verbose flag was used in the validations themselves to prevent accumulating RESULT_DETAIL items. When the WebUI requirements started to lean towards a combination of verbose and non-verbose output on every run, I removed the logic in the validators, but left that option in.  Going forward only the ResultCollection class needs to know about 'verbose' so I will remove this all from the tests.

lib/task/import/validate/csvValidatorResultCollection.php
sbreker: This function is used by the PHPUnit tests to grab exactly one result array without the collection's filename/testname hierarchy.

lib/task/import/validate/csvValidatorResultCollection.php
sbreker: No - will remove. Must have been expecting to put something in there!

lib/task/import/validate/csvValidatorResult.class.php
sbreker: 👍

lib/task/import/validate/csvParentValidator.class.php
sbreker: Nice - TIL...

lib/task/import/validate/csvLegacyIdValidator.class.php
sbreker: It is not. It's up to the Validator if they need the combined row or not on each iteration - there are a couple that don't call this.

lib/task/import/validate/csvLanguageValidator.class.php
sbreker: 😂  - will clean up

lib/task/import/validate/csvFieldLengthValidator.php
sbreker: Sounds good - will apply this change across the project. Didn't know about the perf diff between the two.

lib/task/import/validate/csvColumnNameValidator.class.php
sbreker: Good catch. I assumed that each yml file would _always_ contain a 'columnNames' section. Will check this out....

lib/task/import/validate/csvColumnCountValidator.class.php
sbreker: 👍

lib/task/import/validate/csvValidatorResultCollection.php
djjuhasz: @sbreker I mean that the line should be removed, not the function.  I think this second return is extraneous because the preceding foreach loop does the same thing except without the risk of errors if the $filename and $testname keys don't exist in the array.

test/phpunit/csvImportValidator/CsvValidatorResultTest.php
```diff
@@ -0,0 +1,122 @@
...
+        $csvValidator->setStatusWarn();
+        // Should still be in error state.
+        $this->assertSame(CsvValidatorResult::RESULT_ERROR, $csvValidator->getStatus());
+
+        $csvValidator = new CsvValidatorResult('Title', 'Filename', 'Classname', true);
+        $csvValidator->setStatus(CsvValidatorResult::RESULT_ERROR);
+        $this->assertSame(CsvValidatorResult::RESULT_ERROR, $csvValidator->getStatus());
+    }
+
+    public function testSetResult()
```
djjuhasz: Cosmetic: I think `testAddResult()` is more accurate.

test/phpunit/csvImportValidator/CsvValidatorResultTest.php
```diff
@@ -0,0 +1,122 @@
...
+    {
+        $csvValidator = new CsvValidatorResult('Title', 'Filename', 'Classname', true);
+        $csvValidator->addResult('Result');
+        $this->assertSame(['Result'], $csvValidator->getResults());
```
djjuhasz: It might be good to test adding a second result?

test/phpunit/csvImportValidator/CsvValidatorResultTest.php
```diff
@@ -0,0 +1,122 @@
...
+    }
+
+    public function testSetDetail()
```
djjuhasz: Cosmetic: `testAddDetail()`?

test/phpunit/csvImportValidator/CsvValidatorResultTest.php
```diff
@@ -0,0 +1,122 @@
...
+    {
+        $csvValidator = new CsvValidatorResult('Title', 'Filename', 'Classname', true);
+        $csvValidator->addDetail('Detail');
+        $this->assertSame(['Detail'], $csvValidator->getDetails());
```
djjuhasz: Might be good to test adding a second detail.

test/phpunit/csvImportValidator/CsvValidatorResultTest.php
```diff
@@ -0,0 +1,122 @@
...
+    }
+
+    public function testFormatStatus()
+    {
+        //$csvValidator = new CsvValidatorResult('Title', 'Filename', 'Classname', true);
```
djjuhasz: Remove

test/phpunit/csvImportValidator/CsvValidatorResultTest.php
```diff
@@ -0,0 +1,122 @@
...
+    }
+
+    public function testToArray()
+    {
+        $csvValidator = new CsvValidatorResult('Title', 'Filename', 'Classname', true);
+        $result = $csvValidator->toArray();
+        $this->assertSame(
+            [
+                'title' => 'Title',
+                'status' => 0,
```
djjuhasz: I think `'status' => CsvValidatorResult::RESULT_INFO,` would be more robust.

djjuhasz (changes requested):
@sbreker Wow, looks great overall!  Most of the changes I requested are pretty minor. Some of my opinions could be controversial, but I'm always happy to discuss.

I'm very glad there are unit tests with all of the many validators and options! 😅

lib/task/import/csvCheckImportTask.class.php
sbreker: 👍

lib/task/import/csvCheckImportTask.class.php
sbreker: I am going to remove it as the CsvValidator itself is already responsible for validating these options.

lib/task/import/validate/CsvEventValuesValidator.class.php
sbreker: 👍

lib/task/import/validate/CsvEventValuesValidator.class.php
sbreker: This was mistakenly added somehow. Removed in all Validator classes.

lib/task/import/validate/CsvEventValuesValidator.class.php
sbreker: Maybe PHP-CS-Fixer added these at one point in development? Removed in all cases.

lib/task/import/validate/CsvEventValuesValidator.class.php
sbreker: Blank line added in all cases.

lib/task/import/validate/CsvEventValuesValidator.class.php
sbreker: Added.

lib/task/import/validate/csvBaseValidator.class.php
sbreker: Inserted.

lib/task/import/validate/csvColumnCountValidator.class.php
sbreker: Addressed in mulitple places...

lib/task/import/validate/csvDigitalObjectPathValidator.class.php
sbreker: Added.

lib/task/import/validate/csvFieldLengthValidator.php
sbreker: Added.

lib/task/import/validate/csvLanguageValidator.class.php
sbreker: Updated to use `if (!$this->languageColumnPresent)`

lib/task/import/validate/csvLanguageValidator.class.php
sbreker: Removed.

lib/task/import/validate/csvParentValidator.class.php
sbreker: Completed in this file. Will look around for other instances.

lib/task/import/validate/csvParentValidator.class.php
sbreker: 👍 Changed

lib/task/import/validate/csvParentValidator.class.php
sbreker: Fixed

lib/task/import/validate/csvValidatorResult.class.php
sbreker: Change completed.

lib/task/import/validate/csvValidatorResult.class.php
sbreker: Fixed

lib/task/import/validate/csvValidatorResultCollection.php
sbreker: Removed

lib/task/import/validate/csvValidatorResultCollection.php
sbreker: `--verbose` option removed from the Validator classes.

lib/task/import/validate/csvBaseValidator.class.php
sbreker: Ok - that makes sense. Change made in all affected Validator constructors.

lib/CsvImportValidator.class.php
sbreker: Removed.

lib/CsvImportValidator.class.php
sbreker: Good catch! Fixed...

lib/CsvImportValidator.class.php
sbreker: Line break added.

lib/CsvImportValidator.class.php
sbreker: Fixed as per FoPHP method_argument_space.

lib/CsvImportValidator.class.php
sbreker: Fixed

lib/CsvImportValidator.class.php
sbreker: fixed

lib/CsvImportValidator.class.php
sbreker: Will clean up that initial 'not'. The implode() is necessary though because even when there is only one column, $this->rows[$i] is an array.

lib/CsvImportValidator.class.php
sbreker: Removing.

lib/CsvImportValidator.class.php
sbreker: Yeah - that's better. 😬

lib/task/import/validate/csvDuplicateColumnNameValidator.class.php
sbreker: Addressed in new commit. Replaced uses of `array_key_exists()` with `isset()`.

lib/task/import/validate/csvFieldLengthValidator.php
sbreker: Addressed in new commit. Replaced uses of `array_key_exists()` with `isset()`.

lib/task/import/validate/csvLegacyIdValidator.class.php
sbreker: Addressed in new commit. Replaced uses of `array_key_exists()` with `isset()`.

lib/task/import/validate/csvCultureValidator.class.php
sbreker: Added.

lib/task/import/validate/csvCultureValidator.class.php
sbreker: Added.

lib/task/import/validate/csvCultureValidator.class.php
sbreker: Another great suggestion. Change completed.

lib/task/import/validate/csvCultureValidator.class.php
sbreker: Change completed.

lib/task/import/validate/csvDigitalObjectPathValidator.class.php
sbreker: Agreed. Change completed.

lib/task/import/validate/csvDigitalObjectPathValidator.class.php
sbreker: Change completed.

lib/task/import/validate/csvDigitalObjectPathValidator.class.php
sbreker: This logic (returning $path in the case where realpath() fails) is needed to support the use of the VFS in the PHP Unit tests as realpath() always returns false.

lib/task/import/validate/csvDigitalObjectPathValidator.class.php
sbreker: Good idea. Change completed.

lib/task/import/validate/csvDuplicateColumnNameValidator.class.php
sbreker: No unfortunately. The Validators don't have the data until the first row is passed into testRow(). This test checks the header for dupes once only - this condition prevents it being re-checked on every testRow() iteration. I can definitely update this block to return early if !empty().

lib/task/import/validate/csvFieldLengthValidator.php
sbreker: Fixed. Good catch.

lib/task/import/validate/csvLanguageValidator.class.php
sbreker: Yes, good suggestion. Updated....

lib/task/import/validate/csvLanguageValidator.class.php
sbreker: 👍  Change completed.

lib/task/import/validate/csvLanguageValidator.class.php
sbreker: Good suggestion! Implemented...

lib/task/import/validate/csvLegacyIdValidator.class.php
sbreker: Definitely unintentional. 🦅 👁️

sbreker: Fixed and updated existing unit test for this condition.

lib/task/import/validate/csvValidatorResultCollection.php
sbreker: Oh I see now! Haha. Yes, that's shouldn't be there - leftover from a refactor. Removed.

test/phpunit/csvImportValidator/CsvBaseValidatorTest.php
sbreker: True... removed.

test/phpunit/csvImportValidator/CsvBaseValidatorTest.php
sbreker: 👍  Removed.

test/phpunit/csvImportValidator/CsvValidatorResultTest.php
sbreker: 👍

test/phpunit/csvImportValidator/CsvValidatorResultTest.php
sbreker: Unit test updated.

test/phpunit/csvImportValidator/CsvValidatorResultTest.php
sbreker: Fixed.

test/phpunit/csvImportValidator/CsvValidatorResultTest.php
sbreker: Done.

test/phpunit/csvImportValidator/CsvValidatorResultTest.php
sbreker: Done

test/phpunit/csvImportValidator/CsvValidatorResultTest.php
sbreker: Yes, definitely

lib/task/import/csvCheckImportTask.class.php
sbreker: Nice - fixed.

lib/task/import/csvCheckImportTask.class.php
sbreker: Removed.

lib/task/import/validate/csvBaseValidator.class.php
sbreker: This code was removed completely during a refactor to remove the unused 'verbose' option from the validations.

lib/task/import/validate/csvColumnNameValidator.class.php
sbreker: The CSV data is not available when the constructor is run.

lib/task/import/validate/csvColumnNameValidator.class.php
sbreker: Both suggestions implemented.

lib/task/import/validate/csvCultureValidator.class.php
sbreker: 👍

lib/task/import/validate/csvDigitalObjectUriValidator.class.php
sbreker: 👍

lib/task/import/validate/csvFieldLengthValidator.php
sbreker: Updated to use 11.

lib/CsvImportValidator.class.php
sbreker: I have added a commit containing a bit of a refactor - I didn't really like the way the logic for building and maintaining the list of validators was spread all over inside of CsvImportValidator class. I have collected this logic into a new CsvValidatorCollection that populates itself with the correct tests given the CSV entity type specified.

The Switch statement you commented on here is gone now, since the validators themselves are responsible for declaring which entities they are limited to, if any.

lib/task/import/validate/csvColumnNameValidator.class.php
sbreker: The logic that I've used to parse these files will fallback to assembling the valid columns from other sections in this file. As it is, all existing entity validation files do contain a 'columnNames' section. I have added a new commit that will throw an exception if the yml file for the entity in question is missing entirely.

sbreker:
Hi @djjuhasz ! I've reviewed your and @mcantelon 's comments and I have added a few commits to address everything. The biggest commit creates a new CsvValidatorCollection class which will assemble and contain the correct validator classes for the selected entity type - this allowed me to delete the switch() statement being used to select the correct validators. I think it's all ready for your review once more.

lib/task/import/validate/csvColumnNameValidator.class.php
```diff
@@ -23,17 +23,15 @@
...
  */
 class CsvColumnNameValidator extends CsvBaseValidator
 {
     const TITLE = 'Column Name Validation';
+
     // Do not reset in between multiple CSVs.
     protected $validColumnNames = [];
     protected $validColumnNamesLowercase = [];
-
+    // Reset between files.
```
djjuhasz, line 34: Helpful comment 👍

lib/task/import/csvCheckImportTask.class.php
```diff
@@ -71,14 +71,15 @@ protected function configure()
                 sfCommandOption::PARAMETER_OPTIONAL, 'The application name', 'qubit'
             ),
             new sfCommandOption('env', null, sfCommandOption::PARAMETER_REQUIRED,
```
djjuhasz: Although it takes a lot of vertical space, I think it's better to readability and consistency to put each argument on it's own line when they don't fit on a single line, e.g.

```
new sfCommandOption(
    'env', 
    null,
    sfCommandOption::PARAMETER_REQUIRED,
    'The environment', 
    'cli'
),
```

lib/task/import/validate/csvColumnNameValidator.class.php
```diff
@@ -65,26 +65,31 @@ public function testRow(array $header, array $row)
...
+            $this->testColumnName($columnName);
         }
 
         $this->complete = true;
     }
 
+    protected function testColumnName(string $columnName)
+    {
+        // If $columnName is not in validColumnNames.
+        if (!in_array($columnName, $this->validColumnNames)) {
```
djjuhasz: I'd avoid nested if statements with a return:

```
if (in_array($columnName, $this->validColumnNames)) {
    return;
}
```

lib/task/import/validate/csvValidatorCollection.php
```diff
@@ -0,0 +1,141 @@
...
+     * the correct Validator objects for the specified $className.
+     *
+     * @param string $className             The entity name. e.g. QubitInformationObject
+     * @param array  $options               the options array passed to the validator
+     * @param string $specificValidatorList comma separated string of validator class names
+     */
+    public static function getValidatorCollection(string $className, array &$options, ?string $validatorList = null): CsvValidatorCollection
+    {
+        // Map of available validator classes.
+        $validatorsAvailable = [
```
djjuhasz: I think it would be better to define `$validatorsAvailable` as a (static?) class property to make it more prominent, and to make the list accessible outside this method.  An example use might be to return the list of available validators for output to the command line.

lib/task/import/validate/csvValidatorCollection.php
```diff
@@ -0,0 +1,141 @@
...
+            'CsvParentValidator' => CsvParentValidator::class,
+            'CsvLegacyIdValidator' => CsvLegacyIdValidator::class,
+            'CsvEventValuesValidator' => CsvEventValuesValidator::class,
+            'CsvDigitalObjectPathValidator' => CsvDigitalObjectPathValidator::class,
+            'CsvDigitalObjectUriValidator' => CsvDigitalObjectUriValidator::class,
+        ];
+
+        // Remove tests that are not applicable to this $className.
+        foreach ($validatorsAvailable as $key => $class) {
+            if (!in_array($className, $class::LIMIT_TO) && !empty($class::LIMIT_TO)) {
```
djjuhasz: I think  `if (!empty($class::LIMIT_TO) && !in_array($className, $class::LIMIT_TO) {` would be better to avoid a warning when `$class::LIMIT_TO` is not set.

lib/task/import/validate/csvValidatorCollection.php
```diff
@@ -0,0 +1,141 @@
...
+                unset($validatorsAvailable[$key]);
+            }
+        }
+
+        // If specific tests not used, use default validators for this $className.
+        if (!isset($validatorList)) {
```
djjuhasz: I think `if (empty($validatorList))` might be better here, because it evaluates true when `$validatorList` is an empty string.

lib/task/import/validate/csvColumnNameValidator.class.php
```diff
@@ -123,29 +123,34 @@ protected function testColumnName(string $columnName)
...
+                // Add from property map
+                if (null !== $propertyMap) {
+                    $this->validColumnNames = array_merge($this->validColumnNames, array_values($propertyMap));
+                }
             }
-        }
 
-        $this->validColumnNamesLowercase = array_map('strtolower', $this->validColumnNames);
+            $this->validColumnNamesLowercase = array_map('strtolower', $this->validColumnNames);
+        } catch (Exception $e) {
```
djjuhasz: If this is supposed to catch exceptions from `QubitFlatfileExport::loadResourceConfigFile()` then I think it's better to put the catch right after line 128 to avoid catching unrelated exceptions from the rest of the intervening code.

djjuhasz (changes requested):
@sbreker looking good! I just suggested a few more small changes.

lib/task/import/validate/csvColumnNameValidator.class.php
sbreker: 👍

lib/task/import/csvCheckImportTask.class.php
sbreker: Sounds good. It does read easier - change completed.

lib/task/import/validate/csvColumnNameValidator.class.php
sbreker: This change was already made. csvColumnNameValidator::testColumnName() no longer has this return statement.

lib/task/import/validate/csvValidatorCollection.php
sbreker: This makes sense. I like having this map up at the top visible. Change completed.

lib/task/import/validate/csvValidatorCollection.php
sbreker: It makes sense for that check to be in the reverse order. I will reverse these conditions.

That said, `$class::LIMIT_TO` can't be not set because it is assigned an empty array in `csvBaseValidator`.

lib/task/import/validate/csvValidatorCollection.php
sbreker: Yes -  `empty()` is better and handles this edge case more cleanly. Changed in next commit.

lib/task/import/validate/csvColumnNameValidator.class.php
sbreker: Agreed - will change.

sbreker:
Hi @djjuhasz - I've added one more commit with couple changes addressing your latest feedback. Ready for your review.

lib/task/import/validate/csvValidatorCollection.php
```diff
@@ -80,39 +98,21 @@ public function getResultCollection(): CsvValidatorResultCollection
...
-            'CsvDigitalObjectUriValidator' => CsvDigitalObjectUriValidator::class,
-        ];
-
         // Remove tests that are not applicable to this $className.
-        foreach ($validatorsAvailable as $key => $class) {
-            if (!in_array($className, $class::LIMIT_TO) && !empty($class::LIMIT_TO)) {
-                unset($validatorsAvailable[$key]);
+        foreach (self::$validatorsAvailable as $key => $class) {
+            if (!empty($class::LIMIT_TO) && !in_array($className, $class::LIMIT_TO)) {
+                unset(self::$validatorsAvailable[$key]);
```
djjuhasz: Because `self::$validatorsAvailable` is a static property, I think this will remove the validator from `self::$validatorsAvailable` for ALL instances of `CsvValidatorCollection`.  I think you'll want to create a local working copy of `self::$validatorsAvailable` to modify.

djjuhasz (changes requested):
Hi @sbreker one more change request.  Almost done. :)

lib/task/import/validate/csvValidatorCollection.php
sbreker: Thanks @djjuhasz - Added commit that uses a working copy of the validator list. Also added a couple basic PHPUnit tests for the CsvValidatorCollection class.

djjuhasz (approved):
Looks good to me @sbreker 🎉 :shipit:

PR #1275, opened 2021-03-28
CSV transform text normalization, refs #13494

mcantelon:
Added optional normalization of Windows encoded data in CSV transformation
class. Also added ability to preserve CSV sort order (rather than sorting
data from CSV by level of description).


mcantelon:
Need to rebase

PR #1276, opened 2021-03-28
Various relation dialog fixes and improvements, refs #10682 #13218 #13425 #13499

mInnes-archives:
This is a small bug fix addressing 500 errors being thrown after attempting to save various types of records (eg. ISAAR, ISDF, Accession, etc.) containing incomplete relationships. For example, the related entity is unspecified so the relationship is incomplete:

emptyRelatedEntity

To address this, an additional string validator has been added requiring that the relevant value be set. The incomplete relationship can still be staged as part of the form, like in the image above, but this relationship entry will now be ignored when saving. No error is thrown and all other form values and relationships will save as expected.

### DIFF
--- a/js/dialog.js
+++ b/js/dialog.js
@@ -108,8 +108,6 @@
         // Submit dialog data
         var handleYuiSubmit = function ()
           {
-            this.hide(); // Hide dialog
-
             var data = this.getData();
 
             // The validator must return false to cancel the operation
@@ -119,11 +117,18 @@
               return false;
             };
 
+            this.hide(); // Hide dialog
+
             thisDialog.submit(data); // Save dialog data
           }
 
         var handleYuiCancel = function ()
           {
+            // Optionally execute logic upon canceling dialog
+            if ('undefined' !== typeof thisDialog.options.afterCancelLogic) {
+              thisDialog.options.afterCancelLogic.call(this);
+            }
+
             this.cancel(); // Cancel YUI submit
             thisDialog.clear(); // Clear dialog fields
           }
@@ -345,6 +350,7 @@
             {
               if (null == thisData[fieldname])
               {
+                thisData[fieldname] = '';
                 continue;
               }
 
@@ -382,7 +388,17 @@
                   dataSource.responseType = YAHOO.util.DataSourceBase.TYPE_TEXT;
                   dataSource.parseTextData = function (request, response)
                     {
-                      return { 'results' : [ $(response).find('.label, #main-column h1').text().trim() ] };
+                      var $response = $(response);
+
+                      // For pages decorated with 'layout_2col' or 'layout_3col'
+                      var $result = $response.find('#main-column h1');
+
+                      // Fallback for pages decorated with 'layout_1col'
+                      if (0 == $result.length) {
+                        $result = $response.find("#resource-name");
+                      }
+
+                      return {'results': [$result.first().text().trim()]};
                     };
 
                   // Set visible input field of yui-autocomplete
--- a/plugins/sfIsaarPlugin/modules/sfIsaarPlugin/actions/relatedAuthorityRecordComponent.class.php
+++ b/plugins/sfIsaarPlugin/modules/sfIsaarPlugin/actions/relatedAuthorityRecordComponent.class.php
@@ -93,7 +93,7 @@ class sfIsaarPluginRelatedAuthorityRecordComponent extends RelationEditComponent
                 $value = $this->form->getValue('resource');
                 if (isset($value)) {
                     $params = $this->context->routing->parse(Qubit::pathInfo($value));
-                    if ($this->resource->id != $this->relation->objectId) {
+                    if (!isset($this->resource->id) || $this->resource->id != $this->relation->objectId) {
                         $this->relation->object = $params['_sf_route']->resource;
                     } else {
                         $this->relation->subject = $params['_sf_route']->resource;
--- a/plugins/sfIsdfPlugin/modules/sfIsdfPlugin/templates/_relatedFunction.php
+++ b/plugins/sfIsdfPlugin/modules/sfIsdfPlugin/templates/_relatedFunction.php
@@ -39,7 +39,7 @@
             <?php echo render_value_inline($item->description); ?>
           </td><td>
             <?php echo render_value_inline(Qubit::renderDateStartEnd($item->date, $item->startDate, $item->endDate)); ?>
-          </td><td style="text-align: center">
+          </td><td style="text-align: right">
             <input class="multiDelete" name="deleteRelations[]" type="checkbox" value="<?php echo url_for([$item, 'module' => 'relation']); ?>"/>
           </td>
         </tr>
@@ -91,11 +91,33 @@ var handleFieldRender = function (fname)
 Drupal.behaviors.relatedFunction = {
   attach: function (context)
     {
+      // Add validator to ensure a related function is selected
+      var validator = function(data) {
+          var relation = data["relatedFunction[resource]"];
+
+          if (!relation.length) {
+            // Display error message
+            jQuery('#relatedFunctionError').css('display', 'block');
+
+            return false;
+          } else {
+            // Hide error message until required again
+            jQuery('#relatedFunctionError').css('display', 'none');
+          }
+      }
+
+      // Hide error on cancel
+      var afterCancelLogic = function () {
+          jQuery('#relatedFunctionError').css('display', 'none');
+      }
+
       // Define dialog
       var dialog = new QubitDialog('functionRelation', {
         'displayTable': 'relatedFunctions',
         'newRowTemplate': {$rowTemplate},
         'handleFieldRender': handleFieldRender,
+        'validator': validator,
+        'afterCancelLogic': afterCancelLogic,
         'relationTableMap': function (response)
           {
             response.resource = response.object;
@@ -126,6 +148,12 @@ content
     <h3><?php echo __('Related function'); ?></h3>
 
     <div>
+
+      <div class="messages error" id="relatedFunctionError" style="display: none">
+        <ul>
+          <li><?php echo __('Please complete all required fields.'); ?></li>
+        </ul>
+      </div>
 
       <div class="form-item">
         <?php echo $form->resource
--- a/plugins/sfIsdfPlugin/modules/sfIsdfPlugin/templates/indexSuccess.php
+++ b/plugins/sfIsdfPlugin/modules/sfIsdfPlugin/templates/indexSuccess.php
@@ -1,13 +1,10 @@
 <?php decorate_with('layout_1col'); ?>
 
 <?php slot('title'); ?>
-  <h1 class="multiline">
-    <?php echo __('View ISDF function'); ?>
-    <span class="sub"><?php echo render_title($resource); ?></span>
-  </h1>
-<?php end_slot(); ?>
 
-<?php slot('before-content'); ?>
+  <h1 id="resource-name">
+    <?php echo render_title($resource); ?>
+  </h1>
 
   <?php if (isset($errorSchema)) { ?>
     <div class="messages error">
@@ -19,6 +16,17 @@
       </ul>
     </div>
   <?php } ?>
+
+  <section class="breadcrumb">
+    <ul>
+      <li><?php echo link_to(esc_specialchars(sfConfig::get('app_ui_label_function')), ['module' => 'function', 'action' => 'browse']); ?></li>
+      <li><span><?php echo render_title($resource); ?></span></li>
+    </ul>
+  </section>
+
+<?php end_slot(); ?>
+
+<?php slot('before-content'); ?>
 
   <?php echo get_component('default', 'translationLinks', ['resource' => $resource]); ?>
 
--- a/apps/qubit/modules/contactinformation/templates/_edit.php
+++ b/apps/qubit/modules/contactinformation/templates/_edit.php
@@ -135,7 +135,7 @@ content
         <div id="contactInformationRelation_Tab1">
 
           <?php echo $form->primaryContact
-              ->label(__('Primary contract'))
+              ->label(__('Primary contact'))
               ->renderRow(); ?>
 
           <?php echo $form->contactPerson
--- a/apps/qubit/modules/relation/actions/editComponent.class.php
+++ b/apps/qubit/modules/relation/actions/editComponent.class.php
@@ -113,9 +113,10 @@ class RelationEditComponent extends sfComponent
                 break;
 
             case 'resource':
-                $this->form->setValidator('resource', new QubitValidatorForbiddenValues([
-                    'forbidden_values' => [$this->context->routing->generate(null, $this->resource)],
-                ]));
+                $this->form->setValidator('resource', new sfValidatorAnd([
+                    new sfValidatorString(),
+                    new QubitValidatorForbiddenValues(['forbidden_values' => [$this->context->routing->generate(null, $this->resource)]]),
+                ], ['required' => true]));
                 $this->form->setWidget('resource', new sfWidgetFormSelect(['choices' => []]));
 
                 break;
--- a/js/autocomplete.js
+++ b/js/autocomplete.js
@@ -5,7 +5,7 @@
         // Share <form/> with nested scopes
         var $form = $(this);
 
-        // Keep a list of new releated resource forms that must be submitted
+        // Keep a list of new related resource forms that must be submitted
         // before $form so we can get the URIs after they are created
         var relatedResourceForms = [];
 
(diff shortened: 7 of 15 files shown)


mInnes-archives:
Noticed I needed to rebase and make some minor syntax fixes. I may have caused Syntax checks to bug out with all the force-pushes somehow. :sweat:

jraddaoui:
Hi @mInnes-archives,

Thanks for reporting the issue and submitting a PR. I see you also rebased your branch after the latest changes in qa/2.x, which added the syntax checks I mentioned you in your previous PR. We just announced this change in our coding standards in the forum, where you can find some information:

https://groups.google.com/g/ica-atom-users/c/edNrTGACuv4/m/-pPIhMYoCgAJ

As you can see in the failing check details, it's only complaining about indentation, so it should be an easy fix.

jraddaoui:
About the actual issue and changes, I believe @fiver-watson had a question about it and I'd love to see some sort of notification about the relation not being saved, but I'm not sure how hard would it be in this multi-row component form and it's probably needed in other places. So it looks good to me.

mInnes-archives:
> As you can see in the failing check details, it's only complaining about indentation, so it should be an easy fix.

Okay, now that I'm reading the details correctly, I see that. I'll get the plug-in installed to avoid this in future.

> About the actual issue and changes, I believe @fiver-watson had a question about it and I'd love to see some sort of notification about the relation not being saved, but I'm not sure how hard would it be in this multi-row component form and it's probably needed in other places. So it looks good to me.

Agreed. that would be ideal. This is somewhat of a quick fix for now.

jraddaoui:
> Okay, now that I'm reading the details correctly, I see that. I'll get the plug-in install to avoid this in future.

I see, the details are not very clear. I've tried to improve this in https://github.com/artefactual/atom/pull/1278, using a shorter diff format.

fiver-watson:
Hi @mInnes-archives! Thanks again for your continued contributions to the project! 

I just wanted to clarify the behavior on this. If I'm understanding correctly: 

* Adding an incomplete relationship currently throws a 500 error, which means that any other data entered into the edit page risks being lost
* This proposed fix will make these relationships fail silently - they will be ignored, allowing the page to be saved, and no error or warning will be thrown
* As a consequence, however, the user has no indication that some of the data they entered (even though incomplete) has not been saved

Is that correct?

If so, then while I see the motivation for this (if you spend a lot of time working on a description, it's frustrating to lose all that work because of one bad relationship element), but I also worry about not letting the user know that some data has been dropped. 

AtoM does have a warning notification that can be displayed to authenticated users on view pages following a save - we use it for post-submission validation (AtoM's always had a very permissive approach - you can technically save a completely blank record and it will work). For example, you'll see one of the warnings if lower-level dates are incompatible with those at higher levels, or if one of the elements the related standard considers mandatory are not filled in (for example, leaving the identifier, title, or extent field blank). 

As Radda has mentioned, I think it would be ideal to use one of these to inform the user about what's happened. I don't know if you have the time or inclination to add this - if so, I'm hoping that @jraddaoui can provide an implementation example from elsewhere in the code as a reference, and we might make this a valuable addition to the next major release. 

If not, I'm still inclined for us to accept the PR - I agree that this is probably better than potentially losing a lot of other user data when throwing the error. In that case, I will file a new issue and we'll try to revisit this and add the warning ourselves in the future, time allowing. 

Thoughts?

jraddaoui:
Here an example of a flash alert ...

https://github.com/artefactual/atom/blob/qa/2.x/apps/qubit/modules/informationobject/actions/updatePublicationStatusAction.class.php#L50-L51

I believe "notice" is a yellow notification, "error" is red and "info" is blue.

mInnes-archives:
@fiver-watson 
> I don't know if you have the time or inclination to add this

I'd be happy to dig a bit deeper and find an alternative solution which could provide better feedback for users. I'm just doing this in my spare time and I'm really just learning things as I go, so I'll see what I can find out over the next week or so.

@jraddaoui 
>Here an example of a flash alert ...

Thanks. :)

mInnes-archives:
### Cause of 500 errors

After digging around I found out the cause of the original 500 errors. Attempting to add incomplete relationships to the database entails trying to add `null` values to `NOT NULL` columns. To explore a fully permissive approach to this problem I tested the effects of changing this constraint but found that this ended up causing way more problems. Therefore, stopping these data from being submitted remains the best approach.

### Augmented solution

To augment my original solution I've added additional validation and error messaging as part of the relevant YUI dialogs. This should add sufficient feedback for users and halt the creation of these relationships. This solution has been applied to the following dialogs thus far:

- ISAAR - Related authority records
- ISAAR - Related resources
- ISDF - Related authority records
- ISDF - Related resources

This solution has yet to be added to the following due to the bugs outlined in the next section:

- Accession records - Related donors
- ISDF - Related functions

### Additional issues

I've identified some additional issues in doing this work. Some I've already addressed, such as typos or alignment issues, but the following remain:

- Dialogs which allow for on-the-fly record creation for new related entities causes both the new record to be created (as desired) as well an additional 'Untitled' record. The new dialog validator also seems to cause some problems when dealing with these newly created entities. Affects the following:
  - 'Accession - Related Donors'
  - Actor events for archival descriptions, but note that the 500 error outlined above isn't relevant for this. 

- Clicking the edit icon for 'ISDF - Related functions' relationships ~~when the current record being edited is the object of the relationship~~ seems to result in a failure to pull up some field values in the dialog.

I'd like to try my hand at addressing these. so these fixes should be forthcoming.

### Images
ErrorExample
*Error example from ISAAR - Related resources*

jraddaoui:
Thanks for the update and fixes @mInnes-archives, it's looking great so far!

mInnes-archives:
Issues related to the 'ISDF - Related functions' dialog are resolved. The result returned from the Ajax response didn't accommodate the layout type used for ISDF records, so the name wasn't able to be retrieved. Having to rely on Ajax requests isn't great, but it works now.

I've also modified the title display formatting when viewing ISDF records to be more appropriate for public-facing records, similar to other record types.

js/dialog.js
```diff
@@ -382,7 +387,15 @@
...
                     {
-                      return { 'results' : [ $(response).find('.label, #main-column h1').text().trim() ] };
+                      // For pages decorated with 'layout_2col' or 'layout_3col'
+                      var multiColumnResult = $(response).find('#main-column h1').text().trim();
+
+                      if (0 != multiColumnResult.length) {
+                        return { 'results' : [multiColumnResult] };
+                      } else {
+                        // Fallback for pages decorated with 'layout_1col'
+                        return { 'results' : [ $(response).find('h1').text().trim() ] };
```
jraddaoui: To avoid unexpected `h1` in some customized headers I'd use `#wrapper h1` as selector. I'd also use `.first()` before getting the text to make sure we have only the first one. And, a suggestion to avoid parsing the response twice into a jQuery var and some (really minor) duplication ...

```javascript
var $response = $(response);

// For pages decorated with 'layout_2col' or 'layout_3col'
var $multiColumnResult = $response.find('#main-column h1');

// Fallback for pages decorated with 'layout_1col'
if (0 == $multiColumnResult.length) {
  $multiColumnResult = $response.find('#wrapper h1');
}

return {'results': [$multiColumnResult.first().text().trim()]};
```

plugins/sfIsdfPlugin/modules/sfIsdfPlugin/templates/_relatedFunction.php
```diff
@@ -127,6 +149,12 @@
 
     <div>
 
+     <div class="messages error" id="relatedFunctionError" style="display: none">
```
jraddaoui: Nitpick: indentation missing one space.

plugins/sfIsdfPlugin/modules/sfIsdfPlugin/templates/indexSuccess.php
```diff
@@ -20,6 +17,17 @@
     </div>
   <?php } ?>
 
+  <section class="breadcrumb">
+    <ul>
+    <li><?php echo link_to(esc_specialchars(sfConfig::get('app_ui_label_function')), ['module' => 'function', 'action' => 'browse']); ?></li>
```
jraddaoui: Nitpick: indentation.

plugins/sfIsdfPlugin/modules/sfIsdfPlugin/templates/indexSuccess.php
```diff
@@ -20,6 +17,17 @@
...
+      <li><span><?php echo render_title($resource); ?></span></li>
+    </ul>
+  </section>
+
+<?php end_slot(); ?>
+
+<?php slot('before-content'); ?>
+
```
jraddaoui, line 30: Apart from helping with the implementation I think this is a good change, but I'd like to check with @fiver-watson ...

Basically, it removes the "View ISDF function" from the function's index page heading and adds a breadcrumb below the function title. This breadcrumb uses the UI label for functions and links to the function's browse page.

jraddaoui (approved):
Thanks @mInnes-archives!

I've added some comments below but it looks great to me. I'm not sure if we have a ticket number already and I've pinged @fiver-watson to check the function heading change, but it should be almost ready to squash and merge.

plugins/sfIsdfPlugin/modules/sfIsdfPlugin/templates/indexSuccess.php
fiver-watson: Looks good to me - thanks for noticing and addressing this, @mInnes-archives ! The functions module has been a bit neglected, as it is not frequently used in most North American archival practices (or some European and others - it's most often used in the Australian series system). However, I agree that this is a good change, that brings the page layout more in line with how other entities are presented. Thanks again!

js/dialog.js
mInnes-archives: I tried using the selector `#wrapper h1` prior to making the commit, but it doesn't seem to want to return anything. It still doesn't, and I have no idea why.

Alternatively, we could just add something like `id='resource-name'` to the `<h1>` we are targeting and use `#resource-name` as the selector.

jraddaoui: Right, I faced that issue a while ago and couldn't find the reason at first glance. Adding an id to the header sounds good @mInnes-archives, it may be needed later one in other pages, but we'll figure out. Thanks!

mInnes-archives:
Two bugs still need to be addressed.

Mentioned above:
> Dialogs which allow for on-the-fly record creation for new related entities causes both the new record to be created (as desired) as well an additional 'Untitled' record. The new dialog validator also seems to cause some problems when dealing with these newly created entities. Affects the following:
>
> - 'Accession - Related Donors'
> - Actor events for archival descriptions, but note that the 500 error outlined above isn't relevant for this

And:
 
When trying to edit an existing relation (one already in the database), the following series of actions results in the original `null` value of empty fields being changed to a string containing the word `'null'`:

1. click the relation you want edit
2. dialog pops up
3. cancel the dialog
4. save record

Result:
nullBug


The 'null' strings can be removed easily by editing the relation again. This appears to affect most relation dialogs.

I can address these over the next little bit.

jraddaoui:
Thanks @mInnes-archives.

We have created https://projects.artefactual.com/issues/13499 to reference in the final commit.

mInnes-archives:
Since my last comment, the following issues have been addressed:

### Issue 1
(as above)
>When trying to edit an existing relation (one already in the database), the following series of actions results in the original null value of empty fields being changed to a string containing the word 'null':
>
>1. click the relation you want edit
>2. dialog pops up
>3. cancel the dialog
>4. save record

Solution: 0224883993de95101589c389c55bb784c5af8e4e

### Issue 2
(modified from above)
Dialogs with autocomplete fields which allow for the creation of new related resources causes all new records (n>=1) to be created (as desired) as well an additional 'Untitled' record. This appears to be the result of both dialog.js and autocomplete.js trying to add these new related resources, with dialog.js being the only one to do so correctly. I have added an exception in autocomplete.js for the three instances where this behaviour occurs. I couldn't devise a more generalized solution to this issue, so should an autocomplete field with add functionality find its way into a different YUI, this case would need to be added.

Solution: d585b14a285baa04fed8ea6e77287ed076de7762

### Issue 3
Related to Issue 2 above. With both dialog.js and autocomplete.js firing on the same data, the relation trying to be added to the database would sometimes be missing the object, the newly created related resource, resulting in a 500 error from a `NOT NULL` constraint. 

Solution: d585b14a285baa04fed8ea6e77287ed076de7762 (same as Issue 2)

All other commits should be fairly straightforward. I think this covers all of the YUI dialog related bugs that I've found. I'd like a final review and then I'll prepare things for merge.

jraddaoui:
Thanks @mInnes-archives, this is looking great and almost ready for merge!

Just one question about the autocomplete exceptions ... do you think it would be feasible to check for the input ancestor elements and stop if there is a `div.yui-dialog`? If that's possible it could prevent having to add new dialog autocomplete inputs in there.

mInnes-archives:
> Just one question about the autocomplete exceptions ... do you think it would be feasible to check for the input ancestor elements and stop if there is a `div.yui-dialog`? If that's possible it could prevent having to add new dialog autocomplete inputs in there.

``` javascript
inputAncestor = $input.parent()[0].className;
if (inputAncestor != "form-item yui-ac") {
  $relatedResourceForm = getRelatedResourceForm($input, $hidden, uri, rrFormInputId);
}
```

This seems to work. Are there any issues this might cause elsewhere?

jraddaoui:
I'm not sure that will work, `yui-ac` may refer to the autocomplete, so it would be present in all of them. Maybe we could use ...

```js
if ($input.parents('div.yui-dialog').length == 0) {
  $relatedResourceForm = getRelatedResourceForm($input, $hidden, uri, rrFormInputId);
}
```

Or something like that.

mInnes-archives:
Okay, I did a bit of testing and found some instances where my solution creates problems. Selecting `yui-dialog` is the way to go.

mInnes-archives:
While working on this I found another issue that may be related to dialog/autocomplete. I did a check on the main branch and found the same problem, so it doesn't look like I caused this with anything I've done here.

### Issue 4
When trying to create relationships between the following you receive a 500 error:

- New function [function/add] -> function
- New authority record [actor/add] -> authority record

Establishing these relationships from an existing record [actorSlug/edit; functionSlug/edit] works fine. The fact that these are relationships between the same record types might point to the cause.

I'll look into this since I'd like to fix up all the relation issues in one go, if possible.

mInnes-archives:
Works now. New records don't have IDs at this point.

mInnes-archives:
From what I can tell, this PR addresses the following open issues:

- Cannot create new authority record and add relationship at the same time
  - https://projects.artefactual.com/issues/10682
  - **[Issue 4 above]**
- Error when adding function related to another
  - https://projects.artefactual.com/issues/13218
  - **[Issue 4 above]**
- Adding multiple event actors on the fly in the RAD template can cause the link to the authority record to be lost on save
  - https://projects.artefactual.com/issues/13425
  - **[Issue 3 above]**

plugins/sfIsaarPlugin/modules/sfIsaarPlugin/actions/relatedAuthorityRecordComponent.class.php
```diff
@@ -93,7 +93,7 @@ protected function processField($field)
                 $value = $this->form->getValue('resource');
                 if (isset($value)) {
                     $params = $this->context->routing->parse(Qubit::pathInfo($value));
-                    if ($this->resource->id != $this->relation->objectId) {
+                    if ($this->resource->id != $this->relation->objectId || $this->resource->id == null) {
```
jraddaoui: Not a big deal, but I think the following is clearer ...

```php
if (!isset($this->resource->id) || $this->resource->id != $this->relation->objectId) {
```

Not in this case, but usually checking the existency before comparing may prevent some issues.

jraddaoui (approved):
Thanks a lot @mInnes-archives, those are great news!

Added a minor comment below, but feel free to ignore it. Once you clean-up and get this ready for merge, if you're able to reference the three tickets in the commit subject that would be great, just `, refs #12345 #12345 #12345`. If it's a pain to make that fit, just reference the ticket we created for this in the subject and the other two in the body.

jraddaoui:
Great one @mInnes-archives! Thanks again!

adaxem:
Hi all,

Can you please confirm in which version Issue 2 below has been addressed for new Donors created while editing an Accession record? Has the solution been tested in multilingual instances where English is not the main language?

> ### Issue 2
> 
> (modified from above)
> 
> Dialogs with autocomplete fields which allow for the creation of new related resources causes all new records (n>=1) to be created (as desired) as well an additional 'Untitled' record. This appears to be the result of both dialog.js and autocomplete.js trying to add these new related resources, with dialog.js being the only one to do so correctly. I have added an exception in autocomplete.js for the three instances where this behaviour occurs. I couldn't devise a more generalized solution to this issue, so should an autocomplete field with add functionality find its way into a different YUI, this case would need to be added.
> 
> 
> 
> Solution: d585b14a285baa04fed8ea6e77287ed076de7762
>

anvit:
Hi @adaxem! These fixes were included in [AtoM v2.7](https://wiki.accesstomemory.org/wiki/Releases/Release_announcements/Release_2.7). As for multilingual instances where English isn't the primary language, we haven't personally tested this fix on one, but definitely let us know if you have any updates or open a new issue if you find out that there is a problem.

PR #1282, opened 2021-04-19
Fix video URL deriv creation, refs #13497

sbreker:
Derivatives for videos linked to a description using a URL were not
being created. This commit fixes this by correcting the temp path that
is passed to ffmpeg. In addition, code has been added to save the
derivative assets.

Added option '-y' to ffmpeg so that the 'confirm overwrite' message is
bypassed.

Previously linked videos will have their derivatives created once this
fix is deployed and the regen-derivs CLI task is run.

### DIFF
--- a/lib/model/QubitDigitalObject.php
+++ b/lib/model/QubitDigitalObject.php
@@ -2251,7 +2251,7 @@ class QubitDigitalObject extends BaseDigitalObject
         }
 
         $extension = '.'.self::THUMB_EXTENSION;
-        list($originalNameNoExtension) = explode('.', $this->getName());
+        $originalNameNoExtension = pathinfo($this->getName(), PATHINFO_FILENAME);
         $derivativeName = $originalNameNoExtension.'_'.$usageId.$extension;
 
         // Resize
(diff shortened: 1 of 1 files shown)


lib/model/QubitDigitalObject.php
```diff
@@ -2617,47 +2617,24 @@ public static function convertAudioToMp3($originalPath, $newPath)
     public function createVideoDerivative($usageId, $connection = null)
     {
         // Build new filename and path
-        $originalFullPath = $this->getAbsolutePath();
-        list($originalNameNoExtension) = explode('.', $this->getName());
```
mcantelon, line 2621: Will this work for files with weird filenames where there's a period in a name?

Like "bob.comedy.mp4".

Seems like this might not matter though as the filename's going to be an internal filename anyways.

mcantelon (approved):
I had one question, but otherwise it seems good!

lib/model/QubitDigitalObject.php
sbreker: Good point!  This logic does cause the derivative filenames to be truncated (Your eg would become bob_141.mp4).  This same logic is used for audio files and images currently as well. It should be pretty easy to update to use PHP's pathinfo() to get the filename in each circumstance.

mcantelon: Nice!

PR #1283, opened 2021-04-23
Fix repo upload limit form validation, refs #13450

jraddaoui:
Form validation requires something to be binded to succeed which was not
happening whith CSRF protection disabled. Also, separate validation
from authentication and return diffrent response codes.


djjuhasz (approved):
Looks good to me! 👍

jraddaoui:
Thanks @djjuhasz!

PR #1284, opened 2021-04-24
Fix missing finding aid logo, refs #13458

djjuhasz:
- Fix invalid FO XSL template elements
- Add "{{ app_root }}" template variables to XSL files
- Add "{{ app_root }}" to helper-functions.xsl href
- Replace template vars with AtoM application path at run time


mcantelon (approved):
One line in one of the commit messages has a typo (maybe "in" rather that "to">): `"Use {{ app_root }} to helper-functions.xsl href`.

Otherwise looks good to me! The `app_root` tag it a nice touch.

PR #1285, opened 2021-04-28
Remove duplicate csv col, refs #13502

fiver-watson:
The ISAD CSV example template in `lib/task/import/example/isad` contains a duplicate `accessionNumber` column. This PR removes the unnecessary column.


jraddaoui (approved):
:ok_hand:

PR #1286, opened 2021-04-29
Fix full treeview icons, refs #13501

mcantelon:
Fixed issue with full treeview that prevented the correct icons from being displayed.


djjuhasz (approved):
Looks good to me 👍

PR #1287, opened 2021-05-01
Fix D.O. load task indexing issue, refs #13484

mcantelon:
Fixed issue with the digital object load task not properly updating the search index. The issue was due to the sfConfig settings cache not being populated and, subsequently, the cached i18n_languages setting being empty during indexing and causing i18n data not to be indexed.


jraddaoui (approved):
Thanks @mcantelon!

PR #1288, opened 2021-05-02
Remove 404 if no autocomplete hits, refs #13469

mcantelon:
Removed 404 error generation when a global search autocomplete attempt doesn't return any matches.


jraddaoui (approved):
Thanks @mcantelon!

PR #1289, opened 2021-05-02
Add ISAD pub. notes to ES index, refs #13301

mcantelon:
Added Elasticsearch indexing of publication notes in ISAD descriptions.


jraddaoui (approved):
Thanks @mcantelon!

PR #1290, opened 2021-05-04
Schema version bump, refs #13490

mcantelon:
A migration was created in an earlier commit but the schema wasn't accordingly bumped.


jraddaoui (approved):
Thx!

PR #1291, opened 2021-05-04
Improve job worker debug output, refs #13504

mcantelon:
Changed job worker error handling so it'll handle PHP errors as well as exceptions.

Added logging of file and line number in which the error or exception occurs.


jraddaoui (approved):
Nice! Thanks @mcantelon!

PR #1292, opened 2021-05-06
Replace web installer by CLI task, refs #13452

jraddaoui:
- Remove sfInstallPlugin and move sfInstall class to arInstall.
- Tweak arInstall to throw exceptions instead of returning errors.
- Move DB configuration handler class into arInstall.
- Remove install layout and related styles.
- Create install task to configure and initialize a new instance.
- Normalize purge and install tasks.

Other fixes:

- Fix QubitActor index on save check.
- Allow Propel's default DB name change.
- Install apcu extension in Docker.
- Avoid DB exceptions on routing initialization.
- Avoid PHP notice.


jraddaoui:
More details in:

https://github.com/artefactual/atom-docs/blob/2.7/admin-manual/installation/linux/ubuntu.rst#run-the-installer
https://github.com/artefactual/atom-docs/blob/2.7/admin-manual/maintenance/cli-tools.rst#installer

lib/model/QubitObject.php
```diff
@@ -35,7 +35,10 @@ public function __isset($name)
                     );
                     $statement->execute([$this->id]);
                     $row = $statement->fetch();
-                    $this->values['slug'] = $row[0];
+
+                    if (false !== $row = $statement->fetch()) {
```
sbreker, line 38: Is line 37 redundant?  On line 39 `$row` appears to be immediately overwritten by a second fetch().

lib/routing/QubitMetadataRoute.class.php
```diff
@@ -77,17 +77,8 @@ public function matchesUrl($url, $context = [])
...
-                }
-            }
-            // If for any reason the database can't be accessed, trigger the installer
-            catch (PropelException $e) {
-                $parameters['module'] = 'sfInstallPlugin';
-                $parameters['action'] = 'index';
-
-                return $parameters;
+            if (null === $this->resource = QubitObject::get($criteria)->__get(0)) {
+                return false;
```
sbreker, line 81: What happens in this case? 500 error if arInstall has not completed?

lib/arInstall.class.php
```diff
@@ -0,0 +1,421 @@
...
+        QubitSearch::enable();
+        QubitSearch::getInstance()->populate();
+    }
+
+    public static function createSetting($name, $value)
+    {
+        $setting = new QubitSetting();
+        $setting->name = $name;
+        $setting->value = $value;
+        $setting->save();
```
sbreker, line 419: Nitpick - insert newline before 419.

lib/arInstall.class.php
```diff
@@ -0,0 +1,421 @@
...
+
+class arInstall
+{
+    public static function createDirectories()
+    {
+        if (!is_dir(sfConfig::get('sf_log_dir'))) {
+            mkdir(sfConfig::get('sf_log_dir'), 0777, true);
+        }
+
+        Qubit::createUploadDirsIfNeeded();
```
sbreker, line 28: Nitpick: I think our coding standard states that there should be a blank line preceeding the last line of a function/method. This file has this issue in a number of places.

lib/arInstall.class.php
```diff
@@ -0,0 +1,421 @@
...
+            }
+        }
+
+        // Generate and set CSRF secret
+        $content = str_replace(
+            'change_me',
+            bin2hex(openssl_random_pseudo_bytes(16)),
+            $content
+        );
+        if (false === file_put_contents($settingsYmlPath, $content)) {
```
sbreker, line 135: Nitpick: I believe coding standard states to have a newline following a multi-line statement. There are a few examples of this in this file.

lib/arInstall.class.php
```diff
@@ -0,0 +1,421 @@
...
+class arInstall
+{
+    public static function createDirectories()
+    {
+        if (!is_dir(sfConfig::get('sf_log_dir'))) {
+            mkdir(sfConfig::get('sf_log_dir'), 0777, true);
+        }
+
+        Qubit::createUploadDirsIfNeeded();
+        Qubit::createDownloadsDirIfNeeded();
```
sbreker, line 29: Nitpick: AtoM's coding standard states that there should be a blank line before the last line of a function/method.

lib/arInstall.class.php
```diff
@@ -0,0 +1,421 @@
...
+            }
+        }
+    }
+
+    public static function createDatabasesYml()
+    {
+        $databasesYmlPath = sfConfig::get('sf_config_dir').'/databases.yml';
+
+        // Read databases.yml contents from existing databases.yml,
+        // databases.yml.tmpl (for a Subversion checkout), or symfony
```
sbreker: Subversion

lib/arInstall.class.php
```diff
@@ -0,0 +1,421 @@
...
+                "Can't write configuration file {$settingsYmlPath}"
+            );
+        }
+    }
+
+    public static function createSfSymlink()
+    {
+        $file = sfConfig::get('sf_root_dir').'/vendor/symfony/data/web/sf';
+        $link = sfConfig::get('sf_root_dir').'/sf';
+        if (!file_exists($link) && false === symlink($file, $link)) {
```
sbreker, line 146: Newline before if?

lib/arInstall.class.php
```diff
@@ -0,0 +1,421 @@
...
+
+        $configFile = sfConfig::get('sf_config_dir').'/config.php';
+        $content = sprintf(
+            "<?php\n"
+            ."// auto-generated by arInstall::configureDatabase()\n"
+            ."// date: %s\n\nreturn %s;\n",
+            date('Y/m/d H:i:s'),
+            var_export($config, true)
+        );
+        if (false === file_put_contents($configFile, $content)) {
```
sbreker, line 217: Newline before if.

lib/task/tools/installTask.class.php
```diff
@@ -0,0 +1,536 @@
...
+                    '',
+                    'Elasticsearch connection failure:',
+                    '',
+                    $e->getMessage(),
+                    '',
+                ],
+                'ERROR'
+            );
+
+            exit(2);
```
sbreker: Is there a meaningful difference between `exit(1)` used below and `exit(2)` used above?

lib/model/QubitObject.php
jraddaoui: Right, good catch!

lib/routing/QubitMetadataRoute.class.php
jraddaoui: I have not tested it, but that was my assumption. I'll give it a try.

lib/task/tools/installTask.class.php
```diff
@@ -0,0 +1,536 @@
...
+
+        $this->logSection('install', 'Creating admin user');
+
+        addSuperuserTask::addSuperUser(
+            $adminOptions['username'],
+            $adminOptions
+        );
+
+        $this->logSection('install', 'Installation completed');
+    }
```
sbreker: `execute()` is > 300 lines in length. Would readability be improved if it were broken up?

lib/arInstall.class.php
jraddaoui: Why?

lib/arInstall.class.php
jraddaoui: I don't think that's the case. Or do you mean before control statements? In which case, `if` it's not included in PphCsFixer ruleset:

https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/3.0/doc/rules/whitespace/blank_line_before_statement.rst#rule-sets

lib/arInstall.class.php
jraddaoui: I think that's only if the function ends in a return or other control statements from the ones mentioned in the rule from the previous comment.

lib/arInstall.class.php
jraddaoui: Not sure what do you mean here.

lib/arInstall.class.php
jraddaoui: As said before, not actually considered in the control statements that require an empty line before.

lib/arInstall.class.php
jraddaoui: Same.

lib/task/tools/installTask.class.php
jraddaoui: There are different exit/error codes, not something that we'd use but the general idea would be ...

- 0 = success
- 1 = aborted
- 2 = error

lib/task/tools/installTask.class.php
jraddaoui: One of the downsides of being careful with the 80 chars line length and 4 chars indentation ... I try not to move things out unless they are "reused" somewhere else but I see your point. There is also some duplication between the install and purge task after configuration, so I'll see if we can avoid/improve that.

lib/task/tools/installTask.class.php
sbreker: Opinion: These should be constants if there is meaning behind them - e.g. `exit(self::ERROR)`

jraddaoui: I like that, thanks Steve.

lib/arInstall.class.php
sbreker: Please disregard! I believe I have misinterpreted the PHP-CS-Fixer rule.

lib/arInstall.class.php
sbreker: Thanks for the link - Which ruleset are we using?

lib/arInstall.class.php
sbreker: The comment is out of date ('subversion' instead of 'github') - sorry I should have clarified!  😄

sbreker (changes requested):
This looks really great @jraddaoui - thanks!  I've noted a few small items to review.  

This will be a big improvement - looking forward to no more installation timeouts!  👍

lib/arInstall.class.php
jraddaoui: We use the `@PhpCsFixer` ruleset with a single exception to avoid inconsistencies in templates:

https://github.com/artefactual/atom/blob/qa/2.x/.php_cs#L24-L25

lib/routing/QubitMetadataRoute.class.php
jraddaoui: Good thing you pointed this out as it's quite confusing. Based on the URL you hit, when you go to the site without running the installer you get different errors ...

- Using a slug you get our normal 500 error page and the following error in the logs:

```
[11-Jun-2021 17:27:49] WARNING: [pool atom] child 16 said into stderr: "NOTICE: PHP message: No connection information in your runtime configuration file for datasource [propel]"
```

- Without slug you get a default 500 error page and a bit more confusing log:

```
[11-Jun-2021 17:21:02] WARNING: [pool atom] child 16 said into stderr: "NOTICE: PHP message: Empty module and/or action after parsing the URL "/" (/)."
[11-Jun-2021 17:21:02] WARNING: [pool atom] child 16 said into stderr: "NOTICE: PHP message: PHP Fatal error:  Uncaught sfError404Exception: Action "default/error404" does not exist. in /atom/src/cache/qubit/prod/config/config_core_compile.yml.php:660"
[11-Jun-2021 17:21:02] WARNING: [pool atom] child 16 said into stderr: "Stack trace:"
[11-Jun-2021 17:21:02] WARNING: [pool atom] child 16 said into stderr: "#0 /atom/src/vendor/symfony/lib/exception/sfError404Exception.class.php(49): sfController->forward('default', 'error404')"
[11-Jun-2021 17:21:02] WARNING: [pool atom] child 16 said into stderr: "#1 /atom/src/cache/qubit/prod/config/config_core_compile.yml.php(2396): sfError404Exception->printStackTrace()"
[11-Jun-2021 17:21:02] WARNING: [pool atom] child 16 said into stderr: "#2 /atom/src/vendor/symfony/lib/util/sfContext.class.php(170): sfFrontWebController->dispatch()"
[11-Jun-2021 17:21:02] WARNING: [pool atom] child 16 said into stderr: "#3 /atom/src/index.php(6): sfContext->dispatch()"
[11-Jun-2021 17:21:02] WARNING: [pool atom] child 16 said into stderr: "#4 {main}"
[11-Jun-2021 17:21:02] WARNING: [pool atom] child 16 said into stderr: "  thrown in /atom/src/cache/qubit/prod/config/config_core_compile.yml.php on line 660"
```

I'm trying to normalize this,  but it's not as easy as it looks ... thoughts?

sbreker: Previously we redirected to sfInstall when things weren't set up. Could we detect if the installer has not been run and  redirect to a static "Go run the AtoM CLI installer" page for all routes?

jraddaoui: Ok, still a normal error page but now the error will be the same for all routes. In any case, lets hope they don't forget to run the installer.

lib/model/QubitObject.php
jraddaoui: Addressed in 684e29b.

lib/routing/QubitMetadataRoute.class.php
jraddaoui: Addressed in fa2a1dd and 2df334e.

lib/arInstall.class.php
jraddaoui: Maybe just `git`, guessing it could be checked from other repository ... 4997828.

lib/task/tools/installTask.class.php
jraddaoui: I'm thinking this doesn't add a lot of value considering that the error code could be useful in automated process and the difference between 1 and 2 would require user interaction to abort the task. So I think just `exit(1);` in all cases should be enough.

jraddaoui: Addressed in 0252ed5.

lib/task/tools/installTask.class.php
jraddaoui: Addressed in 40ee253 and 9ca9016.

I hope that makes it a bit clearer. We could go a bit further extending the installTask from purgeTask, but it will require to change one of the task option names and the user input process to normalize them.

sbreker: This looks great!

sbreker (approved):
Looks great @jraddaoui!  👍

jraddaoui:
Thanks @sbreker!

PR #1295, opened 2021-05-17
Add termination logging to worker, refs #13504

mcantelon:
Added signal and shutdown handlers to the job worker so that termination can be logged.

### DIFF
--- a/lib/task/jobs/jobWorkerTask.class.php
+++ b/lib/task/jobs/jobWorkerTask.class.php
@@ -108,6 +108,8 @@ EOF;
         $this->log('Running worker...');
         $this->log('PID '.getmypid());
 
+        $this->activateTerminationHandlers();
+
         $counter = 0;
 
         // The worker loop!
@@ -126,5 +128,35 @@ EOF;
                 }
             }
         );
+    }
+
+    protected function activateTerminationHandlers()
+    {
+        pcntl_async_signals(true);
+
+        // Define signal handler function within the object scope to allow it
+        // to access the object's methods, etc.
+        $signalHandler = function ($signal) {
+            $messages = [
+                SIGINT => 'Job worker termination requested by user.',
+                SIGHUP => 'Job worker hang up requested.',
+                SIGTERM => 'Job worker termination requested.',
+                SIGQUIT => 'Job worker quit requested.',
+            ];
+
+            $this->log($messages[$signal]);
+
+            exit();
+        };
+
+        pcntl_signal(SIGINT, $signalHandler);
+        pcntl_signal(SIGHUP, $signalHandler);
+        pcntl_signal(SIGTERM, $signalHandler);
+        pcntl_signal(SIGQUIT, $signalHandler);
+
+        // Define shutdown function
+        register_shutdown_function(function () {
+            $this->log('Job worker stopped.');
+        });
     }
 }
--- a/Dockerfile
+++ b/Dockerfile
@@ -20,6 +20,7 @@ RUN set -xe \
       mbstring \
       mysqli \
       opcache \
+      pcntl \
       pdo_mysql \
       sockets \
       xsl \
--- a/docker/entrypoint.sh
+++ b/docker/entrypoint.sh
@@ -6,13 +6,10 @@ set -o nounset
 # set -o xtrace
 
 __dir="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
-__file="${__dir}/$(basename "${BASH_SOURCE[0]}")"
-__atom_root="/atom/src"
-
 
 # Clean-ups
 rm -rf /usr/local/etc/php-fpm.d/*
-rm -rf ${__atom_root}/cache/*
+rm -rf ${__dir}/../cache/*
 
 # Populate configuration files
 php ${__dir}/bootstrap.php $@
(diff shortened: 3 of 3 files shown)


mcantelon:
I'm only intercepting INT, HUP, and TERM signals... not sure if I should intercept any others (there's about 30 of them).

lib/task/jobs/jobWorkerTask.class.php
```diff
@@ -119,6 +121,9 @@ function ($handle, $job, $e) {
             // and restablish the connection when needed. Also, the persistent mode
             // could be disabled for this worker. See issue #4182.
             function () use (&$counter) {
+                // Needed for signal handler to work
+                pcntl_signal_dispatch();
```
jraddaoui: Is it needed inside the loop? I'm not sure why it would be pending signals. Could it be solved calling `pcntl_async_signals(true);` only once. I'm not sure if they are related but I had to use the later to make signals work one time.

If not, could it be done with a worker callback?

lib/task/jobs/jobWorkerTask.class.php
```diff
@@ -127,4 +132,43 @@ function () use (&$counter) {
...
         );
     }
+
+    protected function activateTerminationHandlers()
+    {
+        // Define signal handler function within the object scope to allow it
+        // to access the object's methods, etc.
+        $signalHandler = function ($signal) {
+            switch ($signal) {
+                case SIGINT:
```
jraddaoui: I'd include SIGQUIT in here too.

lib/task/jobs/jobWorkerTask.class.php
```diff
@@ -127,4 +132,43 @@ function () use (&$counter) {
...
+                    $this->log('Job worker termination requested by user.');
+
+                    exit();
+
+                    break;
+
+                case SIGHUP:
+                    $this->log('Job worker restart requested.');
```
jraddaoui: Is this logging okay ... it looks to be a termination signal caused by hang up ...

https://www.gnu.org/software/libc/manual/html_node/Termination-Signals.html

jraddaoui: If that's right, we may want to call `exit()` in all cases (removing the breaks and adding a default case).

lib/task/jobs/jobWorkerTask.class.php
```diff
@@ -127,4 +132,43 @@ function () use (&$counter) {
...
+
+                    exit();
+
+                    break;
+            }
+        };
+
+        pcntl_signal(SIGINT, $signalHandler);
+        pcntl_signal(SIGHUP, $signalHandler);
+        pcntl_signal(SIGTERM, $signalHandler);
```
jraddaoui, line 154: Add `SIGQUIT`.

lib/task/jobs/jobWorkerTask.class.php
```diff
@@ -127,4 +132,43 @@ function () use (&$counter) {
...
+
+        // Define shutdown function
+        $shutdownFunction = function () {
+            $this->log('Job worker stopped.');
+        };
+
+        register_shutdown_function($shutdownFunction);
```
jraddaoui: Nitpick: you could declare the function directly in the argument without using a variable.

jraddaoui (changes requested):
Nice @mcantelon, thanks!

I have added a few comments below. Also, this won't work in the Docker environment without installing the `pcntl` extension and changing how the signals are managed in the entrypoint script. I made those changes in the following commit for another work:

https://github.com/artefactual/atom/commit/9005f4551a2ef74366c2ff3f61a3c4e85824de53

You will only need to add the changes from `Dockerfile` and `docker/entrypoint.sh` to make it work, and here more info if you're interested:

https://www.ctl.io/developers/blog/post/gracefully-stopping-docker-containers/
https://medium.com/@gchudnov/trapping-signals-in-docker-containers-7a57fdda7d86
https://blog.forma-pro.com/dockerize-any-php-project-a-pure-bash-script-38dd37baf710
https://hynek.me/articles/docker-signals/
https://unix.stackexchange.com/questions/146756/forward-sigterm-to-child-in-bash
https://petermalmgren.com/signal-handling-docker/

At some point we may want to use `dumb-init`, but it's not required for this.

https://github.com/Yelp/dumb-init

mcantelon:
Hi @radda ... thanks for the tips! That first article "Gracefully Stopping Docker Containers" was super helpful!

I've updated the PR based on your feedback (and simplified the signal handler function a bit) and tested it in Docker... let me know what you think! Should we have different error codes for the different signal types?

jraddaoui:
Thanks @mcantelon, looks great!

If you don't mind, could you grab all the changes from the entrypoint?

https://github.com/artefactual/atom/commit/9005f4551a2ef74366c2ff3f61a3c4e85824de53#diff-4f5cabe26761257a4d685a6edc7a43e0fe0f78762f50eeb48530f2bd3b3ee7ca

It would be nice to have those clean-ups in here too, now that you're changing the worker command, it would be great to consolidate it for the fpm command.

PS. Sorry @Radda, I know it's not the first time, heheh. @mcantelon ... use @jraddaoui :P

jraddaoui (approved):

mcantelon:
Hi @jraddaoui ... I put in all the `entrypoint.sh` changes and retested. Let me know if anything else should be tweaked!

jraddaoui:
Awesome, thanks @mcantelon!

PR #1297, opened 2021-05-19
Fix cherry pick to stable/2.6.x, refs #8096

sbreker:
These commits revert the original cherry pick of 8096 in the 2.6.3 release and reapply it on stable/2.6.x. There was an issue with the conflict resolution when this cherry-pick was originally completed.

### DIFF
--- a/js/dominion.js
+++ b/js/dominion.js
@@ -250,7 +250,7 @@
           .on('blur', $.proxy(this.blur, this))
           .on('keydown', $.proxy(this.keydown, this));
 
-        this.$menu.on('click', 'li', $.proxy(this.click, this));
+        this.$form.on('click', '.search-popover li', $.proxy(this.click, this));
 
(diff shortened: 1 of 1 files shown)


djjuhasz (approved):
Looks good @sbreker.   I don't know if you were planning to make this two commits on stable/2.6.x, but I'd advocate for making it a single commit.

sbreker:
Thanks for reviewing @djjuhasz! I _was_ going to leave it as two commits as I wanted it to be clear that the previous 8096 was reverted, and this is not an additional change or duplicate.

djjuhasz:
@sbreker I think for such a small change the extra commit isn't necessary for understanding the change history, but I'm okay with two commits if you think it's important.

jraddaoui:
I'd say one commit too, just in case someone with 2.6.4 installed wants to get a patch for it.

sbreker:
Good point @jraddaoui - will merge as one commit.

PR #1300, opened 2021-05-23
Varnish culture support for Docker, refs #13405

mcantelon:
Updated Varnish VCL, for use with AtoM running via Docker, to allow the content of non-default cultures to be cached.

### DIFF
--- a/apps/qubit/config/qubitConfiguration.class.php
+++ b/apps/qubit/config/qubitConfiguration.class.php
@@ -31,7 +31,7 @@ class qubitConfiguration extends sfApplicationConfiguration
 
     public function listenToChangeCultureEvent(sfEvent $event)
     {
-        setcookie('atom_culture', $event['culture']);
+        setcookie('atom_culture', $event['culture'], ['path' => '/']);
     }
 
     /**
--- a/docker/docker-compose.varnish.yml
+++ b/docker/docker-compose.varnish.yml
@@ -1,10 +1,10 @@
 ---
-version: "2"
+version: "2.1"
 
 services:
 
   varnish:
-    image: varnish:6.0
+    image: varnish:6.5.1
     volumes:
       - ./etc/varnish/default.vcl:/etc/varnish/default.vcl:ro
     ports:
--- a/docker/etc/varnish/default.vcl
+++ b/docker/etc/varnish/default.vcl
@@ -1,20 +1,40 @@
 vcl 4.1;
+import cookie;
 
 backend default {
   .host = "nginx";
 }
 
 sub vcl_recv {
+  # Do not cache culture change request
+  if (req.url ~ "sf_culture") {
+    return (pass);
+  }
+
+  # Set HTTP header to culture cookie value
+  cookie.parse(req.http.cookie);
+  set req.http.X-Atom-Culture = cookie.get("atom_culture");
+
   # Do not cache clipboard POST request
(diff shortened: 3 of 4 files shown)


mcantelon:
In the Docker compose file for Varnish I changed the version to 6.5.1 because it comes with the cookie VMOD included (6 LTS doesn't and it's only from 6.4 on that the VMODs are included). On Varnish's release page 6.5.1 is the non-LTS version with the longest support window (until 2022-09-15).

jraddaoui:
Thanks @mcantelon!

Now that you mention this, just slightly related, I see in some AtoM pages where the new `atom_culture` cookie gets added multiple times, could you take a quick look now that you're checking this?

It's possible that having that cookie repeated is what requires the following workaround in the e2e tests:

https://github.com/artefactual/atom/blob/qa/2.x/cypress/support/index.js#L22-L29

Not a huge deal, but it would be great if you could check this out.

docker/docker-compose.varnish.yml
```diff
@@ -1,10 +1,10 @@
 ---
-version: "2"
+version: "2.1"
```
jraddaoui, line 2: Oh, good catch! It may be needed in the other docker-compose file for PMM too, so it would be great if you change it in there too.

apps/qubit/config/qubitConfiguration.class.php
```diff
@@ -31,7 +31,7 @@ public function responseFilterContent(sfEvent $event, $content)
 
     public function listenToChangeCultureEvent(sfEvent $event)
     {
-        setcookie('atom_culture', $event['culture']);
+        setcookie('atom_culture', $event['culture'], ['path' => '/']);
```
jraddaoui, line 34: Maybe this avoids the multiple cookie issue ...

docker/etc/varnish/default.vcl
```diff
@@ -1,20 +1,40 @@
...
+import cookie;
 
 backend default {
   .host = "nginx";
 }
 
 sub vcl_recv {
+  # Do not cache culture change request
+  if (req.url ~ "sf_culture") {
+    return(pass);
```
jraddaoui: For consistency, we're adding a whitespace before the parenthesis below.

apps/qubit/config/qubitConfiguration.class.php
mcantelon: Yeah, I'm going to do some more testing to double-check but I think that's the cause of the multiple cookie issue. I hadn't ran into it before.

mcantelon:
@jraddaoui Thanks for the quick review, Radda! I've updated the PR. Yeah, this might fix the e2e test FF multiple cookie issue. Should I remove that FF workaround from the e2e test in this PR, so we can confirm this, or is there a way to test, to confirm that it's safe to remove, some other way?

jraddaoui (approved):
Thanks @mcantelon!

jraddaoui:
> @jraddaoui Thanks for the quick review, Radda! I've updated the PR. Yeah, this might fix the e2e test FF multiple cookie issue. Should I remove that FF workaround from the e2e test in this PR, so we can confirm this, or is there a way to test, to confirm that it's safe to remove, some other way?

Sorry I missed those questions @mcantelon, FYI, that did the trick ... https://github.com/artefactual/atom/pull/1319. Thanks!

Issue #1128, opened 2020-06-18
Language menu label: "português" with first letter in lower case 


ricardodepinho:
https://github.com/artefactual/atom/blob/7b2ab77cc2ac5424abb9cbc17f2df5e231b0ee73/vendor/symfony/lib/i18n/data/pt.dat#L1
Replace the first letter to Upper case in text:
s:10:"português"
with 
s:10:"Português"

fiver-watson:
Hi Ricardo, 

Thank you for raising this issue. It has generated some interesting discussions here at Artefactual!

I've done a bit of initial investigation, and it seems this is something we are inheriting from Symfony's defaults. Looking further, it seems that, while English (and some other languages) typically capitalize proper nouns including languages, Portuguese (and others, like French, Spanish, etc) typically do not capitalize languages when writing?

If you want to make this change right now, you should be able to edit this file locally, if you clear the application cache, restart PHP-FPM, and if in use, memcached after. You may need to re-index as well. However, I haven't tested this myself yet, so make sure you test in a dev environment, and/or make a data backup first if you try this! 

In terms of implementing this change in the public project - given I'm not a Portuguese speaker, and there are also other languages in our language menu to consider, I wonder if perhaps we should try to have a broader discusson in the forum or something to solicit some other opinions first?

Cheers, 

Dan Gillean, MAS/MLIS
AtoM Program Manager

ricardodepinho:
Hi Dan,
I'm glad for generating interesting discussions at Artefactual based on this. ;-)
I'm not the best person to answer to your questions, not even to start a boarder discussion about this. 

In my personal opinion all options on the Language menu should be in capitalized words. 
All web sites/apps I can remember of implement this, no matter what language they are on. 
At this point, the AtoM language menu with some languages capitalized and other not, it just don't look right!
And there is another problem with the AtoM language menu. The languages text should change according to the activated language. As you know, "English" in french is "Anglais" and in portuguese is "Inglês".

I've searched for a reference, and found it at the google translator language selection drop-down: 
https://translate.google.com/?hl=pt (capitalized)
https://translate.google.com/?hl=fr (capitalized)
https://translate.google.com/?hl=es (not capitalized)

Eventually our suggestion (mine and José Anjos) of changing Symfony's defaults in pt.dat for solving the problem was not the best one. I can suggest you to study other better solutions for the particular "AtoM language menu" problem.

I hope this could be of any help!
Cheers,
Ricardo Pinho

[...]

PR #1302, opened 2021-05-26
Change initial letter of languages to uppercase in menu, refs #13515

mInnes-archives:
PR is in response to open issue #1128 regarding inconsistency in upper/lower casing of initial letter for languages in the language menu ~~and the language settings menu~~. Based on the assumption that language names written as a single word in a menu should *probably* start uppercased, it is now consistently uppercase.

If this is still an issue open for a bit of debate, that's cool, but this PR should at least provide an easy solution for those wishing for the consistency without modifying vendor files.

### DIFF
--- a/apps/qubit/modules/menu/templates/_changeLanguageMenu.php
+++ b/apps/qubit/modules/menu/templates/_changeLanguageMenu.php
@@ -16,7 +16,7 @@
       <ul>
         <?php foreach ($langCodes as $value) { ?>
           <li<?php if ($sf_user->getCulture() == $value) { ?> class="active"<?php } ?>>
-            <?php echo link_to(format_language($value, $value), ['sf_culture' => $value] + $sf_data->getRaw('sf_request')->getParameterHolder()->getAll()); ?>
+            <?php echo link_to(ucfirst(format_language($value, $value)), ['sf_culture' => $value] + $sf_data->getRaw('sf_request')->getParameterHolder()->getAll()); ?>
           </li>
         <?php } ?>
       </ul>


jraddaoui:
Thanks @mInnes-archives,

@fiver-watson, would you mind giving it a try to this and create a ticket?

ricardodepinho:
Thanks @mInnes-archives . Excellent!
I will give it a try ASAP.
Cheers

fiver-watson:
lang-menu

I've tested this, and as far as I'm able to tell, it doesn't seem to be negatively impacting non-latin characters. I think it looks good!

fiver-watson:
Redmine issue ticket created here: https://projects.artefactual.com/issues/13515

jraddaoui:
Thanks @mInnes-archives! Just a couple of things ...

I wonder why did you revert the change in the settings page, the ticket doesn't mention that but I think it may also be a good addition (@fiver-watson, thoughts?). And, we just merged another PR so if you could rebase we could merge with `--ff-only` without adding myself to the commit.

mInnes-archives:
>I wonder why did you revert the change in the settings page, the ticket doesn't mention that but I think it may also be a good addition (@fiver-watson, thoughts?).

Yes, I really should have explained that. That commit added uppercasing in the settings menu, which in isolation is a fine idea. However, since the settings menu has both the table with the list of active languages and the select form underneath for adding new languages, I was actually introducing some inconsistency. The main issue being that there isn't an easy way to change the casing on the choices of the select form. So you had initial uppercasing above but below it was back to lowercasing. Given that the spirit of the original issue was about consistency, I felt it best to keep the settings menu consistent within itself. The settings menu also wasn't as much of an issue as the top bar menu, since you aren't displaying information using different i18n cultures side-by-side. It also isn't public facing.

> And, we just merged another PR so if you could rebase we could merge with --ff-only without adding myself to the commit.

I will do this later. I'm just away from my dev space at the moment.

mInnes-archives:
Good to go now.

jraddaoui:
Thanks @mInnes-archives!

PR #1303, opened 2021-05-26
Highlight errors on job detail page, refs #13504

mcantelon:
On the job detail page show error details, highlighted in red, in the
area above the full job output.


mcantelon:
Job time being show beside errors is kind of redundant given that error time is always the same as completed time so not sure if I should keep it in there.

djjuhasz (approved):
Looks good to me 👍

PR #1304, opened 2021-05-27
Remove redundant anchors from cards with images, refs #13535 

mInnes-archives:
PR removes unnecessary anchors included in card view bricks with images. These anchors are intended to provide a text link to the relevant record for text-only bricks, therefore they serve no purpose for bricks with images as these bricks already have an appropriate link.

This PR is motivated by an accessibility concern for the following reasons:

- The redundant anchors introduce unnecessary focus elements when tabbing through card view
- The redundant anchors lack text, so are problematic for screen readers
- The previous point results in these elements being flagged by accessibility audit software (eg. WAVE evaluation tool)

I also cleaned up the logic quite a bit to better separate out image vs. text-only bricks.

### DIFF
--- a/apps/qubit/modules/informationobject/templates/_cardViewResults.php
+++ b/apps/qubit/modules/informationobject/templates/_cardViewResults.php
@@ -7,12 +7,9 @@
     <?php $title = get_search_i18n($doc, 'title', ['allowEmpty' => false, 'culture' => $selectedCulture]); ?>
 
     <?php if (!empty($doc['hasDigitalObject'])) { ?>
-      <div class="brick">
-    <?php } else { ?>
-      <div class="brick brick-only-text">
-    <?php } ?>
 
-      <a href="<?php echo url_for(['module' => 'informationobject', 'slug' => $doc['slug']]); ?>">
+      <div class="brick">
+
         <?php if (
             isset($doc['digitalObject'])
             && !empty($doc['digitalObject']['thumbnailPath'])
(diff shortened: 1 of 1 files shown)


apps/qubit/modules/informationobject/templates/_cardViewResults.php
```diff
@@ -23,23 +20,27 @@
             ['alt' => isset($doc['digitalObject']['digitalObjectAltText']) ? $doc['digitalObject']['digitalObjectAltText'] : truncate_text(strip_markdown($title), 100)]),
             ['module' => 'informationobject', 'slug' => $doc['slug']]); ?>
 
-        <?php } elseif (isset($doc['digitalObject']) && !empty($doc['digitalObject']['mediaTypeId'])) { ?>
+        <?php } else { ?>
 
           <?php echo link_to(image_tag(QubitDigitalObject::getGenericIconPathByMediaTypeId($doc['digitalObject']['mediaTypeId']),
```
jraddaoui, line 25: Without the check over `$doc['digitalObject']['mediaTypeId']` this could raise a `PHP Notice:  Undefined index: mediaTypeId`.

jraddaoui (changes requested):
Thanks @mInnes-archives, this is a great improvement!

Sorry I could not get to these pull requests sooner but we really appreciate them. Since you're addressing these accessibility issues, we should mention that we are also putting some effort on this matter and give you a general idea about the plan.

---

You may have seen https://github.com/artefactual/atom/pull/1326, where we are creating a revamped version of the existing themes using Bootstrap 5 and other (more up-to-date) components. This new themes require a big rewrite (duplication) of the templates and Javascript you are modifying in these pull requests, but our plan is to maintain both versions for a while after we release the new themes (which probably won't happen in 2.7).

At that point (lets call it 2.8) the current themes will enter a deprecated status to give time to update existing custom themes until we fully remove them (lets say 2.9 or 2.10). That being said, these improvements are great additions to the existing themes that will be maintained for some period and, while we are working in the new themes duplicating the templates, I'm keeping an eye on this changes so we don't ignore them in the new versions.

---

About his particular fix, I have added a comment below to prevent an edge case where the media type is not set for some reason, but other than that it looks great. Please, use the following ticket for the reference:

https://projects.artefactual.com/issues/13535

Thanks again! Best,
Radda.

mInnes-archives:
> You may have seen #1326, where we are creating a revamped version of the existing themes using Bootstrap 5 and other (more up-to-date) components.

I have noticed and this is great! I've held off on doing more of these little accessibility tweaks once I saw #1326. I also have a bit less free time these days, but I think I will direct my attention to the accession module when I get a chance. I've heard at conferences and in user groups that they would like things like advanced search functionality, cli tasks, and some additional steps towards implementing CAAIS.

jraddaoui:
Sounds good @mInnes-archives, I've heard some things about CAAIS internally too, so I'd suggest that you check with @fiver-watson before doing any implementation.

fiver-watson:
Hi @mInnes-archives thanks again SO MUCH for all your amazing contributions of late. 

RE: Accessions and CAAIS, the only upcoming work confirmed for 2.7 (finished, and therefore already in qa/2.x) is the addition of Events. See: https://projects.artefactual.com/issues/13315

At this time, I'm not aware of any scheduled work around CAAIS. I'd suggest starting with smaller changes if you go down this route, as our team might prefer to gather more specific input from our community and clients before major revisions, etc. 

I will say that an advanced search module for Accessions would be AMAZING, and yes, I too have heard feedback from other users wishing for this. 

There are likely also some refinements to the indexed fields we might want to make to improve accession searching overall - but I'd suggest avoiding messing with the ES mappings at this time. After the BS5 upgrade is done, our next big dependency upgrade goal is Elasticsearch, and since there are a number of implementation changes, we will hopefully take the opportunity to revisit some of the analyzers and filter used, and some of the indexed fields per entity as well. 

I'll try to QA all of these PRs soon!

mInnes-archives:
@fiver-watson, thanks for the update.

I haven't worked with ES as of yet, so any work I do in that area would be quite small. For anything CAAIS related, if I have an idea for something specific, I can throw it in the Roadmap Committee channel to get you thoughts.

jraddaoui (approved):
Nice! Thanks @mInnes-archives!

PR #1306, opened 2021-06-01
Allow plugins cust. within themes, refs #13518

jraddaoui:
Allows to overwrite controllers and views from modules in other plugins
within a theme plugin.


sbreker (approved):
Looks good @jraddaoui. Too bad there isn't a way to insert the theme plugin at the front of the configuration class plugins list without re-creating it. This looks like it should effectively replace the plugin list with a reordered one!

PR #1308, opened 2021-06-02
Support actor D.O. regeneration, refs #13513

mcantelon:
Added logic to the digital object regeneration task so it will, in addition to its current functionality, also regenerate digital objects linked to authority records.


sbreker (approved):
Looks great @mcantelon! 👍

mcantelon:
Thanks @sbreker !

PR #1311, opened 2021-06-09
Cosmetic: Vertically center privacy message text

djjuhasz:
Add padding to the "privacy-message-content" CSS so the line height
matches the privacy-message-button so the text is vertically centered
in the privacy banner area.


jraddaoui (approved):
:ok_hand:

PR #1312, opened 2021-06-10
Improve accessibility of expander links, refs #13535

mInnes-archives:
PR adds aria-labels ~~and titles~~ to expander more/less links. ~~Titles add hover text to the "»" and "«" icons for general clarity~~. The aria-label redirects screen readers to ignore the link text, which currently reads as "right-pointing double arrow link" or some variation of this.

The only caveat with the added title is that it doesn't work with Symfony's i18n dictionary. However, browser-based translation picks up the slack.

### DIFF
--- a/apps/qubit/templates/_footer.php
+++ b/apps/qubit/templates/_footer.php
@@ -10,6 +10,13 @@
     <?php echo __('Printed: %d%', ['%d%' => date('Y-m-d')]); ?>
   </div>
 
+  <div id="js-i18n">
+    <div id="read-more-less-links"
+      data-read-more-text="<?php echo __('Read more'); ?>" 
+      data-read-less-text="<?php echo __('Read less'); ?>">
+    </div>
+  </div>
+
 </footer>
 
 <?php $gaKey = sfConfig::get('app_google_analytics_api_key', ''); ?>
--- a/js/qubit.js
+++ b/js/qubit.js
@@ -29,6 +29,26 @@ Drupal.behaviors.expander = {
             widow: 4,
             expandEffect: 'show'
           });
+
+          (function ($) {
+            'use strict';
+            // Get i18n text for read more/less links from footer
+            var $i18n = $('#js-i18n #read-more-less-links');
+
+            // Get read more/less link elements
+            var $readMoreLink = $element.find('.read-more a');
+            var $readLessLink = $element.find('.read-less a');
+
+            $(function() {
+              // Add accessibility label to read more link
+              $readMoreLink.attr('aria-label', $i18n.data('read-more-text'));
+
+              // Add accessibility label to read less link
+              $readLessLink.attr('aria-label', $i18n.data('read-less-text'));
+            });
+
+          })(jQuery);
+
         }
       });
     }


jraddaoui (changes requested):
Thanks @mInnes-archives, nice addition!

A couple of details ...

**Translations:** we don't have a system to translate Javascript content in AtoM yet and those labels should be translated somehow. To work around this issue, we normally use data attributes in the related elements to pass the content translated to the JS code, see an example in the clipboard menu:

https://github.com/artefactual/atom/blob/qa/2.x/apps/qubit/modules/menu/templates/_clipboardMenu.php#L3-L7

In this case is hard to know where the expander will be added and therefore where to include those data attributes, as it may also happen in many pages. The following is not ideal in terms of performance but I'd suggest this solution for this particular case (and maybe others we're not aware) ...

[At the end of the footer](https://github.com/artefactual/atom/blob/qa/2.x/apps/qubit/templates/_footer.php#L24), add the following block:

```
<div id="js-i18n">
  <div id="read-more-less-links"
       data-read-more-text="<?php echo __('Read more'); ?>" 
       data-read-less-text="<?php echo __('Read less'); ?>">
  </div>
</div>
```

And from the JS:

```javascript
var $i18n = $('#js-i18n #read-more-less-links');
$element.find('.read-more a').attr('aria-label', $i18n.data('read-more-text'));
$element.find('.read-less a').attr('aria-label', $i18n.data('read-more-less'));
```

**Expanded status:** it would be great if we could add [`aria-expanded` attributes](https://www.w3.org/TR/wai-aria-1.1/#aria-expanded) to represent the status. I don't think it's a huge deal in this case and it doesn't have to happen in this PR, just a comment.

jraddaoui:
Some notes about merging this changes @mInnes-archives ...

Please, use the same ticket for the reference after rebasing and squashing the changes (https://projects.artefactual.com/issues/13535). If you submit all the pull requests ready to merge at the same time we'll need to rebase the others after merging one of them to be able to merge all with `--ff-only`. With our current infrastructure we have two options to solve this issue:

- If you care about being the author and committer (like this commit: 2c8f751), I'll need to ask you to rebase the other PRs and re-submit them after merging one; in which case we better do the rebase and merge one by one.
- If you are okay with being the author but not the committer (like: 6629c1f), feel free to submit all PRs at the same time and I'll take care of the rebase and merge.

mInnes-archives:
@jraddaoui, we'll do it one-by-one. This gives me time to implement the requested changes. I'll keep an eye on each as they get merged and set up the next if it's ready.

mInnes-archives:
@jraddaoui, I've added and tested your suggestion for the i18n data and everything works as it should. I've rebased and prepped for merge.

>Expanded status: it would be great if we could add aria-expanded attributes to represent the status.

In this case, I don't think this would be necessary/appropriate. The more and less links are technically two different elements, and their labeling is descriptive enough to convey the expanded status. Additionally, the SR output for the read-less link with `aria-expanded='true'` is a bit confusing. It reads things like 'read less link expanded' (Orca) or 'read less expanded link' (NVDA).

jraddaoui:
> @jraddaoui, I've added and tested your suggestion for the i18n data and everything works as it should. I've rebased and prepped for merge.
> 
> > Expanded status: it would be great if we could add aria-expanded attributes to represent the status.
> 
> In this case, I don't think this would be necessary/appropriate. The more and less links are technically two different elements, and their labeling is descriptive enough to convey the expanded status. Additionally, the SR output for the read-less link with `aria-expanded='true'` is a bit confusing. It reads things like 'read less link expanded' (Orca) or 'read less expanded link' (NVDA).

Thanks for looking into this Matt, sounds great!

The e2e tests are failing in the CI process, I took a look locally and found the following error in the console, which is preventing the addition of the `aria-label` attributes ...

```
TypeError: $ is not a function at HTMLDivElement.<anonymous> (qubit.js:34)
```

That JS is not wrapped in the usual ...

```js
(function($)
{
  'use strict';

  $(function()
  {
    
  });

})(jQuery);
```

Also, this requires another rebase, after d5c1097 (and another commit that I'll merge in a minute).

mInnes-archives:
Thanks @jraddaoui. It looks like that did the trick. I'll prep #1315 once this is merged.

jraddaoui (approved):

jraddaoui:
Thanks @mInnes-archives!

PR #1313, opened 2021-06-10
Fix CSRF validation, refs #13450

jraddaoui:
Avoid `setValidators` outside of form's `configure` method as it may
overwrite other validators.


sbreker (approved):
Looks great @jraddaoui !

PR #1314, opened 2021-06-10
CAS: make service URL configurable, refs #13524

hakamine:
Make it possible to override the CAS service string (service URL).
This is needed when the hostname does not match the host part of the
AtoM instance URL (as phpCAS seems to set the service URL based on the
hostname by default).


djjuhasz (approved):
Looks good @hakamine 👍

tw4l (approved):
Looks like @djjuhasz beat me to it but this looks great! Thank you, @hakamine :clap:

hakamine:
Thank you @tw4l @djjuhasz !

PR #1301, opened 2021-05-26
Add title attributes to view options, refs #13510

mInnes-archives:
PR adds titles to card and table view buttons. This is a minor accessibility improvement, providing support for screen readers. It also clarifies the functionality of each button more generally.

PR #1315, opened 2021-06-11
Search accessibility improvements, refs #13535

mInnes-archives:
### Description
PR seeks to improve search and advanced search form accessibility and compatibility with screen readers, as well as general clarity of function. This is achieved through the following:

- **Add aria-labels to search forms currently relying on placeholder text.** Not all screen reader use placeholder text.
- **Add ~~titles~~ aria-labels to search buttons.** This adds text for screen readers ~~and hover text for general clarity~~.
- **Associate form labels with existing forms.**
- **Change reset icon in advanced search to 'undo' for general clarity of function**. The 'x' has the same functionality as the reset action button at the bottom of advanced search. I felt that the 'undo' icon was a better representation for reset. 

For a fuller breakdown of the changes, see the commit descriptions.

### Images
Reset button icon changed from 'x' to 'undo'
times_button
undo_button

### DIFF
--- a/apps/qubit/modules/actor/templates/_advancedSearch.php
+++ b/apps/qubit/modules/actor/templates/_advancedSearch.php
@@ -1,4 +1,4 @@
-<section class="advanced-search-section">
+<section class="advanced-search-section" role="search" aria-label="<?php echo __('Advanced %1%', ['%1%' => sfConfig::get('app_ui_label_actor')]); ?>">
 
   <a href="#" class="advanced-search-toggle <?php echo $show ? 'open' : ''; ?>" aria-expanded="<?php echo $show ? 'true' : 'false'; ?>"><?php echo __('Advanced search options'); ?></a>
 
@@ -26,7 +26,7 @@
                 <option value="not"<?php echo 'not' == $item['operator'] ? ' selected="selected"' : ''; ?>><?php echo __('not'); ?></option>
               </select>
 
-              <input class="query" type="text" placeholder="<?php echo __('Search'); ?>" name="sq<?php echo $key; ?>" value="<?php echo $item['query']; ?>"/>
+              <input class="query" type="text" aria-label="<?php echo __('Search'); ?>" placeholder="<?php echo __('Search'); ?>" name="sq<?php echo $key; ?>" value="<?php echo $item['query']; ?>"/>
 
               <span><?php echo __('in'); ?></span>
 
@@ -37,7 +37,7 @@
                 <?php } ?>
               </select>
 
-              <a href="#" class="delete-criterion"><i class="fa fa-times"></i></a>
+              <a href="#" class="delete-criterion" aria-label="<?php echo __('Delete criterion'); ?>"><i aria-hidden="true" class="fa fa-times"></i></a>
 
             </div>
 
@@ -55,7 +55,7 @@
             <option value="not"><?php echo __('not'); ?></option>
           </select>
 
-          <input class="query" type="text" placeholder="<?php echo __('Search'); ?>" name="sq<?php echo $count; ?>"/>
+          <input class="query" type="text" aria-label="<?php echo __('Search'); ?>" placeholder="<?php echo __('Search'); ?>" name="sq<?php echo $count; ?>"/>
 
           <span><?php echo __('in'); ?></span>
 
@@ -66,7 +66,7 @@
             <?php } ?>
           </select>
 
-          <a href="#" class="delete-criterion"><i class="fa fa-times"></i></a>
+          <a href="#" class="delete-criterion" aria-label="<?php echo __('Delete criterion'); ?>"><i aria-hidden="true" class="fa fa-times"></i></a>
 
         </div>
 
--- a/apps/qubit/modules/repository/templates/_advancedFilters.php
+++ b/apps/qubit/modules/repository/templates/_advancedFilters.php
@@ -8,21 +8,8 @@
 
     <div class="row-fluid">
       <div class="span4">
-        <?php echo __('Thematic area:'); ?>
-      </div>
-
-      <div class="span4">
-        <?php echo __('Archive type:'); ?>
-      </div>
-
-      <div class="span4">
-        <?php echo __('Regions:'); ?>
-      </div>
-    </div>
-
-    <div class="row-fluid">
-      <div class="span4">
-        <select name="thematicAreas">
+        <label for="thematicAreas"><?php echo __('Thematic area:'); ?></label>
+        <select name="thematicAreas" id="thematicAreas">
           <option selected="selected"></option>
           <?php foreach ($thematicAreas as $r) { ?>
             <option value="<?php echo $r->getId(); ?>">
@@ -33,7 +20,8 @@
       </div>
 
       <div class="span4">
-        <select name="types">
+        <label for="types"><?php echo __('Archive type:'); ?></label>
+        <select name="types" id="types">
           <option selected="selected"></option>
           <?php foreach ($repositoryTypes as $r) { ?>
             <option value="<?php echo $r->getId(); ?>">
@@ -44,7 +32,8 @@
       </div>
 
       <div class="span4">
-        <select name="regions">
+        <label for="regions"><?php echo __('Region:'); ?></label>
+        <select name="regions" id="regions">
           <option selected="selected"></option>
           <?php $regions = []; ?>
           <?php foreach ($repositories as $r) { ?>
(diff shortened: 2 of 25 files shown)


mInnes-archives:
For reference, 2937d4d05946210de5098b8432c49163d14e021a was an attempt at addressing a routing bug with the "Exclude narrower terms" button in the "Related authority records" tab in term search. It currently routes back to the "Related Archival descriptions tab" rather than staying in the correct tab. I know what's causing it, but I just haven't created a fix yet. I also forgot that I started playing with this issue on this branch.

mInnes-archives:
I'm finding that using aria-labels instead of titles is considered better practice in general. The hover text of titles is only useful for mouse users and even then it's really just a crutch for bad UI design. Assuming icons are well selected, the titles are just excess.

mInnes-archives:
I added the original titles in d73bbd252acacc817a0e2c3b4362cc52600c1a29 in #1301. SInce the view options are also used for search, it's relevant to modify this here.

mInnes-archives:
fd011f3162daa735fa3d41648b3a4c2525b3e7a6  -> 'Region' and 'Region:' are treated as two seperate dictionary entries, so you need to separate out the punctuation for the translation to work. I'll keep an eye out for these as I work on different things. It should help cut down the size of the dictionary as well as redundant work when doing translations.

apps/qubit/modules/actor/templates/_advancedSearch.php
```diff
@@ -37,7 +37,7 @@
                 <?php } ?>
               </select>
 
-              <a href="#" class="delete-criterion"><i class="fa fa-times"></i></a>
+              <a href="#" class="delete-criterion" aria-label="<?php echo __('Delete criterion'); ?>"><i class="fa fa-times"></i></a>
```
jraddaoui: Could you add `aria-hidden="true"` to the `i` element?

apps/qubit/modules/actor/templates/_advancedSearch.php
```diff
@@ -26,7 +26,7 @@
                 <option value="not"<?php echo 'not' == $item['operator'] ? ' selected="selected"' : ''; ?>><?php echo __('not'); ?></option>
               </select>
 
-              <input class="query" type="text" placeholder="<?php echo __('Search'); ?>" name="sq<?php echo $key; ?>" value="<?php echo $item['query']; ?>"/>
+              <input class="query" type="text" aria-label="<?php echo __('Search'); ?>" placeholder="<?php echo __('Search'); ?>" name="sq<?php echo $key; ?>" value="<?php echo $item['query']; ?>"/>
```
jraddaoui, line 29: Not sure if `role="search"` would apply to this input (and all the others below). Feel free to ignore.

apps/qubit/modules/repository/templates/_advancedFilters.php
```diff
@@ -8,21 +8,8 @@
...
-
-      <div class="span4">
-        <?php echo __('Regions:'); ?>
-      </div>
-    </div>
-
-    <div class="row-fluid">
-      <div class="span4">
-        <select name="thematicAreas">
+        <label for="thematicAreas"><?php echo __('Thematic area').':'; ?></label>
```
jraddaoui: I'm not sure if formatting the translations this way may be an issue for RTL languages. Same apply for other translations below.

apps/qubit/modules/actor/templates/_advancedSearch.php
jraddaoui: And if possible to all the others "affected" by this PR ...

This accessibility docs should still apply even if we're using a different Font Awesome version:

https://fontawesome.com/v5.15/how-to-use/on-the-web/other-topics/accessibility

jraddaoui (changes requested):
This is really a great improvement @mInnes-archives!

I have added a few comments below and the same, check https://github.com/artefactual/atom/pull/1312#issuecomment-867393714 for the reference and how to rebase/merge this changes.

Thanks again for all these PRs, great additions!

apps/qubit/modules/repository/templates/_advancedFilters.php
mInnes-archives: That's a good point. I'm making an unjustified assumption on punctuation placement with this change. I'll revert the separation.

apps/qubit/modules/actor/templates/_advancedSearch.php
mInnes-archives: Since `role="search"` is meant to be placed on the element containing all of the objects encompassing the search function, I've tried to implement this suggestion with this in mind (22942e8b633e09f681e0942652367f96bc8a02b8, 4b38262ce56e8b1789e4b5ece907b71bc754d57f).

I've created landmarks on both the inline search section and entirety of the advanced search section. I'm not sure if this is an appropriate approach, so I'd like to know your thoughts.

plugins/qtAccessionPlugin/modules/accession/templates/browseSuccess.php
```diff
@@ -11,7 +11,8 @@
     <div class="row">
       <div class="span6">
         <?php echo get_component('search', 'inlineSearch', [
-            'label' => __('Search %1%', ['%1%' => strtolower(sfConfig::get('app_ui_label_accession'))]), ]); ?>
+            'label' => __('Search accessions'),
+            'landmarkLabel' => __('Accession'), ]); ?>
```
jraddaoui, line 15: This is something that has been puzzling me for a while and I'd like to get @fiver-watson input on it. There are a few places where we need to use the entities' UI labels in plural, like this one. Considering, i18n this is not something you can do manually so we end up not using those labels or using them in singular.

We could add another UI label setting for each entity to include a plural version. I'd consider it another issue if Dan likes the idea, but I'd keep using the UI labels in these cases until we address that one.

plugins/qtAccessionPlugin/modules/donor/templates/browseSuccess.php
```diff
@@ -2,7 +2,7 @@
 <?php use_helper('Date'); ?>
 
 <?php slot('title'); ?>
-  <h1><?php echo __('Browse donor'); ?></h1>
+  <h1><?php echo __('Browse donors'); ?></h1>
```
jraddaoui, line 5: Another example, that wasn't even using the UI label, not even in singular.

jraddaoui (changes requested):
That looks really good to me @mInnes-archives!

I've added some comments below about the UI labels, nothing big and probably something we will want to address separately.

Thanks again, great improvements!

plugins/qtAccessionPlugin/modules/accession/templates/browseSuccess.php
mInnes-archives: Just to clarify, I removed the use of UI labels for 'accession' and 'donor' because, as far as I can tell, these labels were never actually created. I made a note of it and was thinking about adding them in a different PR. I also noticed the plural issue, but I can't immediately think of a better solution than what you've suggested.

fiver-watson: Hey both, 

I do think that any changes around UI labels and plurality should probably be addressed in a separate PR/issue, so the work isn't lost in here, and/or blooms into something much larger that no longer fits the original PR purpose. 

This has certainly been an issue. In the past it's been dealt with on a case by case basis, so the solutions aren't always consistent across entity types. Mainly I think we need to establish a clear solution that can work for all the various page types and headers, and then focus on implementing it consistently going forward, rather than trying to sneak in a fix or two in different places, as this is part of what's led us to some inconsistency across entities. I'll add a few initial thoughts below, however. 

I do think there's a certain amount of mileage we can get from simply structuring strings a bit differently - for example, "Donor browse", "Authority record browse" etc could potentially work fine for most entities, but is not currently the norm. In some cases, adding "records" could solve a lot of issues around plurality (Browse donor records, browse repository records, browse archival description records, etc).... but it's a bit ungainly, and more importantly, it wouldn't work for authority records (Browse authority record records, lol).

Adding a whole set of separate label UI elements for plural cases would be nice, but also a fair bit of work, and likely needs consideration in terms of both ongoing maintenance, and impact on upgrading users (not sure how translations, fixtures, etc will factor into all of this for example). 

All this to say: I think it makes sense to address this as a separate issue, and inform any decision by trying to gather up some concrete examples of different places where we need to consider plurality.

jraddaoui:
Thanks @fiver-watson and thanks again @mInnes-archives, great improvements!

PR #1316, opened 2021-06-15
Improve clipboard accessibility, refs #13535

mInnes-archives:
PR adds accessibility fixes and enhancements for screen readers to the clipboard module, as well as a small clipboard related improvement for table view in repository search. Improvements include:

- **Clipboard ID form no longer sharing label association with user password form.** Depending on the SR you use, this can cause behaviour such as multiple labels being heard for one form, or none at all.
- **Add legend to 'Load clipboard' fieldset.** I've opted for an 'sr-only' legend to avoid adding what could be seen as superfluous information for sighted users.
- **Add aria-label to help button.**
- **Add text to clipboard column header for table view in repository search.** Without header text this column label was either not identified or identified as 'column 5'. I've opted for 'sr-only' to keep the look of the table unchanged.

### DIFF
--- a/apps/qubit/modules/clipboard/actions/loadAction.class.php
+++ b/apps/qubit/modules/clipboard/actions/loadAction.class.php
@@ -21,7 +21,7 @@ class ClipboardLoadAction extends DefaultEditAction
 {
     // Arrays not allowed in class constants
     public static $NAMES = [
-        'password',
+        'clipboardPassword',
         'mode',
     ];
 
@@ -98,9 +98,9 @@ class ClipboardLoadAction extends DefaultEditAction
     protected function addField($name)
     {
         switch ($name) {
-            case 'password':
-                $this->form->setValidator('password', new sfValidatorString(['required' => true]));
-                $this->form->setWidget('password', new sfWidgetFormInput());
+            case 'clipboardPassword':
+                $this->form->setValidator('clipboardPassword', new sfValidatorString(['required' => true]));
+                $this->form->setWidget('clipboardPassword', new sfWidgetFormInput());
 
                 break;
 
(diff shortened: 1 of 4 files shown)


apps/qubit/modules/clipboard/actions/loadAction.class.php
```diff
@@ -98,9 +98,9 @@ public function execute($request)
     protected function addField($name)
     {
         switch ($name) {
-            case 'password':
-                $this->form->setValidator('password', new sfValidatorString(['required' => true]));
-                $this->form->setWidget('password', new sfWidgetFormInput());
+            case 'clipboardPassword':
+                $this->form->setValidator('clipboardPassword', new sfValidatorString(['required' => true]));
+                $this->form->setWidget('clipboardPassword', new sfWidgetFormInput());
```
jraddaoui, line 103: Nice! I believe this was also throwing a warning in the browser's dev tools.

jraddaoui (approved):
Great! Thanks @mInnes-archives!

Please, see https://github.com/artefactual/atom/pull/1312#issuecomment-867393714 and use the same ticket for the reference.

apps/qubit/modules/repository/templates/_browseTableView.php
```diff
@@ -42,6 +42,7 @@
       </th>
 
       <th style="width 3%">
+        <span class="sr-only"><?php echo __('Clipboard'); ?></span>
```
jraddaoui, line 45: I didn't know this was available in Bootstrap 2!

mInnes-archives: Apparently it's also part of Font Awesome.

mInnes-archives:
This one's good to go.

jraddaoui:
Thanks @mInnes-archives!

PR #1317, opened 2021-06-15
CSV Checker WebUI integration, refs #13508

sbreker:
Add logic to CSV import background job to run the validator prior to import if configured. Also adds a dedicated 'Validate CSV' task that can be run from the WebUI. Depending on 'permissive' or 'strict' setting, the validator results will prevent the import from proceeding. The verbose validator report will be zipped and linked as a download from the import and validate job results.


mcantelon (approved):
Looks good to me!

PR #1322, opened 2021-06-18
Use render_value_html in alerts, refs #13495

jraddaoui:
User input should not be used in flash alerts but that's hard to control.
By using `render_value_html` when Markdown support is enabled, even
with Parsedown's safeMode disabled, the content is sligthly sanitized.
This does not fully escape malicious code, specially with Markdown
disabled, but it mnay prevent XSS attacks in some cases.


jraddaoui:
@djjuhasz, I just remembered this minor change you reviewed a while ago in Gitlab.

djjuhasz (approved):
Looks good! 👍

jraddaoui:
Thanks @djjuhasz!

Issue #1323, opened 2021-06-18
Problem: Physical object CSV import unit test failing in Windows


djjuhasz:
### AtoM version
qa/2.x (2e2ce4a)

### Installation environment
Docker compose environment (Docker Desktop for Windows v20.10.7; PHP 7.4.20 (cli); Percona 8.0.22-13; Elasticsearch v5.6.16; Windows 10 Home, version 2004)

### Steps to reproduce the issue

1. Run the PHPunit test suite:
```
docker-compose exec atom composer test
```

###  The resulting error 
```
PHPUnit 9.5.4 by Sebastian Bergmann and contributors.

...........................F...................................  63 / 165 ( 38%)
............................................................... 126 / 165 ( 76%)
.......................................                         165 / 165 (100%)

Time: 00:00.081, Memory: 20.00 MB

There was 1 failure:

1) PhysicalObjectCsvImporterTest::testSetFilenameFileUnreadableException
Failed asserting that exception of type "sfException" is thrown.

FAILURES!
Tests: 165, Assertions: 400, Failures: 1.
Script phpunit handling the test event returned with error code 1
```

### Expected outcome

All tests should pass

jraddaoui:
Thanks @djjuhasz,

This is more related to the Docker environment than Windows, I created [this Redmine ticket](https://projects.artefactual.com/issues/13492) a while ago, but this raises another issue about the containers running by the root user. 

Also, I'm a big fan of using Github issues but, have you discussed creating this in here with someone? @fiver-watson still uses Redmine for the release notes and ticket list and, while one could argue if this should go to the release notes, I believe we have been including this kind of changes in there.

djjuhasz:
@jraddaoui oh great, I'll use the Redmine ticket number instead.

I haven't discussed using Github issues with anyone, but at this point projects.artefactual.com is so slow that I find it really annoying to use it. 😬   I was hoping to get by without a Redmine ticket on this one because it has nothing to do with the actual functionality of the software, it's just a glitch in the unit test.

PR #1324, opened 2021-06-18
Fix test for unreadable filename, refs #13492

djjuhasz:
On Windows phpunit runs as the root user, so the virtual "root.csv" file
was readable, causing the test to fail. I've changed the name of the
virtual file to "unreadable.csv" and set the file owner to
vfsStream::OWNER_USER_1 to trigger the unreadable file exception in the
test.

### DIFF
--- a/test/phpunit/PhysicalobjectCsvImporterTest.php
+++ b/test/phpunit/PhysicalobjectCsvImporterTest.php
@@ -59,7 +59,7 @@ class PhysicalObjectCsvImporterTest extends \PHPUnit\Framework\TestCase
                 "\n",
                 $this->csvData + $this->csvData
             ),
-            'root.csv' => $this->csvData[0],
+            'unreadable.csv' => $this->csvData[0],
             'error.log' => '',
         ];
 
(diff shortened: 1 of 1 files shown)


test/phpunit/PhysicalobjectCsvImporterTest.php
```diff
@@ -59,17 +59,17 @@ public function setUp(): void
...
 
         // setup and cache the virtual file system
         $this->vfs = vfsStream::setup('root', null, $directory);
 
         // Make 'root.csv' owned and readable only by root user
-        $file = $this->vfs->getChild('root/root.csv');
+        $file = $this->vfs->getChild('root/unreadable.csv');
         $file->chmod('0400');
-        $file->chown(vfsStream::OWNER_ROOT);
+        $file->chown(vfsStream::OWNER_USER_1);
```
jraddaoui, line 72: What if the tests are run by user 1?

jraddaoui (commented):
Thanks @djjuhasz!

Based on the outcome from https://github.com/artefactual/atom/issues/1323#issuecomment-864363125, we may want to reference the Redmine ticket ... https://projects.artefactual.com/issues/13492

I added a question bellow, I struggled with this issue when I found it, I'm not sure of the value of this particular case, and my first thoughts were to remove it or address the "containers run by root" issue.

test/phpunit/PhysicalobjectCsvImporterTest.php
djjuhasz: Yeah, the test will fail if run by vfsStream::OWNER_USER_1 (which refs UID 1).  UID is a reserved system UID though, so I assume there is very little chance that the test will ever be run by that user.   If my assumption is wrong, and UID 1 is used to run the test suite in some cases, then I agree that we need a better solution.

It's not ideal, but a lot easier than "fixing" the Docker "run as root" issue.

jraddaoui: Right, lets hope that's okay, I agree than fixing the Docker root issue can wait for another time, heheh.

jraddaoui (approved):
Thanks @djjuhasz! Please don't forget changing the reference.

PR #1325, opened 2021-06-19
Avoid routing exception in purge task, refs #13452

jraddaoui:
Context creation causes route initialization, which hits the database.
As in the install task, this exception can be avoided with the
`app_avoid_routing_propel_exceptions` config setting.


sbreker (approved):
Good catch @jraddaoui!  👍

PR #1327, opened 2021-06-22
Varnish cache bypass support, refs #13519

mcantelon:
Added setting of a cookie, when a user is logged in, to let a reverse proxy know to let the user bypass the cache. Updated Varnish VCL, for use with AtoM running via Docker, to allow cache bypassing.


jraddaoui (approved):
Nice! Thanks @mcantelon!

PR #1328, opened 2021-06-23
Add CLI task to generate finding aids, refs #13500

djjuhasz:
- Add `finding-aid:generate SLUG` CLI task with options for format (PDF or RTF) and "model" ("inventory-summary" or "full-details")
- Refactor general finding aid code into QubitFindingAid class
- Refactor finding aid generation code into QubitFindingAidGenerator
  class
- Use QubitFindingAid and QubitFindingAidGenerator libraries in
  arFindingAidJob
- Add a finding-aid:generate task to generate a new finding aid from
  the CLI using the new libraries
- When generating a new finding aid, delete the old finding aid first
  to avoid having two finding aid files for the same resource (e.g.
  foo.pdf and foo.rtf)
- Add unit tests

### DIFF
--- a/apps/qubit/modules/informationobject/actions/findingAidComponent.class.php
+++ b/apps/qubit/modules/informationobject/actions/findingAidComponent.class.php
@@ -21,18 +21,23 @@ class InformationObjectFindingAidComponent extends sfComponent
 {
     public function execute($request)
     {
+        $this->showDownload = false;
+        $this->showStatus = false;
+        $this->showUpload = false;
+        $this->showGenerate = false;
+        $this->showDelete = false;
+
         // Get finding aid data from top-level
         if (QubitInformationObject::ROOT_ID != $this->resource->parentId) {
             $this->resource = $this->resource->getCollectionRoot();
         }
 
-        $this->showDownload = $this->showStatus = $this->showUpload = $this->showGenerate = $this->showDelete = false;
-
-        $this->path = arFindingAidJob::getFindingAidPathForDownload($this->resource->id);
+        $findingAid = new QubitFindingAid($this->resource);
 
         // Public users can only see the download link if the file exists
         if (!$this->getUser()->isAuthenticated()) {
-            if (isset($this->path)) {
+            if (!empty($findingAid->getPath())) {
+                $this->path = $findingAid->getPath();
                 $this->showDownload = true;
 
                 return;
@@ -47,7 +52,8 @@ class InformationObjectFindingAidComponent extends sfComponent
         // show allowed actions or nothing
         if (!isset($lastJobStatus)) {
             // Edge case where the job status is missing but the file exists,
-            if (isset($this->path)) {
+            if (!empty($findingAid->getPath())) {
+                $this->path = $findingAid->getPath();
                 $this->showDownload = true;
 
                 // Check ACL to show delete option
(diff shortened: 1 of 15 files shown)


apps/qubit/modules/informationobject/actions/findingAidLinkComponent.class.php
```diff
@@ -26,29 +26,32 @@ public function execute($request)
             $this->resource = $this->resource->getCollectionRoot();
         }
 
-        $this->path = arFindingAidJob::getFindingAidPathForDownload($this->resource->id);
+        $findingAid = new QubitFindingAid($this->resource);
+        $this->path = $findingAid->getPath($this->resource->id);
```
mcantelon: The `getPath` method doesn't seem to require an argument.

apps/qubit/modules/informationobject/actions/findingAidComponent.class.php
```diff
@@ -100,30 +102,33 @@ public function execute($request)
...
                     break;
 
-                case arFindingAidJob::UPLOADED_STATUS:
-                    $this->status = $i18n->__('Uploaded');
+                case QubitFindingAid::UPLOADED_STATUS:
+                    $this->status = $this->context->i18n->__('Uploaded');
 
                     break;
-                // It should never get here if we don't add more finding aid statuses
+                // It should never get here if we don't add more finding aid
```
mcantelon, line 128: Should we throw an exception if this is the case given it shouldn't happen?

lib/QubitFindingAid.class.php
```diff
@@ -0,0 +1,460 @@
...
+    }
+
+    /**
+     * Get the Finding Aid file format (pdf or rtf).
+     */
+    public function getFormat()
+    {
+        // If not set, try and determine the format from the file extension
+        if (!isset($this->format) && !empty($this->getPath())) {
+            $parts = pathinfo($this->getPath());
```
mcantelon: This could use `pathinfo($this->getPath(), PATHINFO_EXTENSION)` (trick I learned from Steve).

lib/QubitFindingAid.class.php
```diff
@@ -0,0 +1,460 @@
...
+    }
+
+    /**
+     * Set the Finding Aid status.
+     *
+     * @param $status either self::GENERATED_STATUS or self::UPLOADED_STATUS
+     */
+    public function setStatus(int $status)
+    {
+        // Search for an existing 'findingAidStatus' property
```
mcantelon, line 218: This property setting code seems the same as around line 341... maybe it could be a helper method.

mcantelon (changes requested):
Some possible minor changes but overall nicely done!

apps/qubit/modules/informationobject/actions/findingAidLinkComponent.class.php
djjuhasz: Good catch, thanks!

apps/qubit/modules/informationobject/actions/findingAidComponent.class.php
djjuhasz: The purpose of this code is to show a status message to logged in users on the description page, like this:

Screenshot_finding_aid_status

If there is bad status data in the database I think it's better not to break the page by throwing an exception at this point.

lib/QubitFindingAid.class.php
djjuhasz: Slick! 👍

lib/QubitFindingAid.class.php
djjuhasz: Yeah, good call. :)

apps/qubit/modules/informationobject/actions/findingAidComponent.class.php
mcantelon: Ah cool... makes sense!

mcantelon (approved):
Looks good!

PR #1329, opened 2021-06-24
Make Uppy work in the B5 theme

sbreker:
- Removed reference to javascript 'Qubit' data structure used to pass
data from the template to multiFileUpload.js.
- Add references to vendor/uppy minified js and CSS.

Going forward we would want to install Uppy Core, Dashboard and
XHRUpload via NPM but this will require aidding a .js bundler.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5,7 +5,6 @@
   "requires": true,
   "packages": {
     "": {
-      "name": "atom",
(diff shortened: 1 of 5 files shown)


jraddaoui (approved):
Awesome, thanks @sbreker! Only a few complains from php-cs-fixer about some parenthesis.

PR #1330, opened 2021-06-25
Make MediaElement work on B5 theme

sbreker:
- Added MediaElement to package.json to install via npm.
- Removed refs to MediaElement js and css from the show* actions.
- Added class 'mejs__player' to the show* templates on <video> and
<audio> elements.
- Load options using the <video> and <audio> element attribute
'data-mejsoptions'.
- Added reference to MediaElement minified js and css to B5 plugin start
and end templates.


jraddaoui (approved):
Awesome, really nice with the auto start!

PR #1331, opened 2021-06-28
Rename PHP CS Fixer config file, refs #13537

djjuhasz:
Rename ".php_cs" to ".php-cs-fixer.dist.php" for PHP-CS-Fixer v3
compatibility.

PHP-CS-Fixer version 3 has changed the naming convention for their
configuration files to ".php-cs-fixer.php" and
".php-cs-fixer.dist.php". I believe we should be using the "distributed"
configuration file name for the public AtoM code.

Reference:
https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/3.0/UPGRADE-v3.md

### DIFF
--- a/.php-cs-fixer.dist.php
+++ b/.php-cs-fixer.dist.php
@@ -11,7 +11,7 @@ $finder = PhpCsFixer\Finder::create()
 ;
 
 // Indentation inside switch blocks and some multiline
-// elements (control statements, assignments, strings) 
+// elements (control statements, assignments, strings)
 // is not considered. See:
 // https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/776
 // https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/4502
@@ -19,10 +19,10 @@ $finder = PhpCsFixer\Finder::create()
 // PhpCsFixer's method_argument_space default value,
 // which includes ensure_fully_multiline, causes
 // inconsistencies in templates.
-return PhpCsFixer\Config::create()
+return (new PhpCsFixer\Config())
     ->setRules([
         '@PhpCsFixer' => true,
-        'method_argument_space' => true,
+        'method_argument_space' => ['on_multiline' => 'ignore'],
     ])
     ->setFinder($finder)
 ;
--- a/composer.json
+++ b/composer.json
@@ -1,8 +1,11 @@
 {
     "name": "artefactual/accesstomemory",
     "type": "project",
-    "description" : "A web-based, open source application for standards-based archival description.",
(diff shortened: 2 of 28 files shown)


djjuhasz:
@jraddaoui the dry run is returning error code 8 (some files need fixing).  When I run the fix command locally, it modifies 83 files and some of the modifications to the templates look worse than the originals. :(   How did you deal with autoformatting issues when you implemented php-cs-fixer?  Can you manually tweak some of the formatting, or can we make an exception for the rules creating bad results, or do we just live with the autoformatted code?

jraddaoui:
Thanks for looking into this @djjuhasz!

In the initial implementation I went rule by rule until I found the problematic one (the only one disabled in the config). I kept the entire commit history on merge (https://github.com/artefactual/atom/commit/85c0ba69a5fa414596d0896225e6bf428b0d57d8) but I don't think it will help you now, since it doesn't include the actual rules being addressed in the `php_cs` config file.

I've not looked at the changes locally and the check details don't include any diff so I'm not sure how badly it looks, but I'd try to find the problematic rule in this case if it's really bad, to see if we can tweak the config like before.

.php-cs-fixer.dist.php
```diff
@@ -11,15 +11,16 @@
...
 // is not considered. See:
 // https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/776
 // https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/4502
 //
 // PhpCsFixer's method_argument_space default value,
 // which includes ensure_fully_multiline, causes
 // inconsistencies in templates.
-return PhpCsFixer\Config::create()
+$config = new PhpCsFixer\Config();
+return $config
```
jraddaoui: Nitpick: you could chain the constructor using parenthesis ...

```php
return (new PhpCsFixer\Config())
    ->setRules([...
```

https://www.php.net/manual/en/language.oop5.basic.php#example-173

djjuhasz: @jraddaoui okay, I've implemented your suggestion in more recent commits.

composer.json
```diff
@@ -26,4 +29,4 @@
         "test": "phpunit",
         "php-cs-fixer": "php-cs-fixer"
     }
-}
+}
```
jraddaoui: Nitpick: missing new line.

jraddaoui (approved):
Great! Thanks @djjuhasz!

composer.json
djjuhasz: Thanks, fixed.  Strangely VS Code seems to delete the final newline from `.js`d files by default.  I had to turn on the `files.insertFinalNewline` option.

PR #1333, opened 2021-07-02
Fix CSV transformation encoding, refs #13538

mcantelon:
Fixed problem with CSV transformations defaulting to being converted from using Windows character set.


sbreker (approved):
Looks good @mcantelon!  👍

PR #1334, opened 2021-07-02
Fix job error display, refs #13523

mcantelon:
Fixed, on the job detail page, a minor issue with error display. In the black section of the page, showing all the job's logging output, errors were being shown with the category "[info]". This has been changed so they show as an error category.


jraddaoui (approved):
Nice, thanks @mcantelon!

PR #1335, opened 2021-07-03
Avoid removal of qtAclPlugin, refs #13518

jraddaoui:
If the theme plugin was not yet included in the enabled plugins list,
when it was moved to the top to allow plugin customizations it was
replacing the qtAclPlugin. Only unset when the plugin is found and
use `array_merge` to add it on top.


sbreker (approved):
Great catch @jraddaoui!  👍

PR #1336, opened 2021-07-05
Add CSV validator for repo, refs #13475

sbreker:
Add new validator for the repository column. Checks if the CSV import
will result in new AtoM repository records being created and, if so,
will trigger a warning. This validator is only valid for information
object import.

Includes:
- new phpunit test for the new validator.
- fix to limit the scriptOfDescription check to info obj CSVs only.

### DIFF
--- a/lib/task/import/validate/csvRepoValidator.class.php
+++ b/lib/task/import/validate/csvRepoValidator.class.php
@@ -1,0 +1,110 @@
+<?php
+
+/*
+ * This file is part of the Access to Memory (AtoM) software.
+ *
+ * Access to Memory (AtoM) is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * Access to Memory (AtoM) is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with Access to Memory (AtoM).  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+/**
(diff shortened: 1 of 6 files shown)


lib/task/import/validate/csvRepoValidator.class.php
```diff
@@ -0,0 +1,113 @@
...
+    public function __construct(?array $options = null)
+    {
+        $this->setTitle(self::TITLE);
+
+        parent::__construct($options);
+    }
+
+    public function reset()
+    {
+        //$this->values = [];
```
mcantelon: Dead code

lib/task/import/validate/csvRepoValidator.class.php
```diff
@@ -0,0 +1,113 @@
...
+
+        if (!$this->repositoryExists($row['repository'])) {
+            $this->testData->addDetail(implode(',', $row));
+        }
+    }
+
+    public function getTestResult()
+    {
+        if (!$this->repositoryColumnPresent) {
+            // repository column not present in file.
```
mcantelon: Nitpick: IIRC comments need to be capitalized

lib/task/import/validate/csvRepoValidator.class.php
```diff
@@ -0,0 +1,113 @@
...
+            $this->testData->addResult(sprintf("'repository' column not present in file."));
+
+            return parent::getTestResult();
+        }
+
+        if (0 < count($this->newRepositories)) {
```
mcantelon, line 78: Nitpick: These two conditionals could be combined as an if/else I think.

mcantelon (approved):
Aside from a few nitpicks, looks good!

lib/task/import/validate/csvRepoValidator.class.php
sbreker: Correct - I will combine these.

lib/task/import/validate/csvRepoValidator.class.php
sbreker: Fixed.

lib/task/import/validate/csvRepoValidator.class.php
sbreker: Will remove...

PR #1338, opened 2021-07-13
Replace Imageflow with Accessible-Slick

sbreker:
Using two Accessible-Slick sliders:
- nav slider has large images and is used to select image
- clicking nav slider image will display Description view page.
- second 'for' slider displays digital object title.
- instructionsText is translatable.
- prev next buttons translatable
- Use alt text on carousel images; fall back to title when empty
- Added ability to collapse the carousel section. Open by default.

### DIFF
--- a/COPYRIGHT
+++ b/COPYRIGHT
@@ -73,6 +73,12 @@ Url:       http://imageflow.finnrudolph.de/
 Copyright: Finn Rudolph
 License:   Creative Commons Attribution-NonCommercial 3.0 Unported
 
+Accessible-Slick
+----------------
+Url:       https://accessible360.github.io/accessible-slick
+Copyright: 2013-2016, Accessible360
+License:   MIT
+
 jQuery
 ------
 Url:       http://jquery.com/
--- a/package-lock.json
+++ b/package-lock.json
@@ -5,9 +5,9 @@
   "requires": true,
   "packages": {
     "": {
-      "name": "atom",
       "version": "0.0.0",
       "dependencies": {
+        "@accessible360/accessible-slick": "^1.0.1",
         "@fortawesome/fontawesome-free": "^5.15.3",
         "bootstrap": "^5.0.1",
         "bootstrap-autocomplete": "^2.3.7",
@@ -23,44 +23,12 @@
         "sass": "^1.35.1"
       }
     },
-    "node_modules/@cypress/listr-verbose-renderer": {
-      "version": "0.4.1",
-      "resolved": "https://registry.npmjs.org/@cypress/listr-verbose-renderer/-/listr-verbose-renderer-0.4.1.tgz",
-      "integrity": "sha1-p3SS9LEdzHxEajSz4ochr9M8ZCo=",
-      "dev": true,
-      "dependencies": {
-        "chalk": "^1.1.3",
-        "cli-cursor": "^1.0.2",
-        "date-fns": "^1.27.2",
-        "figures": "^1.7.0"
-      },
-      "engines": {
-        "node": ">=4"
-      }
-    },
-    "node_modules/@cypress/listr-verbose-renderer/node_modules/chalk": {
-      "version": "1.1.3",
-      "resolved": "https://registry.npmjs.org/chalk/-/chalk-1.1.3.tgz",
-      "integrity": "sha1-qBFcVeSnAv5NFQq9OHKCKn4J/Jg=",
-      "dev": true,
-      "dependencies": {
-        "ansi-styles": "^2.2.1",
-        "escape-string-regexp": "^1.0.2",
-        "has-ansi": "^2.0.0",
-        "strip-ansi": "^3.0.0",
-        "supports-color": "^2.0.0"
-      },
-      "engines": {
-        "node": ">=0.10.0"
-      }
-    },
-    "node_modules/@cypress/listr-verbose-renderer/node_modules/supports-color": {
-      "version": "2.0.0",
-      "resolved": "https://registry.npmjs.org/supports-color/-/supports-color-2.0.0.tgz",
-      "integrity": "sha1-U10EXOa2Nj+kARcIRimZXp3zJMc=",
-      "dev": true,
-      "engines": {
-        "node": ">=0.8.0"
+    "node_modules/@accessible360/accessible-slick": {
+      "version": "1.0.1",
+      "resolved": "https://registry.npmjs.org/@accessible360/accessible-slick/-/accessible-slick-1.0.1.tgz",
+      "integrity": "sha512-fAdHKy0CLuApCHFgRs2zcQpzjEdlZUuJND1zIY8T5PiV8bB52gpomjCUFCZwCk9lb87WC0IJGIUaPD6RToocqA==",
+      "peerDependencies": {
+        "jquery": ">=1.8.0"
       }
     },
     "node_modules/@cypress/request": {
@@ -132,22 +100,10 @@
         "url": "https://opencollective.com/popperjs"
       }
     },
-    "node_modules/@samverschueren/stream-to-observable": {
-      "version": "0.3.1",
-      "resolved": "https://registry.npmjs.org/@samverschueren/stream-to-observable/-/stream-to-observable-0.3.1.tgz",
-      "integrity": "sha512-c/qwwcHyafOQuVQJj0IlBjf5yYgBI7YPJ77k4fOJYesb41jio65eaJODRUmfYKhTOFBrIZ66kgvGPlNbjuoRdQ==",
-      "dev": true,
-      "dependencies": {
-        "any-observable": "^0.3.0"
-      },
-      "engines": {
-        "node": ">=6"
-      }
-    },
     "node_modules/@types/node": {
-      "version": "14.14.37",
-      "resolved": "https://registry.npmjs.org/@types/node/-/node-14.14.37.tgz",
-      "integrity": "sha512-XYmBiy+ohOR4Lh5jE379fV2IU+6Jn4g5qASinhitfyO71b/sCo6MKsMLF5tc7Zf2CE8hViVQyYSobJNke8OvUw==",
+      "version": "14.17.4",
+      "resolved": "https://registry.npmjs.org/@types/node/-/node-14.17.4.tgz",
+      "integrity": "sha512-8kQ3+wKGRNN0ghtEn7EGps/B8CzuBz1nXZEIGGLP2GnwbqYn4dbTs7k+VKLTq1HvZLRCIDtN3Snx1Ege8B7L5A==",
       "dev": true
     },
     "node_modules/@types/sinonjs__fake-timers": {
@@ -157,10 +113,33 @@
       "dev": true
     },
     "node_modules/@types/sizzle": {
-      "version": "2.3.2",
-      "resolved": "https://registry.npmjs.org/@types/sizzle/-/sizzle-2.3.2.tgz",
-      "integrity": "sha512-7EJYyKTL7tFR8+gDbB6Wwz/arpGa0Mywk1TJbNzKzHtzbwVmY4HR9WqS5VV7dsBUKQmPNr192jHr/VpBluj/hg==",
+      "version": "2.3.3",
+      "resolved": "https://registry.npmjs.org/@types/sizzle/-/sizzle-2.3.3.tgz",
+      "integrity": "sha512-JYM8x9EGF163bEyhdJBpR2QX1R5naCJHC8ucJylJ3w9/CVBaskdQ8WqBf8MmQrd1kRvp/a4TS8HJ+bxzR7ZJYQ==",
       "dev": true
+    },
+    "node_modules/@types/yauzl": {
+      "version": "2.9.1",
+      "resolved": "https://registry.npmjs.org/@types/yauzl/-/yauzl-2.9.1.tgz",
+      "integrity": "sha512-A1b8SU4D10uoPjwb0lnHmmu8wZhR9d+9o2PKBQT2jU5YPTKsxac6M2qGAdY7VcL+dHHhARVUDmeg0rOrcd9EjA==",
+      "dev": true,
+      "optional": true,
+      "dependencies": {
+        "@types/node": "*"
+      }
+    },
+    "node_modules/aggregate-error": {
+      "version": "3.1.0",
+      "resolved": "https://registry.npmjs.org/aggregate-error/-/aggregate-error-3.1.0.tgz",
+      "integrity": "sha512-4I7Td01quW/RpocfNayFdFVk1qSuoh0E7JrbRJ16nH01HhKFQ88INq9Sd+nd72zqRySlr9BmDA8xlEJ6vJMrYA==",
+      "dev": true,
+      "dependencies": {
+        "clean-stack": "^2.0.0",
+        "indent-string": "^4.0.0"
+      },
+      "engines": {
+        "node": ">=8"
+      }
     },
     "node_modules/ajv": {
       "version": "6.12.6",
@@ -172,6 +151,10 @@
         "fast-json-stable-stringify": "^2.0.0",
         "json-schema-traverse": "^0.4.1",
         "uri-js": "^4.2.2"
+      },
+      "funding": {
+        "type": "github",
+        "url": "https://github.com/sponsors/epoberezkin"
       }
     },
     "node_modules/amdefine": {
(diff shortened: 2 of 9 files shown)


COPYRIGHT
```diff
@@ -73,6 +73,12 @@ Url:       http://imageflow.finnrudolph.de/
 Copyright: Finn Rudolph
 License:   Creative Commons Attribution-NonCommercial 3.0 Unported
 
+Accessible-Slick
+----------------
+Url:       https://accessible360.github.io/accessible-slick
+Copyright: 2013-2016, Accessible360
+License:   MIT
+
```
jraddaoui, line 81: Nice, I didn't do this for Bootstrap Autocomplete and maybe others, so I'll do that some point.

package-lock.json
```diff
@@ -5,9 +5,9 @@
   "requires": true,
   "packages": {
     "": {
-      "name": "atom",
```
jraddaoui, line 8: We keep fighting for this line as my NPM version adds it and yours removes it ...

```
$ dc exec atom npm --version
7.19.1
```

plugins/arDominionB5Plugin/js/imageflow.js
```diff
@@ -0,0 +1,35 @@
+(function ($) {
```
jraddaoui: Nitpick: you could use an arrow function here too.

plugins/arDominionB5Plugin/js/imageflow.js
```diff
@@ -0,0 +1,35 @@
...
+  'use strict';
+  $(() => {
+    var node = $('div.atom-carousel');
```
jraddaoui: The `div` doesn't add a lot of value and I'd like to see this using an id instead of a class, since it's a very particular implementation for the IO page that I doubt we'll replicate.

plugins/arDominionB5Plugin/js/imageflow.js
```diff
@@ -0,0 +1,35 @@
...
+
+    if ($(node).length) {
+      $('.atom-slider-nav').slick({
+        slidesToShow: 3,
+        slidesToScroll: 1,
+        asNavFor: '.atom-slider-for',
```
jraddaoui: Same, maybe `.atom-slider-title` gives a bit more info. And they could be ids too.

plugins/arDominionB5Plugin/js/imageflow.js
```diff
@@ -0,0 +1,35 @@
+(function ($) {
+  'use strict';
+  $(() => {
+    var node = $('div.atom-carousel');
+
+    if ($(node).length) {
+      $('.atom-slider-nav').slick({
```
jraddaoui: Maybe `.atom-slider-images` gives a bit more info.

plugins/arDominionB5Plugin/js/imageflow.js
```diff
@@ -0,0 +1,35 @@
...
+        slidesToShow: 3,
+        slidesToScroll: 1,
+        asNavFor: '.atom-slider-for',
+        dots: true,
+        centerMode: true,
+        instructionsText: $(node).data('carousel-instructions-text-image-link'),
+        regionLabel: 'image carousel',
```
jraddaoui: Oh, this should be translated too ...

jraddaoui: And, maybe "Archival description images carousel"

plugins/arDominionB5Plugin/js/imageflow.js
```diff
@@ -0,0 +1,35 @@
...
+        centerMode: true,
+        slidesToShow: 1,
+        slidesToScroll: 1,
+        draggable: false,
+        swipe: false,
+        arrows: false,
+        fade: true,
+        asNavFor: '.atom-slider-nav',
+        instructionsText: $(node).data('carousel-instructions-text-text-link'),
+        regionLabel: 'image title'
```
jraddaoui: And this should also be translated. And maybe it may be better with "Archival description title display/link".

plugins/arDominionB5Plugin/js/imageflow.js
```diff
@@ -0,0 +1,35 @@
...
+
+      $('.atom-slider-for').slick({
+        centerMode: true,
+        slidesToShow: 1,
+        slidesToScroll: 1,
+        draggable: false,
+        swipe: false,
+        arrows: false,
+        fade: true,
+        asNavFor: '.atom-slider-nav',
```
jraddaoui: Not sure if this is right anymore after we prevented the drag and slide in this carousel. We should check for what is this being used. It may be targeting the other container from within an Aria attribute, which will probably make the use of ids a requirement?

plugins/arDominionB5Plugin/modules/digitalobject/templates/_imageflow.php
```diff
@@ -0,0 +1,45 @@
+<?php use_helper('Text'); ?>
```
jraddaoui: Do you know for what is this needed?

plugins/arDominionB5Plugin/modules/digitalobject/templates/_imageflow.php
```diff
@@ -0,0 +1,45 @@
...
+
+<button class="btn atom-btn-secondary mb-3" type="button" data-bs-toggle="collapse" data-bs-target="#collapseExample" aria-expanded="true" aria-controls="collapseExample">
```
jraddaoui: I'd be great to not use `example` in the ids. Just `collapse-carousel` or something like that. I know they use camel case in the attribute values in some cases, but I'd stick to kebab case when possible for values too.

plugins/arDominionB5Plugin/modules/digitalobject/templates/_imageflow.php
```diff
@@ -0,0 +1,45 @@
...
+    <div class="atom-carousel"
+      data-carousel-instructions-text-text-link="<?php echo __('Clicking this description title link will open the description view page for this digital object. Advancing the carousel above will update this title text.'); ?>"
+      data-carousel-instructions-text-image-link="<?php echo __('Changing the current slide of this carousel will change the description title displayed in the following carousel. Clicking any image in this carousel will open the related description view page.'); ?>"
+      data-carousel-next-arrow-button-text="<?php echo __('Next'); ?>"
+      data-carousel-prev-arrow-button-text="<?php echo __('Previous'); ?>">
+
+      <div class="atom-slider-nav">
+        <?php foreach ($thumbnails as $item) { ?>
+          <a title="<?php echo $item->parent->object; ?>" href="<?php echo url_for([$item->parent->object, 'module' => 'informationobject']); ?>">
+            <?php echo image_tag($item->getFullPath(), ['class' => 'img-thumbnail mx-2', 'longdesc' => url_for([$item->parent->object, 'module' => 'informationobject']), 'alt' => strip_markdown($item->getDigitalObjectAltText() ?: $item->parent->object)]); ?>
```
jraddaoui, line 18: First time I see the Elvis operator in AtoM ... `?:`

plugins/arDominionB5Plugin/modules/digitalobject/templates/_imageflow.php
```diff
@@ -0,0 +1,45 @@
...
+          </a>
+        <?php } ?>
+      </div>
+
+      <div class="atom-slider-for">
+        <?php foreach ($thumbnails as $item) { ?>
+          <a href="<?php echo url_for([$item->parent->object, 'module' => 'informationobject']); ?>" class="text-white">
+            <h6 class="text-center mt-1"><?php echo strip_markdown($item->parent->object); ?></h6>
```
jraddaoui: I'd add the two classes to the link and remove the heading, as it may be confusing for heading navigation and level 6 is the same as normal text.

plugins/arDominionB5Plugin/modules/digitalobject/templates/_imageflow.php
```diff
@@ -0,0 +1,45 @@
+<?php use_helper('Text'); ?>
+
+<button class="btn atom-btn-secondary mb-3" type="button" data-bs-toggle="collapse" data-bs-target="#collapseExample" aria-expanded="true" aria-controls="collapseExample">
+  Expand image carousel
```
jraddaoui: It's already expanded by default and the status is actually indicated with the `aria-expanded` attribute. I'd just say "Images carousel" or "Toggle images carousel".

plugins/arDominionB5Plugin/scss/_layout.scss
```diff
@@ -47,3 +47,36 @@
   margin-top: $spacer;
   padding: $spacer;
 }
+
+.atom-slider-for,
+.atom-slider-nav {
```
jraddaoui: I'd target the entire carousel directly.

plugins/arDominionB5Plugin/scss/_layout.scss
```diff
@@ -47,3 +47,36 @@
...
+  .slick-track {
+    display: flex;
+    align-items: center;
+    justify-content: center;
+  }
+
+  button {
```
jraddaoui: Since we're no longer adding other styles to buttons, I'd target the arrows directly.

plugins/arDominionB5Plugin/scss/_layout.scss
```diff
@@ -47,3 +47,36 @@
...
+      left: -30px;
+    }
+
+    &.slick-arrow.slick-next {
+      right: -30px;
+    }
+  }
+}
+
+.atom-slider-nav {
```
jraddaoui: I'd put this inside of the `.slick-track`.

plugins/arDominionB5Plugin/scss/_layout.scss
```diff
@@ -47,3 +47,36 @@
...
+  img {
+    max-width: 400px;
+    max-height: 300px;
```
jraddaoui: This may require some responsive design changes based on the screen size, but we can revisit later.

plugins/arDominionB5Plugin/scss/_layout.scss
```diff
@@ -47,3 +47,36 @@
...
+  }
+}
+
+.atom-slider-for {
+  .slick-slide {
+    display: flex;
+    justify-content: center;
+  }
+}
```
jraddaoui, line 69: I think this is no longer needed, as it's already being affected by lines 54 to 56.

plugins/arDominionB5Plugin/modules/digitalobject/templates/_imageflow.php
```diff
@@ -0,0 +1,45 @@
...
+          <a href="<?php echo url_for([$item->parent->object, 'module' => 'informationobject']); ?>" class="text-white">
+            <h6 class="text-center mt-1"><?php echo strip_markdown($item->parent->object); ?></h6>
+          </a>
+        <?php } ?>
+      </div>
+
+      <?php if (isset($limit) && $limit < $total) { ?>
+        <div class="text-white text-center mt-1">
+          <?php echo __('Results %1% to %2% of %3%', ['%1%' => 1, '%2%' => $limit, '%3%' => $total]); ?>
+          <a class='btn atom-btn-outline-light btn-sm' href="<?php echo url_for([
```
jraddaoui: Add `ms-2` to separate it a bit from the text.

plugins/arDominionB5Plugin/scss/_layout.scss
```diff
@@ -47,3 +47,36 @@
   margin-top: $spacer;
   padding: $spacer;
 }
+
+.atom-slider-for,
+.atom-slider-nav {
+  .slick-track {
+    display: flex;
+    align-items: center;
+    justify-content: center;
```
jraddaoui: This may not be needed in the end.

plugins/arDominionB5Plugin/scss/_layout.scss
jraddaoui: I gave it a quick try and this is the final SCSS I got ...

```scss
.atom-carousel {
  .slick-track {
    display: flex;
    align-items: center;

    img {
      max-width: 400px;
      max-height: 300px;
    }
  }

  .slick-arrow.slick-prev {
    left: -30px;
  }

  .slick-arrow.slick-next {
    right: -30px;
  }
}
```

plugins/arDominionB5Plugin/scss/_variables.scss
```diff
@@ -24,3 +24,9 @@ $h4-font-size: 1.45rem !default;
 $accordion-bg: #fff !default;
 $accordion-button-active-color: #000 !default;
 $table-bg: #fff !default;
+
+// Accessible-Slick carousel
+$slick-font-path: "../../../../node_modules/@accessible360/accessible-slick/slick/fonts/";
+$slick-dot-color: white;
+$slick-dot-color-active: $primary;
+$slick-dot-size: 14px;
```
jraddaoui, line 32: Nice!!

jraddaoui (changes requested):
Looking great Steve! I've been thorough but nothing major, just minor details and clean-ups.

package-lock.json
sbreker:
```
npm --version
7.19.0
```

plugins/arDominionB5Plugin/js/imageflow.js
sbreker: 👍

COPYRIGHT
sbreker: 👍

plugins/arDominionB5Plugin/js/imageflow.js
sbreker: I have removed the class name from this containing div and replaced it with an id: 

`<div id="atom-digital-object-carousel"`

plugins/arDominionB5Plugin/js/imageflow.js
sbreker: Updated to `#atom-slider-title`

plugins/arDominionB5Plugin/js/imageflow.js
sbreker: Updated to `#atom-slider-images`

plugins/arDominionB5Plugin/js/imageflow.js
sbreker: Using `Archival description title link` - not sure how screen reader will 'say' the forward slash in there.

sbreker: Updated and made translatable.

plugins/arDominionB5Plugin/js/imageflow.js
sbreker: Updated and made translatable.

plugins/arDominionB5Plugin/js/imageflow.js
sbreker: You are correct `asNavFor: '.atom-slider-nav',` seems to not be necessary - removing....

plugins/arDominionB5Plugin/modules/digitalobject/templates/_imageflow.php
sbreker: Good catch - This helper contained the function `truncate_text()`, but now that that call was removed I will eliminate this.

plugins/arDominionB5Plugin/modules/digitalobject/templates/_imageflow.php
sbreker: 😄

plugins/arDominionB5Plugin/modules/digitalobject/templates/_imageflow.php
sbreker: Fixed.

plugins/arDominionB5Plugin/modules/digitalobject/templates/_imageflow.php
sbreker: Thanks for the info - did not realize style h6 was the same as normal text. Change completed 👍

plugins/arDominionB5Plugin/modules/digitalobject/templates/_imageflow.php
sbreker: Updated text string. Also updated to use i18n strings.

plugins/arDominionB5Plugin/modules/digitalobject/templates/_imageflow.php
sbreker: 👍   Looks better with ms-2 in place - good call

plugins/arDominionB5Plugin/scss/_layout.scss
sbreker: Good idea

plugins/arDominionB5Plugin/scss/_layout.scss
sbreker: Makes sense. Thanks!

plugins/arDominionB5Plugin/scss/_layout.scss
sbreker: Good idea.  This makes sense now that there are not 2 sliders both with differently styled images.

plugins/arDominionB5Plugin/scss/_layout.scss
sbreker: I will note this in the spreadsheet.

plugins/arDominionB5Plugin/scss/_layout.scss
sbreker: You are correct - this is a remnant of having two sliders each with images. Removed.

plugins/arDominionB5Plugin/scss/_layout.scss
sbreker: You are correct!  Removed...

plugins/arDominionB5Plugin/scss/_layout.scss
sbreker: Same as what I have after addressing your feedback.  Thanks for your help!

plugins/arDominionB5Plugin/scss/_variables.scss
sbreker: 👍

sbreker:
Feedback implemented in second commit @jraddaoui - ready for your review.

plugins/arDominionB5Plugin/js/imageflow.js
```diff
@@ -1,34 +1,33 @@
-(function ($) {
+(() => {
```
jraddaoui: This should be `($ => {`. Isn't it complaining about line 3 without passing `$`?

jraddaoui (approved):
Thanks @sbreker! Just a minor comment, but looks great! Feel free to squash and merge after checking that comment ;)

PR #1339, opened 2021-07-13
Dev/issue 13530 b5 themes uppy data params title

sbreker:
This commit updates the Uppy data elements to 'kebab-case'. 

A bug was introduced when porting this script to B5 whereby image title
updates were not being saved because the assignment of this.$element
was changed. Revert saving $(element) to this.$element and change the 
data node to use a different class property.

### DIFF
--- a/plugins/arDominionB5Plugin/js/multiFileUpload.js
+++ b/plugins/arDominionB5Plugin/js/multiFileUpload.js
@@ -4,7 +4,7 @@
 
   function MultiFileUpload (element)
   {
-    this.$element = $(element).find('div.multiFileUpload');
+    this.$element = $(element);
 
     this.uppy = new Uppy.Core({ 
       debug: false,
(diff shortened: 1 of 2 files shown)


plugins/arDominionB5Plugin/js/multiFileUpload.js
```diff
@@ -4,7 +4,8 @@
 
   function MultiFileUpload (element)
   {
-    this.$element = $(element).find('div.multiFileUpload');
+    this.$multiFileUpload = $(element).find('div.multiFileUpload');
+    this.$element = $(element);
```
jraddaoui: I don't understand why do you need two vars here ... based on line 386, you already have the correct element and line 8 should be enough to use `this.$element` for everything ... right?

jraddaoui (commented):
Thanks @sbreker!

plugins/arDominionB5Plugin/js/multiFileUpload.js
sbreker: I have fixed this underlying issue allowing me to remove the second variable.

sbreker:
Thanks for your review @jraddaoui!  I have pushed a new commit that better addresses the root issues affecting saving the titles correctly.

The selector in the document ready function was not selecting correctly due to kebab-case issues it seems. I have renamed this. Secondly, I have moved the data elements to a higher div on the page so that the full form is selected - the js needs access to buttons AND the data elements.

plugins/arDominionB5Plugin/js/multiFileUpload.js
```diff
@@ -384,7 +383,8 @@
 
   $(function()
   {
-    var $node = $('.multiFileUpload');
+    var $node = $('.multifileupload-form');
+    //console.log($node.data('multifileupload-i18n-some-files-failed-error'));
```
jraddaoui: Ups ...

jraddaoui (approved):
Awesome, thanks @sbreker!

plugins/arDominionB5Plugin/js/multiFileUpload.js
sbreker: Removed...

PR #1340, opened 2021-07-15
Exclude narrower terms link fix, refs #13544

mInnes-archives:
PR addresses a routing bug with the "Exclude narrower terms" button in the "Related authority records" tab in term search. It currently routes back to the "Related Archival descriptions tab" rather than staying in the correct tab.

Including the correct action when constructing the url fixes the issue.

147279b05dfcfa46735db5c4e305bff6f97587fa -> I was getting a message about the config file name being incorrect so I renamed it, but I see the issue is more complicated after seeing #1331. My changes aren't triggering the fixer anyway.

### DIFF
--- a/apps/qubit/modules/term/templates/_directTerms.php
+++ b/apps/qubit/modules/term/templates/_directTerms.php
@@ -5,7 +5,7 @@
           '%1%' => $aggs['direct']['doc_count'], ]); ?>
       <?php $params = $sf_data->getRaw('sf_request')->getGetParameters(); ?>
       <?php unset($params['page']); ?>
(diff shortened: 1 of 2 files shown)


mInnes-archives:
Just realising this now, but staying on the correct tab seems to be an issue for a number of different link elements. I'll switch this to a WIP so to track them all down.

mInnes-archives:
The links in the treeview menu still reset the tabs, but that's not an easy fix as far as I can tell and I'm not sure if I'd even call this behaviour an 'issue'. Everything else is fine.

jraddaoui (approved):
Nice find @mInnes-archives, thank you!

Looks great, I've created https://projects.artefactual.com/issues/13544 for the reference.

jraddaoui:
Thanks @mInnes-archives!

PR #1341, opened 2021-07-19
Start styling user index page

replaceafill:
This updates the following elements of the user index page:

* The navigation "pills" at the top of the page
* The `Edition history` accordion markup to reuse the existing `/js/editingHistory.js` script and adds the `hidden` CSS class as implemented in BS 2.3.2 and a wrapper for the loading spinner and text
* Makes the table in the accordion responsive and updates the classes of its pager buttons and the background color of the strips


jraddaoui (approved):
Nice!

jraddaoui (approved):

PR #1344, opened 2021-07-21
Remove duplicate foreach() statement

djjuhasz:
Remove and extra copy of the "delete relations" foreach() loop from
inside the "delete children" loop of the QubitDigitalObject::delete()
method.


jraddaoui (approved):
Nice catch @djjuhasz!

PR #1346, opened 2021-07-22
Add treeview functionality to B5

sbreker:
Work in Progress

### DIFF
--- a/apps/qubit/modules/informationobject/actions/indexAction.class.php
+++ b/apps/qubit/modules/informationobject/actions/indexAction.class.php
@@ -42,7 +42,7 @@ class InformationObjectIndexAction extends sfAction
 
         $this->dispatcher->notify(new sfEvent($this, 'access_log.view', ['object' => $this->resource]));
 
-        if ('fullWidth' == sfConfig::get('app_treeview_type__source', 'sidebar')) {
+        if (!sfConfig::get('app_b5_theme', false) && 'fullWidth' == sfConfig::get('app_treeview_type__source', 'sidebar')) {
             $this->getResponse()->addStylesheet('fullWidthTreeView', 'last');
             $this->getResponse()->addStylesheet('/vendor/jstree/themes/default/style.min.css', 'last');
             $this->getResponse()->addJavascript('treeviewTypes', 'last');
--- a/js/fullWidthTreeView.js
+++ b/js/fullWidthTreeView.js
@@ -1,8 +1,16 @@
-"use strict";
-
 (function ($) {
 
-  $(loadTreeView);
+  "use strict";
+
+  $(function()
+  {
+    var $node = $('#treeview');
+
+    if ($node.length)
+    {
+      loadTreeView();
+    }
+  });
 
   function makeFullTreeviewCollapsible ($treeViewConfig, $mainHeader, $fwTreeViewRow)
   {
@@ -24,7 +32,7 @@
     $fwTreeViewRow.appendTo($wrapper);
 
     // Activate toggle button
-    $toggleButton.click(function() {
+    $toggleButton.on('click', function() {
       // Determine appropriate toggle button text
       var toggleText = $treeViewConfig.data('opened-text');
 
@@ -199,7 +207,7 @@
         $('#main-column .breadcrumb').after($(response.find('#main-column > div.messages.error')));
 
         // Attach the Drupal Behaviour so blank.js does its thing
-        Drupal.attachBehaviors(document)
+        Drupal.attachBehaviors(document);
 
         // Update clipboard buttons
         if (jQuery('#clipboard-menu').data('clipboard') !== undefined)
@@ -289,7 +297,7 @@
       .bind('move_node.jstree', moveNodeListener);
 
     // Clicking "more" will add next page of results to tree
-    $moreButton.click(function() {
+    $moreButton.on('click', function() {
       pager.next();
       pager.getAndAppendNodes(function() {
         // Queue is empty so update paging link
@@ -298,12 +306,12 @@
     });
 
     // Clicking reset link will reset paging and tree state
-    $('#fullwidth-treeview-reset-button').click(function()
+    $('#fullwidth-treeview-reset-button').on('click', function()
     {
       pager.reset($moreButton, $resetButton);
     });
 
     // TODO restore window.history states
-    $(window).bind('popstate', function() {});
+    $(window).on('popstate', function() {});
   }
 })(jQuery);
--- a/js/hierarchy.js
+++ b/js/hierarchy.js
@@ -1,8 +1,16 @@
-"use strict";
-
 (function ($) {
 
-  $(loadTreeView);
+  "use strict";
+
+  $(function()
+  {
+    var $node = $('.browse.hierarchy');
+
+    if ($node.length)
+    {
+      loadTreeView();
+    }
+  });
 
   function startActivity()
   {
@@ -135,7 +143,7 @@
       .bind('select_node.jstree', selectNodeListener);
 
     // Clicking "more" will add next page of results to tree
-    $moreButton.click(function() {
+    $moreButton.on("click", function() {
       pager.next();
       startActivity();
       pager.getAndAppendNodes(function() {
@@ -146,7 +154,7 @@
     });
 
     // Clicking reset link will reset paging and tree state
-    $resetButton.click(function()
+    $resetButton.on("click", function()
     {
       pager.reset($moreButton);
     });
--- a/package.json
+++ b/package.json
@@ -9,6 +9,8 @@
     "bootstrap-autocomplete": "^2.3.7",
     "jquery": "^3.6.0",
     "jquery-expander": "^2.0.1",
+    "jquery-ui-dist": "^1.12.1",
+    "jstree": "^3.3.11",
     "masonry-layout": "^4.2.2",
     "mediaelement": "^4.2.16"
   },
--- a/plugins/arDominionB5Plugin/modules/browse/templates/hierarchySuccess.php
+++ b/plugins/arDominionB5Plugin/modules/browse/templates/hierarchySuccess.php
@@ -1,0 +1,25 @@
+<?php decorate_with('layout_1col.php'); ?>
+<?php slot('title'); ?>
+
+  <h1>
+    <?php echo __('Hierarchy'); ?>
+    <span id="fullwidth-treeview-activity-indicator">
+      <i class="fas fa-spinner fa-spin ms-2" aria-hidden="true"></i>
+      <span class="visually-hidden"><?php echo __('Loading ...'); ?></span>
+    </span>
+  </h1>
+
+  <div class="d-flex flex-wrap gap-2 mb-3">
+    <input type="button" id="fullwidth-treeview-reset-button" class="btn btn-sm atom-btn-white" value="<?php echo __('Reset'); ?>" />
+    <input type="button" id="fullwidth-treeview-more-button" class="btn btn-sm atom-btn-white" data-label="<?php echo __('%1% more'); ?>" value="" />
+  </div>
+
+<?php end_slot(); ?>
+
+<?php slot('content'); ?>
+
+<div id='main-column'>
+  <span id="fullwidth-treeview-configuration" data-items-per-page="<?php echo $itemsPerPage; ?>"></span>
+</div>
+
+<?php end_slot(); ?>
--- a/plugins/arDominionB5Plugin/scss/_fullwidthtreeview.scss
+++ b/plugins/arDominionB5Plugin/scss/_fullwidthtreeview.scss
@@ -1,0 +1,76 @@
+// From css/fullWidthTreeView.css
+#fullwidth-treeview-header {
+    border: 1px solid #ddd;
+    border-bottom: none;
+    height: 20px;
+}
+
+#fullwidth-treeview-header div {
+    float: left;
+    background-color: #eee;
+    padding-left: 7px;
+}
+
+#fullwidth-treeview-row {
+    margin-bottom: 30px;
+    border: 1px solid #ddd;
+    clear: both;
+}
+
+#fullwidth-treeview {
+    overflow-x: hidden;
+    height: 100%;
+    background-color: #ffffff;
+}
+
+.ui-resizable-s {
+    background-image: url('/images/resizable_grip.png');
+    height: 9px;
+    border: 1px solid #ddd;
+    cursor: ns-resize;
+    background-repeat: no-repeat;
+    background-position: center center;
+    background-color: #eee;
+    bottom: -11px;
+    left: -1px;
+    margin: -1px;
+}
+
+.fullwidth-treeview-toggle.active {
+    border: 1px inset #eee;
+}
+
+#fullwidth-treeview-loading {
+    float: right;
+    margin-right: 110px;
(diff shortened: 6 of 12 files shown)


package.json
```diff
@@ -10,6 +10,8 @@
     "jquery": "^3.6.0",
     "jquery-expander": "^2.0.1",
     "masonry-layout": "^4.2.2",
+    "jquery-ui-dist": "^1.12.1",
+    "jstree": "^3.3.11",
```
jraddaoui, line 13: Conflict solving messed alphabetic order.

plugins/arDominionB5Plugin/templates/_layout_start.php
```diff
@@ -8,6 +8,8 @@
     <link rel="shortcut icon" href="<?php echo public_path('favicon.ico'); ?>"/>
     <!-- TODO: Move other CSS to SCSS build (some not possible due to https://github.com/sass/sass/issues/2535) -->
     <link rel="stylesheet" href="/node_modules/mediaelement/build/mediaelementplayer.min.css">
+    <link rel="stylesheet" href="/node_modules/jquery-ui-dist/jquery-ui.min.css">
+    <link rel="stylesheet" href="/node_modules/jstree/dist/themes/default/style.min.css">
```
jraddaoui, line 11: Do they use relative paths like mediaelementplayer? Or could they be moved to the SCSS build?

apps/qubit/modules/informationobject/actions/indexAction.class.php
```diff
@@ -42,7 +42,7 @@ public function execute($request)
 
         $this->dispatcher->notify(new sfEvent($this, 'access_log.view', ['object' => $this->resource]));
 
-        if ('fullWidth' == sfConfig::get('app_treeview_type__source', 'sidebar')) {
+        if (!sfConfig::get('app_b5_theme', false) && 'fullWidth' == sfConfig::get('app_treeview_type__source', 'sidebar')) {
```
jraddaoui, line 45: And how are they loaded from the hierarchy page?

plugins/arDominionB5Plugin/js/fullWidthTreeView.js
```diff
@@ -0,0 +1,320 @@
...
+  $(function()
+  {
+    var $node = $('#treeview');
+
+    if ($node.length)
+    {
+      console.log("Loading treeview js");
+      $(loadTreeView);
+    }
+  });
```
jraddaoui: If this would be the only change, I'd consider it safe enough to do it in the existing JS file, but I see there is some Drupal bellow ...

plugins/arDominionB5Plugin/js/fullWidthTreeView.js
```diff
@@ -0,0 +1,320 @@
...
+        $('.translation-links').replaceWith($(response.find('.translation-links')));
+
+        // Replace error message
+        $('#main-column > div.messages.error').remove();
+        $('#main-column .breadcrumb').after($(response.find('#main-column > div.messages.error')));
+
+        // Attach the Drupal Behaviour so blank.js does its thing
+
+        // TODO: commented this out - does it need replacing?
+        //Drupal.attachBehaviors(document)
```
jraddaoui: Agh, I faced a similar issue the other day ... https://github.com/artefactual/atom/blob/dev-test/issue-13530-b5-themes/plugins/arDominionB5Plugin/js/clipboard.js#L127-L136

Don't worry about it yet, the expander and other behaviors won't work for now after changing node, but I hope we can find a global solution at some point. For example, to hide the empty fields in the index page I hope we can live without that JS behavior, avoiding the empty field/heading display directly while rendering (@replaceafill), but we still need to share some things as you know.

plugins/arDominionB5Plugin/js/hierarchy.js
```diff
@@ -0,0 +1,162 @@
...
+(function ($) {
+
+  $(function()
+  {
+    var $node = $('.browse.hierarchy');
+
+    if ($node.length)
+    {
+      $(loadTreeView);
+    }
```
jraddaoui: In this case this is the only change to the entire JS file, and I think it's safe enough to do it in the current one and avoid the duplication.

package.json
sbreker: Fixed.

apps/qubit/modules/informationobject/actions/indexAction.class.php
sbreker: They are loaded using view.yml in the browse module.

plugins/arDominionB5Plugin/templates/_layout_start.php
sbreker: These packages do not include scss files - can they be moved to the SCSS build if they do not @jraddaoui ?

sbreker: I have moved jquery-ui to main.scss but jstree has references to images that fail when bundled with scss.

plugins/arDominionB5Plugin/js/fullWidthTreeView.js
sbreker: Sounds good - some additional changes with respect to deprecated jquery calls too. Tested in the old theme and functionality is still present.  Moving to the root /js folder.

plugins/arDominionB5Plugin/js/hierarchy.js
sbreker: Same as above - some additional changes with respect to deprecated jquery calls. Tested in the old theme and functionality is still present.  Moving to the root /js folder.

js/fullWidthTreeView.js
```diff
@@ -2,7 +2,15 @@
```
jraddaoui: It was already like that but, just following what we chatted yesterday, the `"use strict";` in this file is outside the IIFE (`(function($){})(jQuery)`) and it's polluting the global scope. Hopefully all JS that comes after is also in strict mode, but that may not be the case. If you could move inside the IIFE, that would be great.

js/fullWidthTreeView.js
```diff
@@ -2,7 +2,15 @@
...
 (function ($) {
 
-  $(loadTreeView);
+  $(function()
+  {
+    var $node = $('#treeview');
+
+    if ($node.length)
+    {
+      $(loadTreeView);
```
jraddaoui: Since you added the document ready above (`$(function(){}`), there is no need to call the main function with `$()`, which only does that when the argument is a callback:

https://api.jquery.com/jquery/#jQuery3

You could just call the function `loadTreeView();`.

js/hierarchy.js
```diff
@@ -2,7 +2,15 @@
...
 (function ($) {
 
-  $(loadTreeView);
+  $(function()
+  {
+    var $node = $('.browse.hierarchy');
+
+    if ($node.length)
+    {
+      $(loadTreeView);
```
jraddaoui: Same last two comments for this file.

package.json
```diff
@@ -9,6 +9,8 @@
     "bootstrap-autocomplete": "^2.3.7",
     "jquery": "^3.6.0",
     "jquery-expander": "^2.0.1",
+    "jquery-ui-dist": "^1.12.1",
```
jraddaoui, line 12: We'll see later, if in the end is only used to resize this, we may want to find an alternative.

plugins/arDominionB5Plugin/modules/browse/templates/hierarchySuccess.php
```diff
@@ -0,0 +1,20 @@
+<?php decorate_with('layout_1col.php'); ?>
+<?php slot('title'); ?>
+
+  <div class="d-flex justify-content-end flex-wrap gap-2 ms-auto">
+    <input type="button" id="fullwidth-treeview-reset-button" class="btn btn-sm atom-btn-white" value="<?php echo __('Reset'); ?>" />
+    <input type="button" id="fullwidth-treeview-more-button" class="btn btn-sm atom-btn-white" data-label="<?php echo __('%1% more'); ?>" value="" />
+  </div>
```
jraddaoui, line 15: This will be "unordered", not a huge deal in here as the following element doesn't allow focus but we should move if after the heading if we can, to represent better what is visible to screen readers.

plugins/arDominionB5Plugin/modules/browse/templates/hierarchySuccess.php
```diff
@@ -0,0 +1,20 @@
...
+
+  <?php echo image_tag('/vendor/jstree/themes/default/throbber.gif', ['id' => 'fullwidth-treeview-activity-indicator', 'alt' => __('Loading ...')]); ?>
+  <h1><?php echo __('Hierarchy'); ?></h1>
```
jraddaoui: I'd change the order here too and use the same loading spinner we use in other places. For example ...

https://github.com/artefactual/atom/blob/dev-test/issue-13530-b5-themes/plugins/arDominionB5Plugin/modules/user/templates/indexSuccess.mod_standard.php#L97-L100

Something like this ...

```html
<h1>
  <?php echo __('Hierarchy'); ?>
  <span id="fullwidth-treeview-activity-indicator">
    <i class="fas fa-spinner fa-spin ms-2" aria-hidden="true"></i>
    <span class="visually-hidden"><?php echo __('Loading ...'); ?></span>
  </span>
</h1>
```

plugins/arDominionB5Plugin/modules/browse/templates/hierarchySuccess.php
```diff
@@ -0,0 +1,20 @@
...
+<?php end_slot(); ?>
+
+<?php slot('content'); ?>
+
+<div id='main-column'></div>
+<div id='fullwidth-treeview-hierarchy'>
+  <span id="fullwidth-treeview-configuration" data-items-per-page="<?php echo $itemsPerPage; ?>"></span>
+</div>
```
jraddaoui: Missing indentation and, more importantly, where does the `main-column` end?

plugins/arDominionB5Plugin/scss/_fullwidthtreeview.scss
```diff
@@ -0,0 +1,78 @@
+@use "sass:math";
```
jraddaoui: I think you don't need this.

plugins/arDominionB5Plugin/scss/_fullwidthtreeview.scss
```diff
@@ -0,0 +1,78 @@
...
+
+// From css/fullWidthTreeView.css
```
jraddaoui, line 1: If you're copying the styles as they are, you could import directly the existing CSS file.

plugins/arDominionB5Plugin/scss/_fullwidthtreeview.scss
```diff
@@ -0,0 +1,78 @@
...
+}
+
+// From search.less
+.full-treeview-section {
+    background-color: $gray-200;
+    border: 1px solid $black;
+    .actions {
+      margin-bottom: -20px;
+    }
+}
```
jraddaoui, line 76: I'd love to see if we can improve these styles, but no need to do it now as we chatted.

jraddaoui (changes requested):
Looking great @sbreker, thanks!

Just a couple of important comments below and few notes.

js/fullWidthTreeView.js
sbreker: I have moved this line inside the IIFE

js/fullWidthTreeView.js
sbreker: Fixed.

js/hierarchy.js
sbreker: 👍

package.json
sbreker: Ok

plugins/arDominionB5Plugin/modules/browse/templates/hierarchySuccess.php
sbreker: I see. I have reformatted this section - updated both the order and the B5 classes.

plugins/arDominionB5Plugin/modules/browse/templates/hierarchySuccess.php
sbreker: Fixed. Didn't notice the main-column div issue when I copied it from the apps/qubit folder.

plugins/arDominionB5Plugin/scss/_fullwidthtreeview.scss
sbreker: Removed.

plugins/arDominionB5Plugin/scss/_fullwidthtreeview.scss
sbreker: They were modified a little - some of the original styling was not necessary anymore.

plugins/arDominionB5Plugin/scss/_fullwidthtreeview.scss
sbreker: 👍

plugins/arDominionB5Plugin/modules/browse/templates/hierarchySuccess.php
```diff
@@ -1,19 +1,20 @@
...
 
-  <div class="d-flex justify-content-end flex-wrap gap-2 ms-auto">
+  <h1>
+    <?php echo __('Hierarchy'); ?>    
+  </h1>
+
+  <div class="d-flex flex-wrap gap-2 mb-3">
     <input type="button" id="fullwidth-treeview-reset-button" class="btn btn-sm atom-btn-white" value="<?php echo __('Reset'); ?>" />
-    <input type="button" id="fullwidth-treeview-more-button" class="btn btn-sm atom-btn-white" data-label="<?php echo __('%1% more'); ?>" value="" />
+    <input type="hidden" id="fullwidth-treeview-more-button" class="btn btn-sm atom-btn-white" data-label="<?php echo __('%1% more'); ?>" value="" />
```
jraddaoui: Why type hidden? Maybe just d-none to be updated later from the JS?

plugins/arDominionB5Plugin/modules/browse/templates/hierarchySuccess.php
```diff
@@ -1,19 +1,20 @@
...
   </div>
 
-  <?php echo image_tag('/vendor/jstree/themes/default/throbber.gif', ['id' => 'fullwidth-treeview-activity-indicator', 'alt' => __('Loading ...')]); ?>
```
jraddaoui: Are you adding the spinner differently? Isn't this used from the JS?

jraddaoui (commented):
Awesome, thanks @sbreker!

plugins/arDominionB5Plugin/modules/browse/templates/hierarchySuccess.php
sbreker: Removed. This should not have been in here.

plugins/arDominionB5Plugin/modules/browse/templates/hierarchySuccess.php
sbreker: I have updated this so that during page load it uses the fa throbber by the section title. Still uses the jstree icon within the treeview itself when building the treeview nodes.

jraddaoui (approved):
Great, thanks @sbreker!

PR #1347, opened 2021-07-23
Set up sfTranslatePlugin in arDominionBS5Plugin

sevein:
I have confirmed this is still working in arDominionPlugin.
It assumes that the behaviour is going to be attached from the new qubit.js.


jraddaoui (approved):
Neat!

PR #1349, opened 2021-07-26
arDominionBS5Plugin: fix edit rights behaviour

sevein:
I've updated the form row formatter so the divs have a .form-row class. It's not a BS5 layout requirement but it's necessary in the old edit/rights behaviour to easily target form fields by their names.

### DIFF
--- a/plugins/arDominionB5Plugin/lib/arB5WidgetFormSchemaFormatter.class.php
+++ b/plugins/arDominionB5Plugin/lib/arB5WidgetFormSchemaFormatter.class.php
@@ -19,8 +19,6 @@
 
 class arB5WidgetFormSchemaFormatter extends sfWidgetFormSchemaFormatter
 {
-    protected $rowFormat = "<div class=\"mb-3\">\n  %label%\n  %field%\n"
-        ."  %error%\n  %help%\n  %hidden_fields%\n</div>";
     protected $errorListFormatInARow = '<div class="invalid-feedback"'
         ." id=\"%errors_id%\">\n  %errors%\n</div>\n";
     protected $errorRowFormatInARow = "<span>%error%</span>\n";
@@ -63,5 +61,25 @@ class arB5WidgetFormSchemaFormatter extends sfWidgetFormSchemaFormatter
             $this->helpFormat,
             ['%help_id%' => $this->name.'-help']
         );
+    }
+
+    public function getRowFormat()
+    {
+        // HACK ->formatRow() lacks access to lots of information about the field,
+        // including the name.  So to add the name to the row markup, we must
+        // either,
+        //  * Add it outside ->formatRow(), perhaps using FluentDOM
+        //  * Extract the name from the field markup with regex
+        //  * Take advantage that ->renderRow() always calls ->renderLabel(), which
+        //    calls ->generateLabel(), before ->renderRow()
+        return <<<return
+<div class="mb-3 form-row-{$this->name}">
+  %label%
+  %error%%field%
+  %help%
+  %hidden_fields%
+</div>
+
+return;
     }
 }
--- a/plugins/arDominionB5Plugin/modules/right/templates/editSuccess.php
+++ b/plugins/arDominionB5Plugin/modules/right/templates/editSuccess.php
@@ -56,7 +56,7 @@
                 ->help(__('The country or other political body that has enacted the statute.'))
                 ->renderRow(); ?>
 
-            <div class="form-item form-item-statuteCitation">
+            <div class="form-row form-row-statuteCitation">
               <?php echo $form->statuteCitation->renderLabel(); ?>
               <?php echo $form->statuteCitation->render(['class' => 'form-autocomplete']); ?>
               <?php if (QubitAcl::check(QubitTaxonomy::getById(QubitTaxonomy::RIGHTS_STATUTES_ID), 'createTerm')) { ?>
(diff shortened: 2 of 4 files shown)


plugins/arDominionB5Plugin/lib/arB5WidgetFormSchemaFormatter.class.php
```diff
@@ -64,4 +62,24 @@ public function getHelpFormat()
             ['%help_id%' => $this->name.'-help']
         );
     }
+
+    public function getRowFormat()
```
jraddaoui, line 66: We're trying to not use ->renderRow() to render fields and use the `render_field` helper from QubitHelper. We still need to go over many fields to change that as part of row 33 in the spreadsheet. In there, you have access to the field name to add it as a class ...

https://github.com/artefactual/atom/blob/dev-test/issue-13530-b5-themes/lib/helper/QubitHelper.php#L123-L158

plugins/arDominionB5Plugin/lib/arB5WidgetFormSchemaFormatter.class.php
```diff
@@ -64,4 +62,24 @@ public function getHelpFormat()
...
+        //  * Extract the name from the field markup with regex
+        //  * Take advantage that ->renderRow() always calls ->renderLabel(), which
+        //    calls ->generateLabel(), before ->renderRow()
+        return <<<return
+<div class="mb-3 form-row-{$this->name}">
+  %label%
+  %error%%field%
+  %help%
+  %hidden_fields%
+</div>
```
jraddaoui, line 81: This is changing the layout to the one we had before. As mentioned, we're trying not to use it and normalize all fields rendering in the helper. I added the $rowFormat as part of an initial implementation and thought to leave it in there, but it lacks some logic needed for some inputs.

plugins/arDominionB5Plugin/modules/right/templates/editSuccess.php
```diff
@@ -185,4 +185,89 @@
 
   </form>
 
+  <script type="text/javascript">
```
jraddaoui, line 188: I'd love to see this JS in its own `.js` file so we can include it in the bundle later on. As we chatted, it looks like it only needs one PHP call, which should be managed differently, adding a data attribute to one of the elements to pass that translation or something like that.

plugins/arDominionB5Plugin/modules/right/templates/editSuccess.php
```diff
@@ -185,4 +185,89 @@
...
+    (function(){
+
+      // Basis drop-down.
+
+      let BasisSelect = {
+        fieldsets: {
+          copyright: jQuery('#wrapper form div[class *= form-row-copyright]'),
+          license: jQuery('#wrapper div[class *= form-row-license]'),
+          statute: jQuery('#wrapper div[class *= form-row-statute]'),
```
jraddaoui, line 197: Since we need to "duplicate" this JS, instead of adding those classes to the fields I'd try using a different selector ... `.form-control[name^=license]` (you could use the id attribute instead of the name and it may be `.form-check-input` if they're check boxes/radio buttons).

jraddaoui: Use arrow functions and `$` when moving it to its own JS.

plugins/arDominionB5Plugin/modules/right/templates/editSuccess.php
```diff
@@ -185,4 +185,89 @@
...
+        added.find('legend').replaceWith('<legend><?php echo __('New granted right'); ?></legend>');
+
+        // yank out the fieldset-wrapper dic that collapse adds
+        // because it is about to add another. =(
+        var fswrapper = added.find('.fieldset-wrapper');
+        html = fswrapper.html()
+        fswrapper.replaceWith(html);
+
+        updateGrantedRightsNamesAndIds();
+        added.show(400);
```
jraddaoui, line 245: This is still using the old collapsible sections and I'm not sure if we want to keep it this way. No need to worry now, but I added a note in the spreadsheet (row 32) about this ...

> - Another nested accordion mixed with "multi-row" in edit rights, act/granted rights (still to do, we may want to normalize this with dialogs/tables or tabs when we get to that).

plugins/arDominionB5Plugin/modules/right/templates/editSuccess.php
```diff
@@ -185,4 +185,89 @@
...
+      });
+
+      // Granted Rights Delete X
+      jQuery('#wrapper').on('click', '.c-btn-delete', function(){
```
jraddaoui, line 249: We will probably change this class when we get to style this sections. Maybe add a new class and use that one, like the `newItem` link.

plugins/arDominionB5Plugin/modules/user/templates/indexSuccess.mod_cas.php
```diff
@@ -1,3 +1,17 @@
...
+<?php if (!$resource->active) { ?>
+  <div class="messages error">
+    <ul>
+      <?php if (!$resource->active) { ?>
+        <li><?php echo __('This user is inactive'); ?></li>
+      <?php } ?>
+    </ul>
+  </div>
+<?php } ?>
+
```
jraddaoui, line 14: Oh, is this fixing what we mentioned today about the nav pills? Nice! I guess it should to a different commit.

plugins/arDominionB5Plugin/modules/user/templates/indexSuccess.mod_standard.php
```diff
@@ -1,3 +1,17 @@
...
+<?php if (!$resource->active) { ?>
+  <div class="messages error">
+    <ul>
+      <?php if (!$resource->active) { ?>
+        <li><?php echo __('This user is inactive'); ?></li>
+      <?php } ?>
+    </ul>
+  </div>
+<?php } ?>
+
```
jraddaoui, line 14: Same.

jraddaoui (changes requested):
Thanks @sevein!

Wow, you entered an area that requires quite a few changes for which I was considering a different approach, specially those collapsible sections that are far from accessible and having that Javascript in the template. I've added some comments below but we can chat tomorrow, there is no need to address all the feedback at the moment. I was planning to get to it after most of the forms styling is done (multi rows, dialogs, tabs, etc.) to see if there is a better way to render this section.

sevein:
I'll revisit when forms are further developed.

PR #1354, opened 2021-07-28
CSV Validator output changes, refs #13475

sbreker:
Column specific tests to check if a column is present or is duplicated
moved to the base validator. New php_unit tests added for the case where
a column is duplicated in the CSV file.

This commit includes formatting changes to the output report
- add two lines separating the summary at the top of the report from the
following test specific sections.
- format the test titles so the result is first and in all caps.
- update column name check result so that the list of unknown column
names is returned with the non-verbose version of the output.

### DIFF
--- a/lib/task/import/validate/csvBaseValidator.class.php
+++ b/lib/task/import/validate/csvBaseValidator.class.php
@@ -42,6 +42,11 @@ abstract class CsvBaseValidator
     protected $options = [];
     protected $ormClasses = [];
     protected $testData;
+    protected $columnPresent = [];
+    protected $columnDuplicated = [];
+    protected $requiredColumns = [];
+    protected $header = [];
+    protected $proceedWithRowValidation;
 
     public function __construct(?array $options = null)
     {
@@ -52,16 +57,97 @@ abstract class CsvBaseValidator
         $this->testData = new CsvValidatorResult($this->title, $this->filename, $this->displayFilename, $this->getClassName());
     }
 
+    public function columnPresent(string $columnName)
+    {
+        if (isset($this->columnPresent[$columnName])) {
+            return $this->columnPresent[$columnName];
+        }
+
+        if (null !== $this->header) {
+            return $this->columnPresent[$columnName] = in_array($columnName, $this->header);
+        }
+
+        return false;
+    }
+
+    public function columnDuplicated(string $columnName)
+    {
+        if (!$this->columnPresent($columnName)) {
+            return false;
+        }
+
+        if (isset($this->columnDuplicated[$columnName])) {
+            return $this->columnDuplicated[$columnName];
+        }
+
+        if (null !== $this->header) {
+            return $this->columnDuplicated[$columnName] = count(array_keys($this->header, $columnName)) > 1;
+        }
+
+        return false;
+    }
+
+    public function appendDuplicatedColumnError(string $columnName)
+    {
+        $this->testData->setStatusError();
+        $this->testData->addResult(sprintf("'%s' column appears more than once in file.", $columnName));
+        $this->testData->addResult(sprintf('Unable to validate because of duplicated columns in CSV.'));
+    }
+
+    public function setHeader(array $header)
+    {
+        $this->header = $header;
+    }
+
+    // Specify required columns to trigger basic checks - presence, duplication.
+    public function setRequiredColumns(array $requiredColumns)
+    {
+        $this->requiredColumns = $requiredColumns;
+    }
+
     public function testRow(array $header, array $row)
     {
         ++$this->rowNumber;
+
+        // Save copy of header when processing first row of CSV.
+        if (empty($this->header)) {
+            $this->setHeader($header);
+        }
+
+        // Test required columns for presence and duplication. Return false if
+        // column is absent or duplicated.
+        if (isset($this->proceedWithRowValidation)) {
+            return $this->proceedWithRowValidation;
+        }
+
+        $this->proceedWithRowValidation = true;
+
+        foreach ($this->requiredColumns as $column) {
+            if (!$this->columnPresent($column)) {
+                $this->proceedWithRowValidation = false;
+            }
+
+            if ($this->columnDuplicated($column)) {
+                $this->proceedWithRowValidation = false;
+            }
+        }
+
+        return $this->proceedWithRowValidation;
     }
 
+    /**
+     *  Called between CSV files when more than one CSV file is specified on
+     *  the command line.
+     */
     public function reset()
     {
         $this->filename = '';
         $this->displayFilename = '';
         $this->testData = new CsvValidatorResult($this->title, $this->filename, $this->displayFilename, $this->getClassName());
+        $this->columnPresent = [];
+        $this->columnDuplicated = [];
+        $this->header = [];
+        $this->proceedWithRowValidation = null;
     }
 
     public function setOrmClasses(array $classes)
--- a/lib/task/import/validate/csvDigitalObjectPathValidator.class.php
+++ b/lib/task/import/validate/csvDigitalObjectPathValidator.class.php
@@ -34,8 +34,6 @@ class CsvDigitalObjectPathValidator extends CsvBaseValidator
     protected $fileList = [];
     protected $pathToDigitalObjects = '';
     // Reset between files.
-    protected $digitalObjectPathColumnPresent;
-    protected $digitalObjectUriColumnPresent;
     protected $digitalObjectUses = [];
     protected $overriddenByUriCount = 0;
 
@@ -45,13 +43,12 @@ class CsvDigitalObjectPathValidator extends CsvBaseValidator
         parent::__construct($options);
 
         $this->setPathToDigitalObjects($this->options['pathToDigitalObjects']);
+        $this->setRequiredColumns(['digitalObjectPath']);
     }
 
     public function reset()
     {
-        $this->digitalObjectPathColumnPresent = null;
         $this->digitalObjectUses = [];
-        $this->digitalObjectUriColumnPresent = null;
         $this->overriddenByUriCount = 0;
 
         parent::reset();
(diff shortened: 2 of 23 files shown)


lib/task/import/validate/csvBaseValidator.class.php
```diff
@@ -52,6 +54,41 @@ public function __construct(?array $options = null)
         $this->testData = new CsvValidatorResult($this->title, $this->filename, $this->displayFilename, $this->getClassName());
     }
 
+    public function columnPresent(string $columnName, ?array $header = null)
+    {
+        if (!isset($this->columnPresent[$columnName]) && null !== $header) {
+            $this->columnPresent[$columnName] = in_array($columnName, $header);
+        }
+
+        if (isset($this->columnPresent[$columnName])) {
```
djjuhasz, line 63: Won't this evaluate as true even if `$this->columnPresent[$columnName] === false`?

lib/task/import/validate/csvBaseValidator.class.php
```diff
@@ -52,6 +54,41 @@ public function __construct(?array $options = null)
...
+            return $this->columnPresent[$columnName];
+        }
+
+        return false;
+    }
+
+    public function columnDuplicated(string $columnName, ?array $header = null)
+    {
+        if ($this->columnPresent($columnName, $header)) {
```
djjuhasz: Opinion: To avoid the nested conditionals I'd prefer:

```
if (!$this->columnPresent($columnName, $header)) {
    return false;
}
```

lib/task/import/validate/csvBaseValidator.class.php
```diff
@@ -62,6 +99,8 @@ public function reset()
         $this->filename = '';
         $this->displayFilename = '';
         $this->testData = new CsvValidatorResult($this->title, $this->filename, $this->displayFilename, $this->getClassName());
+        $this->columnPresent = [];
```
djjuhasz, line 147: Is it necessary to reset the `this->columnPresent` and `$this->columnDuplicated` every row?

lib/task/import/validate/csvDigitalObjectPathValidator.class.php
```diff
@@ -62,21 +58,17 @@ public function testRow(array $header, array $row)
...
         $row = $this->combineRow($header, $row);
 
-        if (!isset($this->digitalObjectPathColumnPresent)) {
-            $this->digitalObjectPathColumnPresent = isset($row['digitalObjectPath']);
-        }
-
-        if (!isset($this->digitalObjectUriColumnPresent)) {
-            $this->digitalObjectUriColumnPresent = isset($row['digitalObjectUri']);
-        }
+        if ($this->columnPresent('digitalObjectPath', $header)) {
```
djjuhasz: I think these two conditionals can be combined into:

```
if (
    !$this->columnPresent('digitalObjectPath', $header) 
    || $this->columnDuplicated('digitalObjectPath', $header)
) {
    return;
}
```

And it avoids the long "if" statement.

lib/task/import/validate/csvParentValidator.class.php
```diff
@@ -107,39 +86,64 @@ public function testRow(array $header, array $row)
         }
 
         if (
-            $this->parentIdColumnPresent
+            $this->columnDuplicated('parentId', $header)
+            || $this->columnDuplicated('qubitParentSlug', $header)
+            || $this->columnDuplicated('legacyId', $header)) {
```
djjuhasz: Closing parenthesis and opening curly brace should be on next line.

lib/task/import/validate/csvParentValidator.class.php
```diff
@@ -107,39 +86,64 @@ public function testRow(array $header, array $row)
...
             $this->legacyIdList[] = $row['legacyId'];
         }
     }
 
     public function getTestResult()
     {
-        if (!$this->parentIdColumnPresent && !$this->qubitParentSlugColumnPresent) {
+        if (!$this->columnPresent('parentId') && !$this->columnPresent('qubitParentSlug')) {
             $this->testData->setStatusWarn();
             $this->testData->addResult(sprintf("'parentId' and 'qubitParentSlug' columns not present. CSV contents will be imported as top level records."));
```
djjuhasz: This line looks like it's over 80 chars and should be broken into multiple lines

lib/task/import/validate/csvParentValidator.class.php
```diff
@@ -107,39 +86,64 @@ public function testRow(array $header, array $row)
...
         } else {
```
djjuhasz: Can you return here to avoid the long "else"?

lib/task/import/validate/csvDigitalObjectUriValidator.class.php
```diff
@@ -53,11 +51,11 @@ public function testRow(array $header, array $row)
         parent::testRow($header, $row);
         $row = $this->combineRow($header, $row);
 
-        if (!isset($this->digitalObjectUriColumnPresent)) {
-            $this->digitalObjectUriColumnPresent = isset($row['digitalObjectUri']);
-        }
+        if ($this->columnPresent('digitalObjectUri', $header)) {
```
djjuhasz: I think it's better to combine these conditionals, e.g.

```
if (
   !$this->columnPresent('digitalObjectUri', $header) 
   || $this->columnDuplicated('digitalObjectUri', $header))
) {
    return;
}
```

lib/task/import/validate/csvValidatorResultCollection.php
```diff
@@ -157,7 +157,7 @@ public static function renderResultsAsText(CsvValidatorResultCollection $results
                 }
 
                 // Add double line divider unless there are no results to display.
-                if ($verbose == true || (!empty($warnCount) || !empty($errorCount))) {
+                if (true == $verbose || (!empty($warnCount) || !empty($errorCount))) {
```
djjuhasz: I think this is equivalent and shorter:

```
if ($verbose || !empty($warnCount) || !empty($errorCount)) {
```

djjuhasz (changes requested):
This looks good overall @sbreker, I've just added some style nitpicks.

The one concern I have with the overall design is the need to repeatedly check for duplicate columns in each of the individual validators, and on each row.  Is it possible to check for duplicate columns once at the beginning of the validation run instead?  I think it's only possible to identify duplicate columns if the column name is duplicated in the header, right?

lib/task/import/validate/csvDigitalObjectPathValidator.class.php
sbreker: Good idea! Change completed.

lib/task/import/validate/csvParentValidator.class.php
sbreker: Fixed

lib/task/import/validate/csvBaseValidator.class.php
sbreker: `reset()` is only called between CSV files and only if multiple CSVs are specified on the command line.  I think this is fine to reset in this circumstance.

lib/task/import/validate/csvBaseValidator.class.php
sbreker: 👍

lib/task/import/validate/csvParentValidator.class.php
sbreker: I've shortened this

lib/task/import/validate/csvParentValidator.class.php
sbreker: 👍  Updated

lib/task/import/validate/csvDigitalObjectUriValidator.class.php
sbreker: 👍

lib/task/import/validate/csvValidatorResultCollection.php
sbreker: 😄

sbreker:
Hi @djjuhasz - thanks for the feedback!  

I've  your concern about re-running the columnDuplicated base class function repeatedly - the header values are only checked once on the very first CSV row and the result is stored in an associative array in the base validator. All subsequent calls to columnDuplicated() and columnPresent() during processing of a CSV just read the cached result from the associative array.  You noticed that the base class method reset() dumps these cached values - this is only run in between CSVs, when multiple are specified.

That said, a future refactor could introduce a method that is called only on the first row where these checks could be performed and the results stored similarly to what is described above. These results could be read to prevent testRow() from being run on these tests.

djjuhasz:
@sbreker my other concern with the current duplicate column check method is the need to explicitly call `$this->columnDuplicated('qubitParentSlug', $header)` for each column to check for duplicates.  It seems to me that checking for duplicate columns in the header should be done without having to explictly specify each column name.

sbreker:
@djjuhasz There is a general duplicate column check validator that scans all column header values - this is the main duplicate check and does not rely on columnDuplicated(). 

Specific tests that depend on one (or multiple) csv columns (e.g. CsvCultureValidator) need a way to perform some specific sanity checks on that column - check if present, or check that it's not duplicated.  This gives a way for a validator to check one specific column, or multiple as is needed in the CsvParentValidator which needs to check qubitParentSlug, parentId, and legacyId. This provides a shareable way in the base class for validators to optionally choose which columns to test and react accordingly. The code that was replaced by culturePresent() was not extensible and had to be copied to each validator that needed to test a column.

It was found during testing to be confusing when the duplicate column validator found columns duplicated and produced an error, but the culture validator did not (for example).  Prior to this change the culture column validator would have validated the first of the multiple culture columns in the CSV and if the value was valid, it would return a positive result.

sbreker:
Let me look at it again. I'd like to remove passing the header in each time - easy to save it as an attribute of the base class.

lib/task/import/validate/csvBaseValidator.class.php
```diff
@@ -54,27 +57,27 @@ public function __construct(?array $options = null)
...
-    public function columnPresent(string $columnName, ?array $header = null)
+    public function columnPresent(string $columnName)
     {
         if (isset($this->columnPresent[$columnName])) {
             return $this->columnPresent[$columnName];
         }
 
-        if (!isset($this->columnPresent[$columnName]) && null !== $header) {
-            return $this->columnPresent[$columnName] = in_array($columnName, $header);
+        if (!isset($this->columnPresent[$columnName]) && null !== $this->header) {
```
djjuhasz: `isset($this->columnPresent[$columnName])` has already been tested in line 62, and must be false so it's not necessary to test it here.

lib/task/import/validate/csvBaseValidator.class.php
```diff
@@ -91,9 +94,45 @@ public function appendDuplicatedColumnError(string $columnName)
         $this->testData->addResult(sprintf('Unable to validate because of duplicated columns in CSV.'));
     }
 
+    public function setHeader($header)
```
djjuhasz: Nitpick: It would be nice to specify the type of `$header` parameter.  I'm guessing it's an array?

lib/task/import/validate/csvBaseValidator.class.php
```diff
@@ -91,9 +94,45 @@ public function appendDuplicatedColumnError(string $columnName)
...
+    {
+        $this->requiredColumns = $requiredColumns;
+    }
+
     public function testRow(array $header, array $row)
     {
         ++$this->rowNumber;
+
+        // Save copy of header when processing first row of CSV.
+        if (0 == count($this->header)) {
```
djjuhasz: Opinion: I think it's better to test `!isset($this->header)` or `empty($this->header)` instead of using count.

lib/task/import/validate/csvBaseValidator.class.php
```diff
@@ -91,9 +94,45 @@ public function appendDuplicatedColumnError(string $columnName)
...
+
+        // Test required columns for presence and duplication. Return false if
+        // column is absent or duplicated.
+        if (isset($this->proceedWithRowValidation)) {
+            return $this->proceedWithRowValidation;
+        }
+
+        $this->proceedWithRowValidation = true;
+
+        foreach ($this->requiredColumns as $column) {
```
djjuhasz, line 125: Nice! 😁

djjuhasz (dismissed):
Looks good @sbreker! Nicely done. :)

djjuhasz (changes requested):
Sorry, there were just a couple of minor changes I requested.

lib/task/import/validate/csvBaseValidator.class.php
sbreker: This is correct. I have updated this.

lib/task/import/validate/csvBaseValidator.class.php
sbreker: Good idea - fixed

lib/task/import/validate/csvBaseValidator.class.php
sbreker: 👍

sbreker:
Hi @djjuhasz - see latest commit

lib/task/import/validate/csvBaseValidator.class.php
```diff
@@ -76,14 +76,14 @@ public function columnDuplicated(string $columnName)
...
 
-        if (!isset($this->columnDuplicated[$columnName]) && null !== $this->header) {
-            $this->columnDuplicated[$columnName] = count(array_keys($this->header, $columnName)) > 1;
-        }
-
         if (isset($this->columnDuplicated[$columnName])) {
             return $this->columnDuplicated[$columnName];
         }
 
+        if (null !== $this->header) {
```
djjuhasz, line 83: Nice move. :)

lib/task/import/validate/csvBaseValidator.class.php
```diff
@@ -110,7 +110,8 @@ public function testRow(array $header, array $row)
         ++$this->rowNumber;
 
         // Save copy of header when processing first row of CSV.
-        if (0 == count($this->header)) {
+        //if (0 == count($this->header)) {
```
djjuhasz: I think can be removed?

djjuhasz (approved):
@sbreker looks good! There's one commented line of code that I think is an accident, but other than that it's ready to commit. :)

PR #1358, opened 2021-07-29
arDominionB5Plugin: style repository search box

sevein:
This pull request styles the repository search box.
The _holdingsInstitution template is duplicated to introduce the new layout.

image


jraddaoui (approved):
Neat! We may want to use the style from the inline searches, just make it small and use the `atom-btn-white` class. Thanks!

PR #1359, opened 2021-07-29
Style content container of actor index page

replaceafill:
It also fixes the hover decoration color of some section labels I missed previously.

### DIFF
--- a/plugins/arDominionB5Plugin/modules/sfIsaarPlugin/templates/indexSuccess.php
+++ b/plugins/arDominionB5Plugin/modules/sfIsaarPlugin/templates/indexSuccess.php
@@ -9,8 +9,8 @@
   <h1><?php echo render_title($resource); ?></h1>
 
   <?php if (isset($errorSchema)) { ?>
-    <div class="messages error">
-      <ul>
+    <div class="messages error alert alert-danger" role="alert">
+      <ul class="<?php echo render_b5_show_list_css_classes(); ?>">
         <?php foreach ($errorSchema as $error) { ?>
           <?php $error = sfOutputEscaper::unescape($error); ?>
           <li><?php echo $error->getMessage(); ?></li>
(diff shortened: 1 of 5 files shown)


plugins/arDominionB5Plugin/modules/sfIsaarPlugin/templates/indexSuccess.php
```diff
@@ -60,201 +60,209 @@
...
+      <div class="field <?php echo render_b5_show_field_css_classes(); ?>">
+        <?php echo render_b5_show_label(__('Parallel form(s) of name')); ?>
+        <div class="<?php echo render_b5_show_value_css_classes(); ?>">
+          <ul class="<?php echo render_b5_show_list_css_classes(); ?>">
+            <?php foreach ($resource->getOtherNames(['typeId' => QubitTerm::PARALLEL_FORM_OF_NAME_ID]) as $item) { ?>
+              <li><?php echo render_value_inline($item->__toString()); ?></li>
+            <?php } ?>
+          </ul>
+        </div>
+      </div>
```
jraddaoui, line 88: One suggestion about what we chatted yesterday (`sfOutputEscaper::unescape()` may not be needed in some cases) ...

```php
<?php echo render_show(__('Parallel form(s) of name'), sfOutputEscaper::unescape(
    $resource->getOtherNames(['typeId' => QubitTerm::PARALLEL_FORM_OF_NAME_ID])
)); ?>
```

And in the helper ...

```php
function render_b5_show_value($value, $options = [])
{
    $tag = 'div';
    $cssClasses = render_b5_show_value_css_classes($options);

    $finalValue = $value;
    if (is_array($value)) {
        $finalValue = '<ul class="'.render_b5_show_list_css_classes().'">';
        foreach ($value as $item) {
            $finalValue .= '<li>'.render_value_inline(string($item)).'</li>';
        }
        $finalValue .= '</ul>';
    }

    return render_b5_show_container($tag, $finalValue, $cssClasses, $options);
}
```

plugins/arDominionB5Plugin/modules/sfIsaarPlugin/templates/indexSuccess.php
```diff
@@ -60,201 +60,209 @@
...
+      <div class="field <?php echo render_b5_show_field_css_classes(); ?>">
+        <?php echo render_b5_show_label(__('Language(s)')); ?>
+        <div class="<?php echo render_b5_show_value_css_classes(); ?>">
+          <ul class="<?php echo render_b5_show_list_css_classes(); ?>">
+            <?php foreach ($resource->language as $code) { ?>
+              <li><?php echo format_language($code); ?></li>
+            <?php } ?>
+          </ul>
+        </div>
+      </div>
```
jraddaoui, line 236: For this cases that are not a `string` or a call to `__toString()`, we could format the array in the template ...

```php
<?php
    $list = [];
    foreach ($resource->language as $code) {
        $list[] = format_language($code);
    }
    echo render_show(__('Language(s)'), $list);
?>
```

It may require further tweaks as we go, but this is what I was thinking yesterday.

jraddaoui (approved):

PR #1360, opened 2021-07-29
Carousel dot placement issue

sbreker:
Update the styling so that the Carousel dots are positioned relative to
the images in the carousel. This prevents the images overlaying the dots
when there are a large number of images.


jraddaoui:
Thanks @sbreker!

PR #1361, opened 2021-07-30
Style content container of repository index page

replaceafill:
It also adds JS for making Google Maps work in DO metadata and repository pages.


jraddaoui (approved):
Nice Douglas! This one was bigger than expected.

PR #1367, opened 2021-08-02
Style content container of donor index page

replaceafill:


jraddaoui (approved):
Awesome, thanks @replaceafill!

PR #1326, opened 2021-06-21
Create Bootstrap 5 based plugin theme, refs #13530

jraddaoui:
Add arDominionB5Plugin theme:

- Bootstrap 5 theme based on arDominionPlugin.
- Using NPM to manage almost all dependencies.
- Using Webpack to build the final theme assets.
- Using Prettier for linting, included in CI.
- Using placeholder templates to import hashed CSS/JS assets on build.
- Using Sass.
- New JS versions of many components.
- Sets `app_b5_theme` config for conditional implementations in existing
  actions and other classes.
- Layout and templates using and adapted to Bootstrap 5, and without
  other components (JS and CSS from view. yml files).
- Layout partials (start and end) to avoid duplication.
- Updated logo.
- WIP image for themes form.

Tweaks in existing themes code to accommodate arDominionB5Plugin:

- Check `app_b5_theme` config to:
  - Not add JS from actions.
  - Select B5 render methods and widgets.
- Add `alert-link` class to alert links.
- Update JS selectors to target only the required elements.
- Consider `aria-describedby` and `disabled` attributes in autocomplete.
- Remove unneeded sf_module_dirs modification from existing  themes.

Update Docker environment:

- Update NPM globally.
- Install NPM dependencies.
- Build new themes .
- Add `npm_deps` named volume to avoid overwriting `node_modules`.
- Set `NODE_ENV` env. variable to `development`.

Other changes:

- Add workflow to update a pre-configured test site.
- Update Composer dependencies.
- Do not pass relative URL root to JS code.
- Modify user forum URL from error page.
- Fix `setTemplate` calls when using theme plugins.
- Delete `informationobject` module list action and template. 

Co-authored-by: Douglas Cerna (Soy Douglas) <douglascerna@yahoo.com>
Co-authored-by: Jesús García Crespo <jesus@sevein.com>
Co-authored-by: Steve Breker <sbreker@artefactual.com>

PR #1375, opened 2021-08-03
arDominionB5Plugin: add breadcrumbs

sevein:
B5's breadcrumb doesn't truncate text and it can take multiple lines, which could be seen as an improvement. We can tweak later after some testing.

image


jraddaoui:
Merged in https://github.com/artefactual/atom/pull/1326/commits/d6730b165b52b09e5551039aedbf2fdc3bbbd275

PR #1376, opened 2021-08-03
Style content of job index page

replaceafill:


plugins/arDominionB5Plugin/modules/jobs/templates/reportSuccess.php
```diff
@@ -4,73 +4,53 @@
...
-  </div>
+  <?php
+      $icon = '';
+      if (QubitTerm::JOB_STATUS_COMPLETED_ID == $job->statusId) {
+          $icon = '<i aria-hidden="true" class="fa fa-check-square ms-2 text-success" id="job-check-color"></i>';
+      } elseif (QubitTerm::JOB_STATUS_ERROR_ID == $job->statusId) {
+          $icon = '<i aria-hidden="true" class="fa fa-exclamation-triangle ms-2 text-danger" id="job-warning-color"></i>';
+      } elseif (QubitTerm::JOB_STATUS_IN_PROGRESS_ID == $job->statusId) {
+          $icon = '<i aria-hidden="true" class="fa fa-cogs ms-2 text-secondary" id="job-cogs-color"></i>';
+      }
```
jraddaoui: I believe this ids are only there to set the color from the CSS.

plugins/arDominionB5Plugin/modules/jobs/templates/reportSuccess.php
```diff
@@ -4,73 +4,53 @@
...
+      <div class="alert alert-danger m-2" role="alert">
+        <pre id="job-log-error-output"><?php echo render_value_inline($errorOutput); ?></pre>
+      </div>
     </div>
   <?php } ?>
 
-  <h2><?php echo __('Log'); ?></h2>
-  <div>
+  <?php echo render_b5_section_label(__('Log')); ?>
+  <div class="bg-secondary alert text-white rounded p-3 m-2">
```
jraddaoui: Now that I see the `pre` section bellow ... could we remove the `alert` class here and add `pb-0` to the `pre`? I'd prefer to not depend on that temporary CSS if we can.

replaceafill: `mb-0` :wink:

PR #1377, opened 2021-08-04
Style content of term index page

replaceafill:


jraddaoui (approved):

lib/helper/QubitHelper.php
```diff
@@ -315,6 +315,16 @@ function render_b5_show_list_css_classes($options = [])
...
+function render_b5_show_nav_pills_css_classes($options = [])
+{
+    return 'nav nav-pills mb-2 d-flex gap-2';
+}
+
+function render_b5_show_nav_pills_button_css_classes($options = [])
+{
+    return 'btn atom-btn-white active-orange text-wrap';
+}
+
```
jraddaoui: I lost the comment while approving.

It's not a big deal, but I'm not a big fan of this helper functions for classes. In the end calling them in the templates is even larger that the classes, but the main reason is that I think the may not be the same in all cases. For example the nav pills in the term index page, due to their current location, I think they deserve a bigger margin bottom.

replaceafill: Understood. I removed the helpers and set `mb-3` on the `ul.nav-pills` to match the `ol.breadcrumbs` in the term index page in https://github.com/artefactual/atom/pull/1377/commits/7daa1219b4e39ec4d81d806ff8d4b0167dd1aaa9

Let me know if that works.

PR #1382, opened 2021-08-05
Prevent actor import dupe relations, refs #13512

mcantelon:
Fixed problem where, when in update mode, CSV authority record import was creating duplicate relations to place and subject terms.

### DIFF
--- a/lib/task/import/csvAuthorityRecordImportTask.class.php
+++ b/lib/task/import/csvAuthorityRecordImportTask.class.php
@@ -65,6 +65,42 @@ EOF;
             QubitTaxonomy::DESCRIPTION_DETAIL_LEVEL_ID => 'detailLevelTypes',
         ]);
 
+        // Define helper function used in post-save logic
+        $relateTermFunction = function ($import, $column, $taxonomyId) {
+            if (empty($import->rowStatusVars[$column])) {
+                return;
+            }
+
+            $terms = explode('|', $import->rowStatusVars[$column]);
+
+            for ($i = 0; $i < count($terms); ++$i) {
+                if (empty($terms[$i])) {
+                    continue;
+                }
+
+                // Attempt to create relation to term object
+                $relation = QubitActor::setTermRelationByName(
+                    $terms[$i],
+                    [
+                        'taxonomyId' => $taxonomyId,
+                        'culture' => $import->columnValue('culture'),
+                    ]
+                );
+
+                if (null === $relation) {
+                    continue;
+                }
+
+                // If a relation to term object could be created, associate it with this actor and save
+                $relationAlreadyExists = QubitFlatfileImport::objectTermRelationExists($import->object->id, $relation->termId);
+
+                if (!$relationAlreadyExists) {
+                    $relation->object = $import->object;
+                    $relation->save();
+                }
+            }
+        };
+
         // Define import
         $import = new QubitFlatfileImport([
             // Pass context
@@ -88,6 +124,7 @@ EOF;
                 'detailLevelTypes' => $termData['detailLevelTypes'],
                 'aliases' => $aliases,
                 'actorNames' => [],
+                'relateTermFunction' => $relateTermFunction,
             ],
 
             // Import columns that map directory to QubitActor properties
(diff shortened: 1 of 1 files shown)


lib/task/import/csvAuthorityRecordImportTask.class.php
```diff
@@ -284,7 +284,14 @@ public function execute($arguments = [], $options = [])
                                 continue;
                             }
 
-                            if (null !== $relation = QubitActor::setTermRelationByName($places[$i], $options = ['taxonomyId' => QubitTaxonomy::PLACE_ID, 'culture' => $self->columnValue('culture')])) {
+                            $relation = QubitActor::setTermRelationByName(
+                                            $places[$i],
+                                            ['taxonomyId' => QubitTaxonomy::PLACE_ID,
```
djjuhasz: I think the first array element should be on the next line, e.g.

```
$relation = QubitActor::setTermRelationByName(
    $places[$i],
    [
        'taxonomyId' => QubitTaxonomy::PLACE_ID,
        'culture' => $self->columnValue('culture'), 
    ]
);
```

lib/task/import/csvAuthorityRecordImportTask.class.php
```diff
@@ -284,7 +284,14 @@ public function execute($arguments = [], $options = [])
...
+                                                'culture' => $self->columnValue('culture'), ]);
+
+                            $relationAlreadyExists = QubitFlatfileImport::objectTermRelationExists($self->object->id, $relation->termId);
```
djjuhasz: I think `$relation->term` will throw a fatal error if `$relation === null`, which can happen if a term doesn't exist and the current user doesn't have permission to create a new term in the taxonomy.

lib/task/import/csvAuthorityRecordImportTask.class.php
```diff
@@ -299,7 +306,14 @@ public function execute($arguments = [], $options = [])
                                 continue;
                             }
 
-                            if (null !== $relation = QubitActor::setTermRelationByName($subjects[$i], $options = ['taxonomyId' => QubitTaxonomy::SUBJECT_ID, 'culture' => $self->columnValue('culture')])) {
+                            $relation = QubitActor::setTermRelationByName(
```
djjuhasz: Most of this code is duplicated for places and subjects, and can be refactored into a method called in both places.

djjuhasz (changes requested):
@mcantelon I've requested a few changes.

lib/task/import/csvAuthorityRecordImportTask.class.php
mcantelon: Good catch!

mcantelon:
Good suggestions... PR updated!

lib/task/import/csvAuthorityRecordImportTask.class.php
```diff
@@ -65,6 +65,35 @@ public function execute($arguments = [], $options = [])
...
+
+                    $relation = QubitActor::setTermRelationByName(
+                        $places[$i],
+                        [
+                            'taxonomyId' => $taxonomyId,
+                            'culture' => $import->columnValue('culture'),
+                        ]
+                    );
+
+                    if (null !== $relation) {
```
djjuhasz: Opinion: For code readability I prefer avoiding nested conditionals when possible.  In this case an alternative to the nesting here is to do:

```
if (null === $relation) {
    continue;
}
```

lib/task/import/csvAuthorityRecordImportTask.class.php
```diff
@@ -276,6 +306,7 @@ public function execute($arguments = [], $options = [])
                         $info->save();
                     }
 
+                    /*
```
djjuhasz: This looks like a mistake?

lib/task/import/csvAuthorityRecordImportTask.class.php
```diff
@@ -65,6 +65,37 @@ public function execute($arguments = [], $options = [])
             QubitTaxonomy::DESCRIPTION_DETAIL_LEVEL_ID => 'detailLevelTypes',
         ]);
 
+        // Define helper function used in post-save logic
+        $relateTermFunction = function ($import, $column, $taxonomyId) {
+            if (!empty($import->rowStatusVars[$column])) {
+                $places = explode('|', $import->rowStatusVars[$column]);
```
djjuhasz: Heh, I think `$terms` (or similar) is preferable to `$places` now that this is a generic method. :)

djjuhasz (changes requested):
Thanks for the update @mcantelon.  I added a couple of more minor change requests.

lib/task/import/csvAuthorityRecordImportTask.class.php
djjuhasz: Oops, old code.

lib/task/import/csvAuthorityRecordImportTask.class.php
mcantelon: D'oh! Good point. ;)

lib/task/import/csvAuthorityRecordImportTask.class.php
```diff
@@ -65,6 +65,40 @@ public function execute($arguments = [], $options = [])
             QubitTaxonomy::DESCRIPTION_DETAIL_LEVEL_ID => 'detailLevelTypes',
         ]);
 
+        // Define helper function used in post-save logic
+        $relateTermFunction = function ($import, $column, $taxonomyId) {
+            if (!empty($import->rowStatusVars[$column])) {
```
djjuhasz: Oops, I missed this big conditional last time @mcantelon.  I think it's better to exit the function rather than nest the body in a long conditional, i.e.:

```
if (empty($import->rowStatusVars[$column])) {
    return;
}
```

djjuhasz (changes requested):
Sorry, one more change request!

lib/task/import/csvAuthorityRecordImportTask.class.php
mcantelon: Good thinking!

mcantelon:
Ready for another review... thanks!

djjuhasz (approved):
Looks good  @mcantelon! :shipit: 😁

mcantelon:
Woo... thanks for suggestions! Much cleaner now!

PR #1386, opened 2021-08-10
WIP: Worker memory leak

jraddaoui:
- Install memprof PHP extension:

Requires bsd-compat-headers and judy. Needs
https://gitlab.alpinelinux.org/alpine/aports/-/merge_requests/23089
to be able to install judy with apk.


jraddaoui:
This was addressed by @sbreker in https://github.com/artefactual/atom/pull/1452 and https://github.com/artefactual/atom/pull/1453.

PR #1387, opened 2021-08-11
Add sidebar treeview

sevein:
You may not want to follow the history of the commits and just look at the final diff.

Some things to consider:
* Added a new `render_b5_treeview_node` helper but it's pretty similar, just fitting the new B5 layout where needed. Each row is now made of three high-level elements: arrow + grip + text (span that contains pub status, title and level of description).
* I've used flex in treeview rows, but I've used CSS rules because I thought it'd be more efficient when having hundreds of rows.
* Arrow icons are working but they're still improvable. Still using `<i>`, but sprites were replaced with font-awesome icons via CSS content. Tagged with aria-hidden for now, I suggest to address this specific issue later if we decide we want to.
* Dependency jquery-mousewheel added to package.json (we used to keep it embedded in the old treeview js).
* When there is just a single tab (e.g. Browse Subjects page) we used to hide the tab menu via PHP templating, but it's now done with CSS (`#treeview-menu .nav-item:only-child`), I think it makes the template easier to read.
* The new js/treeview.js is not a big depature, but I've applied some clean-ups along the way (this is where having the commit history may result interesting). Some further refactoring could be done but I resisted the temptation, e.g. factor out search and list tabs functionality into different classes, or a HTTP client.


plugins/arDominionB5Plugin/js/treeView.js
```diff
@@ -0,0 +1,701 @@
...
+        );
+
+      this.$list.on(
+        "click.treeview.atom",
+        ".pagination a",
+        this.clickPagerButton.bind(this)
+      );
+
+      // Search box auto focus.
+      this.$menu.on("shown.bs.tab", "#treeview-search-tab", (event) => {
```
jraddaoui, line 103: Nice!

plugins/arDominionB5Plugin/scss/_layout.scss
```diff
@@ -123,3 +123,93 @@ body > iframe {
...
   height: 200px;
 }
+
+#treeview {
+  max-height: 36rem;
+  overflow: hidden auto;
+
+  // Sortable placeholder.
+  .placeholder {
+    background-color: rgba(255, 255, 255, 0.8);
```
jraddaoui: We could use one of the grays ($gray-100 ... $gray-900) for this rgba colors.

plugins/arDominionB5Plugin/scss/_layout.scss
```diff
@@ -123,3 +123,93 @@ body > iframe {
...
+  // Each sortable/expandable treeview row.
+  .list-group-item {
+    display: flex;
+    flex-flow: row nowrap;
+    align-items: baseline;
+  }
+
+  .grip,
+  .arrow {
+    min-width: 14px;
```
jraddaoui: Use $spacer please, it defaults to the browser font size (usually 16px).

plugins/arDominionB5Plugin/scss/_layout.scss
```diff
@@ -123,3 +123,93 @@ body > iframe {
...
+  }
+
+  // Grip (sortable functionality).
+  .grip {
+    cursor: move;
+    margin-left: auto;
+    @extend .text-muted;
```
jraddaoui, line 167: Nice!

plugins/arDominionB5Plugin/scss/_layout.scss
```diff
@@ -123,3 +123,93 @@ body > iframe {
...
+  }
+
+  // Ancestors show darker rows.
+  .ancestor {
+    background-color: $gray-200;
+  }
+
+  // Compensation when the left arrow is unused.
+  .text:first-child {
+    margin-left: 22px;
```
jraddaoui: Same, use $spacer. If you want it bigger you can multiply `$spacer * 1.25`.

jraddaoui (approved):
Really nice @sevein, thanks!

plugins/arDominionB5Plugin/scss/_layout.scss
sevein: This looks like it can do the trick:
```
@use "sass:color";
background-color: color.adjust($white, $alpha: -0.2);
```

jraddaoui: I was thinking on Bootstrap variables instead of the Sass approach. Looking at their docs they use `rgba` so `rgba($white, 0.8)` may be cleaner in the end, without the need of `@use`.

https://getbootstrap.com/docs/5.0/utilities/colors/#utilities-api

Sorry for the back and forward, and not a big deal of course, feel free to squash and merge ;)

sevein: Cool, that worked. Fixed.

sevein:
Squashed (13018a323).

PR #1390, opened 2021-08-13
Init Masonry with imagesloaded

sevein:
I've finally used [imagesloaded](https://github.com/desandro/imagesloaded), it seems to be the best solution because it fires exactly when all images within the `.masonry` container are loaded. Other solutions have been considered but they fire either too early (`DOMContentLoaded`) or too late (`window.onload`).

* `$(document).ready` or `DOMContentLoaded` may happen too early, i.e. when the DOM tree is built, which is a point where the images may or may not be loaded, but likely the former.
* `defer` seems to [relate](`https://developer.mozilla.org/en-US/docs/Web/HTML/Element/script#attr-defer`) to `DOMContentLoaded`, only preventing the event to fire until the script has been loaded and evaluated, but not the images.
* `data-masonry`, which is looked up by Masonry when `DOMContentLoaded` is fired (https://github.com/desandro/masonry/blob/3b0883cf4a4a046896719b9cf282ea74be7ffecd/dist/masonry.pkgd.js#L703-L711), so the same problem applies.
* `window.onload` fires when all resources are loaded, included fonts, etc... this would work but it's more effective to have Masonry do its rendering as soon as possible, i.e. when all images are loaded.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -12,6 +12,7 @@
         "@fortawesome/fontawesome-free": "^5.15.3",
         "bootstrap": "^5.0.1",
         "bootstrap-autocomplete": "^2.3.7",
+        "imagesloaded": "^4.1.4",
         "jquery": "^3.6.0",
         "jquery-expander": "^2.0.1",
         "jquery-mousewheel": "^3.1.13",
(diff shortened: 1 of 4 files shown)


jraddaoui (approved):
Awesome! Thanks for the research and info!

PR #1393, opened 2021-08-16
arDominionB5Plugin: style edit theme form

sevein:


lib/helper/QubitHelper.php
```diff
@@ -95,6 +95,8 @@ function render_b5_field($field, $translation = null, $options = [])
         $isFormCheck = true;
         $inputClass = 'form-check-input';
         $labelClass = 'form-check-label';
+    } elseif ('color' == $field->type) {
+        $inputClass = 'form-control form-control-color';
```
jraddaoui: Nitpick: you could concatenate here.

jraddaoui (approved):
Looking amazing @sevein!

In Bootstrap 4 there is a way to translate the "Browse"/"Choose file" text in the file input ...

https://getbootstrap.com/docs/4.5/components/forms/#translating-or-customizing-the-strings-with-html

However, I can't find a similar solution for B5, maybe it's using the lang attribute of the html element or it requires its own attribute. Also, I found this ...

https://web3canvas.com/convert-bootstrap-4-to-5-migration-guide/  (look for "file browser")

But I couldn't find anything from B5 docs.

lib/helper/QubitHelper.php
sevein: Fixed, thanks.

sevein:
> In Bootstrap 4 there is a way to translate the "Browse"/"Choose file" text in the file input ...

Let me know if you want me to implement our own field, it'd be easy using the File API. I can see why B5 gave up on that old method, nobody wants to manage translations using Sass variables.

I've found the label of the native button to read in other languages when switching the system language, with the much larger translation dataset provided by the browser. But you won't see the translation just by switching the application language, not even when adding the content-language response header.

image

PR #1400, opened 2021-08-18
CSV validator enhancements, refs #13475

sbreker:
This commit contains enhancements to the CSV Validation utility:
- Add error and warning numbers to output reports.
- Updates the character encoding validator output to include only the first 10 CSV line numbers that triggered the error.
- In some validators, if an issue was detected in a CSV row, the entire row would be appended to the detailed results. When there are several rows this becomes difficult to read. This commit replaces this output with a summary of the row numbers that triggered the errors.
- Add more specific messaging to the parentId validator when QubitParentSlug is the source of an issue (vs parentId).
- if duplicate legacyIds are consecutive and culture is also the same then throw an error as this will trigger a duplicate translation error on import.
- DO Path check will no longer stop when the digitalObjectPath is populated but the path option is not set. Instead it will complete and output the files that were unable to be found.
- digitalObjectURI validator will now check 'digitalObjectURI' and not 'digitalObjectUri'.

### DIFF
--- a/lib/task/import/validate/csvDigitalObjectPathValidator.class.php
+++ b/lib/task/import/validate/csvDigitalObjectPathValidator.class.php
@@ -34,7 +34,7 @@ class CsvDigitalObjectPathValidator extends CsvBaseValidator
     protected $fileList = [];
     protected $pathToDigitalObjects = '';
     // Reset between files.
-    protected $digitalObjectUses = [];
+    protected $digitalObjectUseCountList = [];
     protected $overriddenByUriCount = 0;
 
     public function __construct(?array $options = null)
@@ -48,7 +48,7 @@ class CsvDigitalObjectPathValidator extends CsvBaseValidator
 
     public function reset()
     {
-        $this->digitalObjectUses = [];
+        $this->digitalObjectUseCountList = [];
         $this->overriddenByUriCount = 0;
 
         parent::reset();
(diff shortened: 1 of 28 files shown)


lib/task/import/validate/csvFileEncodingValidator.class.php
```diff
@@ -106,7 +106,10 @@ protected function finalizeTestResults()
             $this->testData->addResult('File encoding is UTF-8 compatible.');
         } else {
             $this->testData->addResult('File encoding does not appear to be UTF-8 compatible.');
+            $this->testData->addResult(sprintf('Count of UTF-8 incompatible CSV rows: %s', count($this->getCsvRowList())));
             $this->testData->setStatusError();
+            // Detail lines: comma sep string of first ten row numbers affected.
+            $this->testData->addDetail(sprintf('Affected row numbers (first 10): %s', implode(', ', array_slice($this->getCsvRowList(), 0, 10, true))));
```
mcantelon: Nitpick: This line's a bit long.

lib/task/import/validate/csvDigitalObjectPathValidator.class.php
```diff
@@ -172,14 +164,14 @@ public function setPathToDigitalObjects(string $path)
 
     protected function addToUsageSummary($value)
     {
-        $this->digitalObjectUses[$value] = (!isset($this->digitalObjectUses[$value])) ? 1 : $this->digitalObjectUses[$value] + 1;
+        $this->digitalObjectUseCountList[$value] = (!isset($this->digitalObjectUseCountList[$value])) ? 1 : $this->digitalObjectUseCountList[$value] + 1;
```
mcantelon: Nitpick: This line's a bit long.

mcantelon (approved):
Looks good other than a couple super minor nitpicks!  👍

mcantelon:
One more nitpick... there's a type in one of the commit messages ("vaalidator ").

PR #1401, opened 2021-08-19
Use aria-invalid when a field has errors

sevein:
[`aria-invalid`](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_aria-invalid_attribute) seems to be an effective way to indicate validity state in server-side validation for assitive technologies. In this [post](https://stackoverflow.com/a/52651709), the author says:

> So to cover all your bases, I would recommend setting aria-invalid. It's great that you don't want to be redundant, but at least being redundant doesn't hurt anything. It's just extra code or typing you have to do.


jraddaoui (approved):
Nice!

PR #1413, opened 2021-08-29
Set manual L.O.D. in CSV transform, refs #13557

mcantelon:
Added logic to allow available levels of description, used to order
the output of CSV rows by level of description hierarchy, to be set
manually rather than being restricted to the default AtoM levels of
description.

Also made numberedFilePathVariation method static so a script can
be used to generate an import Bash script for CSV transformations
that generate a lot of CSV files.

### DIFF
--- a/lib/QubitCsvTransform.class.php
+++ b/lib/QubitCsvTransform.class.php
@@ -23,6 +23,7 @@ class QubitCsvTransform extends QubitFlatfileImport
     public $transformLogic;
     public $rowsPerFile = 1000;
     public $preserveOrder = false;
+    public $levelsOfDescription;
     public $convertWindowsEncoding = false;
 
     private $link;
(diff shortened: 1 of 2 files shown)


lib/QubitCsvTransform.class.php
```diff
@@ -134,13 +144,13 @@ public function addRowToMySQL($sortorder)
...
 
-    public function numberedFilePathVariation($filename, $number)
+    public static function numberedFilePathVariation($filename, $number)
     {
         $parts = pathinfo($filename);
         $base = $parts['filename'];
         $path = $parts['dirname'];
 
-        return $path.'/'.$base.'_'.$number.'.'.$parts['extension'];
+        return $path.'/'.$base.'_'.sprintf('%04d', $number).'.'.$parts['extension'];
```
djjuhasz: Suggestion: Could use `sprintf()` for the whole string, and use `$parts['filename']` and `$parts['dirname']` directly instead of assigning one off variables.

djjuhasz (commented):
Just one suggestion @mcantelon. :)

lib/QubitCsvTransform.class.php
mcantelon: Good call... that definitely ends up looking cleaner.

mcantelon:
I've implemented the PR feedback. Thanks!

lib/QubitCsvTransform.class.php
```diff
@@ -147,10 +147,14 @@ public function addRowToMySQL($sortorder)
...
         $parts = pathinfo($filename);
-        $base = $parts['filename'];
-        $path = $parts['dirname'];
 
-        return $path.'/'.$base.'_'.sprintf('%04d', $number).'.'.$parts['extension'];
+        return sprintf(
+            '%s/%s_%s.%s',
+            $parts['dirname'],
+            $parts['filename'],
+            sprintf('%04d', $number),
```
djjuhasz: This doesn't need a nested sprintf(), you can do:

```
return sprintf(
            '%s/%s_%04d.%s',
            $parts['dirname'],
            $parts['filename'],
            $number,
            $parts['extension']
);
```

djjuhasz (changes requested):

lib/QubitCsvTransform.class.php
mcantelon: Ahhh! Slick! Good call.

mcantelon:
I've implemented that suggestion, @djjuhasz!

djjuhasz (approved):

PR #1414, opened 2021-08-29
Fix CSV invalid parent replacement ID, refs #13558

mcantelon:
Fixed issue where, during CSV import of information objects, the
replacement parent ID of an invalid parent ID wouldn't reflect
the --default-legacy-parent-id option setting.

### DIFF
--- a/lib/task/import/csvImportTask.class.php
+++ b/lib/task/import/csvImportTask.class.php
@@ -482,18 +482,15 @@ EOF;
                         } elseif (null !== QubitInformationObject::getById($self->rowStatusVars['parentId'])) {
                             $parentId = $self->rowStatusVars['parentId'];
                         } else {
+                            $parentId = $self->status['defaultParentId'];
+
                             $error = sprintf(
-                                'legacyId %s: could not find parentId %s in key_map table or existing data. Setting parent to root...',
+                                'legacyId %s: could not find parentId %s in key_map table or existing data. Setting parent to default...',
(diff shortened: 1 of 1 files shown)


lib/task/import/csvImportTask.class.php
```diff
@@ -482,18 +482,15 @@ public function execute($arguments = [], $options = [])
                         } elseif (null !== QubitInformationObject::getById($self->rowStatusVars['parentId'])) {
                             $parentId = $self->rowStatusVars['parentId'];
                         } else {
+                            $parentId = $self->status['defaultParentId'];
```
djjuhasz, line 485: In the original code there is this conditional:

```
        if ($notImportingTranslation) {
            $self->object->parentId = QubitInformationObject::ROOT_ID;
        }
```

Is this no longer necessary?

djjuhasz (commented):
@mcantelon I've got a question about one of the changes.

lib/task/import/csvImportTask.class.php
mcantelon: Yeah, it's redundant as the same check gets done a little later in the code:

https://github.com/artefactual/atom/blob/qa/2.x/lib/task/import/csvImportTask.class.php#L501

djjuhasz: Ah okay. 👍

djjuhasz (approved):
Looks good! :shipit:

PR #1415, opened 2021-08-30
Update Regex on Slug for API Requests, refs #13476

melaniekung:
Update regex on REGEX_SLUG to accommodate valid characters for API Requests

### DIFF
--- a/lib/model/QubitSlug.php
+++ b/lib/model/QubitSlug.php
@@ -25,6 +25,19 @@ class QubitSlug extends BaseSlug
     public const SLUG_BASIS_IDENTIFIER = 3;
     public const SLUG_RESTRICTIVE = 0;
     public const SLUG_PERMISSIVE = 1;
+    public const SLUG_RESTRICTIVE_CHARS = '0-9a-z-';
+
+    // From RFC 3987 IRI allowed chars. Not guaranteed to match \p{L}\p{Nd}.
+    public const SLUG_RFC_3987_CHARS = "\u{00A0}-\u{D7FF}"."\u{F900}-\u{FDCF}"
+        ."\u{FDF0}-\u{FFEF}"."\u{10000}-\u{1FFFD}"."\u{20000}-\u{2FFFD}"
+        ."\u{30000}-\u{3FFFD}"."\u{40000}-\u{4FFFD}"."\u{50000}-\u{5FFFD}"
+        ."\u{60000}-\u{6FFFD}"."\u{70000}-\u{7FFFD}"."\u{80000}-\u{8FFFD}"
+        ."\u{90000}-\u{9FFFD}"."\u{A0000}-\u{AFFFD}"."\u{B0000}-\u{BFFFD}"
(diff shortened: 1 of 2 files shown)


plugins/arRestApiPlugin/config/arRestApiPluginConfiguration.class.php
```diff
@@ -24,7 +24,7 @@ class arRestApiPluginConfiguration extends sfPluginConfiguration
 {
     public const REGEX_UUID = '[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}';
     public const REGEX_ID = '\d+';
-    public const REGEX_SLUG = '[0-9A-Za-z-]+';
+    public const REGEX_SLUG = '[0-9A-Za-z-_~:;,=*@]+';
```
djjuhasz: @melaniekung This looks good, but I'm curious how you determined which characters to add to the Regex?

djjuhasz (commented):
Hi @melaniekung.  I just has one question about the regex you used.

plugins/arRestApiPlugin/config/arRestApiPluginConfiguration.class.php
djjuhasz: Ah, I think the new characters correspond to the "extraSet" values defined in: https://github.com/artefactual/atom/blob/qa/2.x/lib/model/QubitSlug.php#L73

djjuhasz: I think the regex for valid slug character should be defined once in the `QubitSlug` class (via a static method like `validSlugRegex()` or similar?) and then called here, rather than maintaining two different regex strings in two different parts of the code.

djjuhasz (changes requested):
Hi @melaniekung, I've requested a change to avoid defining a regex for testing valid slugs here and in the `QubitSlug` class.

lib/model/QubitSlug.php
```diff
@@ -85,14 +95,14 @@ public static function slugify($slug, $creationType = null)
...
                 $slug = strtolower($slug);
                 // Allow only digits, letters, and dashes.  Replace sequences of other
                 // characters with dash
-                $slug = preg_replace('/[^0-9a-z]+/', '-', $slug);
+                $slug = preg_replace('/[^'.self::SLUG_RESTRICTIVE_CHARS.']+/', '-', $slug);
         }
 
+        // Replace repeating dashes in slug with single dash.
+        $slug = preg_replace('/-+/', '-', $slug);
         $slug = "-{$slug}-";
```
djjuhasz: I can't see any need for `$slug = "-{$slug}-";`, as the hyphens added are immediately trimmed on line 109. 😕  I think this line should be removed.

djjuhasz (changes requested):
@melaniekung one more change request.

djjuhasz (approved):
Looks good @melaniekung. 👍 Please make sure to squash the commits before merging them to qa/2.x.

PR #1417, opened 2021-08-30
Update way of defining master copy, refs #13543

melaniekung:
Defined master copy using QubitTerm::MASTER_ID since getRepresentationByUsage() only searches direct children of current digital object

### DIFF
--- a/apps/qubit/modules/digitalobject/actions/metadataComponent.class.php
+++ b/apps/qubit/modules/digitalobject/actions/metadataComponent.class.php
@@ -26,6 +26,10 @@ class DigitalObjectMetadataComponent extends sfComponent
 {
     public function execute($request)
     {
+        if (!isset($this->resource)) {
+            return;
+        }
+
         // Check related object type to display IO properties in the template
         $this->relatedToIo = $this->resource->object instanceof QubitInformationObject;
         $this->relatedToActor = $this->resource->object instanceof QubitActor;
@@ -36,7 +40,6 @@ class DigitalObjectMetadataComponent extends sfComponent
             && arStorageServiceUtils::getAipDownloadEnabled()
         );
 
-        $this->masterCopy = $this->resource->getRepresentationByUsage(QubitTerm::MASTER_ID);
         $this->referenceCopy = $this->resource->getRepresentationByUsage(QubitTerm::REFERENCE_ID);
         $this->thumbnailCopy = $this->resource->getRepresentationByUsage(QubitTerm::THUMBNAIL_ID);
 
@@ -195,7 +198,7 @@ class DigitalObjectMetadataComponent extends sfComponent
             && !$this->isEmpty($this->masterFileDenyReason)
         );
 
-        return (
+        return
             $this->showMasterFileGoogleMap
             || $this->showMasterFileGeolocation
             || $this->showMasterFileURL
@@ -204,7 +207,7 @@ class DigitalObjectMetadataComponent extends sfComponent
             || $this->showMasterFileMimeType
             || $this->showMasterFileSize
             || $this->showMasterFileCreatedAt
-        ) && null !== $this->masterCopy;
+        ;
     }
 
     protected function setReferenceCopyShowProperties()


apps/qubit/modules/digitalobject/actions/metadataComponent.class.php
```diff
@@ -36,7 +36,8 @@ public function execute($request)
             && arStorageServiceUtils::getAipDownloadEnabled()
         );
 
-        $this->masterCopy = $this->resource->getRepresentationByUsage(QubitTerm::MASTER_ID);
+        $id = QubitTerm::MASTER_ID;
```
djjuhasz: This line is unnecessary, the `$id` variable is never used.

melaniekung: @djjuhasz it is used in line 40

apps/qubit/modules/digitalobject/actions/metadataComponent.class.php
```diff
@@ -36,7 +36,8 @@ public function execute($request)
...
+        $this->masterCopy = $this->resource->id;
```
djjuhasz: Declaring a `$this->masterCopy` variable doesn't seem necessary, as it's only used once (line 208), and the assigned value can be tested directly, e.g.

```
        ) && null !== $this->masterCopy;
```

Can be changed to:

```
        ) && isset($this->resource->id);
```

I believe using `isset()` instead of `null !== $this->resource->id` avoids a raising a warning if `$this->resource` is null.

djjuhasz (changes requested):
Hi @melaniekung, I've requested a few changes. :)

apps/qubit/modules/digitalobject/actions/metadataComponent.class.php
djjuhasz: No, line 40 uses `$this->resource->id`, the "id" references a property of $this->resource not the local variable `$id`.

apps/qubit/modules/digitalobject/actions/metadataComponent.class.php
```diff
@@ -205,7 +203,7 @@ protected function setMasterFileShowProperties()
             || $this->showMasterFileMimeType
             || $this->showMasterFileSize
             || $this->showMasterFileCreatedAt
-        ) && null !== $this->masterCopy;
+        ) && isset($this->resource->id);
```
djjuhasz: @melaniekung sorry, but I re-read the code and if the code makes it this far `isset($this->resource->id)` is always going to be true, so it's unnecessary here.  I think it's better to test `if (!isset($this->resource)` at the beginning of the `execute()` method, and return if `$this->resource` is null.

djjuhasz (changes requested):
Sorry, but I re-read my suggested change, and I think I was wrong.  I believe the test `isset($this->resource->id)` is extraneous and should be removed.

apps/qubit/modules/digitalobject/actions/metadataComponent.class.php
```diff
@@ -36,10 +36,13 @@ public function execute($request)
             && arStorageServiceUtils::getAipDownloadEnabled()
         );
 
-        $this->masterCopy = $this->resource->getRepresentationByUsage(QubitTerm::MASTER_ID);
         $this->referenceCopy = $this->resource->getRepresentationByUsage(QubitTerm::REFERENCE_ID);
         $this->thumbnailCopy = $this->resource->getRepresentationByUsage(QubitTerm::THUMBNAIL_ID);
 
+        if (!isset($this->resource)) {
```
djjuhasz: I think this conditional block should be moved above line 29 because line 29 & 30 will throw an error if `$this->resource` is null.

djjuhasz (changes requested):
One more change request @melaniekung

djjuhasz (approved):
Looks good @melaniekung!  Please make sure to squash all of the commits on the branch before merging to qa/2.x, and to update the commit summary to match the final code changes.

PR #1419, opened 2021-09-01
Add CSV task check dir. of files, refs #13561

mcantelon:
Transformations of large CSV files can generate many files. This adds
the ability to use the CSV check task on a directory full of files
rather than just single files or a list of files.

### DIFF
--- a/lib/task/import/csvCheckImportTask.class.php
+++ b/lib/task/import/csvCheckImportTask.class.php
@@ -62,7 +62,7 @@ class csvCheckImportTask extends arBaseTask
     {
         $this->addArguments([
             new sfCommandArgument('filename', sfCommandArgument::REQUIRED,
-              'The input file name (csv format).'),
+              'The input file name (CSV format) or directory of CSV files.'),
         ]);
 
         $this->addOptions([
@@ -137,6 +137,11 @@ EOF;
 
     protected function setCsvValidatorFilenames($filenameString)
     {
+        // If a directory's provided return an array of file paths
+        if (is_dir($filenameString) && is_readable($filenameString)) {
+            return $this->getFilePathsFromDirectory($filenameString);
+        }
+
         // Could be a comma separated list of filenames or just one.
         foreach (explode(',', $filenameString) as $filename) {
             CsvImportValidator::validateFileName($filename);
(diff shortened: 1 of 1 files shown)


lib/task/import/csvCheckImportTask.class.php
```diff
@@ -137,6 +137,25 @@ protected function getDbConnection()
...
+        // If a directory's provided return an array of file paths
+        if (is_dir($filenameString)) {
+            $files = array_diff(scandir($filenameString), ['..', '.']);
+
+            $filePaths = [];
+
+            foreach ($files as $file) {
+                $filePath = $filenameString.DIRECTORY_SEPARATOR.$file;
+
+                if (is_file($filePath)) {
```
djjuhasz: I think it would be good to check `is_readable($filePath)` as well.  

I also looked at `CsvImportValidator::validateFileName($filePath);` as a possible alternative, but it throws an exception if any of the checks fail and doesn't check `is_file()` specifically.

lib/task/import/csvCheckImportTask.class.php
```diff
@@ -137,6 +137,25 @@ protected function getDbConnection()
 
     protected function setCsvValidatorFilenames($filenameString)
     {
+        // If a directory's provided return an array of file paths
+        if (is_dir($filenameString)) {
+            $files = array_diff(scandir($filenameString), ['..', '.']);
```
djjuhasz: I'd prefer to see the directory scan logic moved to a separate method for readability and testability, but it's not a blocker.

djjuhasz (commented):
@mcantelon I added a couple of suggestions.

lib/task/import/csvCheckImportTask.class.php
mcantelon: Sounds good!

lib/task/import/csvCheckImportTask.class.php
mcantelon: Makes sense!

mcantelon:
Thanks @djjuhasz ! I've updated the PR to address those issues!

djjuhasz (approved):
Looks good! :shipit:

PR #1420, opened 2021-09-01
arDominionB5Plugin: more work in settings pages

sevein:


plugins/arDominionB5Plugin/modules/settings/templates/treeviewSuccess.php
```diff
@@ -82,40 +76,36 @@
...
-                          '%1%' => format_number(10),
-                          '%2%' => format_number(
-                            sfConfig::get('app_treeview_items_per_page_max', 10000)
-                          ),
-                      ]
-                    )
-                  )
-                  ->renderRow(); ?>
-            </p>
+            <?php echo render_field(
```
replaceafill: Should this be wrapped in a `div.col-md-4` and put in the same container of rows as the other fields above?

replaceafill (approved):
:+1:

plugins/arDominionB5Plugin/modules/settings/templates/treeviewSuccess.php
sevein: Thanks, will do. I've used col-md-12 to dedicate a full line, matching the old layout.

PR #1421, opened 2021-09-01
Style error pages

replaceafill:


plugins/arDominionB5Plugin/modules/admin/templates/error404Success.php
```diff
@@ -0,0 +1,20 @@
+<section class="w-50 mx-auto rounded border border-dark mt-3 border-1 bg-white">
+
+  <h2 class="border-bottom px-3 py-2">
```
sevein: We're using h1 in other headings, should we do the same here?

plugins/arDominionB5Plugin/modules/admin/templates/error404Success.php
```diff
@@ -0,0 +1,20 @@
+<section class="w-50 mx-auto rounded border border-dark mt-3 border-1 bg-white">
```
sevein: I'm wondering if [`role="alert"`](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_alert_role) would be a good fit in these error pages.

sevein (approved):
Just a couple of minor comments/suggestions to consider.

PR #1425, opened 2021-09-02
arDominionB5Plugin: more work in settings pages

sevein:
This PR updates all the remaining setting pages not styled yet excepting the permissions page.

### DIFF
--- a/lib/form/arWidgetFormUploadQuota.class.php
+++ b/lib/form/arWidgetFormUploadQuota.class.php
@@ -21,6 +21,10 @@ class arWidgetFormUploadQuota extends sfWidgetFormInput
 {
     public function render($name, $value = null, $attributes = [], $errors = [])
     {
+        if (sfConfig::get('app_b5_theme', false)) {
+            return $this->renderB5($name, $value, $attributes, $errors);
+        }
+
         $uploadLimit = (int) sfConfig::get('app_upload_limit');
 
         if (0 === $uploadLimit) {
@@ -44,6 +48,32 @@ class arWidgetFormUploadQuota extends sfWidgetFormInput
                 ]
             )
             .'</label>';
+    }
+
+    public function renderB5($name, $value = null, $attributes = [], $errors = [])
+    {
+        $uploadLimit = (int) sfConfig::get('app_upload_limit');
+
+        if (0 === $uploadLimit) {
+            $label = __('%1% upload is disabled', ['%1%' => sfConfig::get('app_ui_label_digitalobject')]);
+        } elseif (-1 === $uploadLimit) {
+            $label = __('Unlimited');
+        } else {
+            $label = __('The uploads directory has not been created yet.');
+
+            $size = Qubit::getDirectorySize(sfConfig::get('sf_upload_dir'));
+            if ($size >= 0) {
+                $label = __(
+                    '%1% used of %2%',
+                    [
+                        '%1%' => hr_filesize($size),
+                        '%2%' => $uploadLimit.' GB',
+                    ]
+                );
+            }
+        }
+
+        return '<input class="form-control" type="text" value="'.$label.'" aria-label="'.$label.'" disabled readonly>';
     }
 
     protected function configure($options = [], $attributes = [])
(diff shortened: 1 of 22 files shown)


lib/form/arWidgetFormUploadQuota.class.php
```diff
@@ -46,6 +50,32 @@ public function render($name, $value = null, $attributes = [], $errors = [])
...
+            if ($size >= 0) {
+                $label = __(
+                    '%1% used of %2%',
+                    [
+                        '%1%' => hr_filesize($size),
+                        '%2%' => $uploadLimit.' GB',
+                    ]
+                );
+            }
+        }
```
replaceafill, line 74: What do you think of making this block part of the existing `render` method and leave the `<label>` and `<input>` concatenations at the end for both cases to avoid the duplicated logic?

lib/helper/QubitHelper.php
```diff
@@ -99,7 +99,7 @@ function render_b5_field($field, $translation = null, $options = [])
     if (
         in_array($field->type, ['checkbox', 'radio'])
         || $widget instanceof sfWidgetFormSelectRadio
-        || $widget instanceof sfWidgetFormChoice
+        || ($widget instanceof sfWidgetFormChoice && !$widget instanceof sfWidgetFormI18nChoiceLanguage)
```
replaceafill: Should `sfWidgetFormI18nChoiceCountry` be excluded here also?

plugins/arDominionB5Plugin/modules/settings/templates/interfaceLabelSuccess.php
```diff
@@ -20,31 +20,22 @@
...
-          <tr>
-            <th><?php echo __('Name'); ?></th>
-            <th><?php echo __('Value'); ?></th>
-          </tr>
-        </thead>
-        <tbody>
-          <?php foreach ($uiLabelForm->getSettings() as $setting) { ?>
-            <tr>
-              <td>
-                <?php if ($sf_user->getCulture() != $setting->getSourceCulture() && !strlen($setting->getValue())) { ?>
```
replaceafill, line 35: Is this styling based on the culture check not needed anymore?

plugins/arDominionB5Plugin/modules/settings/templates/languageSuccess.php
```diff
@@ -25,38 +25,42 @@
...
+        <h2 class="accordion-header" id="language-heading">
+          <button class="accordion-button" type="button" data-bs-toggle="collapse" data-bs-target="#language-collapse" aria-expanded="true" aria-controls="language-collapse">
+            <?php echo __('i18n language settings'); ?>
+          </button>
+        </h2>
+        <div id="language-collapse" class="accordion-collapse collapse show" aria-labelledby="language-heading">
+          <div class="accordion-body">
+            <?php foreach ($i18nLanguages as $setting) { ?>
+              <div class="row mb-3">
+                <label for="inputEmail3" class="col-11 col-form-label">
```
replaceafill: What `inputEmail3` id this points to?

replaceafill (approved):
Nice @sevein! Left a few minor questions but this is good to go :+1:

lib/form/arWidgetFormUploadQuota.class.php
sevein: It is intentional - since we expect the B5 form to take over soon I'm trying to make that process easier for the future developer, i.e. more repetition but less time refactoring hopefully. Do you think that's a good idea?

lib/helper/QubitHelper.php
sevein: Good catch, will do.

plugins/arDominionB5Plugin/modules/settings/templates/interfaceLabelSuccess.php
sevein: Yes, render_field shows the original string nicely. I think it's doing a better job now, the old version isn't really showing the original string for some reason.

plugins/arDominionB5Plugin/modules/settings/templates/languageSuccess.php
sevein: Good catch! Will fix.

PR #1427, opened 2021-09-03
Fix issues with report page, refs #13553

mcantelon:
Fixed two issues with the report page:
* "Continue" button was showing even when no reports where available to generate
* Physical object-related reports were available even for descriptions with no physical objects


djjuhasz (approved):

mcantelon:
Thanks @djjuhasz !

PR #1431, opened 2021-09-10
Fullwidth treeview fixes, refs #13530

sbreker:
These commits address various issues with the Treeview in B5:
- add imageflow.js to common behaviours - fixes carousel after treeview reload.
- add mediaelement.js to common behaviours - fixes mediaelement player after treeview reload.
- fix ambiguous reference to "#main-column .row".
- tooltip styling updated
- modified Treeview styling
- changed the id that is used to allow the display of the treeview on an index page to be "#fullwidth-treeview-capable" instead of "#fullwidth-treeview" which was already in use.

Treeview alert box handling:
- leave "job initiated" messages in place unless another move is started
- if another move is started, remove job initiated alerts as they will be recreated.

### DIFF
--- a/plugins/arDominionB5Plugin/js/fullWidthTreeView.js
+++ b/plugins/arDominionB5Plugin/js/fullWidthTreeView.js
@@ -2,7 +2,7 @@
   "use strict";
 
   $(function () {
-    var $node = $(".index #fullwidth-treeview");
+    var $node = $(".index #fullwidth-treeview-capable");
 
     if ($node.length) {
       loadTreeView();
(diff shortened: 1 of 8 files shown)


plugins/arDominionB5Plugin/js/mediaelement.js
```diff
@@ -0,0 +1,14 @@
+(function ($) {
```
jraddaoui, line 1: Missing `"use strict";`

plugins/arDominionB5Plugin/scss/_fullwidthtreeview.scss
```diff
@@ -13,27 +13,28 @@
...
   clear: both;
 }
 
 #fullwidth-treeview {
   overflow-x: hidden;
   height: 100%;
-  background-color: #ffffff;
+  background-color: $white;
+  border: $border-width solid $border-color;
+  border-radius: $border-radius;
```
jraddaoui: Could we use the "border rounded" classes on the element directly?

jraddaoui: Also `bg-white` and `mb-3`.

plugins/arDominionB5Plugin/js/fullWidthTreeView.js
```diff
@@ -163,18 +163,50 @@
...
+
+      var $alert = $(
+        '<div class="alert ' +
+          type +
+          ' alert-dismissible fade show" role="alert">'
+      ).append(message);
+
+      var closeButton =
+        '<button type="button" class="btn-close" data-bs-dismiss="alert" aria-label="' +
+        $(".index #fullwidth-treeview-present").data("treeview-alert-close") +
```
jraddaoui: Should this be `#fullwidth-treeview-capable` instead of `#fullwidth-treeview-present`?

jraddaoui (approved):
Nice, thanks @sbreker!

I think we still need to deal with the tooltips, which I'll do in a different PR as they also require some normalization. I added a few minor comments bellow but feel free to squash and merge.

plugins/arDominionB5Plugin/js/mediaelement.js
sbreker: 👍

plugins/arDominionB5Plugin/js/fullWidthTreeView.js
sbreker: Good catch! Fixed.

plugins/arDominionB5Plugin/scss/_fullwidthtreeview.scss
sbreker: Leaving these in the scss file to avoid having to copy hierarchy.js into the B5 theme to make the equivalent changes.

PR #1434, opened 2021-09-16
Add index page section button arrows

sbreker:
Add arrow icons pointing to the right for each AtoM index page section
when it is a link.

Arrow icon matches what is used for the B5 accordion header button.

### DIFF
--- a/plugins/arDominionB5Plugin/scss/_buttons.scss
+++ b/plugins/arDominionB5Plugin/scss/_buttons.scss
@@ -92,3 +92,24 @@
     }
   }
 }
+
+// Index page section right facing arrows.
+#content a .atom-section-button {
+  position: relative;
+  display: flex;
+  align-items: center;
+  width: 100%;
+  padding: $spacer $spacer * 1.25;
+  text-align: left;
+  overflow-anchor: none;
+
+  &::after {
+    background-image: url("data:image/svg+xml,%3csvg xmlns='http://www.w3.org/2000/svg' viewBox='0 0 16 16' fill='%23212529'%3e%3cpath fill-rule='evenodd' d='M1.646 4.646a.5.5 0 0 1 .708 0L8 10.293l5.646-5.647a.5.5 0 0 1 .708.708l-6 6a.5.5 0 0 1-.708 0l-6-6a.5.5 0 0 1 0-.708z'/%3e%3c/svg%3e");
+    transform: rotate(-90deg);
+    width: $spacer * 1.25;
+    height: $spacer * 1.25;
+    margin-left: auto;
+    content: "";
+    background-repeat: no-repeat;
+  }
+}
--- a/lib/helper/QubitHelper.php
+++ b/lib/helper/QubitHelper.php
@@ -325,7 +325,7 @@ function render_b5_show_value($value, $options = [])
 
 function render_b5_section_label_css_classes($options = [])
 {
-    return 'border-bottom h5 m-0 p-3 text-primary';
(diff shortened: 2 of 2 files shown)


plugins/arDominionB5Plugin/scss/_buttons.scss
```diff
@@ -79,3 +79,24 @@
...
+
+#content a .atom-section-button::after {
+  background-image: url("data:image/svg+xml,%3csvg xmlns='http://www.w3.org/2000/svg' viewBox='0 0 16 16' fill='%23212529'%3e%3cpath fill-rule='evenodd' d='M1.646 4.646a.5.5 0 0 1 .708 0L8 10.293l5.646-5.647a.5.5 0 0 1 .708.708l-6 6a.5.5 0 0 1-.708 0l-6-6a.5.5 0 0 1 0-.708z'/%3e%3c/svg%3e");
+  transform: rotate(-90deg);
+  width: 1.25rem;
+  height: 1.25rem;
+  margin-left: auto;
+  content: "";
+  background-repeat: no-repeat;
+}
```
jraddaoui: This could be nested in the other block using `&::after`.

plugins/arDominionB5Plugin/scss/_buttons.scss
```diff
@@ -79,3 +79,24 @@
...
   }
 }
+
+// Index page section right facing arrows.
+#content a .atom-section-button {
+  position: relative;
+  display: flex;
+  align-items: center;
+  width: 100%;
+  padding: 1rem 1.25rem;
```
jraddaoui: Use `$spacer $spacer * 1.25` please.

plugins/arDominionB5Plugin/scss/_buttons.scss
```diff
@@ -79,3 +79,24 @@
...
+  text-align: left;
+  overflow-anchor: none;
+}
+
+#content a .atom-section-button::after {
+  background-image: url("data:image/svg+xml,%3csvg xmlns='http://www.w3.org/2000/svg' viewBox='0 0 16 16' fill='%23212529'%3e%3cpath fill-rule='evenodd' d='M1.646 4.646a.5.5 0 0 1 .708 0L8 10.293l5.646-5.647a.5.5 0 0 1 .708.708l-6 6a.5.5 0 0 1-.708 0l-6-6a.5.5 0 0 1 0-.708z'/%3e%3c/svg%3e");
+  transform: rotate(-90deg);
+  width: 1.25rem;
+  height: 1.25rem;
```
jraddaoui: Same ... `$spacer * 1.25`,

jraddaoui (changes requested):
Awesome @sbreker, thanks!

This may change a bit when we normalize the accordion headings and add the hover/focus effect, but it's a great start. I've added some minor comments bellow.

Also, @replaceafill, this may affect the other theme you are working on, I'm not sure what were you using for the underline in those headings, so just double checking ;)

plugins/arDominionB5Plugin/scss/_buttons.scss
sbreker: 👍

plugins/arDominionB5Plugin/scss/_buttons.scss
sbreker: Change completed

plugins/arDominionB5Plugin/scss/_buttons.scss
sbreker: Fixed.

jraddaoui (approved):
Thanks!!

PR #1435, opened 2021-09-19
Add CSV transformation custom order, refs #13565

mcantelon:
Added the ability to define a custom sort order, for CSV row output,
using logic defined in a new property, "sortOrderLogic", in the
QubitCsvTransform and QubitCsvTransformFactory classes.

### DIFF
--- a/lib/QubitCsvTransform.class.php
+++ b/lib/QubitCsvTransform.class.php
@@ -24,6 +24,7 @@ class QubitCsvTransform extends QubitFlatfileImport
     public $rowsPerFile = 1000;
     public $preserveOrder = false;
     public $levelsOfDescription;
+    public $sortOrderCallback;
     public $convertWindowsEncoding = false;
 
     private $link;
--- a/lib/QubitCsvTransformFactory.class.php
+++ b/lib/QubitCsvTransformFactory.class.php
@@ -30,6 +30,7 @@ class QubitCsvTransformFactory
     public $setupLogic;
     public $transformLogic;
     public $preserveOrder;
+    public $sortOrderCallback;
     public $levelsOfDescription;
     public $convertWindowsEncoding;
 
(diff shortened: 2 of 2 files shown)


lib/QubitCsvTransform.class.php
```diff
@@ -24,6 +24,7 @@ class QubitCsvTransform extends QubitFlatfileImport
     public $rowsPerFile = 1000;
     public $preserveOrder = false;
     public $levelsOfDescription;
+    public $sortOrderLogic;
```
djjuhasz: Suggestion: Maybe `$sortCallback` or `$sortFunction` to make it more explicit that a callable function or method  is expected?

lib/QubitCsvTransformFactory.class.php
```diff
@@ -174,6 +178,8 @@ public function make()
                         if ($levelOfDescriptionAvailable) {
                             if (!empty($self->preserveOrder)) {
                                 $sortorder = $self->getStatus('rows');
+                            } elseif (!empty($self->sortOrderLogic)) {
+                                $sortorder = call_user_func_array($self->sortOrderLogic, [$self]);
```
djjuhasz: I think this will throw an fatal error if `$self->sortOrderLogic` isn't callable in the current scope, which may be the desired behaviour.  It could be made a bit more user friendly by testing `is_callable($self->sortOrderLogic)` and throwing a more user friendly error message, but this might be overkill if we are expecting the transformation script to be written by an experienced PHP developer.

djjuhasz (approved):
@mcantelon I added a couple of comments, but I don't think either is a blocker to committing this as is.

lib/QubitCsvTransform.class.php
mcantelon: Good thinkin'.

PR #1436, opened 2021-09-21
Fix Digital Object filter tag behaviour, refs #13547

melaniekung:
Unset hasDigitalObject/onlyMedia tag when "Digital Object Available" option is left blank in Advanced Search of Authority Records and Archival Descriptions.

### DIFF
--- a/apps/qubit/modules/actor/actions/browseAction.class.php
+++ b/apps/qubit/modules/actor/actions/browseAction.class.php
@@ -375,8 +375,10 @@ class ActorBrowseAction extends DefaultBrowseAction
         // Set label for has digital object filter tag
         if (filter_var($request->hasDigitalObject, FILTER_VALIDATE_BOOLEAN)) {
             $this->setFilterTagLabel('hasDigitalObject', $this->i18n->__('With digital objects'));
-        } else {
+        } elseif ('0' === $request->hasDigitalObject) {
             $this->setFilterTagLabel('hasDigitalObject', $this->i18n->__('Without digital objects'));
+        } else {
+            unset($request->hasDigitalObject);
         }
 
         if (!empty($request->emptyField)) {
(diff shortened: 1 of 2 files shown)


apps/qubit/modules/actor/actions/browseAction.class.php
```diff
@@ -375,8 +375,10 @@ protected function setFilterTags($request)
         // Set label for has digital object filter tag
         if (filter_var($request->hasDigitalObject, FILTER_VALIDATE_BOOLEAN)) {
             $this->setFilterTagLabel('hasDigitalObject', $this->i18n->__('With digital objects'));
-        } else {
+        } elseif (!filter_var($request->hasDigitalObject, FILTER_VALIDATE_BOOLEAN) && '0' == $request->hasDigitalObject) {
```
djjuhasz: The `!filter_var($request->hasDigitalObject, FILTER_VALIDATE_BOOLEAN) &&` should be removed from this condition. If `filter_var($request->hasDigitalObject, FILTER_VALIDATE_BOOLEAN)` is true, then this condition will not be tested because the first `if (filter_var($request->hasDigitalObject, FILTER_VALIDATE_BOOLEAN)) {` will evaluate true.

Also, the strict comparison operator (===) should be used in `'0' === $request->hasDigitalObject`

apps/qubit/modules/informationobject/actions/browseAction.class.php
```diff
@@ -503,8 +503,10 @@ protected function setFilterTags($request)
         // Set label for has digital object filter tag
         if (filter_var($request->onlyMedia, FILTER_VALIDATE_BOOLEAN)) {
             $this->setFilterTagLabel('onlyMedia', $i18n->__('With digital objects'));
-        } else {
+        } elseif (!filter_var($request->onlyMedia, FILTER_VALIDATE_BOOLEAN) && '0' == $request->onlyMedia) {
```
djjuhasz: Again, `!filter_var($request->onlyMedia, FILTER_VALIDATE_BOOLEAN)` is redundant here.

djjuhasz (changes requested):

apps/qubit/modules/informationobject/actions/browseAction.class.php
djjuhasz: And the strict comparison operator should be used for `'0' === $request->onlyMedia`

djjuhasz (approved):
Looks good. :shipit:

PR #1439, opened 2021-09-24
Fix ListIdentifiers timeframe params, refs #13568

djjuhasz:
Ensure `$this->getUpdates($options);` argument is an array to avoid
`array_merge($presetOptions, $options);` returning a null value and
ignoring all `$options` values.


jraddaoui (commented):
:ok_hand:

jraddaoui (approved):

PR #1440, opened 2021-09-29
Restart atom-worker job count, refs #xxxxx

sbreker:
Add new parameter to the `symfony jobs:worker` task to set the number of
jobs that will be run before the worker automatically exits. When set,
the worker will output log messages indicating what the job count is,
and what the max number of jobs is.

e.g. php symfony jobs:worker --max-job-count=10


vendor/net_gearman/Net/Gearman/AtomWorker.php
```diff
@@ -0,0 +1,152 @@
...
+     *
+     * Overrides net_gearman's Worker.php beginWork method. Adds check
+     * $this->maxJobCountReached() and exits work loop if true.
+     *
+     * @param callable $monitor Function to monitor work
+     *
+     * @see Net_Gearman_Connection::send(), Net_Gearman_Connection::connect()
+     * @see Net_Gearman_Worker::doWork(), Net_Gearman_Worker::addAbility()
+     */
+    public function beginWork($monitor = null)
```
djjuhasz: Rather than duplicating the `Net_Gearman_Worker::beginWork()` code, what about using the `$monitor` callback to stop the `$working` loop and kill the worker?  It would require integrating the MySQL connection keep-alive anonymous function from https://github.com/artefactual/atom/blob/077b17da608339dceedaefc68b95bc8f59e50d2c/lib/task/jobs/jobWorkerTask.class.php#L123 but I think it's still worthwhile.

vendor/net_gearman/Net/Gearman/AtomWorker.php
```diff
@@ -0,0 +1,152 @@
...
+    }
+
+    protected function incrementJobsCompleted()
+    {
+        ++$this->jobsCompleted;
+    }
+
+    protected function complete($handle, $job, array $result)
+    {
+        $this->incrementJobsCompleted();
```
djjuhasz: Is there an advantage to using a `$this->incrementJobsCompleted()` method here instead of just using `++$this->jobsCompleted` directly?

vendor/net_gearman/Net/Gearman/AtomWorker.php
```diff
@@ -0,0 +1,152 @@
...
+        }
+    }
+
+    protected function maxJobCountReached()
+    {
+        if ($this->maxJobCount <= 0) {
+            return false;
+        }
+
+        if ($this->jobsCompleted >= $this->maxJobCount) {
```
djjuhasz: This condition and the above could be combined into:
```
if ($this->maxJobCount > 0 && $this->jobsCompleted >= $this->maxJobCount) {
```

lib/task/jobs/jobWorkerTask.class.php
```diff
@@ -128,6 +140,17 @@ function () use (&$counter) {
                 }
             }
         );
+
+        $jobsCompleted = $worker->getJobsCompleted();
+
+        // Force worker's __destruct() to run.
+        $worker = null;
+
+        if ($this->maxJobCount > 0 && $this->maxJobCount <= $jobsCompleted) {
```
djjuhasz: This duplicates the logic in `AtomNetGearmanWorker::maxJobCountReached()`.  I think it would be better to use `if ($worker->maxJobCountReached()) {` and move the lines 146-147 that unset `$worker` to the end of this method, after this conditional.

lib/task/jobs/jobWorkerTask.class.php
```diff
@@ -128,6 +140,17 @@ function () use (&$counter) {
                 }
             }
         );
+
+        $jobsCompleted = $worker->getJobsCompleted();
+
+        // Force worker's __destruct() to run.
+        $worker = null;
```
djjuhasz: I prefer `unset($worker);` but I think it's entirely a stylistic choice.

djjuhasz: Huh, it looks like `unset($var)` removes `$var` from the symbol table too, which may be slightly slower, but frees slightly more memory? https://stackoverflow.com/questions/13667137/the-difference-between-unset-and-null

Probably too small a difference to really matter for either speed or memory utilization.

djjuhasz (changes requested):
Looks good overall @sbreker, I just had a few suggestions.

vendor/net_gearman/Net/Gearman/AtomWorker.php
sbreker: This would still require overriding beginWork() as currently the monitor callback only has 2 vars available to it: an idle flag and the last job time available to it. The vendor Worker.php file would still need to be modified here to pass in `$this` in addition to the other two vars:

https://github.com/artefactual/atom/blob/qa/2.x/vendor/net_gearman/Net/Gearman/Worker.php#L298

....then in the callback I could call `maxJobCountReached()` and return false exiting the loop.

I think your suggestion makes a lot of sense, but it will still require overriding beginWork() I think.

vendor/net_gearman/Net/Gearman/AtomWorker.php
sbreker: Just personal pref... Similar to using getters and setters within an object, it provides a unified interface for performing this function. Granted, incrementing an int is pretty simple and probably not necessary!  😄

lib/task/jobs/jobWorkerTask.class.php
sbreker: Hmm... I had thought they were pretty equivalent, but it turns out there are differences.  Good idea - I will update to use unset().

vendor/net_gearman/Net/Gearman/AtomWorker.php
sbreker: I agree.

vendor/net_gearman/Net/Gearman/AtomWorker.php
djjuhasz: My understanding, based on https://www.php.net/manual/en/functions.anonymous.php#example-162 (Example 5: Automatic binding of $this), is `$this` is automatically available within a closure when the closure is defined within the class context.  I think `function () use ($var1, $var2)` can also be used to pass extra variables into the closure scope.

djjuhasz: And I don't object to overriding `Net_Gearman_Worker::beginWork()`, my objection is with copying and pasting all the code from `Net_Gearman_Worker::beginWork()` into `AtomWorker::beginWork()`. 😄

sbreker: This is extremely useful feedback - thanks @djjuhasz.  Expect another commit soon...  😄

lib/task/jobs/jobWorkerTask.class.php
```diff
@@ -105,6 +142,14 @@ function ($handle, $job, $e) {
             Net_Gearman_Worker::JOB_FAIL
         );
 
+        $worker->attachCallback(
```
djjuhasz, line 145: This is cool 😎

lib/task/jobs/jobWorkerTask.class.php
```diff
@@ -120,14 +165,33 @@ function ($handle, $job, $e) {
...
-            function () use (&$counter) {
+            function ($idle, $lastJob) use (&$counter) {
+                if ($this->maxJobCountReached()) {
+                    $this->log(sprintf('Max job count reached: %u jobs completed.', $this->getJobsCompleted()), sfLogger::INFO);
+
+                    // Notify the worker that beginWork() work loop should exit.
+                    return true;
+                }
+
                 if (30 == $counter++) {
```
djjuhasz, line 177: I think comments about the MySQL keep-alive (lines 162-167) should be moved to just before this conditional.

lib/task/jobs/jobWorkerTask.class.php
```diff
@@ -120,14 +165,33 @@ function ($handle, $job, $e) {
...
                     $counter = 0;
 
                     QubitPdo::prepareAndExecute('SELECT 1');
                 }
             }
         );
+
+        // Return code '111' when worker shuts down due to reaching max job count.
+        if ($this->getMaxJobCount() > 0 && $this->maxJobCountReached()) {
```
djjuhasz: `$this->getMaxJobCount() > 0` is redundant here.

lib/task/jobs/jobWorkerTask.class.php
```diff
@@ -120,14 +165,33 @@ function ($handle, $job, $e) {
...
+            $this->log('Worker shutting down - max-job-count reached.');
+            // Force worker's __destruct() to run so gearman connection is closed nicely.
+            unset($worker);
```
djjuhasz, line 189: I don't think there's any advantage to `unset($worker);` before `exit()` - all the memory should be freed when the script is killed, AFAIK.

lib/task/jobs/jobWorkerTask.class.php
```diff
@@ -120,14 +165,33 @@ function ($handle, $job, $e) {
...
+
+            exit(self::JOB_LIMIT_RETURN_STATUS);
+        }
+
+        // Force worker's __destruct() to run so gearman connection is closed nicely.
+        unset($worker);
```
djjuhasz, line 195: 👍

djjuhasz (changes requested):
@sbreker I love the new callback implementation for the $jobsComplete counter! :)  I suggested a few minor changes.

lib/task/jobs/jobWorkerTask.class.php
sbreker: The same comment applies here as with the second call to `unset($worker)` below. This forces the worker's destructor to run which triggers a clean disconnect from Gearman.

https://github.com/artefactual/atom/blob/qa/2.x/vendor/net_gearman/Net/Gearman/Worker.php#L456-L472

Without this line, debug statements indicate that the worker's destructor is not run on exit() - I'd rather have the worker disconnect 'nicely' from gearman to avoid issues with open ports etc.

lib/task/jobs/jobWorkerTask.class.php
sbreker: 👍  I've moved most of this comment to line 175.

lib/task/jobs/jobWorkerTask.class.php
sbreker: 👍

lib/task/jobs/jobWorkerTask.class.php
djjuhasz: Okay, this makes sense to me. 👍

lib/task/jobs/jobWorkerTask.class.php
```diff
@@ -159,20 +159,21 @@ function ($handle, $job, $e) {
...
+            // job is not being processed.
             function ($idle, $lastJob) use (&$counter) {
                 if ($this->maxJobCountReached()) {
                     $this->log(sprintf('Max job count reached: %u jobs completed.', $this->getJobsCompleted()), sfLogger::INFO);
 
                     // Notify the worker that beginWork() work loop should exit.
                     return true;
                 }
-
+                // Ping the database every ~30 seconds in order to keep the
```
djjuhasz, line 171: Nitpick: blank line after closing curly bracket.

djjuhasz (approved):
Looks good @sbreker! Just one more minor nitpick. ;)

PR #1442, opened 2021-09-30
Fix CLI SKOS import issue, refs #13571

mcantelon:
Turns relative paths of import files into absolute paths so the import of SKOS files doesn't fail.


djjuhasz (approved):
Looks good @mcantelon 👍

mcantelon:
Thanks @djjuhasz !

PR #1444, opened 2021-10-01
Remove get_magic_quotes_gpc() calls, refs #13462

djjuhasz:
The get_magic_quotes_gpc() function is deprecated as of PHP 7.4.0 and
removed as of PHP 8.0.0.  As the function always returns false in PHP
7.4.0+ replace it with a false value in all conditionals.

### DIFF
--- a/vendor/htmlpurifier/library/HTMLPurifier/Config.php
+++ b/vendor/htmlpurifier/library/HTMLPurifier/Config.php
@@ -751,14 +751,13 @@ class HTMLPurifier_Config
      * @param array $array $_GET or $_POST array to import
      * @param string|bool $index Index/name that the config variables are in
      * @param array|bool $allowed List of allowed namespaces/directives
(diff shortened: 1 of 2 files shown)


sbreker:
Looks great @djjuhasz! This looks like it should clear up those deprecation warnings.

I did create a ticket specifically for this issue: 

https://projects.artefactual.com/issues/13531

Wondering if you explored updating the HTML Purifier via composer?

djjuhasz:
@sbreker thanks for the link to https://projects.artefactual.com/issues/13531, I'll update the reference in the commit message.

I didn't explore updating HTML Purifier via composer, but it's a good idea.  It looks like the most recent stable release of HTML Purifier is 4.13.0, and AtoM is currently running 4.6.0, so hopefully there aren't any breaking changes. 🤞

djjuhasz:
I've replaced this PR with https://github.com/artefactual/atom/pull/1445

PR #1445, opened 2021-10-04
Remove get_magic_quotes_gpc() calls, refs #13531

djjuhasz:
The get_magic_quotes_gpc() function is deprecated as of PHP 7.4.0 and
removed as of PHP 8.0.0. As the function always returns false in PHP
7.4.0+ replace it with a false value in all conditionals.

* Upgrade to HTMLPurifier 4.13 to remove get_magic_quotes_gpc() calls
* Install HTMLPurifier via Composer
* Remove get_magic_quotes_gpc() from 
  /vendor/symfony/lib/request/sfWebRequest.class.php


djjuhasz:
@sbreker I closed https://github.com/artefactual/atom/pull/1444, and opened this PR against issue #13531 and switched to loading HTMLPurifier with Composer.

sbreker (approved):
This looks great @djjuhasz - Thanks for looking into that HTML Purifier/composer update!  👍

PR #1446, opened 2021-10-06
Restart atom-worker job count, refs #13573

sbreker:
Add new parameter to the `symfony jobs:worker` task to set the number of
jobs that will be run before the worker automatically exits. When set,
the worker will output log messages indicating what the job count is,
and what the max number of jobs is.

e.g. php symfony jobs:worker --max-job-count=10

If the worker task hits it's execution limit, it will return a code of
'111' which systemd can use to identify the reason the worker exited.

### DIFF
--- a/lib/task/jobs/jobWorkerTask.class.php
+++ b/lib/task/jobs/jobWorkerTask.class.php
@@ -22,6 +22,10 @@
  */
 class jobWorkerTask extends arBaseTask
 {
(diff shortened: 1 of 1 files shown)


sbreker:
@djjuhasz - Closing PR [#1440](https://github.com/artefactual/atom/pull/1440) and opening this PR to replace it. This was to correct the branch name but the PR is otherwise identical.

djjuhasz (approved):
Looks good @sbreker :shipit:

PR #1449, opened 2021-10-08
Fix OAI-PMH resumptionToken dates, refs #13574

djjuhasz:
This also fixes issue #13025 by correctly converting OAI-PMH UTC "from"
and "until" dates to local MySQL dates for ORM queries.

- Use ISO 8601 UTC format to represent from & until dates internally and
  in OAI-PMH output (e.g. "2021-10-08T21:53:17Z")
- Pass datetimes to MySQL in MySQL date format using local app time
  (e.g. "2021-10-08 14:53:17" for default_timezone="America/Vancouver")

### DIFF
--- a/lib/oai/QubitOai.class.php
+++ b/lib/oai/QubitOai.class.php
@@ -147,7 +147,7 @@ class QubitOai
     }
 
     /**
-     * Returns formated date.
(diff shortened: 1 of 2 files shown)


mcantelon (approved):
Looks good to me!

PR #1451, opened 2021-10-12
Fix actor CSV import error, refs #13576

mcantelon:
Fixes error, during authority record CSV import, that occurs when the description identifier column isn't populated.


djjuhasz (approved):
Looks good @mcantelon 👍

PR #1452, opened 2021-10-13
Memory profiler atom-worker, refs #13575

sbreker:
Add class for interacting with arnaud-lb/php-memory-profiler. Set up for
use in the atom-worker.

https://github.com/arnaud-lb/php-memory-profiler.

1) install arnaud-lb/php-memory-profiler
2) configure atom-worker.service to include
"Environment=MEMPROF_PROFILE=native"
3) OPTIONAL: set the basepath and name using env var:
"Environment=MEMPROF_OUTPUT_BASENAME=/vagrant/atom_memprof_file.grind"
4) run: sudo systemctl daemon-reload
5) restart the atom-worker

If MEMPROF_OUTPUT_BASENAME is not set, the grind files will be output
to the AtoM folder (atom_memprof_file.grind.<timestamp>).

When enabled, a memory profiler grind file will be output after each
job completes, and when the atom-worker process is shut down.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -36,6 +36,8 @@ RUN set -xe \
       libxslt \
       libmemcached-libs \
       libzip \
+    && apk add --update-cache --repository http://dl-3.alpinelinux.org/alpine/edge/testing/ --allow-untrusted \
+      php8-pecl-memprof \
     && apk del .phpext-builddeps \
     && pecl clear-cache \
     && apk add --no-cache --virtual .atom-deps \
--- a/lib/task/jobs/jobWorkerTask.class.php
+++ b/lib/task/jobs/jobWorkerTask.class.php
@@ -25,6 +25,7 @@ class jobWorkerTask extends arBaseTask
     public const JOB_LIMIT_RETURN_STATUS = 111;
     private $maxJobCount = 0;
     private $jobsCompleted = 0;
+    private $memoryProfiler;
 
     public function gearmanWorkerLogger(sfEvent $event)
     {
@@ -65,6 +66,39 @@ class jobWorkerTask extends arBaseTask
     public function getJobsCompleted()
     {
         return $this->jobsCompleted;
+    }
+
+    public function getMemoryUsageString()
+    {
+        return sprintf(
+            "Worker memory usage (PHP - VmRSS): %skB - %dkB\n",
+            $this->getPhpReportedMemoryUsage(),
+            $this->getLinuxReportedMemoryUsage()
+        );
+    }
+
+    /* Get OS reported mem usage from /prod/self/status.
+     * Grabs the memory value from the VmRSS entry in the status file.
+     * VmRSS is the resident mem set size in kB.
+     *
+     * Returns the resident memory in kB. If the lookup fails this function will
+     * return zero and the caller should deal with that as an error.
+     */
+    public function getLinuxReportedMemoryUsage()
+    {
+        try {
+            preg_match('/^VmRSS:\s(.*)/m', file_get_contents('/proc/self/status'), $matches);
+            $memUsage = (int) trim($matches[1]);
+        } catch (Exception $e) {
+            return 0;
+        }
+
+        return $memUsage;
+    }
+
+    public function getPhpReportedMemoryUsage()
+    {
+        return sprintf('%.0f', memory_get_usage(true) / 1024);
     }
 
     protected function configure()
@@ -123,6 +157,10 @@ EOF;
 
         $worker = new Net_Gearman_Worker($servers);
 
+        if (arPhpMemoryProfiler::getMemprofEnabled()) {
+            $this->memoryProfiler = new arPhpMemoryProfiler();
+        }
+
         // Register abilities (jobs)
         foreach ($abilities as $ability) {
             if (!class_exists($ability)) {
@@ -146,12 +184,22 @@ EOF;
             function ($handle, $job, $e) {
                 ++$this->jobsCompleted;
                 $this->log(sprintf('Jobs completed: %u', $this->getJobsCompleted()));
+
+                if (arPhpMemoryProfiler::getMemprofEnabled()) {
+                    $this->log(self::getMemoryUsageString());
+                    $this->log(sprintf('Memprof enabled. Dumping grind file: %s', $this->memoryProfiler->createMemprofGrindFile()));
+                }
             },
             Net_Gearman_Worker::JOB_COMPLETE
         );
 
         $this->log('Running worker...');
         $this->log('PID '.getmypid());
+
+        if (arPhpMemoryProfiler::getMemprofEnabled()) {
+            $this->log(sprintf('Memprof profile: %s', $this->memoryProfiler->getMemprofProfile()));
+            $this->log(self::getMemoryUsageString());
+        }
 
         $this->activateTerminationHandlers();
 
@@ -223,6 +271,10 @@ EOF;
         // Define shutdown function
         register_shutdown_function(function () {
             $this->log('Job worker stopped.');
+
+            if (arPhpMemoryProfiler::getMemprofEnabled()) {
+                $this->log(sprintf('Memprof enabled. Dumping shutdown grind file: %s', $this->memoryProfiler->createMemprofGrindFile()));
+            }
         });
     }
 }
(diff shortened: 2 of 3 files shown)


lib/arMemprofUtils.class.php
```diff
@@ -0,0 +1,73 @@
...
+ * 5) restart the atom-worker
+ *
+ * If MEMPROF_OUTPUT_BASENAME is not set, the grind files will be output
+ * to the AtoM folder (atom_memprof_file.grind.<timestamp>)
+ */
+class arMemprofUtils
+{
+    public static function createMemprofGrindFile(string $basename = 'atom_memprof_file.grind')
+    {
+        if (self::getMemprofEnabled()) {
```
djjuhasz: Opinion: I'd prefer to return if `!self::getMemprofEnabled()` to nesting all the following code inside a conditional.

lib/task/jobs/jobWorkerTask.class.php
```diff
@@ -67,6 +67,39 @@ public function getJobsCompleted()
         return $this->jobsCompleted;
     }
 
+    public static function getMemoryUsageString()
```
djjuhasz: I think this is better as a regular (non-static) method so it can be overridden, unit tested, etc.

lib/task/jobs/jobWorkerTask.class.php
```diff
@@ -67,6 +67,39 @@ public function getJobsCompleted()
...
+    {
+        return sprintf(
+            "Worker memory usage (PHP - VmRSS): %.2fMb - %dkb\n",
```
djjuhasz: I think it would be better to standardize the units here, rather than mixing Mb and kb.

lib/task/jobs/jobWorkerTask.class.php
```diff
@@ -67,6 +67,39 @@ public function getJobsCompleted()
...
+    }
+
+    /* Get OS reported mem usage from /prod/self/status.
+     * Grabs the memory value from the VmRSS entry in the status file.
+     * VmRSS is the resident mem set size in kB.
+     *
+     * Returns the resident memory in kB. If the lookup fails this function will
+     * return zero and the caller should deal with that as an error.
+     */
+    public static function getLinuxReportedMemoryUsage()
```
djjuhasz: Again, I think it's better to use a standard (non-static) method unless there's a specific reason to make it static.  I won't keep reposting the same comment, but I think this applies to the other static methods in the PR as well.

lib/task/jobs/jobWorkerTask.class.php
```diff
@@ -146,13 +179,24 @@ function ($handle, $job, $e) {
             function ($handle, $job, $e) {
                 ++$this->jobsCompleted;
                 $this->log(sprintf('Jobs completed: %u', $this->getJobsCompleted()));
+
+                if (arMemprofUtils::getMemprofEnabled()) {
```
djjuhasz: I think it's better to instantiate an arMemprof class instead of using static methods.  Using standard classes allows the possibility of defining a standard interface and extending it for specific profiling tools, as well as allowing unit testing.  Using a class instance also allows injecting the memory profiler class, which would make it easier to swap in a different profiler.

lib/task/jobs/jobWorkerTask.class.php
```diff
@@ -223,6 +267,11 @@ protected function activateTerminationHandlers()
         // Define shutdown function
         register_shutdown_function(function () {
             $this->log('Job worker stopped.');
+
+            if (arMemprofUtils::getMemprofEnabled()) {
+                $memprof_filename = arMemprofUtils::createMemprofGrindFile();
+                $this->log(sprintf('Memprof enabled. Dumping shutdown grind file: %s', $memprof_filename));
```
djjuhasz: Nitpick: you could call arMemprofUtils::createMemprofGrindFile() directly here, instead of creating a single use variable.

lib/arMemprofUtils.class.php
```diff
@@ -0,0 +1,73 @@
...
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with Access to Memory (AtoM).  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+/**
+ * Utilities for interacting with arnaud-lb/php-memory-profiler
+ * https://github.com/arnaud-lb/php-memory-profiler.
+ *
+ * 1) install arnaud-lb/php-memory-profiler
```
djjuhasz: Good documentation on how to get the profiler working 👍

djjuhasz (changes requested):
@sbreker I requested a few changes in the comments.

lib/task/jobs/jobWorkerTask.class.php
sbreker: Good idea - done

lib/arMemprofUtils.class.php
sbreker: Thanks!

lib/arMemprofUtils.class.php
sbreker: Done

lib/task/jobs/jobWorkerTask.class.php
sbreker: We chatted about this - I will convert these jobWorkerTask statics to standard class methods. There is one in the profiler class that it does not make sense to make non-static.

lib/task/jobs/jobWorkerTask.class.php
sbreker: See above.

lib/task/jobs/jobWorkerTask.class.php
sbreker: See above.

lib/task/jobs/jobWorkerTask.class.php
sbreker: 👍

Dockerfile
```diff
@@ -36,6 +36,8 @@ RUN set -xe \
       libxslt \
       libmemcached-libs \
       libzip \
+    && apk add --update-cache --repository http://dl-3.alpinelinux.org/alpine/edge/testing/ --allow-untrusted \
```
djjuhasz, line 39: Glad to see there is an alpine package for memprof. :)

lib/task/jobs/jobWorkerTask.class.php
```diff
@@ -67,6 +68,39 @@ public function getJobsCompleted()
         return $this->jobsCompleted;
     }
 
+    public function getMemoryUsageString()
+    {
+        return sprintf(
+            "Worker memory usage (PHP - VmRSS): %.2fkb - %dkb\n",
```
djjuhasz: I think decimal kb for the PHP reported memory usage is probably overkill here.

lib/task/jobs/jobWorkerTask.class.php
```diff
@@ -67,6 +68,39 @@ public function getJobsCompleted()
...
+            $this->getPhpReportedMemoryUsage(),
```
djjuhasz, line 75: `getPhpReportedMemoryUsage()` returns a formatted string, but `sprintf` is expecting a float.

lib/task/jobs/jobWorkerTask.class.php
```diff
@@ -123,6 +157,10 @@ protected function execute($arguments = [], $options = [])
 
         $worker = new Net_Gearman_Worker($servers);
 
+        if (arPhpMemoryProfiler::getMemprofEnabled()) {
+            $this->memory_profiler = new arPhpMemoryProfiler();
```
djjuhasz: Nitpick: camel case for variable names

djjuhasz (changes requested):
@sbreker I've requested a few more minor changes

lib/task/jobs/jobWorkerTask.class.php
sbreker: 🤦

lib/task/jobs/jobWorkerTask.class.php
sbreker: Fixed.

lib/task/jobs/jobWorkerTask.class.php
sbreker: Agreed.

Dockerfile
sbreker: Thanks to @sevein for pointing out this apk package

djjuhasz (approved):
Looks good @sbreker 👍

PR #1453, opened 2021-10-18
AtoM worker max-mem-usage opt, refs #13573

sbreker:
Similar to the '--max-job-count' option, this new option
'--max-mem-usage' will cause the atom-worker process to exit with return
code '111' when the atom-worker's allocated memory exceeds set value.

- max-mem-usage is expected to be in kB.
- max-mem-usage value will be compared to the atom-worker's resident
memory as reported by the operating system.

### DIFF
--- a/lib/task/jobs/jobWorkerTask.class.php
+++ b/lib/task/jobs/jobWorkerTask.class.php
@@ -22,10 +22,11 @@
  */
 class jobWorkerTask extends arBaseTask
 {
-    public const JOB_LIMIT_RETURN_STATUS = 111;
+    public const LIMIT_RETURN_STATUS = 111;
     private $maxJobCount = 0;
     private $jobsCompleted = 0;
     private $memoryProfiler;
+    private $maxMemUsage = 0;
 
     public function gearmanWorkerLogger(sfEvent $event)
     {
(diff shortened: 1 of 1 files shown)


lib/task/jobs/jobWorkerTask.class.php
```diff
@@ -101,6 +102,27 @@ public function getPhpReportedMemoryUsage()
...
     }
 
+    public function getMaxMemUsage()
+    {
+        return $this->maxMemUsage;
+    }
+
+    public function setMaxMemUsage(int $maxMemUsage)
+    {
+        if ($maxMemUsage > 0) {
```
mcantelon, line 112: Nitpick: this defaults to 0 so maybe it should be settable to 0? Or maybe it should throw an error, rather than fail silently, if it's given the value 0?

lib/task/jobs/jobWorkerTask.class.php
```diff
@@ -217,6 +246,13 @@ function ($idle, $lastJob) use (&$counter) {
                     return true;
                 }
 
+                if ($this->maxMemUsageReached()) {
+                    $this->log(sprintf('Max memory usage reached (%ukB): %ukB in use.', $this->getMaxMemUsage(), $this->getLinuxReportedMemoryUsage()), sfLogger::INFO);
```
mcantelon: Nitpick: this line's a bit long.

mcantelon (approved):
Looks good, other than a couple nitpicks!

lib/task/jobs/jobWorkerTask.class.php
sbreker: Updated 👍

lib/task/jobs/jobWorkerTask.class.php
sbreker: I think I'm going to leave this - the default value of 0 equates to the setting being not in use. Allowing it to be set to zero has no benefit. If the value set is greater than zero, a message will be displayed in the log on atom-worker startup indicating 'Worker will shut down if memory consumption exceeds %ukB.' - this message will be absent if the setting is set to a nonsense value: 0 or negative. This mirrors what was done for `--max-jobs-count`.  In most cases this will be set once in `/usr/lib/systemd/system/atom-worker.service` and left.

PR #1454, opened 2021-10-19
Treeview move index issue, refs #13567

sbreker:
This commit fixes an issue where full width treeview move no longer
functions due to the ES index (which populates the treeview) lft values
becoming out of sync with the database lft values when nodes are moved
via the UI. This action was not triggering re-index of affected nodes
resulting in treeview errors: 'Mismatch in current position' when the
database lft ordering did not match the ES index lft ordering. When this
occurs the only fix is to run a search:populate.

The range of treeview node objects that a move action encompasses need
to be re-indexed after a treeview move event. Since reindexing the
entire parent node could involve a large number of records, this fix
only reindexes those objects affected by the move.

The performance of this fix is directly related to the number of nodes
affected by the move event - the 'further' the move, the more objects
will need reindexing, and the longer the move job will take. If a user
refreshes the treeview while the move job is re-indexing it is possible
for the UI ordering to appear out of sync temporarily with the DB.

### DIFF
--- a/lib/job/arObjectMoveJob.class.php
+++ b/lib/job/arObjectMoveJob.class.php
@@ -97,12 +97,14 @@ class arObjectMoveJob extends arBaseJob
                 case 'before':
                     $this->info($this->i18n->__('Moving object before sibling (id: %1)', ['%1' => $targetSiblingId]));
                     $object->moveToPrevSiblingOf($targetSibling);
+                    $this->reindexObjectRange($children, $parameters['newPosition'], $parameters['oldPosition']);
 
                     break;
 
                 case 'after':
                     $this->info($this->i18n->__('Moving object after sibling (id: %1)', ['%1' => $targetSiblingId]));
                     $object->moveToNextSiblingOf($targetSibling);
+                    $this->reindexObjectRange($children, $parameters['oldPosition'], $parameters['newPosition']);
 
                     break;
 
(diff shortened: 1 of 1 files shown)


lib/job/arObjectMoveJob.class.php
```diff
@@ -120,4 +122,13 @@ public function runJob($parameters)
 
         return true;
     }
+
+    public function reindexObjectRange(array $children, int $start, int $end)
+    {
+        for ($i = $start; $i <= $end; ++$i) {
+            $objectId = $children[$i]['id'];
+            $this->info(sprintf('Reindexing object id: %d', $objectId));
+            QubitSearch::getInstance()->update(QubitObject::getById($objectId));
```
djjuhasz: If `QubitObject::getById($objectId)` returns `null` because the row has been deleted from the database, this will throw an error.

lib/job/arObjectMoveJob.class.php
```diff
@@ -97,12 +97,14 @@ public function runJob($parameters)
                 case 'before':
                     $this->info($this->i18n->__('Moving object before sibling (id: %1)', ['%1' => $targetSiblingId]));
                     $object->moveToPrevSiblingOf($targetSibling);
+                    $this->reindexObjectRange($children, $parameters['newPosition'], $parameters['oldPosition'], $children);
```
djjuhasz: You've got an extra argument here. 😉

djjuhasz (changes requested):
Hi @sbreker, couple of minor change requests.

lib/job/arObjectMoveJob.class.php
sbreker: 👍

lib/job/arObjectMoveJob.class.php
sbreker: Missed it during cleanup... Fixed!

lib/job/arObjectMoveJob.class.php
```diff
@@ -120,4 +122,16 @@ public function runJob($parameters)
 
         return true;
     }
+
+    public function reindexObjectRange(array $children, int $start, int $end)
+    {
+        for ($i = $start; $i <= $end; ++$i) {
+            $objectId = $children[$i]['id'];
+            $this->info($this->i18n->__('Reindexing object id: %1', ['%1' => $objectId]));
```
djjuhasz: I think it's better to move this into the conditional, so it only logs for ids that actually get reindexed.

djjuhasz (changes requested):
@sbreker one more change request.

lib/job/arObjectMoveJob.class.php
sbreker: Makes sense - Fixed.

djjuhasz (approved):
👍

jraddaoui (approved):
Neat!

PR #1455, opened 2021-10-21
Add access_log to tools:expire-data, refs #13578

djjuhasz:
- Add "access_log" type to `tools:expire-data` task
- Add "default_max_age" and "method_name" to type specification array
- Wrap long lines

### DIFF
--- a/lib/model/QubitAccessLog.php
+++ b/lib/model/QubitAccessLog.php
@@ -48,4 +48,22 @@ class QubitAccessLog extends BaseAccessLog
 
         return $stmt->fetchAll();
     }
+
+    /**
+     * Delete old access_log records.
+     *
+     * @param string $expiryDate "YYYY-MM-DD" date format; rows with an
+     *                           access_date before this will be deleted
+     *
+     * @return int number of rows deleted
+     */
+    public static function expire(string $expiryDate): int
+    {
+        $sql = sprintf(
+            'DELETE FROM %s WHERE access_date < ?;',
+            self::TABLE_NAME
+        );
+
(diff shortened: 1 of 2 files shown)


lib/model/QubitAccessLog.php
```diff
@@ -48,4 +48,19 @@ public static function getPopularThisWeek(array $options = [])
 
         return $stmt->fetchAll();
     }
+
+    /**
+     * Delete old access_log records.
+     *
+     * @return int number of rows deleted
+     */
+    public static function expire(string $expiryDate): int
```
mcantelon, line 60: Good thinking moving expiry logic to the data type's class!

lib/model/QubitAccessLog.php
```diff
@@ -48,4 +48,19 @@ public static function getPopularThisWeek(array $options = [])
 
         return $stmt->fetchAll();
     }
+
+    /**
+     * Delete old access_log records.
+     *
```
mcantelon, line 54: Nitpick: omment doesn't document the `$expiryData` argument using `@param`.

mcantelon (approved):
One nitpick, but otherwise 💯 !

lib/model/QubitAccessLog.php
djjuhasz: Good call.  I've added a comment for the `$expiryDate` parameter.

lib/model/QubitAccessLog.php
djjuhasz: Thanks! :)

PR #1457, opened 2021-10-22
Calc expiry dates separately by type, refs #13578

djjuhasz:
Calculate expiry dates separately for each resource type when --older-than
option is not used.


sbreker:
👍  Looks great @djjuhasz. A note regarding this new option should perhaps be added to the expiry CLI task docs section:

https://www.accesstomemory.org/en/docs/2.7/admin-manual/maintenance/cli-tools/#delete-temporary-data-saved-clipboards-and-downloads

sbreker (approved):
👍  Looks great @djjuhasz!

PR #1458, opened 2021-10-22
Term notes i18n issues, refs #13563

sbreker:
This commit fixes two issues affecting term notes i18n strings:

1) Translated scope notes will no longer be hidden when browsing a term.

2) Scope notes, source notes and display notes now use the
notesComponent class. Notes will no longer be deleted when a term is
edited in another language if a translation is not supplied.


apps/qubit/modules/object/actions/notesComponent.class.php
```diff
@@ -210,7 +237,7 @@ protected function addField($name)
     {
         switch ($name) {
             case 'content':
-                $this->form->setValidator('content', new sfValidatorString());
+                $this->form->setValidator('content', new sfValidatorString(['max_length' => 65535]));
```
djjuhasz, line 239: 👍

djjuhasz (approved):
Code looks good to me @sbreker 👍 

Please change the commit message to use the [imperative mood](https://chris.beams.io/posts/git-commit/#imperative).

sbreker:
Commit message updated @djjuhasz.  Thanks for the review! 👍

PR #1460, opened 2021-10-25
Embedded form constructor params, Refs #13450

sbreker:
Embedded sfForm objects must be created with params "[], [], false" to
avoid inadvertently creating a new CSRF token. Default is "[], [], null"
which sets a CSRF token - when this occurs an exception is trigggered
preventing the field from being saved. Subsequent instantiations of
sfForm within the same form succeed because the sfForm::bind() method
removes the static token. If the constuctor is set with these params,
calling disableCSRFProtection() is not required.


jraddaoui (approved):
Thanks @sbreker!

If you could tweak the commit message to use [imperative mood](https://chris.beams.io/posts/git-commit/#imperative) and not capitalize the "refs", that would be great.

sbreker:
Thanks @jraddaoui - I have updated the commit message.

PR #1463, opened 2021-10-28
Stop fwtv move when not auth, refs # 12628

sbreker:
Disable fullwidth treeview drag and drop when user is unauthenticated.


js/fullWidthTreeView.js
```diff
@@ -85,7 +86,9 @@
         'open_timeout': 0,
         'drag_selection': false,
         'is_draggable': function (nodes) {
-          return nodes[0].parent !== '#';
+          return dndEnabled ?
```
jraddaoui: `return dndEnabled && nodes[0].parent !== '#';`

jraddaoui (approved):
Just one minor suggestion ;)

js/fullWidthTreeView.js
sbreker: Nice! Updated... 👍

PR #1464, opened 2021-10-29
Fix term edit i18n note fields, refs #13563

sbreker:
Do not hide Translated scope notes when browsing a term.

Use the information object notes component to display notes fields on
the terms page. Using this component avoids issues with i18n
translations where notes were being deleted.


jraddaoui (approved):
Neat, thanks @sbreker! I'll merge these PRs and rebase qa/2.x.

PR #1466, opened 2021-10-29
Fix fwtv move on touchscreen, refs #12628

sbreker:
Fix touchscreen on fullwidth treeview move by adding the 'dnd' touch
option 'selected'.

Previous to this fix, touching the treeview always resulted in a
treeview move event for authenticated users. Adding this option makes it
so only selected nodes can be dragged on touch devices otherwise touch
will scroll the treeview.


jraddaoui (approved):
Thanks @sbreker!

PR #1468, opened 2021-11-02
Fix digitalObjectURI export col, refs #13572

sbreker:
Remote digital objects linked via URL will not have the system Base URL
prepended on export.


jraddaoui (approved):
:ok_hand:

PR #1469, opened 2021-11-04
Fix fwtv move and button issues

sbreker:
Fix fwtv dnd move by changing the 'is_draggable' callback to correctly
use arrow function.

Fix the 'More' and 'Reset' buttons so they will now appear by removing
B5 button class names from the id section of the button definitions in
the _treewView.php template.

Wrap 'More' and 'Reset' in a div so they could be positioned and styled.


jraddaoui (approved):
Nice, thanks @sbreker!

PR #1471, opened 2021-11-04
Fix terms useFor display, refs #13580

sbreker:
Do not hide translated 'use for' names when browsing a term.


djjuhasz (approved):
Looks good @sbreker.  I don't understand why the original conditional + continue block was there, but taking it out seems like the right move.

PR #1472, opened 2021-11-05
Remove unnecessary data from ES, refs #13386

sbreker:
Creators:
- Keep authorized form of name, parallel names, other forms of name.
- Keep admin/bio history
- Remove everything else

Place, subject, genre terms:
- Keep term name.
- Remove everything else.

Removed all physical storage fields from information object ES index document.

Removed all 'Part of' details from the information object ES index document.


jraddaoui (approved):
:ok_hand:

PR #1474, opened 2021-11-15
Fix multi-page file uploads, refs #13526

tw4l:
This commit adds the `name` property while creating `QubitDigitalObjects` for multi-part digital object uploads, avoiding a 500 error, and revises outdated comments in the upload code.


jraddaoui (approved):
Nice, thanks @tw4l! I think this may be related to the STRICT_TRANS_TABLES sql_mode, so I've related the tickets in Redmine.

PR #1475, opened 2021-11-19
Trigger data load from autocomplete, refs #13584

djjuhasz:
Fire the "itemSelect" event from autocomplete.js when an suggested item
is selected to notify the dialog.js listener to load related resource
data.

Also: change id of YUI dialog.js wrapper to avoid a name collision with
the original <div> id.


sbreker (approved):
Looks great @djjuhasz! 👍

djjuhasz:
There's still a bug that prevents the donor data from being cleared when the donor name is cleared.

PR #1476, opened 2021-11-22
Dev/issue 13584 fix donor info load

djjuhasz:
Fire the "itemSelect" event from autocomplete.js when an suggested item
is selected to notify the dialog.js listener to load related resource
data.

Also: 
- change id of YUI dialog.js wrapper to avoid a name collision with
  the original <div> id
- Move JavaScript code from _relatedDonor.php template file to it's own
  relatedDonor.js file to make it easier to read and debug and to allow
  automatic code formatting
- When a pre-existing donor is selected in the relatedDonor dialog, their
  contact information is loaded.  If the donor is then changed, clear
  their contact information from the dialog

### DIFF
--- a/js/autocomplete.js
+++ b/js/autocomplete.js
@@ -509,6 +509,12 @@
               // On single <select/> item select, simply update the
               // value of this input
               $hidden.val(data[1]).trigger('change');
+
+              // Trigger event to load item data if it's needed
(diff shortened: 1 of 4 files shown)


mcantelon (approved):
Looks good to me!

PR #1478, opened 2021-12-02
CSV Terms Import Fix, refs# 13287

sbreker:
Repository CSV import will no longer trigger errors when the
'descriptionStatus' or 'levelOfDetail' columns are blank in the CSV.

### DIFF
--- a/lib/QubitFlatfileImport.class.php
+++ b/lib/QubitFlatfileImport.class.php
@@ -1104,7 +1104,7 @@ class QubitFlatfileImport
             if (null !== $key = QubitFlatfileImport::getTermIndex($rows, $name)) {
                 $terms[] = QubitTerm::getById($rows[$key]['id']);
             } elseif (!isset($termsCreated) || !in_array($name, $termsCreated)) {
-                $terms[] = QubitFlatfileImport::createTerm($taxonomyId, $name, $culture);
+                $terms[] = QubitTerm::createTerm($taxonomyId, $name, $culture);
                 // Don't create duplicates
                 $termsCreated[] = $name;
             }
@@ -1115,27 +1115,6 @@ class QubitFlatfileImport
         }
 
         return $terms;
-    }
-
-    /**
-     * Create a Qubit term.
-     *
-     * @param int    $taxonomyId term taxonomy
-     * @param string $name       name of term
-     * @param string $culture    culture code (defaulting to English)
-     *
-     * @return QubitTerm created term
-     */
-    public static function createTerm($taxonomyId, $name, $culture = 'en')
-    {
-        $term = new QubitTerm();
-        $term->name = $name;
-        $term->taxonomyId = $taxonomyId;
-        $term->parentId = QubitTerm::ROOT_ID;
-        $term->culture = $culture;
-        $term->save();
-
-        return $term;
     }
 
     /**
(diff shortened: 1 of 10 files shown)


lib/task/import/csvRepositoryImportTask.class.php
```diff
@@ -176,20 +176,30 @@ public function execute($arguments = [], $options = [])
                 }
 
                 // Handle description status
-                $self->object->descStatusId = $self->translateNameToTermId(
-                    'description status',
-                    $self->rowStatusVars['descriptionStatus'],
-                    [],
-                    $self->status['descriptionStatusTypes'][$self->columnValue('culture')]
-                );
+                if (
```
djjuhasz: Rather than adding this conditional around all calls to `translateNameToTermId()` I think it would be better to do the null handling inside the `translateNameToTermId()` method to prevent errors like this in the future.

djjuhasz (changes requested):
@sbreker one change request.

lib/task/import/csvAccessionImportTask.class.php
```diff
@@ -416,18 +416,17 @@ public function execute($arguments = [], $options = [])
                     $self->status['resourceTypes'] = self::refreshTaxonomyTerms(QubitTaxonomy::ACCESSION_RESOURCE_TYPE_ID);
                 }
 
-                $this->setObjectPropertyToTermIdLookedUpFromTermNameArray(
-                    $self,
-                    'resourceTypeId',
+                $self->object->resourceTypeId = $self->translateNameToTermId(
                     'resource type',
                     $data,
```
djjuhasz: Sorry, I'm a bit confused, the signature for `QubitFlatfileImport::translateNameToTermId()` expects a string for the second argument.  Is `$data` a string value? If so, it's a very confusing variable name. :(

lib/task/import/csvAccessionImportTask.class.php
```diff
@@ -416,18 +416,17 @@ public function execute($arguments = [], $options = [])
...
+                    [],
```
djjuhasz: The third argument passed to `translateNameToTermId()` is always an empty array, so it seems like an unnecessary parameter, and the method logic can be simplified if it's removed.

djjuhasz (changes requested):
@sbreker one question, and one change request.

lib/task/import/csvAccessionImportTask.class.php
sbreker: Yes it seems to be and yes I agree - I had to take a good look at what was going on there.

lib/task/import/csvAccessionImportTask.class.php
sbreker: I noticed this. Was confused at it not being used anywhere.

lib/QubitFlatfileImport.class.php
```diff
@@ -1416,28 +1416,45 @@ public static function fetchKeymapEntryBySourceAndTargetName($sourceId, $sourceN
...
+            $value,
+            $terms[$culture]
+        );
 
-            throw new sfException('Could not find "'.$termName.'" in '.$description.' terms array.');
-        } else {
-            throw new sfException('Could not find a way to handle '.$description.' value "'.$value.'".');
+        if (false === $key) {
+            echo "\nTerm {$value} not found in {$taxonomyDescription} taxonomy, creating it...\n";
+            QubitFlatfileImport::createTerm(
```
djjuhasz: I think this approach is too risky - if the `$terms` value doesn't match the expected data or array structure (e.g. it's an empty array )  `$value` will be added to the database, without any check that it's not a duplicate. I think it would be better to query the database to double check that there is no existing term with the same taxonomyId, culture, and value, before adding a new term.  IMHO the benefit of ensuring duplicate terms aren't created is worth the cost of the extra database query.

lib/task/import/csvAccessionImportTask.class.php
```diff
@@ -406,65 +406,51 @@ public function execute($arguments = [], $options = [])
...
-                    echo "\nTerm {$data} not found in resource type taxonomy, creating it...\n";
-                    $newTerm = QubitFlatfileImport::createTerm(QubitTaxonomy::ACCESSION_RESOURCE_TYPE_ID, $data, $self->columnValue('culture'));
-                    $self->status['resourceTypes'] = self::refreshTaxonomyTerms(QubitTaxonomy::ACCESSION_RESOURCE_TYPE_ID);
-                }
-
-                $this->setObjectPropertyToTermIdLookedUpFromTermNameArray(
-                    $self,
-                    'resourceTypeId',
+            if ($data && isset($self->object) && $self->object instanceof QubitAccession) {
+                $self->object->resourceTypeId = $self->createOrFetchTermIdFromName(
```
djjuhasz, line 410: Ooh, I like that you moved the "create term" logic into a method to avoid repetition. :)

lib/QubitFlatfileImport.class.php
djjuhasz: I think `QubitFlatfileImport::createTerm()` should be moved to `lib/model/QubitTerm.php` too, as it's directly related to manipulating the model IMHO.

lib/QubitFlatfileImport.class.php
```diff
@@ -1416,28 +1416,45 @@ public static function fetchKeymapEntryBySourceAndTargetName($sourceId, $sourceN
...
+        if (empty($value)) {
+            return;
+        }
 
-            if (in_array($termName, $terms)) {
-                return array_search($termName, $terms);
-            }
+        $key = csvImportBaseTask::arraySearchCaseInsensitive(
+            $value,
+            $terms[$culture]
```
djjuhasz: If the `$terms[$culture]` key doesn't exist I think this will throw an error.

djjuhasz (changes requested):
@sbreker I think adding the `createOrFetchTermIdFromName()` method to avoid duplicating the term creation logic is a big improvement, but I'm worried about the impact of bad data in the $terms argument.

lib/QubitFlatfileImport.class.php
sbreker: The $terms array lookup by taxonomy was implemented in the past to avoid going to the database for every term since looking up terms by name is expensive. I think we can rely on the terms array to be updated, but in looking at this again, I realize I've broken the update of the terms array (csvImportBaseTask::refreshTaxonomyTerms) when a new term is created - the update is not carried out back to the status array.  I may not be able to move this logic into the new QubitFlatfileImport::createOrFetchTermIdFromName() as I had planned.

djjuhasz: @sbreker I have a few comments on this.

1. This is a public method, even if the `$terms` argument is 100% correct all the time in this class, it may not be used correctly when externally called.
2. This is a really complicated and big class, and I think the probability is high of somebody passing the wrong `$terms` data into the method. All it would take is a copy & paste of an existing call and forgetting to change the `$self->status['actorTypes']` key.
3. The final database safety check for the term can happen only if the cached `$terms` lookup misses.  Most imports tend to use existing terms a lot, so the ratio of cache hits to fails should be pretty high.

lib/QubitFlatfileImport.class.php
```diff
@@ -1416,28 +1395,60 @@ public static function fetchKeymapEntryBySourceAndTargetName($sourceId, $sourceN
...
+        if (isset($terms[$culture])) {
+            $key = csvImportBaseTask::arraySearchCaseInsensitive($value, $terms[$culture]);
+        }
 
-            if (in_array($termName, $terms)) {
-                return array_search($termName, $terms);
+        if (false === $key) {
+            // Test if term is present in DB. Retrieve taxonomy terms from term table.
+            $query = "SELECT t.id, ti.name FROM term t LEFT JOIN term_i18n ti ON t.id=ti.id \r
+                WHERE t.taxonomy_id=? AND ti.culture=?";
```
djjuhasz: I think it's better to add the `$value` to the search query (term_i18n.name) so only the one row is returned on a match and zero rows if there is no matching term.

sbreker: See latest commit @djjuhasz.  I have confirmed same number of rows read for both queries and have updated the code.

lib/QubitFlatfileImport.class.php
```diff
@@ -1416,28 +1395,59 @@ public static function fetchKeymapEntryBySourceAndTargetName($sourceId, $sourceN
...
-                return array_search($termName, $terms);
+        if (false === $key) {
+            // Test if term is present in DB. Retrieve taxonomy terms from term table.
+            $query = "SELECT t.id FROM term t LEFT JOIN term_i18n ti ON t.id=ti.id \r
+                WHERE t.taxonomy_id=? AND ti.culture=? AND ti.name=?";
+
+            $result = QubitPdo::fetchOne($query, [$taxonomyId, $culture, $value]);
+
+            if (false !== $result) {
+                return $result->id;
```
djjuhasz, line 1429: In this conditional `$value` was not found in the `$terms` array, but was in the database, so I think it would be good to add the `$value` to the `$terms` array to prevent subsequent database queries for the same `$value`?

djjuhasz (changes requested):
Sorry @sbreker, one more change request 😬

lib/QubitFlatfileImport.class.php
sbreker: I think this makes sense. At one point during dev, I had a call to csvImportBaseTask::refreshTaxonomyTerms() right before this return but removed it! Added a new commit adding it back in.

djjuhasz (approved):
Looks good @sbreker! 🎉 :shipit:

PR #1481, opened 2021-12-13
Update composer dependencies

sbreker:
Update composer dependencies to prevent PHPUnit errors in CI.

Post-update, PHP CS Fixer found new code formatting issues. Fixes for these are contained in second commit.

### DIFF
--- a/lib/task/migrate/arUpgradeSqlTask.class.php
+++ b/lib/task/migrate/arUpgradeSqlTask.class.php
@@ -507,7 +507,7 @@ EOF;
             throw new Exception('This task needs the PHP readline extension.');
         }
 
-        for (;;) {
+        while (true) {
             $this->logSection('upgrade-sql', 'Please enter a new theme choice:');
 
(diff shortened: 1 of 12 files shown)


lib/task/migrate/arUpgradeSqlTask.class.php
```diff
@@ -507,7 +507,7 @@ private function getNewTheme($themes)
             throw new Exception('This task needs the PHP readline extension.');
         }
 
-        for (;;) {
+          while (true){
```
jraddaoui: Indentation? :thinking:

jraddaoui (approved):
:ok_hand:

lib/task/migrate/arUpgradeSqlTask.class.php
sbreker: Weird! Check the CI error in qa/2.x @jraddaoui. Should it be using PHP 8.1?

PR #1483, opened 2021-12-20
Save generated finding aid transcript, #13596

djjuhasz:
When generating a PDF finding aid, save the transcript to the database
and Elasticsearch index, so the text is searchable.


djjuhasz:
Saving generated finding aid text is not a wanted feature.

PR #1484, opened 2021-12-21
Refactor Pub Status Update, refs #13274

sbreker:
Refactor the publication status update job to use ES update by query.

arElasticSearchPlugin::updateByQuery() updates fields specified in
script for matching terms in one ES request.


apps/qubit/modules/informationobject/actions/updatePublicationStatusAction.class.php
```diff
@@ -33,23 +33,19 @@ public function execute($request)
             $this->form->bind($request->getPostParameters());
 
             if ($this->form->isValid()) {
-                // Update resource synchronously
                 $publicationStatusId = $this->form->getValue('publicationStatus');
-                $this->resource->setPublicationStatus($publicationStatusId);
-                $this->resource->save();
```
jraddaoui, line 39: Doing the resource and descendants asynchronously introduces some complexities later and triggers the job in all cases. I'd keep this part as it is and just modify the descendants update in the job.

apps/qubit/modules/informationobject/actions/updatePublicationStatusAction.class.php
```diff
@@ -33,23 +33,19 @@ public function execute($request)
...
-                    $i18n = $this->context->i18n;
-                    $this->context->getConfiguration()->loadHelpers(['Url']);
-                    $jobsUrl = url_for(['module' => 'jobs', 'action' => 'browse']);
-                    $message = $i18n->__('Your description has been updated. Lower level descriptions are being updated now – check the <a href="%1">job scheduler page</a> for status and details.', ['%1' => $jobsUrl]);
-                    $this->getUser()->setFlash('notice', $message);
-                }
+                $updateDescendants = filter_var($this->form->getValue('updateDescendants'), FILTER_VALIDATE_BOOLEAN);
+
+                // Update using job scheduler
+                $options = ['objectId' => $this->resource->id, 'publicationStatusId' => $publicationStatusId, 'updateDescendants' => $updateDescendants];
```
jraddaoui: Nitpick: split into multiple lines if you keep the addition of `updateDescendants`.

apps/qubit/modules/informationobject/actions/updatePublicationStatusAction.class.php
```diff
@@ -33,23 +33,19 @@ public function execute($request)
...
+                QubitJob::runJob('arUpdatePublicationStatusJob', $options);
+
+                // Let user know descendants update has started
+                $i18n = $this->context->i18n;
+                $this->context->getConfiguration()->loadHelpers(['Url']);
+                $jobsUrl = url_for(['module' => 'jobs', 'action' => 'browse']);
+                $message = $i18n->__('Your description has been updated. Lower level descriptions are being updated now – check the <a href="%1">job scheduler page</a> for status and details.', ['%1' => $jobsUrl]);
```
jraddaoui: If you keep the new logic this message should be changed to indicate the resource is also being saved asynchronously. And now that you are here, it would be nice if you could split it in multiple lines.

lib/job/arUpdatePublicationStatusJob.class.php
```diff
@@ -18,39 +18,129 @@
...
      */
-    protected $extraRequiredParameters = ['objectId', 'publicationStatusId'];
+    protected $extraRequiredParameters = ['objectId', 'publicationStatusId', 'updateDescendants'];
 
     public function runJob($parameters)
     {
+        $validPublicationStatusList = [
+            QubitTerm::PUBLICATION_STATUS_DRAFT_ID,
+            QubitTerm::PUBLICATION_STATUS_PUBLISHED_ID,
+        ];
```
jraddaoui: Is this used?

lib/job/arUpdatePublicationStatusJob.class.php
```diff
@@ -18,39 +18,129 @@
...
+                FROM information_object io
+                JOIN slug
+                    ON io.id = slug.object_id
+                WHERE slug.slug = :slug
+            )';
+
+            $message = $this->i18n->__(
+                'Updating publication status for "%1" to "%2".',
+                ['%1' => $resource->getTitle(['cultureFallback' => true]), '%2' => $publicationStatus->name]
+            );
```
jraddaoui: I see you took the query from the gist, which was created to facilitate the work in support (using the slug) and without a lot of care.

If you keep the resource update outside the job and only trigger it for the descendants update this could be a lot more straight forward. And I wouldn't use a subquery, something like ...

```sql
UPDATE status
JOIN information_object io ON status.object_id = io.id
SET status.status_id = :publicationStatus
WHERE status.type_id = :publicationStatusType
AND io.lft > :lft AND io.rgt < :rgt;
```

If you want to keep the resource update here you could use `AND io.lft >= :lft AND io.rgt <= :rgt;`.

lib/job/arUpdatePublicationStatusJob.class.php
```diff
@@ -18,39 +18,129 @@
...
+
+        $descriptionsUpdated = QubitPdo::modify(
+            $sql,
+            $params
+        );
+
+        $response = QubitSearch::getInstance()->updateByQuery(
+            [
+                'ancestors' => $resource->id,
+                '_id' => $resource->id,
```
jraddaoui: Similar here, keeping the resource update outside will remove this clause.

lib/job/arUpdatePublicationStatusJob.class.php
```diff
@@ -18,39 +18,129 @@
...
+                    'inline' => 'ctx._source.publicationStatusId = '.$publicationStatus->id,
+                    'lang' => 'painless',
+                ],
+            ],
+            [
+                // Set to 'proceed' so index update does not abort if a document version
+                // conflict occurs. This happens if an ES doc is updated during this update.
+                // Ignore conflicts since conflict documents would still be indexed with the
+                // updated publication status from the DB.
+                'conflicts' => 'proceed',
```
jraddaoui: Nice.

plugins/arElasticSearchPlugin/lib/arElasticSearchIndexDecorator.class.php
```diff
@@ -0,0 +1,78 @@
...
+            'query' => \is_array($q) ? $q : $q->toArray(),
+            'script' => $script->toArray()['script'],
+        ];
+
+        $endpoint->setBody($body);
+        $endpoint->setParams($options);
+
+        return $this->requestEndpoint($endpoint);
+    }
+}
```
jraddaoui, line 78: Nice! I thought it would be a little more simple.

plugins/arElasticSearchPlugin/lib/arElasticSearchPlugin.class.php
```diff
@@ -365,6 +367,30 @@ public function partialUpdate($object, $data)
...
+            $query->setTerm($k, $v);
+            $queryBool->addShould($query);
+        }
+
+        $queryScript = \Elastica\Script\AbstractScript::create($script);
+        $response = $this->index->updateByQuery($queryBool, $queryScript, $options);
+
+        return $response->getData();
+    }
+
```
jraddaoui: I'm not sure this wrapper ads a lot and I think it could get easily out of hand (allowing other type of queries, etc.). For now, building the query/script and calling `QubitSearch::getInstance()->index->updateByQuery()` from the job directly could be enough.

jraddaoui (changes requested):
Thanks @sbreker!

Nice work bringing Elastica 6.x features before we upgrade to achieve the update by query.

apps/qubit/modules/informationobject/actions/updatePublicationStatusAction.class.php
sbreker: 👍

apps/qubit/modules/informationobject/actions/updatePublicationStatusAction.class.php
sbreker: Good point. I will move the resource update back to the action.

apps/qubit/modules/informationobject/actions/updatePublicationStatusAction.class.php
sbreker: I have reverted this so that the resource update occurs synchronously in the action.

lib/job/arUpdatePublicationStatusJob.class.php
sbreker: Good catch

lib/job/arUpdatePublicationStatusJob.class.php
sbreker: I have refactored this in latest commit.

lib/job/arUpdatePublicationStatusJob.class.php
sbreker: 😊

apps/qubit/modules/informationobject/actions/updatePublicationStatusAction.class.php
```diff
@@ -35,19 +35,60 @@ public function execute($request)
...
                 $publicationStatusId = $this->form->getValue('publicationStatus');
-                $this->resource->setPublicationStatus($publicationStatusId);
-                $this->resource->save();
 
-                // Update descendants using job scheduler
+                $sql = 'UPDATE status
+                    JOIN information_object io ON status.object_id = io.id
+                    SET status.status_id = :publicationStatus
+                    WHERE status.type_id = :publicationStatusType
+                    AND io.id = :id';
```
jraddaoui: No need to join here, just add another where clause for the `status.object_id`.

apps/qubit/modules/informationobject/actions/updatePublicationStatusAction.class.php
```diff
@@ -35,19 +35,60 @@ public function execute($request)
...
+                    ],
+                    [
+                        // Set to 'proceed' so index update does not abort if a document version
+                        // conflict occurs. This happens if an ES doc is updated during this update.
+                        // Ignore conflicts since conflict documents would still be indexed with the
+                        // updated publication status from the DB.
+                        'conflicts' => 'proceed',
+                    ]
+                );
+
```
jraddaoui, line 46: I think it's better to use a partial update here since it's only a single document update ...

```php
QubitSearch::getInstance()->partialUpdate(
    $this->resource,
    ['publicationStatusId' => $publicationStatusId]
);
```

config/ProjectConfiguration.class.php
```diff
@@ -24,6 +24,8 @@
 
 require_once __DIR__.'/../lib/QubitApcUniversalClassLoader.php';
 
+require_once __DIR__.'/../vendor/composer/autoload.php';
+
```
jraddaoui, line 28: Now that this is added to the main project configuration, I think it could be removed from other places (like lib/CsvImportValidator.class.php).

lib/job/arUpdatePublicationStatusJob.class.php
```diff
@@ -18,42 +18,113 @@
...
+
+        $descriptionsUpdated = QubitPdo::modify(
+            $sql,
+            $params
+        );
+
+        $response = QubitSearch::getInstance()->updateByQuery(
+            [
+                'ancestors' => $resource->id,
+            ],
```
jraddaoui, line 83: Nitpick: no need to split this array into multiple lines.

plugins/arElasticSearchPlugin/lib/arElasticSearchPlugin.class.php
jraddaoui: If you use partial update for the resource update I think we still don't need this abstraction. Even if we keep it, we don't need the bool query yet.

apps/qubit/modules/informationobject/actions/updatePublicationStatusAction.class.php
```diff
@@ -35,19 +35,60 @@ public function execute($request)
...
+                $descriptionsUpdated = QubitPdo::modify(
+                    $sql,
+                    $params
+                );
+
+                $response = QubitSearch::getInstance()->updateByQuery(
+                    [
+                        '_id' => $this->resource->id,
+                    ],
```
jraddaoui: Nitpick: no need to split this array into multiple lines.

jraddaoui (changes requested):
Thanks @sbreker!

apps/qubit/modules/informationobject/actions/updatePublicationStatusAction.class.php
sbreker: I have removed this in favour of using $resource->save() with indexOnSave = false

apps/qubit/modules/informationobject/actions/updatePublicationStatusAction.class.php
sbreker: 👍

apps/qubit/modules/informationobject/actions/updatePublicationStatusAction.class.php
sbreker: This has been removed completely so no longer an issue.

lib/job/arUpdatePublicationStatusJob.class.php
sbreker: 👍

plugins/arElasticSearchPlugin/lib/arElasticSearchPlugin.class.php
sbreker: I have removed the wrapper and removed the bool query.

lib/job/arUpdatePublicationStatusJob.class.php
sbreker: This has been removed.

jraddaoui (approved):
Really nice @sbreker, thanks!!

PR #1488, opened 2022-01-06
Add CSV import audit CLI task, refs #13569

mcantelon:
Adds CLI task to check legacy ID values from a CSV import file and list any that haven't been imported yet (creating the list by checking keymap data).

### DIFF
--- a/lib/CsvImportAuditer.class.php
+++ b/lib/CsvImportAuditer.class.php
@@ -1,0 +1,307 @@
+<?php
+
+/*
+ * This file is part of the Access to Memory (AtoM) software.
+ *
+ * Access to Memory (AtoM) is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * Access to Memory (AtoM) is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with Access to Memory (AtoM).  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+require_once __DIR__.'/../vendor/composer/autoload.php';
+
+/**
+ * Auditer for CSV imports.
+ *
+ * @author  Mike Cantelon <mike@artefactual.com>
+ */
+class CsvImportAuditer
+{
+    protected $sourceName;
+    protected $targetName = 'information_object';
+    protected $filename;
+    protected $ormClasses;
+    protected $errorLogHandle;
+    protected $rowsAudited = 0;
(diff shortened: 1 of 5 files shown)


lib/CsvImportAuditer.class.php
```diff
@@ -0,0 +1,308 @@
...
+
+                continue;
+            }
+
+            ++$this->rowsAudited;
+            $this->log($this->progressUpdate($this->rowsAudited));
+        }
+
+        if (!empty($this->missingIds)) {
+            $this->log('');
```
djjuhasz: Is logging an empty string here necessary?

lib/CsvImportAuditer.class.php
```diff
@@ -0,0 +1,308 @@
...
+    {
+        // Write to error log (but not STDERR)
+        if (STDERR != $this->getErrorLogHandle()) {
+            fwrite($this->getErrorLogHandle(), $msg.PHP_EOL);
+        }
+    }
+
+    protected function getErrorLogHandle(): resource
+    {
+        if (null === $filename = $this->getOption('errorLog')) {
```
djjuhasz, line 274: Looks like this is expecting a `$filename` argument for this method that doesn't exist?

lib/task/import/csvAuditImportTask.class.php
```diff
@@ -0,0 +1,153 @@
...
+        $this->namespace = 'csv';
+        $this->name = 'audit-import';
+        $this->briefDescription = 'Audit CSV import.';
+        $this->detailedDescription = <<<'EOF'
+Audit CSV import by checking to make sure a keymap has been created for each
+row.
+EOF;
+    }
+
+    protected function getDbConnection()
```
djjuhasz: It doesn't look like this method is called?

test/phpunit/CsvImportAuditerTest.php
```diff
@@ -0,0 +1,368 @@
+<?php
+
+use org\bovigo\vfs\vfsStream;
+
+/**
+ * @internal
+ * @covers \CsvImportAuditer
+ */
+class CsvImportAuditerTest extends \PHPUnit\Framework\TestCase
```
djjuhasz, line 9: YAY UNIT TESTS!!! 🥇 🎉 👍

djjuhasz (changes requested):
Looks good @mcantelon, just a few minor change requests.  Love the unit tests! 😁

lib/CsvImportAuditer.class.php
mcantelon: I was hacking in a blank line but I'll see if I can do that but explicitly adding one instead.

lib/CsvImportAuditer.class.php
mcantelon: It's getting the value, if set, from the "errorLog" option, if set.

mcantelon:
Thanks @djjuhasz ! I've updated the PR.

lib/CsvImportAuditer.class.php
djjuhasz: Oh right. :)

djjuhasz (approved):

PR #1489, opened 2022-01-09
Fix CSV check multi-file output, refs #13561

mcantelon:
Fixes issues with the output of the CSV check task when checking multiple files.

Now shows CSV filename corresponding to results, rather than the count of CSV file.

Also resets counters for errors/warnings when changing which CSV file's results are being shown.


djjuhasz (approved):
Looks good @mcantelon. :shipit: :)

mcantelon:
Thanks @djjuhasz !

PR #1490, opened 2022-01-09
Fix D.O. path CSV column check tool, refs #13588

mcantelon:
Fixed issue with the digital object path CSV column check tool where checks for absolute digital object paths weren't working correctly.

Fix was suggested by Scott Breeden.


djjuhasz (approved):
Looks good @mcantelon. 😁

mcantelon:
Thanks @djjuhasz !

PR #1491, opened 2022-01-10
Fix HTML title escaping, refs #13586

mcantelon:
Fixed issue with premature escaping of HTML titles.

### DIFF
--- a/lib/QubitWebResponse.class.php
+++ b/lib/QubitWebResponse.class.php
@@ -17,18 +17,18 @@
  * along with Access to Memory (AtoM).  If not, see <http://www.gnu.org/licenses/>.
  */
 
- class QubitWebResponse extends sfWebResponse
- {
-     /**
-      * Sets title for the current web response.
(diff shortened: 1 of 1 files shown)


lib/QubitWebResponse.class.php
```diff
@@ -23,9 +23,9 @@ class QubitWebResponse extends sfWebResponse
       * Sets title for the current web response.
       *
       * @param string  $title   Title name
-      * @param bool    $escape  true, for escaping the title
+      * @param bool    $escape  false, for escaping the title
       */
-     public function setTitle($title, $escape = true)
+     public function setTitle($title, $escape = false)
```
djjuhasz: Nitpick: I think the indenting is wonky, it's using 5 and 9 space indents in my editor. 😕

djjuhasz (changes requested):
One nitpick, but looks good other than that @mcantelon. :)

PR #1493, opened 2022-01-23
Clean up CSV transformation code, refs #13599

mcantelon:
[This PR is to clean up some old code I wrote a long time ago. I'd like to document this functionality so it's easier for others to use, but wanted to clean it up first.]

Cleaned up CSV transformation code, got rid of need for a separate database configuration for it, and added a cleanup function to get rid of temporarily needed data.

### DIFF
--- a/lib/QubitCsvTransform.class.php
+++ b/lib/QubitCsvTransform.class.php
@@ -19,6 +19,11 @@
 
 class QubitCsvTransform extends QubitFlatfileImport
 {
+    /**
+     * Name of MySQL table to create as a place to temporarily store row data.
+     */
+    public const WORKING_AREA_TABLE = 'import_rows';
+
     public $setupLogic;
     public $transformLogic;
     public $rowsPerFile = 1000;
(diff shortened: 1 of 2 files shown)


lib/QubitCsvTransform.class.php
```diff
@@ -99,27 +84,26 @@ public function writeHeadersOnFirstPass()
...
-        if (false === mysqli_query($link, $sql)) {
-            throw new sfException('MySQL delete from import_descriptions failed.');
-        }
+        // Delete contents of work area
+        $sql = 'DELETE FROM '.self::WORKING_AREA_TABLE;
+
+        QubitPdo::prepareAndExecute($sql);
+    }
+
+    public static function dropMySQLtemp()
```
djjuhasz, line 107: I wonder if it's worth adding a class destructor that calls this method, in case the script is halted unexpectedly?

djjuhasz (commented):
Looks great @mcantelon. I just had one optional suggestion.

lib/QubitCsvTransform.class.php
mcantelon: Good thinking! I'll add that.

mcantelon:
Thanks @djjuhasz ! I've updated the PR to add the destructor.

djjuhasz (approved):

PR #1495, opened 2022-02-22
Tweak DC templates, refs #13600

jraddaoui:
- Remove format and type metadata from digital object.
- Move description before subject terms in index and edit pages.


sbreker (approved):
Looks great! Thanks @jraddaoui!  :+1:

PR #1496, opened 2022-03-07
Fix search-box autocomplete, refs #13274

jraddaoui:
Pass the actual ES index instance to `\Elastica\Multi\Search` instead of
the new `arElasticSearchIndexDecorator`.


sbreker (approved):
Thanks! Looks great @jraddaoui 👍

PR #1498, opened 2022-03-30
Normalize holdings label usage, refs #13610

jraddaoui:
- Use "results" instead of harcoded "holdings" in browse link.
- Use `app_ui_label_institutionSearchHoldings` as form arial-label.
- Use only "Search" as input placeholder and aria-label.


djjuhasz (approved):
LGTM 👍

PR #1500, opened 2022-03-30
Add destructor to clean up temp dir, refs #13598

djjuhasz:
- Add arZipFileDownload destructor to automatically clean up the
  temporary job directory when the script completes or fails
- Pass the job `logger` object to arZipFileDownload to log deletion of
  temp job directory
- Use `create_date()` to get a valid DateTime object to pass to
  `date_timestamp_get()` in arZipFileDownload


jraddaoui (approved):
Neat!

PR #1501, opened 2022-04-04
Fix settings culture discrepencies, refs #13607

djjuhasz:
Always use source_culture value for non-translatable settings (e.g.
'hits_per_page') so value doesn't change with the current UI language.

Settings modified to be non-translatable:
- Digital object derivatives > PDF page number for image derivative
- Inventory > Levels of description
- Site information > Base URL
- Treeview settings > Sidebar > Sort (information object)

Also:

Use source_culture value when checking whether to disable the 
"Settings > Default page elements > Digital object map" checkbox.

Update the `tools:purge` task to set the culture of the site title and
description as follows:
- Use the settings.yml `default_culture` by default,
- Use the `--culture` option to override the default culture,
- For the demo data (`--demo`) always use 'en' because the demo strings
  are English

### DIFF
--- a/lib/model/QubitInformationObject.php
+++ b/lib/model/QubitInformationObject.php
@@ -479,7 +479,7 @@ class QubitInformationObject extends BaseInformationObject
 
     public function getMediaTypes()
     {
-        //TO DO: get via linked digital objects & physical objects
+        // TO DO: get via linked digital objects & physical objects
     }
 
     public function getRepositoryCountry()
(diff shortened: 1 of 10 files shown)


lib/model/QubitInformationObject.php
```diff
@@ -2462,7 +2462,9 @@ public function getTreeViewSiblings(array $options = [], &$siblingsRemaining = n
 
     public static function addTreeViewSortCriteria($criteria)
     {
-        switch (sfConfig::get('app_sort_treeview_informationobject')) {
+        $setting = QubitSetting::getByName('sort_treeview_informationobject');
+
+        switch ($setting->getValue(['sourceCulture' => true])) {
```
jraddaoui: I think we could use `sfConfig::get('app_sort_treeview_informationobject__source')` to avoid hitting the DB. See:

https://github.com/artefactual/atom/blob/qa/2.x/lib/model/QubitSetting.php#L102

jraddaoui (approved):
Looks great @djjuhasz, thanks!

lib/model/QubitInformationObject.php
djjuhasz: Oh cool! I did not know about the sfConfig setting `__source` vars.  Thanks @jraddaoui. 👍

PR #1503, opened 2022-04-13
Fix edit Donor typo, refs #13612

melaniekung:
Checkbox for Primary Contract should say Primary Contact, instead of Primary Contract, when editing Contact person for a Donor.

### DIFF
--- a/apps/qubit/i18n/ar/messages.xml
+++ b/apps/qubit/i18n/ar/messages.xml
@@ -3466,7 +3466,7 @@
         <note>https://github.com/artefactual/atom/blob/master/apps/qubit/modules/contactinformation/templates/_contactInformation.php</note>
       </trans-unit>
       <trans-unit id="487b19b411631ca1a19742f365becbcac67042ec">
-        <source>Primary contract</source>
+        <source>Primary Contact</source>
         <target>&#x627;&#x644;&#x639;&#x642;&#x62F; &#x627;&#x644;&#x623;&#x633;&#x627;&#x633;&#x64A;</target>
         <note>https://github.com/artefactual/atom/blob/master/apps/qubit/modules/contactinformation/templates/_edit.php</note>
       </trans-unit>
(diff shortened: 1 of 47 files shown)


djjuhasz:
@melaniekung I can't reproduce the reported bug in qa/2.x, I see "Primary contact":
Screenshot 2022-04-13 142949

Are you looking at the donor form in a culture other than English?  In English the source string for label should be used, and the source string seems to correctly use "Primary contact" (see https://github.com/artefactual/atom/blob/qa/2.x/apps/qubit/modules/contactinformation/templates/_contactInformation.php#L10)

I'm not sure how the "Primary cont*r*act" source string got into the i18n data, but with the source string being "Primary contact" those translations should not be loaded.

djjuhasz (changes requested):
Please see my earlier comment.

djjuhasz:
Ah, looks like the source string was changed from "Primary contract" to "Primary contact" in this commit: https://github.com/artefactual/atom/commit/6d1542a40abc06f25a498df9cb5a21e8fc08c12e

djjuhasz:
@melaniekung I don't think changing the source string in the i18n message.xml files is the way to go here, because it results in two translation units for "Primary contact":

```
      <trans-unit id="55daee81e80bbdfd34ef3aa11e529b197631ebc0">
        <source>Primary contact</source>
        <target>Contacte principal </target>
        <note>https://github.com/artefactual/atom/blob/master/apps/qubit/modules/contactinformation/templates/_contactInformation.php</note>
      </trans-unit>
      <trans-unit id="487b19b411631ca1a19742f365becbcac67042ec">
        <source>Primary Contact</source>
        <target/>
        <note>https://github.com/artefactual/atom/blob/master/apps/qubit/modules/contactinformation/templates/_edit.php</note>
      </trans-unit>
```

At best the second unit will be ignored, at worst it will replace the correct <target> string (Contacte principal).

PR #1504, opened 2022-04-13
Delete DIP after processing, refs #13396

djjuhasz:
- Add `\AccessToMemory\Path` class and tests
- Call qtPackageExtractorBase::clean() after processing metadata
- Use `Path::delete()` to recursively delete the package directory
- Pass logger to qtPackageExtratorBase to allow logging DIP deletion
- Remove unnecessary autoloader from `test/bootstrap/phpunit.php`


djjuhasz:
I'm going to change this fix to also delete the uploaded DIP when processing files (via a destructor)

PR #1505, opened 2022-04-15
Add a composer test coverage report script

djjuhasz:
Add a `composer test-cov` script that generates a phpunit text format
coverage report (using pcov) and sends it to STDOUT.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -29,6 +29,7 @@
     },
     "scripts": {
         "test": "phpunit",
+        "test-cov": "@php -dpcov.enabled=1 ./vendor/composer/bin/phpunit --coverage-html=.coverage/html",
         "php-cs-fixer": "php-cs-fixer"
     }
 }


jraddaoui (approved):
Looks good @djjuhasz!

Maybe we could reference https://projects.artefactual.com/issues/13493 in the commit message and add some notes in https://wiki.accesstomemory.org/wiki/Development/Testing#Unit_tests.

djjuhasz:
@jraddaoui I did this because I didn't know about the documented method of running the [AtoM Unit tests in Docker](https://wiki.accesstomemory.org/wiki/Development/Testing#Docker) using the "ATOM_COVERAGE" environment variable.

After reading the docs, I think it may be better to use the `--coverage-html=.coverage/html` option in the `test-cov` script too. What do you think?

jraddaoui:
I agree @djjuhasz, the text format with that many files is quite long and hard to read. The `ATOM_COVERAGE` env. var was initially added to be able to generate coverage reports from the integration tests, this way is a bit easier so I'd replace the documented method from the Wiki after merging this PR.

Also, I just remembered that we added a coverage section in https://github.com/artefactual/atom/blob/qa/2.x/phpunit.xml#L7-L16 which is already excluding a few directories, so `-dpcov.exclude=\"~vendor~\"` shouldn't be needed (or should replace the vendor entry is you're looking for other vendor directories deep in the tree).

djjuhasz:
https://wiki.accesstomemory.org/wiki/Development/Testing#Docker updated

PR #1506, opened 2022-04-21
Delete DIP after processing, refs #13396

djjuhasz:
- Add a `qtPackageExtractorBase` destructor to delete DIP files after
  processing or when extraction fails
- Add `\AccessToMemory\Path` class and tests
- Use `Path::delete()` to recursively delete the DIP directory
- Pass a logger to `qtPackageExtratorBase` to log DIP deletion
- Remove unused `qtPackageExtractorBase::getFilesFromDirectory()`
  method
- Remove unnecessary autoloader from `test/bootstrap/phpunit.php`

### DIFF
--- a/lib/AccessToMemory/Path.php
+++ b/lib/AccessToMemory/Path.php
@@ -1,0 +1,175 @@
+<?php
+
+/*
+ * This file is part of the Access to Memory (AtoM) software.
+ *
+ * Access to Memory (AtoM) is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * Access to Memory (AtoM) is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with Access to Memory (AtoM).  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+namespace AccessToMemory;
+
+/**
+ * Filesystem path abstraction class.
+ */
+class Path
+{
+    public $path;
+
+    public function __construct($path)
+    {
+        $this->path = $path;
+    }
+
+    public function __toString()
+    {
+        return $this->path;
+    }
+
+    /**
+     * Check if path exists.
+     *
+     * @return bool true if path exists
+     */
+    public function exists()
+    {
+        return file_exists($this->path);
+    }
+
+    /**
+     * Check if path references a directory.
+     *
+     * @return bool true if path references a directory
+     */
+    public function isDir()
+    {
+        $this->throwExceptionIfFileNotFound();
+
+        return is_dir($this->path);
+    }
+
+    /**
+     * Check if path references a file.
+     *
+     * @return bool true if path references a file
+     */
+    public function isFile()
+    {
+        $this->throwExceptionIfFileNotFound();
+
+        return is_file($this->path);
+    }
+
+    /**
+     * Return directory listing as an array of Path objects.
+     *
+     * @param array $exclude from list
+     *
+     * @return array directory contents as Path objects
+     */
+    public function ls($exclude = ['.', '..'])
+    {
+        $this->throwExceptionIfNotADir();
+        $items = [];
+
+        foreach (scandir($this->path) as $item) {
+            // Remove $excluded items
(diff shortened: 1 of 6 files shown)


test/phpunit/PathTest.php
```diff
@@ -0,0 +1,204 @@
...
+        $directory = [
+            'subdir' => [
+                'foo.txt' => 'foo',
+            ],
+            'empty_dir' => [],
+            'bar.txt' => 'bar',
+        ];
+
+        // setup and cache the virtual file system
+        $this->vfs = vfsStream::setup('root', null, $directory);
```
jraddaoui, line 44: Cool!

lib/AccessToMemory/Path.php
```diff
@@ -0,0 +1,175 @@
...
+
+    protected function throwExceptionIfNotAFile()
+    {
+        if (!$this->isFile()) {
+            throw new \RuntimeException(
+                sprintf('%s is not a file', $this->path)
+            );
+        }
+    }
+}
```
jraddaoui, line 175: You went deep in this abstraction to just delete the directory, hopefully we give it more use in the future.

plugins/qtSwordPlugin/lib/qtPackageExtractorBase.class.php
```diff
@@ -55,6 +57,15 @@ public function __construct(array $options = [])
...
         }
+
+        if (isset($options['logger'])) {
+            $this->logger = $options['logger'];
+        }
+    }
+
+    public function __destruct()
+    {
+        $this->cleanUp();
```
jraddaoui: Nitpick: I don't see a point (yet) to have another clean-up method that is only called here.

test/bootstrap/phpunit.php
```diff
@@ -14,14 +14,6 @@
...
 $configuration = ProjectConfiguration::hasActive() ? ProjectConfiguration::getActive() : new ProjectConfiguration(realpath($_test_dir.'/..'));
 
-// autoloader
-$autoload = sfSimpleAutoload::getInstance(sfConfig::get('sf_cache_dir').'/project_autoload.cache');
-$autoload->loadConfiguration(sfFinder::type('file')->name('autoload.yml')->in([
-    sfConfig::get('sf_symfony_lib_dir').'/config/config',
-    sfConfig::get('sf_config_dir'),
-]));
-$autoload->register();
-
```
jraddaoui, line 24: Nice! I always wondered why was this needed.

test/phpunit/PathTest.php
```diff
@@ -0,0 +1,204 @@
...
+        $path = new Path($this->vfs->url().'/bar.txt');
+        $path->delete();
+        $this->assertFalse($this->vfs->hasChild('bar.txt'));
+
+        // Delete a dir recursively
+        $path = new Path($this->vfs->url().'/subdir');
+        $path->delete(true);
+        $this->assertFalse($this->vfs->hasChild('subdir'));
+    }
+}
```
jraddaoui, line 204: Neat!

jraddaoui (approved):
Looks great @djjuhasz! Maybe a bit too much, but hopefully we'll give a good use to that path abstraction in the future.

lib/AccessToMemory/Path.php
djjuhasz: Yes, I also needed to recursively delete a directory in PR https://github.com/artefactual/atom/pull/1500.  

In PR #1500 I used [sfToolkit::clearDirectory()](https://github.com/artefactual/atom/blob/938ee91eeff42c168aa82c641b7e979785596634/lib/job/arZipFileDownload.class.php#L84) which I copied from the previous code. When I was working on this fix I examined `sfToolkit::clearDirectory()` and realized that it won't delete a directory that contains `.svn` or `CVS` files anywhere in it's subtree. :(   It's pretty unlikely that a DIP or export directory would have a CVS or .svn file, but it seems to me that it's a non-intuitive behaviour that could be a problem in some cases.

I also found in writing the Path library that the PHP functions like `rmdir()` and `unlink()` will fail without any error if e.g. you try and `unlink()` a directory, or vice versa.   In some cases a silent failure is fine, but there are cases where a it causes a bug. I think it's better in general to consistently throw an exception on failure, and give the developer the choice to catch the exception if they don't want the script to halt.

plugins/qtSwordPlugin/lib/qtPackageExtractorBase.class.php
djjuhasz: Yeah, fair. I'll move the code into the `__destruct()` method.

djjuhasz:
> Looks great @djjuhasz! Maybe a bit too much, but hopefully we'll give a good use to that path abstraction in the future.

Probably too much. But have you seen the test coverage report for `\AccessToMemory\Path`? 100%! 😆

PR #1507, opened 2022-04-21
Show link to DIP upload target, refs #13504

djjuhasz:
Use `objectId` job parameter for qtSwordPluginWorker so a link to the
target archival description is show on the "Manage jobs" page.


jraddaoui (approved):
:ok_hand:

PR #1508, opened 2022-04-21
Use Path::delete() to delete temp dir, refs #13598

djjuhasz:
`sfToolkit::clearDirectory()` will not delete `CVS` or `.svn` sub-directories, which can prevent the target directory from being deleted.


jraddaoui (approved):
Thanks!

PR #1510, opened 2022-05-05
Add CLI task to get/set settings, refs #13604

mcantelon:
Added CLI task to get or set settings, optionally to or from files.

### DIFF
--- a/lib/task/tools/settingsTask.class.php
+++ b/lib/task/tools/settingsTask.class.php
@@ -1,0 +1,347 @@
+<?php
+
+/*
+ * This file is part of the Access to Memory (AtoM) software.
+ *
+ * Access to Memory (AtoM) is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * Access to Memory (AtoM) is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with Access to Memory (AtoM).  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+/**
+ * Set or get setting values.
+ *
+ * @author  Mike Cantelon <mike@artefactual.com>
+ */
+class settingsTask extends arBaseTask
+{
+    protected $ormClasses;
+
+    public function __construct(sfEventDispatcher $dispatcher, sfFormatter $formatter)
+    {
+        $this->setOrmClasses([
+            'setting' => QubitSetting::class,
+        ]);
+
+        parent::__construct($dispatcher, $formatter);
+    }
+
+    public function setOrmClasses(array $classes): void
+    {
+        $this->ormClasses = $classes;
+    }
+
+    /**
+     * @see sfTask
+     *
+     * @param mixed $arguments
+     * @param mixed $options
+     */
+    public function execute($arguments = [], $options = [])
+    {
+        parent::execute($arguments, $options);
+
+        $this->validateOptions($arguments, $options);
+
+        try {
+            $this->dispatchOperation($arguments, $options);
+        } catch (ValueError $e) {
+            throw new sfException($e->getMessage());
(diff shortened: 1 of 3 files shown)


lib/task/tools/settingsTask.class.php
```diff
@@ -0,0 +1,351 @@
...
+     * @param mixed $arguments
+     * @param mixed $options
+     */
+    public function execute($arguments = [], $options = [])
+    {
+        parent::execute($arguments, $options);
+
+        $this->validateOptions($arguments, $options);
+
+        $this->displayCurrentSettings();
```
djjuhasz: When I run the task I always get the full list of settings, e.g.

```
[dev/issue-13604-settings-task][~/atom]$ symfony tools:setting get version
------------------------------------------------------------------------------------
Name                                                              Scope
------------------------------------------------------------------------------------
_csrf_token                                                       element_visibility
accession_counter
accession_mask
accession_mask_enabled
actor                                                             default_template
...
```

Is this intentional? It makes it hard to actually see the value for the selected setting.

jraddaoui:
Nice one @mcantelon!

@mamedin recently asked for something similar to get setting and config values, I created this quick "tools:run" task ...

```php
<?php

$config = sfConfig::getAll();
$config['factories'] = sfFactoryConfigHandler::getConfiguration(
    $this->configuration->getConfigPaths('config/factories.yml')
);

echo json_encode($config, JSON_PRETTY_PRINT) . "\n";
```

I think it could be a good addition to include a way to get the values from the factories config, maybe not in this task or without a set option. And also to add an option to output as JSON when printing multiple values, so you can then pipe the output a bit easily ...

`php symfony tools:run output_config_json.php | jq -r '.factories.user.param.timeout'`

lib/task/tools/settingsTask.class.php
djjuhasz: The "_csrf_token" setting is due to a bug (https://projects.artefactual.com/issues/13613) I just discovered while testing this task. 🤦

jraddaoui: Oh, good catch @djjuhasz!

mcantelon: Weird one, that "_csrf_token". 😨 

Yeah, I added that display of a list of settings to make the settings easier to work with, but displaying it the way I'm displaying it - with every command - is kludge-y. I'll take it out for now... I could make a new CLI task to display settings if need be. That'd be less weird.

mcantelon:
Thanks @radda! Good thinking... I've kind of shoehorned settings display into this task so I think I'm going to remove it and add the logic to a new task. Adding display of factory settings makes sense, too, so I'll add a Redmine issue to describe this new task (and specify that it should have JSON output too).

lib/task/tools/settingsTask.class.php
```diff
@@ -0,0 +1,351 @@
...
+                'en'
+            ),
+            new sfCommandOption(
+                'scope',
+                null,
+                sfCommandOption::PARAMETER_OPTIONAL,
+                'Scope of setting to get or set',
+                null
+            ),
+            new sfCommandOption(
```
djjuhasz, line 324: Opinion: I think it's more intuitive to make the 'value' an optional argument, so the command can be called like `symfony tools:setting set [name] [value]`

lib/task/tools/settingsTask.class.php
```diff
@@ -0,0 +1,351 @@
...
+                'strict',
+                null,
+                sfCommandOption::PARAMETER_NONE,
+                'Prevent creation of new settings when performing "set" operation',
+                null
+            ),
+        ]);
+
+        $this->namespace = 'tools';
+        $this->name = 'setting';
```
djjuhasz: Nitpick: I think it's better for the class name to be `SettingTask` or to make `$this->name = 'settings';` for consistency.

djjuhasz (approved):
@mcantelon a few comments but no blockers.  I'm impressed that you simulated the database interactions for the tests. :)

lib/task/tools/settingsTask.class.php
mcantelon: Good call. Changing the name to "settings" makes sense.

mcantelon:
@djjuhasz I've updated the PR and did a bit of minor refactoring, fixed some errors I found, added a couple of new tests, and got rid of the `to-file` and `from-file` options (replacing them with a single `file` option).

(Oh you approved it earlier... thanks! I'll merge it!)

PR #1511, opened 2022-05-10
Add RAD template actor visibility, refs #13424

mcantelon:
Improve visibility of non-creation actors in the RAD template by:

1) Making the related actor name in the Dates area a hyperlink to the
   authority record for any actor relation

2) Displaying the related actor name for non-creation events in the
   Name Access Points area


djjuhasz (approved):
Looks good @mcantelon. :)

PR #1513, opened 2022-05-12
Search for terms across all fields, refs #13608

djjuhasz:
Instead of adding one "should" clause for each search field use a
single "should" clause that searches all relevant fields.  This change
allows multiple search terms (e.g. "foo bar") to match "foo" in one
field and "bar" in a different field.

Thank you to https://gist.github.com/bardocuteam for submitting the fix.


djjuhasz:
This is breaking some searches, so I'm withdrawing the PR. :(

PR #1514, opened 2022-05-16
Remove redundancy from actor page, refs #12718

mcantelon:
On the left sidebar of the actor detail page, descriptions related by events were showing in the "subject of" section in addition to in the event type's section. Removed display of these events in the "subject of" section.

### DIFF
--- a/apps/qubit/modules/actor/actions/relatedInformationObjectsAction.class.php
+++ b/apps/qubit/modules/actor/actions/relatedInformationObjectsAction.class.php
@@ -90,18 +90,13 @@ class ActorRelatedInformationObjectsAction extends sfAction
             $queryTerm = new \Elastica\Query\Term(['names.id' => $actorId]);
 
             $queryBool->addMust($queryTerm);
+
+            foreach (QubitTerm::getEventTypes() as $eventType) {
+                $queryNested = self::nestedActorAndEventTypeQuery($actorId, $eventType->id);
+                $queryBool->addMustNot($queryNested);
+            }
         } else {
-            // Get related by event IOs
-            $queryBoolDates = new \Elastica\Query\BoolQuery();
-            $queryBoolDates->addMust(new \Elastica\Query\Term(['dates.actorId' => $actorId]));
-            $queryBoolDates->addMust(new \Elastica\Query\Term(['dates.typeId' => $eventTypeId]));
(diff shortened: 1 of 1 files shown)


apps/qubit/modules/actor/actions/relatedInformationObjectsAction.class.php
```diff
@@ -90,18 +90,13 @@ public static function getRelatedInformationObjects($actorId, $page, $limit, $ev
...
-
-            // Use nested query and mapping object to allow querying
-            // over the actor and event ids from the same event
-            $queryNested = new \Elastica\Query\Nested();
-            $queryNested->setPath('dates');
-            $queryNested->setQuery($queryBoolDates);
 
+            foreach (QubitTerm::getEventTypes() as $eventType) {
+                $queryNested = self::nestedActorAndEventTypeQuery($actorId, $eventType->id);
+                $queryBool->addMustNot($queryNested);
```
djjuhasz, line 96: Is it necessary to explicitly exclude the event type relationships with a "MustNot" query?  I would expect simply not adding the events (like before this change) to the query to be sufficient. 😕

djjuhasz (approved):
I have one question about the functionality @mcantelon, but other than that the PR looks good to me.

apps/qubit/modules/actor/actions/relatedInformationObjectsAction.class.php
mcantelon: It's a bit convoluted. ;)

For issue 12718 it was requested to add the display of non-creator actors (created via the RAD template) in the name access points section of the description's page and to make them searchable, as name access points, in the advanced search. Commit 8ac0c7551a9e771df5ccba30ceca2c4429aa0efc (December of 2019) implemented this by indexing non-creator actors as name access points in the ElasticSearch index.

The only problem with this, though, is that the left sidebar of the actor page now ends up displaying these actors twice: once as a name access point and once grouped by its event type. This PR fixes this display issue by suppressing the display of them in the name access point section of the sidebar.

djjuhasz: Ah, thanks for the explanation.  Sounds like this is the best solution for the current search data model.

PR #1515, opened 2022-05-16
Add search:query command line task

djjuhasz:
Add a command line `search:query` task to return a list of
Elasticsearch results for the given query.

### DIFF
--- a/lib/task/search/arSearchQueryTask.class.php
+++ b/lib/task/search/arSearchQueryTask.class.php
@@ -1,0 +1,176 @@
+<?php
+
+/*
+ * This file is part of the Access to Memory (AtoM) software.
+ *
+ * Access to Memory (AtoM) is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * Access to Memory (AtoM) is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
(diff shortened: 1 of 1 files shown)


lib/task/search/arSearchQueryTask.class.php
```diff
@@ -0,0 +1,173 @@
...
+{
+    /**
+     * @see sfTask
+     *
+     * @param mixed $arguments
+     * @param mixed $options
+     */
+    public function execute($arguments = [], $options = [])
+    {
+        $count = 0;
```
mcantelon: This seems unused.

lib/task/search/arSearchQueryTask.class.php
```diff
@@ -0,0 +1,173 @@
...
+                $i + 1,
+                $resultSet->getTotalHits()
+            )
+        );
+    }
+
+    protected function outputSearchFields($className)
+    {
+        $count = 0;
+        $fields = $this->getFields($className);
```
mcantelon: This function doesn't seem to exist (I get an error when I try to use this option).

mcantelon (commented):
This'll be a handy tool! Found one issue, though.

lib/task/search/arSearchQueryTask.class.php
djjuhasz: Thanks, good catch. :)

lib/task/search/arSearchQueryTask.class.php
djjuhasz: Oops, yeah I updated the method name to `getSearchFields()` and forgot to update it here. 😊

mcantelon: Ah, cool cool!

mcantelon (approved):
Looks good to me!

djjuhasz:
This script is intended for development / analysis so it's probably better not to include it in the public release.

PR #1516, opened 2022-05-18
Fix issues with bulk:export task, refs #13615

mcantelon:
Fixed a few issues with bulk export query task.


djjuhasz (approved):
Looks good!

PR #1517, opened 2022-05-19
Update ES mappings during indexing, refs #13468

mcantelon:
Fixed issue with Elasticsearch indexing task where indexing only select types would "skip" the updating of mappings.

### DIFF
--- a/plugins/arElasticSearchPlugin/lib/arElasticSearchPlugin.class.php
+++ b/plugins/arElasticSearchPlugin/lib/arElasticSearchPlugin.class.php
@@ -214,6 +214,9 @@ class arElasticSearchPlugin extends QubitSearchEngine
 
             // Load mappings if index initialization wasn't needed
             $this->loadAndNormalizeMappings();
+
+            // Update mappings (in case they've been changed during an upgrade, etc.)
+            $this->updateMappings();
         }
 
         // Display what types will be indexed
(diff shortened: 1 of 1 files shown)


djjuhasz (changes requested):
According to the Elasticsearch documentation, most mappings can't be updated in a live index: https://www.elastic.co/guide/en/elasticsearch/reference/5.6/indices-put-mapping.html#updating-field-mappings

I confirmed that I got an error when I changed an information object mapping in `plugins/arElasticSearchPlugin/config/mapping.yml` (`slug: { type: keyword }` -> `slug: { type: text }`), then ran `search:populate` while excluding the informationObject type.

```
*[dev/issue-13468-update-napping-when-excluding-types][~/atom]$ symfony search:populate --exclude-types="accession,aip,functionObject,informationObject,repository,term"
Defining mapping QubitAip...
Defining mapping QubitTerm...
Defining mapping QubitActor...
Defining mapping QubitAccession...
Defining mapping QubitRepository...
Defining mapping QubitFunctionObject...
Defining mapping QubitInformationObject...


  mapper [slug] of different type, current_type [keyword], merged_type [text]
```

djjuhasz:
The error message doesn't stand out in the text paste, so here's a screenshot
Screenshot 2022-05-20 091551

mcantelon:
@djjuhasz Good catch! I'll see if I can take a new approach to it.

PR #1518, opened 2022-05-20
Search for terms across all fields, refs #13608

djjuhasz:
Instead of adding one "should" clause for each search field use a
single "query string" that searches all relevant fields.  This change
allows multiple search terms (e.g. "foo bar") to match "foo" in one
field and "bar" in a different field.  This makes
arElasticSearchPluginUtil::generateBoolQueryString()
unnecessary, so I've replace calls to it with generateBoolQueryString().

Thank you to https://gist.github.com/bardocuteam for submitting the
initial suggestion to remove the multiple "should" search clauses.

Unfortantely this change reintroduced bug #12186: "Searching for titles
with stop words will not give results". To get stop words working again
requires two additional changes:

- use the stop word analyzer for the current user culture in the
  Elastica QueryString object
- remove "keyword" fields from the "_all" search field list - query
  string searches don't filter stop words when keyword fields are
  included in the search fields

The three keyword fields removed from the general search are: slug,
referenceCodeWithoutCountryAndRepo, and digitalObject.thumbnailPath.

### DIFF
--- a/apps/qubit/modules/default/actions/moveAction.class.php
+++ b/apps/qubit/modules/default/actions/moveAction.class.php
@@ -129,7 +129,9 @@ class DefaultMoveAction extends sfAction
                 sprintf('i18n.%s.title', sfContext::getInstance()->user->getCulture()) => 1,
             ];
             $this->queryBool->addMust(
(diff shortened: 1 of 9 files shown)


mcantelon (approved):
Looks good to me... very cool!

PR #1519, opened 2022-05-25
Update content of static pages, refs #13090

mcantelon:
Added a migration to, if about page/home/privacy page content hasn't been changed from the original content, update it to new content.


lib/task/migrate/migrations/arMigration0193.class.php
```diff
@@ -0,0 +1,165 @@
...
+            $culture = $translation['culture'];
+
+            $criteria = new Criteria();
+            $criteria->add(QubitStaticPageI18n::ID, $pageIds[$slug]);
+            $criteria->add(QubitStaticPageI18n::CULTURE, $culture);
+
+            $page = QubitStaticPageI18n::getOne($criteria);
+
+            // Attempt to update static page's title and content if they've never been changed
+            if ($translation['md5'] == md5($page->title.$page->content)) {
```
djjuhasz, line 103: I like the md5 hash comparison :)

data/fixtures/settings.yml
```diff
@@ -12,7 +12,7 @@ QubitSetting:
     name: version
     editable: 0
     deleteable: 0
-    value: 192
+    value: 193
```
djjuhasz, line 15: Changing the data schema version doesn't seem necessary in this case, because the actual table schema isn't changing, just the data in the table. 🤔

lib/task/migrate/migrations/arMigration0193.class.php
```diff
@@ -0,0 +1,165 @@
...
+     * Parse static page fixture data into a simnplified form that's easier
+     * to work with.
+     *
+     * Example of the structure:
+     *
+     *  [
+     *    'pageslug' => [
+     *      'en' => [
+     *        'title' => 'English Title',
+     *        'content' => 'Eglish content'
```
djjuhasz, line 125: Typo: "Eglish" :)

djjuhasz (commented):
I think I understand why you used a data migration to update the text for these static pages - the migrations are always run on upgrade, so it doesn't require running the script independently.  Changing the data schema version can be disruptive though, especially with custom branches, and I don't think we should update the schema version unless the database *schema* is changing.

From a "bigger picture" perspective, I'm also uncomfortable with changing the data in an existing AtoM site without explicit approval from the owner of the site.  I recognize that this script is very carefully designed to avoid overwriting customizations, but it's possible that an AtoM owner *prefers* the original text to the new text.  I also think that the AtoM site owner should have full control over the data in their site.

All that said, I'm fine with updating the fixtures.yml with the new text, but I think changing the text in an existing AtoM installation should require explicitly running a script (not a migration) and should not change the data schema version.

mcantelon:
Good points... I'll create a version that's a Symfony CLI task. 👍

PR #1521, opened 2022-06-03
Fix double escaped donor name, refs #13616

mcantelon:
Fixed issue where donor names, when viewed in accession pages, where getting double escaped.


djjuhasz (approved):
Looks good 👍

PR #1522, opened 2022-06-06
Restore finding aid phys. container, refs #13583

djjuhasz:
- Create a separate EAD physical object template
- Add authorization level to finding aid generator
- Default to showing ead:container data because they are usually
  visible to the public
- Create an XSL key to lookup an ead:physloc element by id
- Conditionally prepend ead:physloc data if present in the EAD
- Update display of ead:container @type and @label attributes
- Change label for phyloc + container data to "Physical storage"
- Add a XSL "ucfirst" function to capitalize the first letter of a
  string
- Remove unused logic for printing origination/persname@role

### DIFF
--- a/plugins/sfEadPlugin/modules/sfEadPlugin/templates/indexSuccessBody.xml.php
+++ b/plugins/sfEadPlugin/modules/sfEadPlugin/templates/indexSuccessBody.xml.php
@@ -81,23 +81,18 @@
 
 <archdesc <?php echo $ead->renderLOD($resource, $eadLevels); ?> relatedencoding="<?php echo $ead->getMetadataParameter('relatedencoding'); ?>">
   <?php
+    $resourceVar = 'resource';
+    $creators = ${$resourceVar}->getCreators();
+    $events = ${$resourceVar}->getActorEvents(['eventTypeId' => QubitTerm::CREATION_ID]);
 
-  $resourceVar = 'resource';
-  $counter = 0;
-  $counterVar = 'counter';
+    $topLevelDid = true;
 
-  $creators = ${$resourceVar}->getCreators();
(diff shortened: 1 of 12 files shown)


andykoch (approved):
I approve the copious use of 💲signs here.

mcantelon:
@andykoch Using PHP is like casting a spell to summon great wealth. ;)

plugins/sfEadPlugin/modules/sfEadPlugin/templates/indexSuccessBody.xml.php
```diff
@@ -81,23 +81,18 @@
 
 <archdesc <?php echo $ead->renderLOD($resource, $eadLevels); ?> relatedencoding="<?php echo $ead->getMetadataParameter('relatedencoding'); ?>">
   <?php
+    $resourceVar = 'resource';
```
mcantelon, line 84: Is this to avoid HTML Purifier from messing things up?

djjuhasz: @mcantelon are you asking if the `$resourceVar` variable is to avoid HTML Purifier messing things up?  If so, I'm not really sure why it uses a variable variable pattern here, but it predates HTML Purifier.   It looks like it was introduced by _you_ 10 years ago: https://github.com/artefactual/atom/blame/0971f2634c3371418fa28f1e2e2c8a9292477758/plugins/sfEadPlugin/modules/sfEadPlugin/templates/indexSuccessBody.xml.php 😆

mcantelon: Lolol..., I was probably trying to avoid some kind of escaping-ish issue then. ;)

mcantelon (approved):
Looks good to me! 👍

PR #1523, opened 2022-06-07
Fix finding aid cosmetic issues, refs #13618

djjuhasz:
General:
- Don't show the field label for "Dates of creation, revision and
  deletion" when there is no content.
- Generalize logic for displaying specialized field labels and values
  (e.g. dates, physical storage)
- Remove admin / bio history "Note" subtitle

Cover page:
- Remove extra line breaks from second line of title
- Remove vertical height setting for logo to avoid extra vertical whitespace
- Add a 1.5cm margin to the top of the page

Improve field type display:
- Move ead:odd type attribute display value lookups to a function
- Use field label lookup function for all field labels
- Add display value lookups for ead:unittitle and ead:note type values

Title notes:
- Don't show "Title Notes" heading when there are no title notes
- Use titleNoteLabel lookup function to display human friendly labels
- Remove the series level call to titleNotes template, as the EAD XML
  only includes title notes data for the top (archdesc) level

### DIFF
--- a/lib/task/pdf/ead-pdf-inventory-summary.xsl
+++ b/lib/task/pdf/ead-pdf-inventory-summary.xsl
@@ -18,9 +18,11 @@
     <xsl:output method="xml" encoding="utf-8" indent="yes"/>
     <!-- Calls a stylesheet with local functions and lookup lists for languages and subject authorities -->
     <xsl:include href="{{ app_root }}/lib/task/pdf/helper-functions.xsl"/>
+    <xsl:strip-space elements="*"/>
+
     <!-- Define keys -->
     <xsl:key name="physlocId" match="ead:physloc" use="@id"/>
-    <xsl:strip-space elements="*"/>
+
     <!-- The following attribute sets are reusabe styles used throughout the stylesheet. -->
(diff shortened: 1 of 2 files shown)


lib/task/pdf/ead-pdf-inventory-summary.xsl
```diff
@@ -1128,8 +1064,8 @@
     <xsl:template match="ead:note">
         <xsl:choose>
```
mcantelon: Could the <xsl:apply-templates> happen after the end of the choose statement? It looks like it always gets fired, no matter what condition within the choose structure, but I am not versed in this stuff so I'm probably wrong. ;)

mcantelon (approved):
I had one question - which could be due to me not understanding much XSL-fu - but all in all looks like a great cleanup/refactor PR!

lib/task/pdf/ead-pdf-inventory-summary.xsl
djjuhasz: @mcantelon I think you're right that <xsl:apply-templates/> should come after the <choose> to avoid repetition. :)

PR #1526, opened 2022-06-18
Add legacy ID to actor CSV export, refs #13620

mcantelon:
Added the internal ID to the actor CSV export so translations can be associated, etc.


djjuhasz (approved):
Looks good 👍

mcantelon:
Thanks @djjuhasz !

PR #1527, opened 2022-06-22
Fix JSON in digitalobject browse API, refs #13621

djjuhasz:
Make digitalobject id a property of JSON object, rather than an array
key. JSON does not allow keyed arrays.


tw4l (approved):
Good change!

Do you know of any existing clients using this API endpoint, considering this will be a breaking change for those? Though given the nature of the problem, maybe this endpoint hasn't been getting much use...

djjuhasz:
@tw4l I'm not sure if any clients are using the digitalobject/browse endpoint.  Hopefully it's an easy update for any clients that are using it.

PR #1529, opened 2022-06-28
Update REST API ACL checks, refs #13445

djjuhasz:
- Log each ACL check and result to the symfony log file (at "info"
  level) to make debugging easier.
- Add ACL checks to informationobject read (GET) endpoint to include
  digital object URLs (master, reference, and thumbnail)
- Add ACL checks to the informationobject create (POST) endpoint
- Add an ACL check to the informationobject update (PUT) endpoint
- Add ACL checks to the informationobject delete (DELETE) endpoint
- Add ACL checks to the informationobject tree (GET) endpoint
- Change informationobject/tree endpoint "parent_slug" paramater name
  to "slug" to be consistent with other endpoints
- Add an ACL check to the digitalobjects browse (GET) endpoint,
  require system-wide "readMaster" authorization

Add ACL checks to the digitalboject create (POST) endpoint:
- Require digitalobject create (POST) request to include one of:
  information_object_slug, information_object_id, or parent_id
- If using `information_object_slug` require "update" permission on the
  information object to add the digital object
- If using `information_object_id` require "create" permission on the
  information object (a child IO is created for the digital object)
- If using `parent_id` require "update" permission on the parent DO's
  information_object

### DIFF
--- a/lib/myUser.class.php
+++ b/lib/myUser.class.php
@@ -24,6 +24,15 @@ class myUser extends sfBasicSecurityUser implements Zend_Acl_Role_Interface
     // Module-specific permissions get temporarily stored here for access checks
     protected $security = [];
 
+    public function __toString()
+    {
+        if (isset($this->user)) {
+            return $this->user->username;
+        }
+
+        return 'NULL';
+    }
+
     /**
      * Required for Zend_Acl_Role_Interface.
      */
--- a/plugins/arRestApiPlugin/config/arRestApiPluginConfiguration.class.php
+++ b/plugins/arRestApiPlugin/config/arRestApiPluginConfiguration.class.php
@@ -67,10 +67,10 @@ class arRestApiPluginConfiguration extends sfPluginConfiguration
             'params' => ['slug' => '['.QubitSlug::getValidSlugChars().']+'],
         ]);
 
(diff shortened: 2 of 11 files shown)


plugins/arRestApiPlugin/lib/exception/QubitApiForbiddenException.class.php
```diff
@@ -19,7 +19,6 @@
 
 class QubitApiForbiddenException extends QubitApiException
 {
-    protected $message = 'Forbidden';
```
mcantelon: This ends up returning a blank message property (example below triggers this by an attempting to delete the root information object).

```
$ curl -XDELETE -u demo@example.com:demo http://127.0.0.1/api/informationobjects/q4ez-yr97-anxe
   {"id":"forbidden","message":""}
```

mcantelon (changes requested):
Found one very minor issue, but otherwise looks good!

plugins/arRestApiPlugin/lib/exception/QubitApiForbiddenException.class.php
djjuhasz: Good catch @mcantelon, I've restored that line.

mcantelon (approved):
Looks good to me! 👍

PR #1536, opened 2022-08-11
Add skip to main content link, refs #13530

jraddaoui:
First element inside the body and only visible on focus (first tab).

skip_to_main_link

PS. First refresh - continue tabbing, second refresh - enter on link and then continue tabbing.


sbreker (approved):
Looks great!  :+1:

PR #1543, opened 2022-10-26
Check language on FA generation, refs #13633

jraddaoui:
Do not include language of material if there are no language elements within it.


sbreker (approved):
Looks great @jraddaoui! :+1:

PR #1546, opened 2022-11-24
i18n pull from Weblate

sbreker:
Integrates translations from Weblate 2.x branch.

New i18n codes:
- Irish: 'ga'
- Macedonian: 'mk'


djjuhasz (approved):
Looks good @sbreker.  I just did a grep to check that the only files changed were "messages.xml" and the `data/fixtures` yml files.

PR #1547, opened 2022-12-05
Consider B5 theme in tarball creation, refs #13530

jraddaoui:
Also added some modifications to the internal Wiki that go alongside this change:

https://internalwiki.artefactual.com/wiki/AtoM_release_procedure#Packaging_the_release


djjuhasz (approved):
LGTM @jraddaoui :+1:

PR #1558, opened 2023-01-21
Fix remote digital objects import error via csv

willian95cs:
When importing digital objects via command line (csv) after importing the records, you get the error "Could not read the file" due to checking the path to the local directory. However, remote digital objects have no local path.

Fixes the bug by commenting out these validations and replacing importFromFile with importFromURI when creating digital objects so we can import to remote digital objects.

Allows the sending of local and remote digital objects in the same csv, validating the URI by the presence of the host. If the host does not exist, import the object with the local path.

### DIFF
--- a/lib/task/digitalobject/digitalObjectLoadTask.class.php
+++ b/lib/task/digitalobject/digitalObjectLoadTask.class.php
@@ -186,12 +186,12 @@ class digitalObjectLoadTask extends arBaseTask
                     continue;
                 }
 
-                if (!file_exists($path = self::getPath($item, $options))) {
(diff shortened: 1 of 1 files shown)


willian95cs:
This change makes the --link-source option work.

sarah-mason:
Hi @willian95cs - I want to apologise for the incredibly long delay in hearing from us. We have been making efforts to improve our response time to PRs, so you should never have to wait years in future. Thank you so much for your contribution! We appreciate our active contributor community around AtoM. After reviewing your PR, we have decided not to accept it at this time. It will break existing functionality for some use cases, so we have to deem it out of scope for the main project. We hope you consider contributing to the project again in future!

PR #1559, opened 2023-01-23
Changes the default behavior of creating new child descriptions

willian95cs:
Changes the default behavior of creating new child descriptions for descriptions 
that already have a digital object

When importing digital objects via the command line, the default behavior is to create 
new child descriptions when multiple lines point to the same description. However, 
there may be cases where the user has mistakenly duplicated identifiers and new descriptions 
start to pollute the database. Therefore, not adding the --attach-only option has the 
same effect as adding it.

Changing this, the default behavior is now to inform the user about duplicate lines 
that point to a description that already has a digital object, indicating that a 
digital object already exists for that identifier. If the user wants, he can use 
the --attach-only option to change the behavior of the task that will add child 
descriptions to attach the digital object.

### DIFF
--- a/lib/task/digitalobject/digitalObjectLoadTask.class.php
+++ b/lib/task/digitalobject/digitalObjectLoadTask.class.php
@@ -154,18 +154,17 @@ class digitalObjectLoadTask extends arBaseTask
 
                 continue;
             }
-
+            //Replace existing digital objects
             if ($options['replace']) {
                 $digitalObjectName = !is_array($item) ? $item : end($item);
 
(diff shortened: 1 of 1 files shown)


sarah-mason:
Hi @willian95cs - again, thank you for your contribution to the AtoM project! We do apologise for the significant delay. Much like your [PR 1558](https://github.com/artefactual/atom/pull/1558) this pull request breaks some existing functionality for a number of other use cases and so for the main project we have to deem it out of scope. We hope you will consider contributing in future and thank you again for your support of the project!

PR #1560, opened 2023-01-31
Add check for existing username in add super user task, ref #13643

anvit:
The CLI task to create users allows the creation of duplicate users.

`php symfony tools:add-superuser --email=email@domain.tld --password=password dupe_user_name`

Adding a check to prevent creating a user if one already exists.

### DIFF
--- a/lib/task/tools/addSuperuserTask.class.php
+++ b/lib/task/tools/addSuperuserTask.class.php
@@ -46,8 +46,6 @@ class addSuperuserTask extends sfBaseTask
         }
 
         sfContext::createInstance($this->configuration);
-        $databaseManager = new sfDatabaseManager($this->configuration);
-        $conn = $databaseManager->getDatabase('propel')->getConnection();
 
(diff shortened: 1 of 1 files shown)


lib/task/tools/addSuperuserTask.class.php
```diff
@@ -46,8 +46,6 @@ public function execute($arguments = [], $options = [])
         }
 
         sfContext::createInstance($this->configuration);
-        $databaseManager = new sfDatabaseManager($this->configuration);
-        $conn = $databaseManager->getDatabase('propel')->getConnection();
```
anvit, line 50: Unsure why this exists since this variable is unused. Please let me know if there's a reason for this, and I'll revert this particular change.

mcantelon (approved):
Looks good to me!

Only thing I see that needs to be fix is the commit header (first line) should be capitalized and should be 50 characters or less long (subsequent links should be 72 characters or less).

https://www.gitkraken.com/learn/git/best-practices/git-commit-message

PR #1561, opened 2023-02-03
Remove referenceCode column, refs #13626

Mbtj:
referenceCode column, only given in exported CSVs, is incompatible with the import format.


mcantelon (approved):
Looks good to me (only nitpick is commit message body needs to be fixed as each commit body line should be 72 characters or less)!

mcantelon:
Only nitpick is that one of the commit message body line's more than 72 characters.

PR #1563, opened 2023-02-08
Handle sourceCulture option for notes, ref#13631

anvit:
Add a __get method to QubitNote to handle sourceCulture option. Currently these calls go to BaseNote which ignores this option, and this leads to hidden fields such as Rights or Conservation showing up when language is changed.

See **line 265-273** in `plugins/sfRadPlugin/modules/sfRadPlugin/templates/indexSuccess.php` for a scenario where the `['sourceCulture' => true]` is passed correctly but would being ignored.

### DIFF
--- a/lib/model/QubitNote.php
+++ b/lib/model/QubitNote.php
@@ -22,6 +22,27 @@ class QubitNote extends BaseNote
     // Flag for updating search index on save or delete
     public $indexOnSave = true;
 
+    public function __get($name)
+    {
+        $args = func_get_args();
+
+        $options = [];
+        if (1 < count($args)) {
+            $options = $args[1];
+        }
+
+        if ('type' === $name && true === $options['sourceCulture']) {
+            $sql = 'SELECT term_i18n.name FROM note
+                JOIN term_i18n ON term_i18n.id = note.type_id
+                WHERE term_i18n.id=?
+                AND note.id=? AND term_i18n.culture=?';
+
+            return QubitPdo::fetchColumn($sql, [$this->typeId, $this->id, $this->sourceCulture]);
+        }
+
+        return call_user_func_array([$this, 'BaseNote::__get'], $args);
+    }
+
     public function __toString()
     {
         if (null === $content = $this->getContent()) {
--- a/plugins/arDominionB5Plugin/modules/sfRadPlugin/templates/indexSuccess.php
+++ b/plugins/arDominionB5Plugin/modules/sfRadPlugin/templates/indexSuccess.php
@@ -305,13 +305,11 @@
 
   <?php foreach ($resource->getNotesByTaxonomy(['taxonomyId' => QubitTaxonomy::RAD_NOTE_ID]) as $item) { ?>
 
-    <?php $type = $item->getType(['sourceCulture' => true]); ?>
-
-    <?php if ('Conservation' == $type && !check_field_visibility('app_element_visibility_rad_conservation_notes')) { ?>
+    <?php if ($conservationTermID === $item->type->id && !check_field_visibility('app_element_visibility_rad_conservation_notes')) { ?>
       <?php continue; ?>
     <?php } ?>
 
-    <?php if ('Rights' == $type && !check_field_visibility('app_element_visibility_rad_rights_notes')) { ?>
+    <?php if ($rightsTermID === $item->type->id && !check_field_visibility('app_element_visibility_rad_rights_notes')) { ?>
       <?php continue; ?>
     <?php } ?>
 
--- a/plugins/sfRadPlugin/modules/sfRadPlugin/actions/indexAction.class.php
+++ b/plugins/sfRadPlugin/modules/sfRadPlugin/actions/indexAction.class.php
@@ -187,5 +187,20 @@ class sfRadPluginIndexAction extends InformationObjectIndexAction
                 $this->errorSchema = $e;
             }
         }
+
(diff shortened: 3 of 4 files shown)


lib/model/QubitNote.php
```diff
@@ -22,6 +22,37 @@ class QubitNote extends BaseNote
...
+            //$criteria->addSelectColumn(QubitNote::TYPE_ID);
+            //$criteria->addJoin(QubitTermI18n::ID, QubitNote::TYPE_ID);
+            //$criteria->add(QubitNote::ID, $this->id);
+            //$criteria->add(QubitTermI18n::ID, $this->typeId);
+            //$criteria->add(QubitTermI18n::CULTURE, $this->sourceCulture);
+            //if (1 == count($query = QubitQuery::createFromCriteria($criteria, 'QubitNote'))) {
+                //echo $this->typeId;
+                //echo $query[0]->row[0];
+            //}
+            $sql = 'SELECT term_i18n.name from note join term_i18n on  term_i18n.id = note.type_id where term_i18n.id=? and note.id=? and term_i18n.culture=?';
```
anvit, line 49: This query works

lib/model/QubitNote.php
```diff
@@ -22,6 +22,37 @@ class QubitNote extends BaseNote
...
+        if (1 < count($args)) {
+            $options = $args[1];
+        }
+
+        if ($name === 'type' && $options['sourceCulture'] === true) {
+            //echo 'note id: ' . $this->id;
+            //echo 'type id: ' . $this->typeId;
+            //echo 'source culture: ' . $this->sourceCulture;
+
+            //$criteria = new Criteria();
```
anvit, line 38: This commented out block was supposed to achieve the same result as the query below but I'm getting the incorrect record, and the result I get is for a different typeId for some reason.

lib/model/QubitNote.php
```diff
@@ -22,6 +22,36 @@ class QubitNote extends BaseNote
     // Flag for updating search index on save or delete
     public $indexOnSave = true;
 
+    public function __get($name)
+    {
+        $args = func_get_args();
+
+        if ('type' === $name && true === $options['sourceCulture']) {
+            /*
```
anvit: This block should get the same output as the sql query below but it seems to be getting results for a record with a different typeId?

sbreker: I would drop this ORM query in favour of the QubitPdo query you have below - the QubitPdo query will be faster, and will deliver just the result you need instead of a full instantiated object.

lib/model/QubitNote.php
```diff
@@ -22,6 +22,24 @@ class QubitNote extends BaseNote
...
+    {
+        $args = func_get_args();
+
+        $options = [];
+        if (1 < count($args)) {
+            $options = $args[1];
+        }
+
+        if ('type' === $name && true === $options['sourceCulture']) {
+            $sql = 'SELECT term_i18n.name from note join term_i18n on  term_i18n.id = note.type_id where term_i18n.id=? and note.id=? and term_i18n.culture=?';
```
mcantelon: Nitpicks: there's a double space in the SQL query ("on  term_i18n.id") and breaking the SQL into multiple lines and capitalizing SQL keywords ("from", "where", etc.) could make it more readable.

```
$sql = 'SELECT term_i18n.name FROM note
    JOIN term_i18n ON term_i18n.id = note.type_id
    WHERE term_i18n.id=?
    AND note.id=? AND term_i18n.culture=?';
```

mcantelon (approved):
Some formatting nitpicks that could be fixed, but otherwise seems good!

lib/model/QubitNote.php
anvit: Fixed these!

lib/model/QubitNote.php
```diff
@@ -22,6 +22,27 @@ class QubitNote extends BaseNote
...
+    public function __get($name)
+    {
+        $args = func_get_args();
+
+        $options = [];
+        if (1 < count($args)) {
+            $options = $args[1];
+        }
+
+        if ('type' === $name && true === $options['sourceCulture']) {
```
sbreker, line 34: Will this trigger an error if the 'sourceCulture' key is not present in the array?

anvit: No, it just skips the if since `$options['sourceCulture']` would just end up being NULL.

sbreker:
I've been thinking about these two comparisons: https://github.com/artefactual/atom/blob/qa/2.x/plugins/sfRadPlugin/modules/sfRadPlugin/templates/indexSuccess.php#L267-L273

This will still break if the sourceCulture is not 'en' (or rather the exact sourceCulture phrases are not 'Rights' and 'Conservation').

Perhaps the comparison should be note.type_id matches the term.id for RAD NOTE "Conservation term"?

anvit:
> I've been thinking about these two comparisons: https://github.com/artefactual/atom/blob/qa/2.x/plugins/sfRadPlugin/modules/sfRadPlugin/templates/indexSuccess.php#L267-L273
> 
> This will still break if the sourceCulture is not 'en' (or rather the exact sourceCulture phrases are not 'Rights' and 'Conservation').
> 
> Perhaps the comparison should be note.type_id matches the term.id for RAD NOTE "Conservation term"?

Updated the condition to use IDs instead of strings in the comparison

sbreker (approved):
Looks great @anvit! Thanks!  :+1:

PR #1564, opened 2023-02-13
Remove user privileges for translator, ref #13647

anvit:
Translator user role should not have the ability to add or edit users so removing this privilege.

To update this policy, clear cache needs to be run after applying this commit `php symfony cc` or `php symfony cache:clear`


sbreker (approved):
Thanks for the fix @anvit! Looks great! :+1:

PR #1566, opened 2023-02-16
Full treeview move error fix, refs #13640

mcantelon:
Reindex nested list "lft" values to fix desynced values.

### DIFF
--- a/lib/task/propel/propelBuildNestedSetTask.class.php
+++ b/lib/task/propel/propelBuildNestedSetTask.class.php
@@ -22,7 +22,7 @@
  *
  * @author     David Juhasz <david@artefactual.com>
  */
-class propelBuildNestedSetTask extends sfBaseTask
+class propelBuildNestedSetTask extends arBaseTask
 {
     private $children;
     private $conn;
@@ -35,6 +35,8 @@ class propelBuildNestedSetTask extends sfBaseTask
      */
     public function execute($arguments = [], $options = [])
     {
+        parent::execute($arguments, $options);
+
         $databaseManager = new sfDatabaseManager($this->configuration);
         $this->conn = $databaseManager->getDatabase('propel')->getConnection();
 
@@ -93,11 +95,6 @@ class propelBuildNestedSetTask extends sfBaseTask
             $this->conn->commit();
         }
 
-        $this->logSection(
-            'propel',
-            'Note: you will need to rebuild your search index for updates to show up properly in search results.'
-        );
-
         $this->logSection('propel', 'Done!');
     }
 
@@ -114,6 +111,7 @@ class propelBuildNestedSetTask extends sfBaseTask
             new sfCommandOption('env', null, sfCommandOption::PARAMETER_REQUIRED, 'The environment', 'cli'),
             new sfCommandOption('connection', null, sfCommandOption::PARAMETER_REQUIRED, 'The connection name', 'propel'),
             new sfCommandOption('exclude-tables', null, sfCommandOption::PARAMETER_OPTIONAL, 'Exclude tables (comma-separated). Options: information_object, term, menu'),
+            new sfCommandOption('index', 'i', sfCommandOption::PARAMETER_OPTIONAL, 'Update search index (defaults to false)', false),
         ]);
 
         $this->namespace = 'propel';
@@ -155,6 +153,24 @@ EOF;
 
         $this->conn->exec($sql);
 
+        if ($options['index']) {
+            if ($node['id'] != $classname::ROOT_ID) {
+                $this->reindexLft($classname, $node['id'], $node['lft']);
+            }
+        } else {
+            $this->logSection(
+                'propel',
+                'Note: you will need to rebuild your search index for updates to show up properly in search results.'
+            );
+        }
+
         return $width;
+    }
+
+    protected function reindexLft(string $classname, int $id, int $lft)
+    {
+        QubitSearch::getInstance()->partialUpdateById(
+            $classname, $id, ['lft' => $lft]
+        );
     }
 }
--- a/plugins/arElasticSearchPlugin/lib/arElasticSearchPlugin.class.php
+++ b/plugins/arElasticSearchPlugin/lib/arElasticSearchPlugin.class.php
@@ -368,6 +368,31 @@ class arElasticSearchPlugin extends QubitSearchEngine
         }
     }
 
+    public function partialUpdateById(string $className, int $id, array $data)
+    {
+        if (!$this->enabled) {
+            return;
+        }
+
+        if (0 == strcmp($className, 'QubitUser')) {
+            return;
+        }
+
+        $document = new \Elastica\Document($id, $data);
+
+        try {
+            $this->index->getType($className)->updateDocument($document);
+        } catch (\Elastica\Exception\ResponseException $e) {
+            // Create document if none exists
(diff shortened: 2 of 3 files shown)


lib/task/propel/propelBuildNestedSetTask.class.php
```diff
@@ -35,6 +35,8 @@ class propelBuildNestedSetTask extends sfBaseTask
      */
     public function execute($arguments = [], $options = [])
     {
+        sfContext::createInstance($this->configuration);
+
```
jraddaoui, line 39: Since this task requires an initiated database, we could extend arBaseTask.

lib/task/propel/propelBuildNestedSetTask.class.php
```diff
@@ -93,11 +95,6 @@ public function execute($arguments = [], $options = [])
             $this->conn->commit();
         }
 
-        $this->logSection(
-            'propel',
-            'Note: you will need to rebuild your search index for updates to show up properly in search results.'
-        );
-
```
jraddaoui, line 100: It would be nice to have an index flag/option for this task now that it involves ES updates. In some cases it may be nice to index directly like this, but in others it may not be desirable (thinking about maintenance tasks where this is run right before indexing).

I'd set it to false to keep the current behavior by default and change it from the other tasks/jobs calling this one or with a CLI flag.

lib/task/propel/propelBuildNestedSetTask.class.php
```diff
@@ -155,6 +152,20 @@ protected function recursivelyUpdateTree($node, $classname)
...
+        }
+
         return $width;
     }
+
+    protected function reindexLft(string $classname, int $id, int $lft)
+    {
+        $this->logSection('propel', 'Classname '.$classname);
+        $this->logSection('propel', 'Id '.$id);
+        $this->logSection('propel', 'Lft '.$lft);
```
jraddaoui: Some debugging left behind? :stuck_out_tongue_closed_eyes:

plugins/arElasticSearchPlugin/lib/arElasticSearchPlugin.class.php
```diff
@@ -368,6 +368,31 @@ public function partialUpdate($object, $data)
...
+        }
+
+        if (0 == strcmp($className, 'QubitUser')) {
+            return;
+        }
+
+        $document = new \Elastica\Document($id, $data);
+
+        try {
+            $this->index->getType($className)->updateDocument($document);
```
jraddaoui, line 384: It doesn't have to happen in this PR but it would be nice to have `batchUpdateDocs` and do batch updates from both partial update methods, like it's done with `batchAddDocs` and `batchDeleteDocs`.

plugins/arElasticSearchPlugin/lib/arElasticSearchPlugin.class.php
```diff
@@ -368,6 +368,31 @@ public function partialUpdate($object, $data)
...
+        } catch (\Elastica\Exception\ResponseException $e) {
+            // Create document if no existing document exists
```
jraddaoui: Not my first language but that feels like a lot of wording :P

jraddaoui (changes requested):
Looking great @mcantelon, thanks!

Some minor comments below, the only request would be making indexing optional in the nested set task.

lib/task/propel/propelBuildNestedSetTask.class.php
mcantelon: Good call!

plugins/arElasticSearchPlugin/lib/arElasticSearchPlugin.class.php
mcantelon: Good thinking... I'll make a Redmine issue for that.

plugins/arElasticSearchPlugin/lib/arElasticSearchPlugin.class.php
mcantelon: True, "if none exists" would suffice.

lib/task/propel/propelBuildNestedSetTask.class.php
jraddaoui: After extending from arBaseTask, we could replace the first four lines from this method with a call to the parent method. Check other tasks doing the same, like:

https://github.com/artefactual/atom/blob/7d52ca72de98df50b64f86bcd9f2b7a0fc805113/lib/task/tools/updatePublicationStatusTask.class.php#L53

lib/task/propel/propelBuildNestedSetTask.class.php
```diff
@@ -114,6 +111,7 @@ protected function configure()
             new sfCommandOption('env', null, sfCommandOption::PARAMETER_REQUIRED, 'The environment', 'cli'),
             new sfCommandOption('connection', null, sfCommandOption::PARAMETER_REQUIRED, 'The connection name', 'propel'),
             new sfCommandOption('exclude-tables', null, sfCommandOption::PARAMETER_OPTIONAL, 'Exclude tables (comma-separated). Options: information_object, term, menu'),
+            new sfCommandOption('no-index', null, sfCommandOption::PARAMETER_OPTIONAL, 'Do not re-index after building nested set', false),
```
jraddaoui: I'd do it the other way around to maintain the current behavior. Like in:

https://github.com/artefactual/atom/blob/7d52ca72de98df50b64f86bcd9f2b7a0fc805113/lib/task/digitalobject/digitalObjectLoadTask.class.php#L252

Then you can set it to true in:

https://github.com/artefactual/atom/blob/7d52ca72de98df50b64f86bcd9f2b7a0fc805113/lib/task/import/csvImportTask.class.php#L1010

jraddaoui (changes requested):
Thanks @anvit!

Minor suggestion for the code review process, try not to amend commits and force push after a code review has started. It was not an issue in this case but it usually makes things harder for the reviewer, having to go over all diff again.

lib/task/propel/propelBuildNestedSetTask.class.php
jraddaoui: @anvit, I think you could also remove the two lines for the connection creation below:

https://github.com/artefactual/atom/blob/f4d2c1c502e88926c7124dfb91468e90446986fb/lib/task/propel/propelBuildNestedSetTask.class.php#L40-L41

jraddaoui (approved):
Thanks @anvit, make sure to squash and include other devs that worked on this in the commit message as co-authors.

lib/task/propel/propelBuildNestedSetTask.class.php
anvit: @jraddaoui I'd done that initially, but then noticed that these weren't actually being set up in arBaseTask's execute method, and also removing it made our `csv_import.cy.js` cypress test which checks for sibling order being preserved on import fail. 
Is there something I might be missing?

jraddaoui: I see, we are using transactions in this task which requires to create the connection like that instead of using the default one. Good thing you tested that out, thanks!

mcantelon:
Redmine issue to note that it would be good to, as Radda suggested, add bulk updating support at some point: https://projects.artefactual.com/issues/13654

PR #1567, opened 2023-02-23
fix README links

Mbtj:
replace the deadlink for linux installation with the updated link on the AtoM wiki


fiver-watson (approved):
👍 LGTM

PR #1568, opened 2023-02-24
Hide locations info from CSV reports, ref #13646

anvit:
When generating reports, location information should be hidden unless the generate report as public user is set to false. We have a condition check for this in our PHP template but adding in this condition check for the entire job so that it affects generated CSVs as well.

### DIFF
--- a/lib/job/arGenerateReportJob.class.php
+++ b/lib/job/arGenerateReportJob.class.php
@@ -187,7 +187,7 @@ class arGenerateReportJob extends arBaseJob
             $parentTitle = QubitInformationObject::getStandardsBasedInstance($item->parent)->__toString();
             $creationDates = $this->getCreationDates($item);
 
-            $results[$parentFile][] = [
+            $columns = [
                 'resource' => $item,
                 'referenceCode' => QubitInformationObject::getStandardsBasedInstance($item)->referenceCode,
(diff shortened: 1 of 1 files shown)


mcantelon:
Nicely done once again!

The commit message body can be up to 72 characters in length (only the commit header needs to be 50 characters or less).

One possible change to the commit message body would be changing `unless the generate report as public user is set to false` to `unless the "generate report as public user" setting is set to false` to make it clearer.

lib/job/arGenerateReportJob.class.php
```diff
@@ -198,8 +198,13 @@ private function getFileOrItemListResults($levelOfDescription)
                 ) : '',
                 'startDate' => isset($creationDates) ? $creationDates->startDate : null,
                 'accessConditions' => $item->getAccessConditions(['cultureFallback' => true]),
-                'locations' => $this->getLocationString($item),
             ];
+
+            if (0 == sfConfig::get('app_generate_reports_as_pub_user', 1)) {
+                $columns += ['locations' => $this->getLocationString($item)];
```
mcantelon: Nitpick... this could be clearer:

```$columns["locations"] = $this->getLocationString($item);```

mcantelon (approved):
I added a couple nitpicks, but otherwise seems good!

PR #1569, opened 2023-02-27
Increase PDF indexing limit, ref #13650

anvit:
Increase the max amount of PDF text that can be stored from 64KB to 16MB. Also update the upgrade-sql task to set the property value column to use the MEDIUMTEXT type instead of TEXT.

### DIFF
--- a/data/fixtures/settings.yml
+++ b/data/fixtures/settings.yml
@@ -3,7 +3,7 @@ QubitSetting:
     name: version
     editable: 0
     deleteable: 0
-    value: 192
+    value: 193
   milestone:
     name: milestone
     editable: 0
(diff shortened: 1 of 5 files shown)


mcantelon:
AtoM, confusingly, has two migration systems: an older one, that's now unused, and a newer one. The migration in this PR will need to be changed so a migration is added in the `lib/task/migrate/migrations` directory as `arMigration0193.class.php`. The `version` value in `data/fixtures/settings.yml` will have to be bumped up to 193.

Otherwise, looking good in general!

mcantelon (changes requested):
The migration needs to be changed (I've added comments to the PR).

mcantelon (approved):
This is looking good! Changes to `lib/task/migrate/arUpgrader120.class.php` should be removed but otherwise seems good to go.

PR #1570, opened 2023-02-28
Fix missing dc:format field in DC XML, ref #13651

anvit:
Update the DC XML template to correctly display dc:format field and remove `foreach` since `$dc->format` is a string instead of an array after the changes to the DC template in #13600.


jraddaoui (approved):
:ok_hand:

PR #1572, opened 2023-03-16
remove markdown rendering on DO files, refs #13503

Mbtj:
Remove function calls with render_value and render_value_inline on digital object files, preventing markdown elements to break the file names.


mcantelon (approved):
Looks good to me! You might want to squash the two commits into one (and capitalize the first line of the commit message).

PR #1574, opened 2023-03-20
Add authority record warnings in BS5, ref #13652

anvit:
Add a warning section to edit authority record BS5 template since these warnings are not easy to find when submitting the form, and don't auto expand the accordion section in BS5 like they do in BS2 themes.

### DIFF
--- a/plugins/arDominionB5Plugin/modules/sfIsaarPlugin/templates/editSuccess.php
+++ b/plugins/arDominionB5Plugin/modules/sfIsaarPlugin/templates/editSuccess.php
@@ -24,6 +24,17 @@
 
     <?php echo $form->renderHiddenFields(); ?>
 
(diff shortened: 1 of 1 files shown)


anvit:
Going with this solution since it makes it obvious that there's action required on a user's part and is a cleaner solution compared to the fix needed for auto-expanding the collapsed section in BS5. Discussed this with the maintainers and we decided that we can revisit this if we continue to get feedback from users about this issue.

jraddaoui (approved):
I see this being used in other templates, it doesn't have to happen in this PR but it may be nice to convert it into a partial that receives the form, so it could be easily reused in the future.

anvit:
> I see this being used in other templates, it doesn't have to happen in this PR but it may be nice to convert it into a partial that receives the form, so it could be easily reused in the future.

That makes sense. Will make a note and plan to do that in the future!

PR #1577, opened 2023-03-30
Sync lft values between DB and ES, refs #13640

mcantelon:
Before providing treeview data from Elasticsearch check for discrepancies between database and Elasticsearch nested set data and, if found, repair.

### DIFF
--- a/apps/qubit/modules/informationobject/actions/fullWidthTreeViewAction.class.php
+++ b/apps/qubit/modules/informationobject/actions/fullWidthTreeViewAction.class.php
@@ -66,6 +66,10 @@ class InformationObjectFullWidthTreeViewAction extends DefaultFullTreeViewAction
         ) {
             $data = $this->getAncestorsAndSiblings($options);
         } else {
+            // Copy DB lft values to Elasticsearch if need be
+            $syncer = new QubitLftSyncer($this->resource->id);
+            $syncer->sync();
+
             $data = $this->getChildren($this->resource->id, $options);
         }
 
--- a/apps/qubit/modules/informationobject/actions/fullWidthTreeViewAction.class.php
+++ b/apps/qubit/modules/informationobject/actions/fullWidthTreeViewAction.class.php
@@ -56,6 +56,10 @@ class InformationObjectFullWidthTreeViewAction extends DefaultFullTreeViewAction
             'limit' => $request->nodeLimit,
         ];
 
+        // Sync Elasticsearch values with DB if need be
+        $syncer = new QubitLftSyncer($this->resource->id);
+        $syncer->sync();
+
         // On first load, retrieve the ancestors of the selected resource, the
         // resource and its siblings, otherwise get only the resource's siblings
         if (
@@ -66,10 +70,6 @@ class InformationObjectFullWidthTreeViewAction extends DefaultFullTreeViewAction
         ) {
             $data = $this->getAncestorsAndSiblings($options);
         } else {
-            // Copy DB lft values to Elasticsearch if need be
-            $syncer = new QubitLftSyncer($this->resource->id);
-            $syncer->sync();
-
             $data = $this->getChildren($this->resource->id, $options);
         }
 
--- a/apps/qubit/modules/informationobject/actions/fullWidthTreeViewAction.class.php
+++ b/apps/qubit/modules/informationobject/actions/fullWidthTreeViewAction.class.php
@@ -56,10 +56,6 @@ class InformationObjectFullWidthTreeViewAction extends DefaultFullTreeViewAction
             'limit' => $request->nodeLimit,
         ];
 
-        // Sync Elasticsearch values with DB if need be
-        $syncer = new QubitLftSyncer($this->resource->id);
-        $syncer->sync();
-
         // On first load, retrieve the ancestors of the selected resource, the
         // resource and its siblings, otherwise get only the resource's siblings
         if (
--- a/lib/QubitLftSyncer.class.php
+++ b/lib/QubitLftSyncer.class.php
@@ -40,13 +40,13 @@ class QubitLftSyncer
 
     private function getChildLftChecksumForDB()
     {
-        $sql = 'SELECT lft
+        $sql = sprintf('SELECT lft
             FROM information_object
             WHERE parent_id=:parentId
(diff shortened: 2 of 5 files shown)


mcantelon:
Any feedback on this general approach (or suggestions of alternatives) are welcome!

apps/qubit/modules/informationobject/actions/fullWidthTreeViewAction.class.php
```diff
@@ -66,6 +66,10 @@ public function execute($request)
         ) {
             $data = $this->getAncestorsAndSiblings($options);
         } else {
+            // Copy DB lft values to Elasticsearch if need be
+            $syncer = new QubitLftSyncer($this->resource->id);
+            $syncer->sync();
+
```
jraddaoui: Why not on first load? There could be an issue in the target resource and siblings too.

lib/QubitLftSyncer.class.php
```diff
@@ -0,0 +1,111 @@
...
+            ORDER BY lft ASC
+            LIMIT :limit';
+
+        $params = [':parentId' => $this->parentId, ':limit' => $this->limit];
+        $result = QubitPdo::fetchAll($sql, $params, ['fetchMode' => PDO::FETCH_ASSOC]);
+
+        $lft = [];
+        foreach ($result as $row) {
+            $lft[] = $row['lft'];
+        }
```
jraddaoui:
```php
$lft = QubitPdo::fetchAll($sql, $params, ['fetchMode' => PDO::FETCH_COLUMN]);
```

Should do the same and avoid the loop.

lib/QubitLftSyncer.class.php
```diff
@@ -0,0 +1,111 @@
...
+        $term = new \Elastica\Query\Term(['parentId' => $this->parentId]);
+        $query->queryBool->addMust($term);
+        $query->query->addSort(['lft' => 'asc']);
+
+        // Get results
+        $result = QubitSearch::getInstance()
+            ->index
+            ->getType('QubitInformationObject')
+            ->search($query->getQuery(false, false))
+        ;
```
jraddaoui, line 72: Was this done by the auto-formatting?

lib/QubitLftSyncer.class.php
```diff
@@ -0,0 +1,111 @@
...
+            $bulk->addAction(
+                new Elastica\Bulk\Action\UpdateDocument(
+                    new Elastica\Document($row['id'], ['lft' => $row['lft']])
+                )
+            );
+        }
+
+        $bulk->send();
+    }
+}
```
jraddaoui, line 108: Really nice class, "simple" and understandable. It would be great to add some unit tests for it, but it may be hard to mock the DB and ES queries.

lib/QubitLftSyncer.class.php
```diff
@@ -0,0 +1,111 @@
+<?php
+
+class QubitLftSyncer
+{
+    private $parentId;
+    private $limit;
+
+    public function __construct($parentId, $limit = 10000)
```
jraddaoui, line 8: This ES limitation may be an issue in here. It would be great to see an implementation using ES' scroll API to overcome this limit.

jraddaoui (commented):
Hi @mcantelon!

Why is this needed after https://github.com/artefactual/atom/pull/1566? I'm a bit concerned about the performance implications of this changes and I think we should try to fix the root of the issue.

If you think this is really needed this way, I'd like to see it as an async implementation that doesn't affect the treeview load time (which is already slow enough). I have a couple of suggestions for it:

- Create a new endpoint and make a request to it from the JS code after treeview load, do the check and trigger an async job if needed, adding a notification in the UI to let the user know about the mismatch found.
- Create it as a task that runs over the entire dataset that could be scheduled. This would have deployment implications so I would rather see the previous one.

Either way, I think we should try to find what's causing the issue.

lib/QubitLftSyncer.class.php
mcantelon: Yeah, php-cs-fixer does it. I tried dragging the semicolon to the end of the previous line but php-cs-fixer ends up reformatting it the same way when run again.

lib/QubitLftSyncer.class.php
mcantelon: Thanks! Yeah, if we end up going this route I'll add tests.

lib/QubitLftSyncer.class.php
mcantelon: I'm not sure if users would end up manually sorting sets of children that large via the treeview so it might not be worth doing.

jraddaoui: I'm not sure I understand Mike. So, this is only needed for manual sorting in the treeview? One thing you didn't address is why is this needed and if we can fix the root of the issue, like in https://github.com/artefactual/atom/pull/1566. If this is only needed after a user moves a node in the tree we should not do it on each node load.

apps/qubit/modules/informationobject/actions/fullWidthTreeViewAction.class.php
mcantelon: Good point. I've added a sync check to the first load as well.

lib/QubitLftSyncer.class.php
mcantelon: Fixed!

lib/QubitLftSyncer.class.php
mcantelon: Yeah, we're not sure about the actual root of the issue. Apparently even with #1566 we're still running into the issue so it could be that there's another issue causing Elasticsearch's lft values to "desync" with the values in the database. Or, given that ES updates asynchronously, ES might not be updating quick enough to keep in sync with the DB changes made while a user moves nodes.

It needs to be done before the user moves a node because otherwise the node positions (provided to the move) will be invalid. One thing I'm going to look at is triggering a one-time sync check/repair of a parent when a user hovers over any of a parent's children. Then, theoretically, if a sync repair had to be done then we can tell the treeview to refresh the parent's children.

mcantelon:
I've opened a new PR with the new approach: https://github.com/artefactual/atom/pull/1595

PR #1578, opened 2023-03-31
Fix upgrade task issue for AtoM 2.5 db, ref #13641

anvit:
Update migration 175 to use raw SQL query to update user's password hash only since the ORM tries to save the users in a DB where digital object language column doesn't exist yet. Also add a log statement that will print the migration being currently applied when the verbose option is used, to make it easier to find migration issues.


jraddaoui (approved):
Nice!

PR #1580, opened 2023-04-05
Update google analytics 4 tags, refs #13658

Mbtj:
Update the scripting for Google Analytics to be compatible with GA4 tags (tags starting with 'G', rather than 'UA').


anvit (approved):
Looks good to me!

Issue #1581, opened 2023-04-14
OAI resumption presents duplicate records


rferris:
Hi,

I discovered an issue where records will be sent multiple times over the course of an OAI harvest. It turns out that this is because the sorting of objects is performed only on `object.updated_at` within the function` lib/model/QubitInformationObject.php->getUpdatedRecords.`

A large amount of our data in Atom was imported from csv, and, as such, most of our information objects have the same updated_at timestamp. This means that the sorting of the query is not guaranteed, which leads to oai missing items and duplicating others.

Adding an additional order by on object.id resolves this problem by providing a secondary column to fall back on when updated_at is the same.

```php
public static function getUpdatedRecords($options = array())
  {
    $criteria = new Criteria;
    $criteria->addJoin(QubitInformationObject::ID, QubitStatus::OBJECT_ID);
    $criteria->add(QubitStatus::STATUS_ID, QubitTerm::PUBLICATION_STATUS_PUBLISHED_ID);
    $criteria->addJoin(QubitInformationObject::ID, QubitObject::ID);

    if (!empty($options['from']))
    {
      $criteria->add(QubitObject::UPDATED_AT, $options['from'], Criteria::GREATER_EQUAL);
    }

    if (!empty($options['until']))
    {
      $criteria->addAnd(QubitObject::UPDATED_AT, $options['until'], Criteria::LESS_EQUAL);
    }

    if (!empty($options['set']))
    {
      $options['set']->apply($criteria);
    }

    if (!empty($options['topLevel']))
    {
      $criteria->add(QubitInformationObject::PARENT_ID, QubitInformationObject::ROOT_ID);
    }

    $criteria->addAscendingOrderByColumn(QubitObject::UPDATED_AT);

    // ############## Additional sort
    $criteria->addAscendingOrderByColumn(QubitObject::ID);

    if (empty($options['offset']))
    {
      $options['offset'] = 0;
    }

    if (empty($options['limit']))
    {
      $options['limit'] = 10;
    }

    $c2 = clone $criteria;
    $count = BasePeer::doCount($c2)->fetchColumn(0);
    $remaining = $count - ($options['offset'] + $options['limit']);

    $criteria->setOffset($options['offset']);
    $criteria->setLimit($options['limit']);

    return array(
      'data'      => QubitInformationObject::get($criteria),
      'remaining' => $remaining
    );
  }
```

rferris:
Created pull request https://github.com/artefactual/atom/pull/1582

[...]

PR #1582, opened 2023-04-14
resolve duplicate oai results

rferris:
Resolve duplicate OAI results when updated_at dates are the same. Issue #1581.

### DIFF
--- a/lib/model/QubitInformationObject.php
+++ b/lib/model/QubitInformationObject.php
@@ -437,6 +437,8 @@ class QubitInformationObject extends BaseInformationObject
         }
 
         $criteria->addAscendingOrderByColumn(QubitObject::UPDATED_AT);
+	// Secondarily filter by ID in case UPDATED_AT has duplicate entries.
+        $criteria->addAscendingOrderByColumn(QubitObject::ID);
 
         if (empty($options['offset'])) {
             $options['offset'] = 0;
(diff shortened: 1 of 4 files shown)


lib/model/QubitInformationObject.php
```diff
@@ -437,6 +437,8 @@ public static function getUpdatedRecords($options = [])
         }
 
         $criteria->addAscendingOrderByColumn(QubitObject::UPDATED_AT);
+	// Secondarily filter by ID in case UPDATED_AT has duplicate entries.
```
anvit, line 440: Hi @rferris! Thanks so much for your contribution! I just finished looking at your changes and this definitely makes sense to me, and should fix the issue!

I have a few super quick changes I'd like to request before we can accept this PR —
1. Could you rebase this to the branch `qa/2.x` instead of `stable/2.7`. We use that branch for any bug fixes or features intended for the next release, and as you might've noticed here there's a few extra changes it picked up that we don't need, when we changed the base to `qa/2.x` here on the PR.
2. Line `#440` seems to be using tabs for indentation which is making PHP Coding Standards fixer throw an error in the tests. Could you make sure this uses spaces instead?
3. Small nitpick, but this repo uses uppercase first letter for the commit message.

Let us know if you have the time to make these changes. If not, we're happy to do these and accept this fix. Thanks again!

anvit (changes requested):

lib/model/QubitInformationObject.php
rferris: Hi Anvit,

Can do!

Rohan

anvit: Awesome! Thanks Rohan! Really appreciate it!

rferris: Closed this PR and reissued over at https://github.com/artefactual/atom/pull/1593

PR #1585, opened 2023-04-14
Fix levels option for CSV export, ref #13659

anvit:
CSV information object was not respecting the levels of description options that were selected during export since the options variable used is null. Using params instead fixes this.

### DIFF
--- a/lib/job/arInformationObjectCsvExportJob.class.php
+++ b/lib/job/arInformationObjectCsvExportJob.class.php
@@ -58,7 +58,7 @@ class arInformationObjectCsvExportJob extends arInformationObjectExportJob
 
         // Export descendants if option was selected
         if (!$this->params['current-level-only']) {
-            foreach ($resource->getDescendantsForExport($options) as $item) {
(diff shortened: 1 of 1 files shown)


anvit:
I noticed that the XML export option also uses `$this->params`, https://github.com/artefactual/atom/blob/22dc59d54e0b78bdd354b4de1421b364bab19afe/lib/job/arInformationObjectXmlExportJob.class.php#L77-L81 and `$options` probably crept in during a past refactor since other `exportResource` methods in a couple other classes such as https://github.com/artefactual/atom/blob/22dc59d54e0b78bdd354b4de1421b364bab19afe/lib/job/arActorCsvExportJob.class.php#L49 and https://github.com/artefactual/atom/blob/22dc59d54e0b78bdd354b4de1421b364bab19afe/lib/job/arActorXmlExportJob.class.php#L49 use that instead.

mcantelon (approved):
Code looks good and it worked for me when I tested it!

PR #1586, opened 2023-04-19
Fix missing institution edit button, ref #13662

anvit:
Update the permissions for archival institutions to have the edit button show up even after language is changed. This change mirrors the permissions from qubit/modules/informationobject/templates/_actions.php.

### DIFF
--- a/plugins/arDominionB5Plugin/modules/sfIsdiahPlugin/templates/indexSuccess.php
+++ b/plugins/arDominionB5Plugin/modules/sfIsdiahPlugin/templates/indexSuccess.php
@@ -268,7 +268,7 @@
   <?php slot('after-content'); ?>
 
     <ul class="actions mb-3 nav gap-2">
-      <?php if (QubitAcl::check($resource, 'update')) { ?>
+      <?php if (QubitAcl::check($resource, 'update') || (QubitAcl::check($resource, 'translate'))) { ?>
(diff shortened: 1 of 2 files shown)


anvit:
I noticed that neither these two, nor `qubit/modules/informationobject/templates/_actions.php` had their permissions updated during #741 - [ref #11410](https://projects.artefactual.com/issues/11410). Let me know if it feels that the more accurate thing to do would be to check if the user belongs to the editor group (`sfContext::getInstance()->getUser()->hasGroup(QubitAclGroup::EDITOR_ID)`) instead of checking for checking for the translate permission (`QubitAcl::check($resource, 'translate')`).

jraddaoui (approved):
I think your approach is correct, editor and translator doesn't have to be the same. I think the PR you linked above should be looking to the translator group/permission. I'd suggest to check with an analyst to be sure and normalize both fixes, but this PR looks good to me.

anvit:
Ran this through Anthea and confirmed that this would be the expected behaviour for this fix.

For the fix in the PR linked in my last comment, I think the main difference is that actions like moving or importing/linking physical objects are language agnostic, so just belonging to an editor group would suffice whereas editing Archival Institution info (or Information objects, which this mirrors) in a different language should require the translator privileges as that isn't a language agnostic action.

PR #1590, opened 2023-04-26
Remove markdown issue template

anvit:
Remove the markdown issue template since the form version is more user friendly. Also updated the form template to make the expected behaviour field required, and make the email field optional since GitHub requires signing up to submit issues using the form as well.


anvit:
You can browse the current form template here: https://github.com/artefactual/atom/issues/new?assignees=&labels=Type%3A+bug&template=BUG-REPORT.yml&title=Problem%3A+

and the proposed updated version here: https://github.com/artefactual/atom/blob/dev/issue-template-fix/.github/ISSUE_TEMPLATE/BUG-REPORT.yml

replaceafill (approved):

PR #1591, opened 2023-05-01
Fix static page quicklinks, ref #13663

anvit:
Update Quicklinks menu logic to strip index.php from the start of the url before testing if an action exists for it as Quicklinks would be hidden in AtoM if the config is set to show index.php in the path name.

### DIFF
--- a/lib/model/QubitMenu.php
+++ b/lib/model/QubitMenu.php
@@ -46,9 +46,11 @@ class QubitMenu extends BaseMenu
     }
 
     /**
-     * Wrapper for BaseMenu::getPath() call to allow additional functionality
-     *  option 'resolveAlias' - resolve aliases into full path
-     *  option 'getUrl' - resolve path to internal or external URL.
+     * Wrapper for BaseMenu::getPath() call to allow additional functionality.
+     *
+     * option 'resolveAlias' - resolve aliases into full path
+     * option 'getUrl' - resolve path to internal or external URL
+     * option 'removeIndex' - trim index.php from the start of path name
      *
      * @param array $options Optional parameters
      *
@@ -101,6 +103,11 @@ class QubitMenu extends BaseMenu
             }
 
             $path = $url;
+        }
+
+        if (isset($options['removeIndex']) && true === $options['removeIndex']) {
+            $regexForIndex = '/\A(\/(index|qubit_dev).php)/';
+            $path = preg_replace($regexForIndex, '', $path);
         }
 
         return $path;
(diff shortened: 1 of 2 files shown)


lib/model/QubitMenu.php
```diff
@@ -48,7 +48,8 @@ public function __toString()
     /**
      * Wrapper for BaseMenu::getPath() call to allow additional functionality
      *  option 'resolveAlias' - resolve aliases into full path
-     *  option 'getUrl' - resolve path to internal or external URL.
+     *  option 'getUrl' - resolve path to internal or external URL
+     *  option 'removeIndex' - trim index.php from the start of path name.
```
jraddaoui: Point at the end or not :P? I think it should be the same for all lines.

lib/model/QubitMenu.php
```diff
@@ -103,6 +104,11 @@ public function getPath($options = [])
             $path = $url;
         }
 
+        if (isset($options['removeIndex']) && true === $options['removeIndex']) {
+            $regex_for_index = '/\A(\/index.php)/';
```
jraddaoui: Camel case for variables.

jraddaoui: Not a big deal, but we could consider `qubit_dev.php` here too.

lib/model/QubitMenu.php
```diff
@@ -103,6 +104,11 @@ public function getPath($options = [])
...
+            $path = preg_replace($regex_for_index, '', $path);
+        }
+
```
jraddaoui, line 112: I wonder if it would be better to do this in https://github.com/artefactual/atom/blob/qa/2.x/lib/routing/QubitMetadataRoute.class.php#L42-L213 or if that would have other implications. I wouldn't know why, but maybe sfRoute is not considering the script name in `matchesUrl`.

lib/model/QubitMenu.php
jraddaoui: E.g.: https://github.com/artefactual/atom/blob/qa/2.x/plugins/sfEacPlugin/lib/sfEacPlugin.class.php#L604

lib/model/QubitMenu.php
anvit: Ah, good catch. I'd gotten a CS fixer error for a missing period at the end which is how I'd ended up here, but I've updated this now! There doesn't seem to be a convention for the options array (we don't seem to be documenting this elsewhere) but to me it makes sense to not have the period at the end for these since these are like params. This should be fixed now!

lib/model/QubitMenu.php
anvit: Fixed!

anvit: Done!

lib/model/QubitMenu.php
anvit: Yup, I can confirm that sfRoute isn't considering the script name in it's implementation of `matchesUrl`. I wasn't sure if that would have impact elsewhere so I ended up making the change in the menu to limit which places would trim `index.php` from the URL.

jraddaoui (approved):
Nice, thanks!

PR #1593, opened 2023-05-03
Resolve duplicate OAI results

rferris:
Resolve duplicate OAI results when updated_at dates are the same. Issue https://github.com/artefactual/atom/issues/1581.


anvit (approved):
Looks good to me!

PR #1595, opened 2023-05-09
Sync lft values between DB and ES, refs #13640

mcantelon:
Before providing treeview data from Elasticsearch check for discrepancies between database and Elasticsearch nested set data and, if found, repair.

### DIFF
--- a/apps/qubit/modules/informationobject/actions/fullWidthTreeViewSyncAction.class.php
+++ b/apps/qubit/modules/informationobject/actions/fullWidthTreeViewSyncAction.class.php
@@ -1,0 +1,64 @@
+<?php
+
+/*
+ * This file is part of the Access to Memory (AtoM) software.
+ *
+ * Access to Memory (AtoM) is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * Access to Memory (AtoM) is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with Access to Memory (AtoM).  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+class InformationObjectFullWidthTreeViewSyncAction extends sfAction
+{
+    public function execute($request)
+    {
+        $this->resource = $this->getRoute()->resource;
+        $i18n = sfContext::getInstance()->i18n;
+
+        $this->response->setHttpHeader('Content-Type', 'application/json; charset=utf-8');
+
+        // Check that this isn't the root
+        if (!isset($this->resource->parent)) {
+            $this->response->setStatusCode(404);
+
+            return $this->renderText(json_encode(['error' => $i18n->__('Sync failed: resource not found')]));
+        }
+
+        // Check user authorization
+        if (!$this->getUser()->isAuthenticated()) {
+            $this->response->setStatusCode(401);
+
+            return $this->renderText(json_encode(['error' => $i18n->__('Sync not allowed: log in required')]));
+        }
+
+        // Check user permissions
+        if (!QubitAcl::check($this->resource, 'update')) {
+            $this->response->setStatusCode(403);
+
+            return $this->renderText(json_encode(['error' => $i18n->__('Sync not allowed: not enough permissions')]));
+        }
+
+        // Sync Elasticsearch values with DB if need be
+        $syncer = new QubitLftSyncer($this->resource->id);
+
+        $responseData = ['checked' => true];
+
+        // If a repair was attempted during a check, note whether it was successful
+        if (is_bool($result = $syncer->sync())) {
+            $responseData['repair_successful'] = $result;
+        }
+
+        $this->response->setStatusCode(200);
+
+        return $this->renderText(json_encode($responseData));
+    }
+}
--- a/js/fullWidthTreeView.js
+++ b/js/fullWidthTreeView.js
@@ -39,6 +39,14 @@
     });
   }
 
+  function commandNodeAndChildren($fwTreeView, id, command) {
+    var parent = $fwTreeView.jstree().get_node(id);
+
+    $fwTreeView.jstree(command, id);
+
+    parent.children.forEach((child) => $fwTreeView.jstree(command, child));
+  }
+
   function loadTreeView ()
   {
     var $treeViewConfig = $('#fullwidth-treeview-configuration');
@@ -129,6 +137,11 @@
           },
         },
         'check_callback': function (operation, node, node_parent, node_position, more) {
+          // Do nothing if node is disabled
+          if (node.state.disabled) {
+            return false;
+          }
+
           // Operations allowed:
           // - Before and after drag and drop between siblings
           // - Move core operations (node drop event)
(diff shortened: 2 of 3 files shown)


apps/qubit/modules/informationobject/actions/fullWidthTreeViewMoveAction.class.php
```diff
@@ -63,6 +63,11 @@ public function execute($request)
             return $this->renderText(json_encode(['error' => $i18n->__('Move not needed: new and old positions are the same')]));
         }
 
+        // Sync Elasticsearch values with DB if need be
+        $syncer = new QubitLftSyncer($this->resource->parentId);
+        $syncer->sync();
```
jraddaoui: Could pass the parent id as a parameter and do this inside the move job?

js/fullWidthTreeView.js
```diff
@@ -39,6 +39,18 @@
...
+    var parent = $fwTreeView.jstree().get_node(id);
+
+    $fwTreeView.jstree(command, id);
+
+    var index = 0;
+    while (index < parent.children.length) {
+      $fwTreeView.jstree(command, parent.children[index]);
+      index++;
+    }
+  }
```
jraddaoui: You could loop over `parent.children` directly.

js/fullWidthTreeView.js
```diff
@@ -218,12 +235,39 @@
     // a node is hovered to make it appear after node changes. It must
     // use the #fullwidth-treeview container to allow a higher
     // height than the node in multiple lines tooltips
+    var syncedParents = {};
     var hoverNodeListener = function (e, data)
     {
+      // Do nothing if node is disabled
+      if ($fwTreeView.jstree("is_disabled", data.node.id)) {
+        return;
+      }
```
jraddaoui: We may want show the tooltip even if the node is disabled.

js/fullWidthTreeView.js
```diff
@@ -218,12 +235,39 @@
...
+
+        $.get(url, function (response)
+        {
+          if (response["repaired"]) {
+            $fwTreeView.jstree("refresh_node", parent);
+          }
+
+          commandNodeAndChildren($fwTreeView, parent, "enable_node");
+        });
+      }
```
jraddaoui, line 265: I'd try not to sync on hover as it may happen a lot of times, if possible I'd do it when ready and when the node is selected/loaded.

lib/QubitLftSyncer.class.php
```diff
@@ -0,0 +1,108 @@
...
+                $maxWaitAttempts = 3;
+
+                do {
+                    sleep(1);
+                    ++$waitCount;
+                    $esChecksum = $this->getChildLftChecksumForElasticsearch();
+                } while ($waitCount < $maxWaitAttempts && $esChecksum != $dbChecksum);
+            }
+
+            return true;
```
jraddaoui: This will return true even if the repair could not be verified after the retries.

js/fullWidthTreeView.js
```diff
@@ -218,12 +235,39 @@
...
         delay: 250,
         container: '#fullwidth-treeview'
       });
+
+      // Check sync of treeview data
+      var parent = data.node.parent;
+      var parentNode = $fwTreeView.jstree("get_json", parent);
+
+      if (parent != "#" && !(parent in syncedParents) && ("href" in parentNode.a_attr)) {
+        syncedParents[parent] = true;
```
jraddaoui: I'd set this only on success, when the node is being refreshed, that way it could be retried if the repair didn't work.

apps/qubit/modules/informationobject/actions/fullWidthTreeViewSyncAction.class.php
```diff
@@ -0,0 +1,63 @@
...
+        // Sync Elasticsearch values with DB if need be
+        $syncer = new QubitLftSyncer($this->resource->id);
+
+        $responseData = ['success' => true];
+
+        if ($syncer->sync()) {
+            $responseData['repaired'] = true;
+        }
+
+        $this->response->setStatusCode(201);
```
jraddaoui: Everything has been done and even checked, this should be a 200.

jraddaoui (changes requested):
Thanks @mcantelon! I think this is a lot better approach.

js/fullWidthTreeView.js
mcantelon: There's a hash `syncInitiated`, with node parent ID as the key, that gets set to `true` once a hover has triggered a sync check. Unless it's later set to `false` (or anything other than `true`) a sync check won't happen on subsequent hovers over the children of the same parent.

I considered the expanding of a parent's children as a potential event type to use to trigger the sync check, but in some cases a user will expand a parent's children but not interact with the children (making an immediate sync not necessary). By triggering a sync check only when a user hovers over a node's children there's no delay to get results when opening a node and less sync checks likely to occur.

js/fullWidthTreeView.js
```diff
@@ -39,6 +39,16 @@
...
   }
 
+  function commandNodeAndChildren($fwTreeView, id, command) {
+    var parent = $fwTreeView.jstree().get_node(id);
+
+    $fwTreeView.jstree(command, id);
+
+    for (let i = 0; i < parent.children.length; i++) {
+      $fwTreeView.jstree(command, parent.children[i]);
+    }
```
jraddaoui: Do you need to access the children with the index? Would one of these examples without index work the same?

```js
parent.children.forEach((child) => $fwTreeView.jstree(command, child));
```

```js
for (child of parent.children) {
  $fwTreeView.jstree(command, child);
}
```

apps/qubit/modules/informationobject/actions/fullWidthTreeViewMoveAction.class.php
jraddaoui: Can we do this? Other than that it's looking great, thanks for addressing the feedback. Let me know and I'll approve.

jraddaoui:
Syntax checks are complaining about lib/QubitLftSyncer.class.php. Maybe it's just that it doesn't like the `++` position in `$waitCount++;`

mcantelon:
@Radda Should be fixed now! Thanks!

js/fullWidthTreeView.js
mcantelon: Ah, good call! I'll implement that.

mcantelon:
@jraddaoui Oops, one more fix I need to make.

apps/qubit/modules/informationobject/actions/fullWidthTreeViewMoveAction.class.php
mcantelon: I've removed that as I don't think it's actually needed.

js/fullWidthTreeView.js
```diff
@@ -218,12 +231,38 @@
...
       $('a.jstree-anchor').tooltip({
         delay: 250,
         container: '#fullwidth-treeview'
       });
+
+      // Check sync of parent's children (if one hasn't yet been started)
+      var parent = data.node.parent;
+      var parentNode = $fwTreeView.jstree("get_json", parent);
+
+      if (parent != "#" && !(parent in syncInitiated) && ("href" in parentNode.a_attr)) {
```
jraddaoui, line 247: This if should also consider `!syncInitiated[parent]`, right?

js/fullWidthTreeView.js
```diff
@@ -218,12 +231,38 @@
...
+        var url = parentNode.a_attr.href + '/informationobject/fullWidthTreeViewSync';
+
+        $.get(url, function (response)
+        {
+          if (response["repair_successful"] === true) {
+            // Refresh parent's child nodes if a repair was needed and was successful
+            $fwTreeView.jstree("refresh_node", parent);
+          } else if (response["repair_successful"] === false) {
+            // Allow for syncing to be attempted again if a repair was needed, but failed
+            syncInitiated[parent] = false;
```
jraddaoui: Or instead of setting it to false here, remove it entirely from `syncInitiated`.

jraddaoui (commented):
Just a couple of comments to make sure it's retried on failure.

js/fullWidthTreeView.js
mcantelon: Good catch! I've changed it to delete the value rather than set it to false.

mcantelon:
@jraddaoui I've updated the PR... thanks!

jraddaoui (approved):
Thanks @mcantelon!

Issue #1596, opened 2023-05-11
Problem: RAD template fields don't appear when switching languages


anvit:
### Current Behavior

**Steps to reproduce the behavior**
1. Set the default `Archival Description` template to RAD under `Admin>Settings>Default template`.
2. Open or create an Archival Description which has "Other title information" and "Title statement of responsibility" fields
3. Switch to another language
4. Notice that the "Other title information" and "Title statement of responsibility" fields have disappeared.

### Expected Behavior

Those fields should still show up when the language is changed

### Version used

AtoM 2.6.4

### Operating System and version

Ubuntu 20.04

### Default installation culture

en

anvit:
This seems to affect pretty much every RAD specific field except `Reference Code`. Also noticed that this affects the `Technical Access` DACS field

anvit:
The problem seems to be that the `cultureFallback` argument (or pretty much the entire `$options` array) is completely ignored in [sfRadPlugin Class](https://github.com/artefactual/atom/blob/8b2d1d44a679745713deaaf3bc6c0128e6fbcc66/plugins/sfRadPlugin/lib/sfRadPlugin.class.php#L53)

PR #1597, opened 2023-05-11
Fix missing rad and dacs fields, ref #1596

anvit:
Changing the overridden `__get` magic method to a new function definition that expects an `$options` array as a second parameter for both RAD and DACS classes since the `__get` method does not expect additional arguments and will ignore them unless the function call passing them is from inside the class itself.

### DIFF
--- a/lib/helper/QubitHelper.php
+++ b/lib/helper/QubitHelper.php
@@ -44,8 +44,13 @@ function render_field($field, $resource = null, array $options = [])
         }
 
         try {
-            $source = $resourceRaw->__get($options['name'], ['sourceCulture' => true]);
-            $fallback = $resourceRaw->__get($options['name']);
+            if ($resourceRaw instanceof sfRadPlugin || $resourceRaw instanceof arDacsPlugin) {
+                $source = $resourceRaw->getProperty($options['name'], ['sourceCulture' => true]);
+                $fallback = $resourceRaw->getProperty($options['name']);
+            } else {
+                $source = $resourceRaw->__get($options['name'], ['sourceCulture' => true]);
+                $fallback = $resourceRaw->__get($options['name']);
+            }
         } catch (Exception $e) {
             if ('Unknown record property' !== substr($e->getMessage(), 0, 23)) {
                 throw $e;
--- a/plugins/sfRadPlugin/lib/sfRadPlugin.class.php
+++ b/plugins/sfRadPlugin/lib/sfRadPlugin.class.php
@@ -23,7 +23,7 @@
  *
  * @author     Peter Van Garderen <peter@artefactual.com>
  */
-class sfRadPlugin implements ArrayAccess
+class sfRadPlugin
 {
     protected $resource;
     protected $property;
@@ -50,15 +50,8 @@ class sfRadPlugin implements ArrayAccess
         return $string;
     }
 
-    public function __get($name)
+    public function getProperty($name, $options = [])
     {
-        $args = func_get_args();
-
-        $options = [];
-        if (1 < count($args)) {
-            $options = $args[1];
-        }
-
         switch ($name) {
             case 'editionStatementOfResponsibility':
             case 'issuingJurisdictionAndDenomination':
@@ -88,7 +81,7 @@ class sfRadPlugin implements ArrayAccess
         }
     }
 
-    public function __set($name, $value)
+    public function setProperty($name, $value)
     {
         switch ($name) {
             case 'editionStatementOfResponsibility':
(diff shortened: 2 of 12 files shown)


anvit:
The diff for [arDacsPlugin.class.php](https://github.com/artefactual/atom/blob/9fb8250196693bc6570bb7fe4de61ee7f6f69b29/plugins/arDacsPlugin/lib/arDacsPlugin.class.php#L40) isn't easy to read but the change is the same as the one for `sfRadPlugin.class.php` — changing the function name to a different name than `__get` and updating its argument list to include an options array.

jraddaoui:
Hi @anvit,

While I like this approach and I don't like those magic methods (specially the way they are being used) I think this is breaking the edit page (and maybe others) where we still access those properties using that magic method. The main case is in [render_field](https://github.com/artefactual/atom/blob/qa/2.x/lib/helper/QubitHelper.php#L33-L87), which receives the RAD resource. But there are others like [this](https://github.com/artefactual/atom/blob/qa/2.x/plugins/sfRadPlugin/modules/sfRadPlugin/templates/editSuccess.php#L87) or [this one](https://github.com/artefactual/atom/blob/qa/2.x/plugins/sfRadPlugin/modules/sfRadPlugin/actions/editAction.class.php#L155) using array access.

I think this issue was introduced quite a lot of time ago when we changed the escaping strategy and that it may be solved unescaping the RAD resource in the template (like it's done in the `render_field` function). However, that may cause the content to be unescaped and may require to be escaped again, I always get confused with how this works internally. If you go this way, make sure to check the content is escaped with the Markdown setting enabled and disabled, as it changes the escaping behavior.

Again, I'd like to see your approach, also with the `__set` magic method, and stop using array access in those classes, but that may be a lot of work.

anvit:
@jraddaoui I spent some time doing some refactoring at all of the places that were using array access earlier. Let me know if this makes sense, or if there is anything else I might've missed? (I looked through the code for instances of `$this->rad`,`$rad` and the `$dacs` equivalents).

Alternatively, we could go with just unescaping the RAD/DACS resource as that does indeed work. I have that change up in another branch (`dev/issue-1596-missing-rad-fields-unescape`). Here's a [diff of those changes](https://github.com/artefactual/atom/compare/qa/2.x...dev/issue-1596-missing-rad-fields-unescape).

jraddaoui:
Thanks @anvit!

I don't trust that unescaping, even if it works I'm not sure it would break the escaping strategy entirely for those fields. And I like the removal of the magic methods, when I said it could be at lot of work I meant doing it in all places, but you already started here. I'll add a couple of suggestions, but it looks good to me.

lib/helper/QubitHelper.php
```diff
@@ -44,8 +44,16 @@ function render_field($field, $resource = null, array $options = [])
...
+            if ($resourceRaw instanceof sfRadPlugin) {
+                $source = $resourceRaw->getRADProperty($options['name'], ['sourceCulture' => true]);
+                $fallback = $resourceRaw->getRADProperty($options['name']);
+            } elseif ($resourceRaw instanceof arDacsPlugin) {
+                $source = $resourceRaw->getDACSProperty($options['name'], ['sourceCulture' => true]);
+                $fallback = $resourceRaw->getDACSProperty($options['name']);
+            } else {
+                $source = $resourceRaw->__get($options['name'], ['sourceCulture' => true]);
+                $fallback = $resourceRaw->__get($options['name']);
+            }
```
jraddaoui: Can we have the same function name in both classes? E.g.: `getProperty`. That will reduce the amount of conditions here, specially if the same pattern is implemented in other resources.

plugins/sfRadPlugin/lib/sfRadPlugin.class.php
```diff
@@ -50,45 +50,7 @@ public function __toString()
         return $string;
     }
 
-    public function __get($name)
```
jraddaoui, line 53: Can we remove the `ArrayAccess` implementation in this class after this changes?

jraddaoui (approved):

lib/helper/QubitHelper.php
anvit: Done!

plugins/sfRadPlugin/lib/sfRadPlugin.class.php
anvit: Done, removed the offset methods and the `ArrayAccess` implementation. We don't seem to be calling the offset methods anywhere and this doesn't seem to have affected the index or the edit pages.

jraddaoui:
Thanks @anvit!

Issue #1599, opened 2023-05-23
Problem: Access control for contact info doesn't check edit/admin access


anvit:
### Current Behavior

**Steps to reproduce the behavior**
1. Log into the the demo site with a non editor or admin user.
2. Try to access `http://demo.accesstomemory.org/contactinformation/index/id/185`.
3. You should be able to see the page even as regular authenticated user or a translator.

### Expected Behavior

You should not be able to see the page unless you have edit access.

### Possible Solution

Updating the permissions in `apps/qubit/modules/contactinformation/actions/indexAction.class.php` should fix this.

### Version used

AtoM 2.6.4

### Default installation culture

en

PR #1600, opened 2023-05-23
Update contact information access, ref #1599

anvit:
Update user authorization condition check to also confirm that the authenticated user is an editor or admin.

### DIFF
--- a/apps/qubit/modules/contactinformation/actions/indexAction.class.php
+++ b/apps/qubit/modules/contactinformation/actions/indexAction.class.php
@@ -21,8 +21,10 @@ class ContactInformationIndexAction extends sfAction
 {
     public function execute($request)
     {
+        $hasEditAccess = QubitAcl::check($this->resource, 'update')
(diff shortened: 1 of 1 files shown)


jraddaoui (approved):

sbreker:
Looks great @anvit ! 

One question: Should similar logic be added to the edit action? Wondering if it has the same problem?

jraddaoui:
I thought the same @sbreker, but there is no edit action, just a component.

anvit:
> Looks great @anvit !
> 
> One question: Should similar logic be added to the edit action? Wondering if it has the same problem?

@sbreker Contact info editing seemed to be through edit actions for donor, rights holder, and repository (there is an `editComponent` for contact info which is used in edit actions for those. Also trying to make `PUT` or `POST` requests to `/contactinformation/edit/id/185` just got me 404). They all have similar user authorization checks, so I think that should be covered! (Unless I'm missing something?)

sbreker (approved):

PR #1617, opened 2023-06-22
Add treeview sync code to the BS5 theme (#1616)

### LINKED ISSUE #1616: Problem: Tree view bug still persists in bootstrap 5 theme
### Current Behavior

Moving or editing records in the treeview used to sometimes result in a "Mismatch in current positions" error (https://projects.artefactual.com/issues/13640). This was fixed in the last release, but the problem seems to still persist in bootstrap 5 themes.

### Expected Behavior

When using a BS5 theme, if the user hovers over records in treeview that have not been synced, it should momentarily lock as a sync is performed, and then unlock (similar to the current behaviour in BS2 themes).

### Possible Solution

This is likely due to the fix being applied only to the `js/fullWidthTreeView.js` and not to `plugins/arDominionB5Plugin/js/fullWidthTreeView.js`

### Version used

AtoM 2.7.2

anvit:
Add the treeview sync fix on hover javascript code to the Bootstrap 5 dominion theme.

### DIFF
--- a/plugins/arDominionB5Plugin/js/fullWidthTreeView.js
+++ b/plugins/arDominionB5Plugin/js/fullWidthTreeView.js
@@ -34,6 +34,9 @@
       this.addButtonSection();
       this.$accordionWrapper.after(this.$buttonSection);
 
+      // Keep track of which syncs have been initiated
+      this.syncInitiated = {};
+
       // Declare jsTree options
       this.options = {
         plugins: ["types", "dnd"],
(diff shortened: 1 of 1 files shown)


plugins/arDominionB5Plugin/js/fullWidthTreeView.js
```diff
@@ -368,6 +372,42 @@
         this.showAlert(moveResponse.success, "alert-info");
       }
     };
+
+    hoverNodeListener = (e, data) => {
+      var parent = data.node.parent;
+      var parentNode = this.$fwTreeView.jstree("get_json", parent);
```
jraddaoui: Not a big deal in this case, but we should start using `let` instead of `var` more often.

jraddaoui (approved):
Thanks!

plugins/arDominionB5Plugin/js/fullWidthTreeView.js
anvit: Updated! And yes, all in favour of going for `const` and `let` instead of `var` moving forward.

PR #1619, opened 2023-06-23
Add a CLI rename slug tool (#1609)

### LINKED ISSUE #1609: Feature: Command line tool for renaming slugs
Having a command line tool to rename slugs (say something like `php symfony tools:rename-slug  old_slug_name new_slug_name`)
can speed up things for users who want to batch rename them using a shell script, and don't want to manually update them one at a time using the UI or run SQL queries.

anvit:
Add a CLI task which lets users update slugs. Slugs can be provided in a CSV file with the --csv option, or be provided as command line arguments.

### DIFF
--- a/lib/task/tools/renameSlugTask.class.php
+++ b/lib/task/tools/renameSlugTask.class.php
@@ -1,0 +1,107 @@
+<?php
+
+/*
+ * This file is part of the Access to Memory (AtoM) software.
+ *
+ * Access to Memory (AtoM) is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * Access to Memory (AtoM) is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with Access to Memory (AtoM).  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+class renameSlugTask extends arBaseTask
+{
+    protected function configure()
+    {
+        $this->addArguments([
+            new sfCommandArgument('oldSlug', sfCommandArgument::OPTIONAL, 'The slug to update'),
+            new sfCommandArgument('newSlug', sfCommandArgument::OPTIONAL, 'Updated slug text'),
+        ]);
+
+        $this->addOptions([
+            new sfCommandOption('application', null, sfCommandOption::PARAMETER_OPTIONAL, 'The application name', true),
+            new sfCommandOption('env', null, sfCommandOption::PARAMETER_REQUIRED, 'The environment', 'cli'),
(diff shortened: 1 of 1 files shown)


lib/task/tools/renameSlugTask.class.php
```diff
@@ -0,0 +1,108 @@
...
+ * Access to Memory (AtoM) is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with Access to Memory (AtoM).  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+class renameSlugTask extends sfBaseTask
```
mcantelon: Should extend arBaseTask if possible.

lib/task/tools/renameSlugTask.class.php
```diff
@@ -0,0 +1,108 @@
...
+{
+    protected function configure()
+    {
+        $this->addArguments([
+            new sfCommandArgument('oldSlug', sfCommandArgument::OPTIONAL, 'The slug to update'),
+            new sfCommandArgument('newSlug', sfCommandArgument::OPTIONAL, 'Update name for the slug'),
```
mcantelon: Nitpick: should "slug text" be used rather than "name"? The text "Update name for the slug" could be changed to "Updated slug text". I could be wrong though. ;)

lib/task/tools/renameSlugTask.class.php
```diff
@@ -0,0 +1,108 @@
...
+    {
+        $databaseManager = new sfDatabaseManager($this->configuration);
+        $conn = $databaseManager->getDatabase('propel')->getConnection();
+
+        if ($arguments['oldSlug'] && $arguments['newSlug']) {
+            $this->renameSlug($arguments['oldSlug'], $arguments['newSlug']);
+        } elseif ($options['csv']) {
+            $this->updateSlugsFromCSV($options['csv']);
+        } else {
+            throw new Exception('Either provide old and new slug values, or use the csv option and supply a csv file containing those values.');
```
mcantelon: "csv" should be capitalized.

lib/task/tools/renameSlugTask.class.php
```diff
@@ -0,0 +1,108 @@
...
+To do a batch update via CSV, run:
+    php symfony tools:rename --csv=file/path/to/csv-file.csv
+
+The supplied CSV file must contain an oldSlug and a newSlug column.
+EOF;
+    }
+
+    protected function execute($arguments = [], $options = [])
+    {
+        $databaseManager = new sfDatabaseManager($this->configuration);
```
mcantelon: I don't think these lines are needed:

```
$databaseManager = new sfDatabaseManager($this->configuration);
$conn = $databaseManager->getDatabase('propel')->getConnection();
```

mcantelon (changes requested):
A few minor changes needed, but seems good overall!

lib/task/tools/renameSlugTask.class.php
anvit: Yep, that does sound better (and I'm realizing that a typo had crept in, It was supposed to say updated).

lib/task/tools/renameSlugTask.class.php
anvit: Done!

mcantelon (approved):
Looks good to me!

PR #1622, opened 2023-06-29
Restore missing french translations

anvit:
Restoring french translations that were dropped during the 2.7.0 release. Running the weblate import task wipes these translations so this is only meant as a temporary fix for the release.


Mbtj (approved):
Looks good to me!

Issue #1624, opened 2023-07-05
Feature: Allow changing of an information object's publication status via the REST API


mcantelon:
Currently a REST API user can't change a description's publication status using the information object update endpoint (`plugins/arRestApiPlugin/modules/api/actions/informationobjectsUpdateAction.class.php`).

We could support these update request conditions to allow publication status to be changed:

1) "published" isn't set in the payload: no change

2) "published" is set to true: set to published

3) "published" is set to false: set to draft

PR #1625, opened 2023-07-05
Add IO pub. status update via REST API. (#1624)

mcantelon:
Added functionality to the information object update REST API endpoint to allow an information object's publication status to be changed.


mcantelon:
(I did the conditional as a ternary originally - `$publicationStatus = ($value) ? "Published" : "Draft";` - but PHP CS Fixer seemed to choke on that, exiting with code 8 "aka Some files need fixing", when checking, but not actually changing the file, when fixing so I just made it a simple conditional.)

mcantelon:
Something's weird. I force push but the commits stay the same. Closing then will try to reopen.

PR #1626, opened 2023-07-05
Add IO pub. status update via REST API. (#1624)

### LINKED ISSUE #1624: Feature: Allow changing of an information object's publication status via the REST API

mcantelon:
Added functionality to the information object update REST API endpoint to allow an information object's publication status to be changed.

### DIFF
--- a/plugins/arRestApiPlugin/modules/api/actions/informationobjectsUpdateAction.class.php
+++ b/plugins/arRestApiPlugin/modules/api/actions/informationobjectsUpdateAction.class.php
@@ -240,6 +240,16 @@ class ApiInformationObjectsUpdateAction extends QubitApiAction
                 }
 
                 break;
+
+            case 'published':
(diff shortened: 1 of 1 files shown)


mcantelon:
(I did the conditional as a ternary originally - `$publicationStatus = ($value) ? "Published" : "Draft";` - but PHP CS Fixer seemed to choke on that, exiting with code 8 aka "Some files need fixing", when checking, but not actually changing the file, when fixing so I just made it a simple conditional.)

anvit (approved):
Looks good!

PR #1631, opened 2023-07-08
Fix D.O. create REST endpoint ACL check. (#1630)

### LINKED ISSUE #1630: Problem: an ACL check in the REST API's digital object create endpoint is malformed
### Current Behavior

When attempted to create a digital object via the digital object create REST API endpoint you'll get an error that shows up like this in Nginx's error log which is caused by a malformed ACL check:

```[error] 1085519#1085519: *203892 FastCGI sent in stderr: "PHP message: PHP Fatal error:  Uncaught Error: Call to undefined function QubitAcl() in /usr/share/nginx/atomplugins/arRestApiPlugin/modules/api/actions/physicalobjectsCreateAction.class.php:112```

### Expected Behavior

The ACL check shouldn't generate an error.

### Possible Solution

_No response_

### Context and Notes

_No response_

### Version used

2.8.0 v193

Ubuntu 20.04

_No response_

### PHP version

7.4.3

### Contact details

_No response_

mcantelon:
Fixed malformed ACL check in REST API's digital object create endpoint.

### DIFF
--- a/plugins/arRestApiPlugin/modules/api/actions/digitalobjectsCreateAction.class.php
+++ b/plugins/arRestApiPlugin/modules/api/actions/digitalobjectsCreateAction.class.php
@@ -102,7 +102,7 @@ class ApiDigitalObjectsCreateAction extends QubitApiAction
 
         // Check that user has permission to update the parent digital object's
         // "object" (information object or actor)
-        if (!QubitAcl($parent->object, 'update')) {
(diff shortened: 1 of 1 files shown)


anvit:
Looks good to me!

anvit (approved):

mcantelon:
Thanks @anvit !

PR #1636, opened 2023-07-12
Add diacritics setting

### LINKED ISSUE #1684: Add Diacritics setting funtionality
New setting for users to enable further char_filter mappings for diacritics. Requires a yaml file to enable the setting, example attached to this gist for reference: https://gist.github.com/melaniekung/6014e100767019b1f85637e5a7cdf9df.

melaniekung:
New setting for users to enable further char_filter mappings for diacritics, assigned to the default, autocomplete, english and french analyzers. Requires a yaml file to enable the setting, example attached to this gist for reference: https://gist.github.com/melaniekung/6014e100767019b1f85637e5a7cdf9df. Commit includes an empty diacritics_mapping.yml (default) which will be replaced with the submitted yaml to map out the required diacritics.

### DIFF
--- a/apps/qubit/modules/settings/actions/diacriticsAction.class.php
+++ b/apps/qubit/modules/settings/actions/diacriticsAction.class.php
@@ -1,0 +1,103 @@
+<?php
+/*
+ * This file is part of the Access to Memory (AtoM) software.
+ *
+ * Access to Memory (AtoM) is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * Access to Memory (AtoM) is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with Access to Memory (AtoM).  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+class SettingsDiacriticsAction extends SettingsEditAction
+{
+    // Arrays not allowed in class constants
+    public static $NAMES = [
+        'diacritics',
+        'mappings',
+    ];
+
+    public function earlyExecute()
+    {
+        parent::earlyExecute();
+
+        $this->updateMessage = $this->i18n->__('Diacritics settings saved.');
+
+        $this->settingDefaults = [
+            'diacritics' => 0,
+        ];
+    }
+
+    public function processForm()
+    {
+        foreach ($this->form as $field) {
+            $this->processField($field);
+        }
+    }
+
+    protected function addField($name)
+    {
+        switch ($name) {
+            case 'diacritics':
+                $this->form->setDefault($name, $this->settingDefaults[$name]);
+                $this->form->setWidget($name, new sfWidgetFormSelectRadio(['choices' => [0 => $this->i18n->__('Disabled'), 1 => $this->i18n->__('Enabled')]], ['class' => 'radio']));
+                $this->form->setValidator($name, new sfValidatorChoice(['choices' => [1, 0]]));
+
+                break;
+
+            case 'mappings':
+                $this->form->setWidget($name, new sfWidgetFormInputFile([], ['accept' => '.yml,.yaml']));
+                $this->form->setValidator($name, new sfValidatorFile(['mime_types' => ['text/plain']]));
+
+                break;
+        }
+    }
+
+    protected function processField($field)
+    {
+        switch ($name = $field->getName()) {
+            case 'diacritics':
+                parent::processField($field);
+
+                break;
+
+            case 'mappings':
+                $file = $this->form->getValue('mappings');
+
+                $diacriticsMappingPath = sfConfig::get('sf_config_dir').DIRECTORY_SEPARATOR.'diacritics_mapping.yml';
+
+                if (null !== $file) {
+                    try {
+                        sfYaml::load($file->getTempName());
+
+                        if (!move_uploaded_file($file->getTempName(), $diacriticsMappingPath)) {
+                            $this->getUser()->setFlash('error', $this->context->i18n->__('Unable to upload diacritics mapping yaml file.'));
+                            unset($this->updateMessage);
+
+                            return;
+                        }
+                    } catch (Exception $e) {
+                        QubitSetting::findAndSave('diacritics', 0, ['sourceCulture' => true]);
+                        unlink($diacriticsMappingPath);
+                        $this->getUser()->setFlash('error', $this->context->i18n->__('Unable to upload diacritis mapping yaml file.'));
+                        unset($this->updateMessage);
+                    }
+                } else {
+                    // Reset diacritics settings when uploading yaml fails
+                    QubitSetting::findAndSave('diacritics', 0, ['sourceCulture' => true]);
+                    unlink($diacriticsMappingPath);
+                    $this->getUser()->setFlash('error', $this->context->i18n->__('Unable to upload diacritis mapping yaml file.'));
+                    unset($this->updateMessage);
+                }
+
+                break;
+        }
+    }
+}
--- a/apps/qubit/modules/settings/actions/editAction.class.php
+++ b/apps/qubit/modules/settings/actions/editAction.class.php
@@ -1,5 +1,4 @@
 <?php
-
 /*
  * This file is part of the Access to Memory (AtoM) software.
  *
@@ -28,7 +27,7 @@ class SettingsEditAction extends DefaultEditAction
 
         // Handle posted data
         if ($request->isMethod('post')) {
-            $this->form->bind($request->getPostParameters());
+            $this->form->bind($request->getPostParameters(), $request->getFiles());
 
             if ($this->form->isValid()) {
                 $this->processForm();
--- a/apps/qubit/modules/settings/templates/diacriticsSuccess.php
+++ b/apps/qubit/modules/settings/templates/diacriticsSuccess.php
@@ -1,0 +1,66 @@
+<?php decorate_with('layout_2col.php'); ?>
+
+<?php slot('sidebar'); ?>
+
+<?php echo get_component('settings', 'menu'); ?>
+
+<?php end_slot(); ?>
+
+<?php slot('title'); ?>
+
+<h1>
+  <?php echo __('Diacritics'); ?>
+</h1>
+
+<?php end_slot(); ?>
+
+<?php slot('content'); ?>
+<div class="alert alert-info">
+  <p>
+    <?php echo __('Please rebuild the search index after uploading diacritics mappings.'); ?>
+  </p>
+  <pre>$ php symfony search:populate</pre>
+</div>
+
+<div class="alert alert-info">
+  <p>
+    <?php echo __('Example CSV:'); ?>
+  </p>
+  <pre>type: mapping<br/>mappings:<br/>  - À => A<br/>  - Á => A</pre>
+</div>
+
+<?php echo $form->renderGlobalErrors(); ?>
+
+<?php echo $form->renderFormTag(url_for(['module' => 'settings', 'action' => 'diacritics'])); ?>
+
+<?php echo $form->renderHiddenFields(); ?>
+
+<div id="content">
+
+  <fieldset class="collapsible">
+    <legend>
+      <?php echo __('Diacritics settings'); ?>
+    </legend>
+
+    <?php echo $form->diacritics->label(__('Diacritics'))->renderRow(); ?>
+  </fieldset>
+
+  <fieldset class="collapsible">
+    <legend>
+      <?php echo __('CSV Mapping YAML'); ?>
+    </legend>
+
+    <?php echo $form->mappings->label(__('Mappings YAML'))->renderRow(); ?>
+  </fieldset>
+
+</div>
+
+<section class="actions">
+  <ul>
+    <li><input class="c-btn c-btn-submit" type="submit" value="<?php echo __('Save'); ?>" /></li>
+  </ul>
+</section>
+
+</form>
+
+<?php end_slot(); ?>
--- a/plugins/arDominionB5Plugin/modules/settings/templates/diacriticsSuccess.php
+++ b/plugins/arDominionB5Plugin/modules/settings/templates/diacriticsSuccess.php
@@ -1,0 +1,74 @@
+<?php decorate_with('layout_2col.php'); ?>
+
+<?php slot('sidebar'); ?>
+
+<?php echo get_component('settings', 'menu'); ?>
+
+<?php end_slot(); ?>
+
+<?php slot('title'); ?>
+<h1>
+  <?php echo __('Diacritics settings'); ?>
+</h1>
+<?php end_slot(); ?>
+
+<?php slot('content'); ?>
+
+<div class="alert alert-info">
+  <p>
+    <?php echo __('Please rebuild the search index after uploading diacritics mappings.'); ?>
+  </p>
+  <pre>$ php symfony search:populate</pre>
+</div>
+
+<?php echo $form->renderGlobalErrors(); ?>
+
+<?php echo $form->renderFormTag(url_for(['module' => 'settings', 'action' => 'diacritics'])); ?>
+
+<?php echo $form->renderHiddenFields(); ?>
+
+<div class="accordion mb-3">
+  <div class="accordion-item">
+    <h2 class="accordion-header" id="diacritics-heading">
+      <button class="accordion-button" type="button" data-bs-toggle="collapse" data-bs-target="#diacritics-collapse"
+        aria-expanded="true" aria-controls="diacritics-collapse">
+        <?php echo __('Diacritics Settings'); ?>
+      </button>
+    </h2>
+    <div id="diacritics-collapse" class="accordion-collapse collapse show" aria-labelledby="diacritics-heading">
+      <div class="accordion-body">
+        <?php echo render_field($form->diacritics->label(__('Diacritics'))); ?>
+      </div>
+    </div>
+  </div>
+  <div class="accordion-item">
+    <h2 class="accordion-header" id="mappings-heading">
+      <button class="accordion-button" type="button" data-bs-toggle="collapse" data-bs-target="#mappings-collapse"
+        aria-expanded="false" aria-controls="mappings-collapse">
+        <?php echo __('CSV Mapping YAML'); ?>
+      </button>
+    </h2>
+
+    <div id="mappings-collapse" class="accordion-collapse collapse show" aria-labelledby="sending-heading">
+
+      <div class="alert alert-info m-3 mb-0">
+        <p>
+          <?php echo __('Example CSV:'); ?>
+        </p>
+        <pre>type: mapping<br/>mappings:<br/>  - À => A<br/>  - Á => A</pre>
+      </div>
+
+      <div class="accordion-body">
+        <?php echo render_field($form->mappings->label(__('Mappings YAML'))); ?>
+      </div>
+    </div>
+  </div>
+</div>
+
+<section class="actions">
+  <input class="btn atom-btn-outline-success" type="submit" value="<?php echo __('Save'); ?>" />
+</section>
+
+</form>
+
+<?php end_slot(); ?>
--- a/plugins/arElasticSearchPlugin/lib/arElasticSearchPlugin.class.php
+++ b/plugins/arElasticSearchPlugin/lib/arElasticSearchPlugin.class.php
@@ -119,10 +119,12 @@ class arElasticSearchPlugin extends QubitSearchEngine
     {
         // Find mapping.yml
         $finder = sfFinder::type('file')->name('mapping.yml');
-        $files = array_unique(array_merge(
-            $finder->in(sfConfig::get('sf_config_dir')),
-            $finder->in(ProjectConfiguration::getActive()->getPluginSubPaths('/config'))
-        ));
+        $files = array_unique(
+            array_merge(
+                $finder->in(sfConfig::get('sf_config_dir')),
+                $finder->in(ProjectConfiguration::getActive()->getPluginSubPaths('/config'))
+            )
+        );
 
         if (!count($files)) {
             throw new sfException('You must create a mapping.xml file.');
@@ -133,6 +135,24 @@ class arElasticSearchPlugin extends QubitSearchEngine
         $esMapping->loadYAML(array_shift($files));
 
         return $esMapping;
+    }
+
+    public function loadDiacriticsMappings()
+    {
+        // Find diacritics_mapping.yml
+        $diacriticsFinder = sfFinder::type('file')->name('diacritics_mapping.yml');
+        $diacriticsFiles = array_unique(
+            array_merge(
+                $diacriticsFinder->in(sfConfig::get('sf_config_dir')),
+                $diacriticsFinder->in(ProjectConfiguration::getActive()->getPluginSubPaths('/config'))
+            )
+        );
+
+        if (!count($diacriticsFiles)) {
+            throw new sfException('You must create a diacritics_mapping.yml file.');
+        }
+
+        return sfYaml::load(array_shift($diacriticsFiles));
     }
 
     /**
@@ -272,10 +292,12 @@ class arElasticSearchPlugin extends QubitSearchEngine
             }
         }
 
-        $this->log(vsprintf(
-            'Index populated with %s documents in %s seconds.',
-            [$total, $timer->elapsed()]
-        ));
+        $this->log(
+            vsprintf(
+                'Index populated with %s documents in %s seconds.',
+                [$total, $timer->elapsed()]
+            )
+        );
 
         if (!$showErrors) {
             return;
@@ -468,6 +490,10 @@ class arElasticSearchPlugin extends QubitSearchEngine
      */
     protected function initialize()
     {
+        if (sfConfig::get('app_diacritics')) {
+            $this->config['index']['configuration']['analysis']['char_filter']['diacritics_lowercase'] = $this->loadDiacriticsMappings();
+        }
+
         try {
             $this->index->open();
         } catch (Exception $e) {
@@ -479,7 +505,17 @@ class arElasticSearchPlugin extends QubitSearchEngine
                     && isset($this->config['index']['configuration']['analysis']['char_filter']['strip_md'])
                 ) {
                     foreach ($this->config['index']['configuration']['analysis']['analyzer'] as $key => $analyzer) {
-                        $this->config['index']['configuration']['analysis']['analyzer'][$key]['char_filter'] = ['strip_md'];
+                        $filters = ['strip_md'];
+
+                        if ($this->config['index']['configuration']['analysis']['analyzer'][$key]['char_filter']) {
+                            $filters = array_merge($filters, $this->config['index']['configuration']['analysis']['analyzer'][$key]['char_filter']);
+                        }
+
+                        if (sfConfig::get('app_diacritics')) {
+                            $filters = array_merge($filters, ['diacritics_lowercase']);
+                        }
+
+                        $this->config['index']['configuration']['analysis']['analyzer'][$key]['char_filter'] = $filters;
                     }
                 }
 
--- a/.gitignore
+++ b/.gitignore
@@ -8,6 +8,7 @@
 /config/config.php
 /config/databases.yml
 /config/propel.ini
+/config/diacritics_mapping.yml
 
 # Internal use
 /cache/
--- a/apps/qubit/modules/settings/actions/menuComponent.class.php
+++ b/apps/qubit/modules/settings/actions/menuComponent.class.php
@@ -40,6 +40,10 @@ class SettingsMenuComponent extends sfComponent
                 'action' => 'template',
             ],
             [
+                'label' => $i18n->__('Diacritics'),
+                'action' => 'diacritics',
+            ],
+            [
                 'label' => $i18n->__('Digital object derivatives'),
                 'action' => 'digitalObjectDerivatives',
             ],


plugins/arElasticSearchPlugin/lib/arElasticSearchPlugin.class.php
```diff
@@ -132,6 +140,22 @@ public static function loadMappings()
...
+        if (sfConfig::get('app_diacritics')) {
+            // Find diacritics_mapping.yml
+            $diacriticsFinder = sfFinder::type('file')->name('diacritics_mapping.yml');
+            $diacriticsFiles = array_unique(array_merge(
+                $diacriticsFinder->in(sfConfig::get('sf_config_dir')),
+                $diacriticsFinder->in(ProjectConfiguration::getActive()->getPluginSubPaths('/config'))
+            ));
+
+            if (!count($diacriticsFiles)) {
+                throw new sfException('You must create a diacritics_mapping.xml file.');
```
mcantelon: Should the "xml" in this line be "yml"?

apps/qubit/modules/settings/actions/diacriticsAction.class.php
```diff
@@ -0,0 +1,104 @@
...
+
+    public function execute($request)
+    {
+        DefaultEditAction::execute($request);
+
+        if ($request->isMethod('post')) {
+            $this->form->bind($request->getPostParameters(), $request->getFiles());
+
+            if ($this->form->isValid()) {
+                if (0 === intval($request->getPostParameters()['diacritics']) || (1 === intval($request->getPostParameters()['diacritics']) && 0 !== $request->getFiles()['mappings']['size'])) {
```
mcantelon: Nitpick: Instead of `0 === intval` you could also use `empty"`for the check (and insted of `1 === intval` you could use `empty`.

apps/qubit/modules/settings/actions/diacriticsAction.class.php
```diff
@@ -0,0 +1,104 @@
...
+        QubitSetting::findAndSave('diacritics', $request->getPostParameters()['diacritics'], ['sourceCulture' => true]);
+    }
+
+    public function uploadDiacritics($request)
+    {
+        $file = $request->getFiles('mappings');
+
+        if (1 === (int) $request->getPostParameters()['diacritics']) {
+            if ('application/x-yaml' === $request->getFiles()['mappings']['type']) {
+                // Move uploaded file to new location to pass off to background arFileImportJob.
```
mcantelon: Is this handled by a background job? I don't see that in the cde.

apps/qubit/modules/settings/actions/diacriticsAction.class.php
```diff
@@ -0,0 +1,104 @@
...
+                try {
+                    $file = Qubit::moveUploadFile($file);
+                } catch (sfException $e) {
+                    $this->getUser()->setFlash('error', $e->getMessage());
+                }
+
+                // Replace /config/diacritics_mapping.yml with uploaded file
+                $diacriticsMappingPath = '/atom/src/plugins/arElasticSearchPlugin/config/diacritics_mapping.yml';
```
mcantelon: Should this be an absolute path given that where AtoM's code lives may vary? Also I'm wondering if the `config` dir in the root of the AtoM installation might be better given that files in plugin conf directories usually don't get changed.

apps/qubit/modules/settings/actions/diacriticsAction.class.php
```diff
@@ -0,0 +1,104 @@
...
+                $uploadMappingPath = $file['tmp_name'];
+                $diacriticsMappings = file_get_contents($uploadMappingPath);
+                file_put_contents($diacriticsMappingPath, $diacriticsMappings);
+            }
+        } else {
+            // Reset diacritics yaml when disabling the setting
+            $diacriticsMappingPath = '/atom/src/plugins/arElasticSearchPlugin/config/diacritics_mapping.yml';
```
mcantelon: This variable could maybe be set earlier in code as it ends up being set to the same thing twice.

apps/qubit/modules/settings/templates/diacriticsSuccess.php
```diff
@@ -0,0 +1,59 @@
...
+<?php end_slot(); ?>
+
+<?php slot('title'); ?>
+    <h1><?php echo __('Diacritics settings'); ?></h1>
+<?php end_slot(); ?>
+
+<?php slot('content'); ?>
+
+  <div class="alert alert-info">
+    <p><?php echo __('Please rebuild the search index if you are using diacritics mappings.'); ?></p>
```
mcantelon: Good thinking to add this as a reminder given that folks may miss the flash message about it. Maybe specify that this needs to be done after the mapping is uploaded.

apps/qubit/modules/settings/templates/diacriticsSuccess.php
```diff
@@ -0,0 +1,59 @@
...
+        </thead>
+        <tbody>
+          <tr>
+            <td><?php echo __('Diacritics'); ?></td>
+            <td><?php echo $form->diacritics; ?></td>
+          </tr>
+          <tr>
+            <td><?php echo __('Mappings YAML'); ?></td>
+            <td>
+              <span><?php echo 'YAML file containing char_filter mappings.'; ?></span></br>
```
mcantelon: Should the text being echoed here be wrapped in a call to the `__` function?

data/fixtures/settings.yml
```diff
@@ -3,7 +3,7 @@ QubitSetting:
     name: version
     editable: 0
     deleteable: 0
-    value: 193
```
mcantelon: I don't know if a migration's actually needed for this as I think your call to `parent::earlyExecute();` will end up creating the new setting in the DB.

plugins/arElasticSearchPlugin/lib/arElasticSearchPlugin.class.php
```diff
@@ -117,6 +123,8 @@ public function __destruct()
 
     public static function loadMappings()
     {
+        global $globalDiacritics;
```
mcantelon: Could you just use the protected `$globalDiacritics` property declared earlier?

mcantelon (changes requested):
I left some feedback, but looks good overall!

plugins/arElasticSearchPlugin/lib/arElasticSearchPlugin.class.php
melaniekung: since it's a static function, i can't use `$this` so i'm not sure how else to access `$globalDiacritics`, suggestions?

data/fixtures/settings.yml
melaniekung: initially, we started out with a global setting so we created a migration. if it's not needed, what changes do i need to make?

apps/qubit/modules/settings/actions/diacriticsAction.class.php
melaniekung: is there a way to get AtoM's base code location instead of using `/atom/src/`? i chose to keep it in `arElasticSearchPlugin/config` since that's where `mapping.yml` is located and I reused some of the code for that in `arElasticSearchPlugin.class.php`, but could change that if you think it's better.

plugins/arElasticSearchPlugin/lib/arElasticSearchPlugin.class.php
mcantelon: It seems like something like this could work:

1. Move "Check for diacritics_mapping.yml" code into a new method (called `loadCharMapping` or `loadCharFilter` whatever makes sense)
2. Have this method return the results of `sfYaml::load`
3. In the `initialize` method set `$this->config['index']['configuration']['analysis']['char_filter']['diacritics_lowercase']` to the results of a call to this method

data/fixtures/settings.yml
mcantelon: I'd try using the setting in a fresh install of AtoM without running the migration and see if it works. If it works without the migration then all you need to do is remove the changes settings.yml from your dev change (and remove the actual migration file).

apps/qubit/modules/settings/actions/diacriticsAction.class.php
mcantelon: Yeah, you might be right about putting it in the plugin directory so I'd say go with that (I'm a bit fuzzy on Symfony's config best practices, tbh).

To get the plugin's config directory path dynamically you can do:

    sfConfig::get('sf_plugins_dir').DIRECTORY_SEPARATOR.'arElasticSearchPlugin'.DIRECTORY_SEPARATOR.'config';

data/fixtures/settings.yml
melaniekung: nice, thanks mike! i _think_ all the mentioned issues have been resolved now

apps/qubit/modules/settings/actions/diacriticsAction.class.php
```diff
@@ -0,0 +1,104 @@
...
+            $this->form->bind($request->getPostParameters(), $request->getFiles());
+
+            if ($this->form->isValid()) {
+                // Check that a diacritics mapping yaml file is uploaded when enabling
+                if (0 === intval($request->getPostParameters()['diacritics']) || (1 === intval($request->getPostParameters()['diacritics']) && 0 !== $request->getFiles()['mappings']['size'])) {
+                    $this->processForm($request);
+                    $this->uploadDiacritics($request);
+
+                    $this->getUser()->setFlash('notice', $this->context->i18n->__('Diacritics setting updated. Rebuild the search index for the changes to be applied.'));
+                    $this->redirect(['module' => 'settings', 'action' => 'diacritics']);
```
mcantelon: The same redirect is done in both `if` conditions. You could do it after the `if` statement.

apps/qubit/modules/settings/actions/diacriticsAction.class.php
```diff
@@ -0,0 +1,104 @@
...
+                }
+
+                // Replace /config/diacritics_mapping.yml with uploaded file
+                $uploadMappingPath = $file['tmp_name'];
+                $diacriticsMappings = file_get_contents($uploadMappingPath);
+                file_put_contents($diacriticsMappingPath, $diacriticsMappings);
+            }
+        } else {
+            // Reset diacritics yaml when disabling the setting
+            file_put_contents($diacriticsMappingPath, '');
```
mcantelon: Maybe just delete the file via `unlink`?

apps/qubit/modules/settings/actions/diacriticsAction.class.php
```diff
@@ -0,0 +1,104 @@
...
+        $file = $request->getFiles('mappings');
+        $diacriticsMappingPath = sfConfig::get('sf_plugins_dir').DIRECTORY_SEPARATOR.'arElasticSearchPlugin'.DIRECTORY_SEPARATOR.'config'.DIRECTORY_SEPARATOR.'diacritics_mapping.yml';
+
+        if (1 === (int) $request->getPostParameters()['diacritics']) {
+            if ('application/x-yaml' === $request->getFiles()['mappings']['type']) {
+                // Move uploaded file to replace diacritics_mapping.yml.
+                try {
+                    $file = Qubit::moveUploadFile($file);
+                } catch (sfException $e) {
+                    $this->getUser()->setFlash('error', $e->getMessage());
```
mcantelon: It might be better, for security reasons, to provide a generic error like "Error: unable to upload" or some such thing. Otherwise the error message might divulge system information.

data/fixtures/settings.yml
```diff
@@ -1351,6 +1351,11 @@ QubitSetting:
   Qubit_Settings_defaultArchivalDescriptionBrowseView:
     name: default_archival_description_browse_view
     value: table
+  Qubit_Settings_diacritics:
```
mcantelon: You shouldn't need to add anything to this file. I think the setting will be dynamically created no matter what.

mcantelon (changes requested):
Looking good! I've added a bit of feedback, but nothing major!

data/fixtures/settings.yml
melaniekung: if i remove these lines, i get a 500 error when i try to open up the diacritics settings page.

mcantelon: Ah weird... I'll take a look.

mcantelon: Can you try just changing line 33 to `"diacritics" => "0"`?

mcantelon: (Hmm that doesn't seem to work - gets rid of 500 error but causes other issue. I'll poke around more.)

mcantelon: Okay, this worked for me:

1. Change line 33 to `'diacritics' => '0'`
2. Change `DefaultEditAction::execute($request);` to `parent::execute($request);`
3. Remove the entire `processForm` function

apps/qubit/modules/settings/actions/diacriticsAction.class.php
```diff
@@ -0,0 +1,104 @@
...
+                }
+
+                // Replace /config/diacritics_mapping.yml with uploaded file
+                $uploadMappingPath = $file['tmp_name'];
+                $diacriticsMappings = file_get_contents($uploadMappingPath);
+                file_put_contents($diacriticsMappingPath, $diacriticsMappings);
+            }
```
mcantelon: Might be good to put an else in this logic to flash an error at the user if file isn't "application/x-yaml".

mcantelon (changes requested):
Getting close! Thanks for your patience!

apps/qubit/modules/settings/actions/diacriticsAction.class.php
melaniekung: thank you for your patience with my work XD i also updated diacriticsSuccess.php to make the save button match those in BS5, but that changed the look on BS2 theme. any suggestions on making them match other save buttons in both themes?

data/fixtures/settings.yml
melaniekung: hmm were you able to upload a yaml file with those changes? i get the following error when uploading: 
```
atom_1           | [18-Jul-2023 17:25:16] WARNING: [pool atom] child 15 said into stderr: "NOTICE: PHP message: This form is multipart, which means you need to supply a files array as the bind() method second argument."
``` 
and this error was the reason why I initially used `DefaultEditAction::execute($request);` as opposed to `parent::execute($request);`

lib/task/migrate/migrations/arMigration0193.class.php
```diff
@@ -1,38 +0,0 @@
...
-    public const MIN_MILESTONE = 2;
-
-    public function up($configuration)
-    {
-        $sql = 'ALTER TABLE `property_i18n` MODIFY `value` MEDIUMTEXT;';
-        QubitPdo::modify($sql);
-
-        return true;
-    }
-}
```
jraddaoui: Be careful, don't delete this migration. I think you modified it initially instead of creating arMigration0194.class.php and then it was decided the new migration was not needed, but this one should be kept ;)

plugins/arElasticSearchPlugin/lib/arElasticSearchPlugin.class.php
```diff
@@ -479,7 +507,11 @@ protected function initialize()
                     && isset($this->config['index']['configuration']['analysis']['char_filter']['strip_md'])
                 ) {
                     foreach ($this->config['index']['configuration']['analysis']['analyzer'] as $key => $analyzer) {
-                        $this->config['index']['configuration']['analysis']['analyzer'][$key]['char_filter'] = ['strip_md'];
+                        if ((in_array($key, ['english', 'french', 'default', 'autocomplete'])) && sfConfig::get('app_diacritics')) {
+                            $this->config['index']['configuration']['analysis']['analyzer'][$key]['char_filter'] = ['strip_md', 'diacritics_lowercase'];
+                        } else {
+                            $this->config['index']['configuration']['analysis']['analyzer'][$key]['char_filter'] = ['strip_md'];
+                        }
```
jraddaoui: Why only English and French? I think it will be more useful for other cultures if the language settings are properly set. Also, as we discussed a while ago, we should improve the previous implementation and not replace the char filter entirely, but merge it with the existing one if one is set; it won't be an issue right now but it could be one in the future if someone adds a char filer in the configuration.

plugins/arElasticSearchPlugin/lib/arElasticSearchPlugin.class.php
```diff
@@ -51,6 +50,13 @@ class arElasticSearchPlugin extends QubitSearchEngine
      */
     protected $mappings;
 
+    /**
+     * Diacritics mappings configuration, diacritics_mapping.yml.
+     *
+     * @var mixed defaults to null
+     */
+    protected $globalDiacritics;
```
jraddaoui: Is this being used in the end?

plugins/arElasticSearchPlugin/lib/arElasticSearchPlugin.class.php
melaniekung: i've updated code to add `diacritics_lowercase` to all analyzers now!

jraddaoui: Thanks @melaniekung! But it's still replacing the existing char filters set in the configuration (`$this->config['index']['configuration']['analysis']['analyzer'][$key]['char_filter']`).

Right now we don't have any in the default config so it won't be an issue, but they may be added in the future and it will be hard to troubleshot why are they not being applied, as it happened to you when you started this work.

melaniekung: how about replacing it after checking for current values ??

jraddaoui: I'd check if it's set and merge the arrays if it's. This is what I shared with you at that time, the current implementation has changed a bit but it should be similar:

```php
foreach ($this->config['index']['configuration']['analysis']['analyzer'] as $key => $analyzer) {
    $filters = ['strip_md'];
    if ($this->config['index']['configuration']['analysis']['analyzer'][$key]['char_filter']) {
        $filters = array_merge($filters, $this->config['index']['configuration']['analysis']['analyzer'][$key]['char_filter']);
    }
    $this->config['index']['configuration']['analysis']['analyzer'][$key]['char_filter'] = $filters;
}
```

melaniekung: sorry , i thought i already implemented that :sweat_smile:

jraddaoui: Thanks!

apps/qubit/modules/settings/actions/diacriticsAction.class.php
mcantelon: I'm not too up-to-speed on BS5 but I'll take a look!

data/fixtures/settings.yml
mcantelon: I was able to upload... I did run into that error but there was a change I made that resolved that. It's entirely possible there was a change I made that I forgot to add to the list, though, so I'll re-create the changes to see if I missed something!

mcantelon: Yeah, I think I tweaked `apps/qubit/modules/settings/actions/editAction.class.php` to get it to work and forgot about it. Changing line 31 of that file to `$this->form->bind($request->getPostParameters(), $request->getFiles());` fixes things (basically just adding the second argument in the function call so the SettingsEditAction can work with multipart form data). So I'd say include that fix in your PR as well.

melaniekung: ah right, i've tried that before but then it runs the parent functions and skips uploadDiaricits(), which is where all my error handling is, i'll play around more and see if there's another way for it to run the function! thanks mike!

melaniekung: yay okay, i found a solution to that! lmk what you think of this solution!

apps/qubit/modules/settings/actions/diacriticsAction.class.php
melaniekung: i added a success page to the BS5 theme plugin to resolve this issue!

data/fixtures/settings.yml
mcantelon: Nicely done! Looks good!

apps/qubit/modules/settings/actions/diacriticsAction.class.php
```diff
@@ -0,0 +1,84 @@
...
+
+    public function execute($request)
+    {
+        parent::execute($request);
+    }
+
+    public function uploadDiacritics($request)
+    {
+        $file = $request->getFiles('mappings');
+        $diacriticsMappingPath = sfConfig::get('sf_plugins_dir').DIRECTORY_SEPARATOR.'arElasticSearchPlugin'.DIRECTORY_SEPARATOR.'config'.DIRECTORY_SEPARATOR.'diacritics_mapping.yml';
```
mcantelon: Could be good to make this multi-line... something like:

```$diacriticsMappingPath = sfConfig::get('sf_plugins_dir') . DIRECTORY_SEPARATOR .  
    arElasticSearchPlugin' . DIRECTORY_SEPARATOR . 
    'config' . DIRECTORY_SEPARATOR .
    'diacritics_mapping.yml';```

**Comment by mcantelon** [comment]
It's looking great... on the home stretch!

Sorry for not looking at the BS5 earlier, but seems you worked out a nice fix. 

There's a couple minor issues left (including one nitpick I just added) but seems, unless @jraddaoui spots anything, like it should be good to go after that:
* https://github.com/artefactual/atom/pull/1636/files#r1272766735
* https://github.com/artefactual/atom/pull/1636/files#r1263130145

**Comment by melaniekung** [comment]
> It's looking great... on the home stretch!
> 
> Sorry for not looking at the BS5 earlier, but seems you worked out a nice fix.
> 
> There's a couple minor issues left (including one nitpick I just added) but seems, unless @jraddaoui spots anything, like it should be good to go after that:
> 
> * https://github.com/artefactual/atom/pull/1636/files#r1272766735
> * https://github.com/artefactual/atom/pull/1636/files#r1263130145

**Comment by melaniekung** [comment]
> It's looking great... on the home stretch!
> 
> Sorry for not looking at the BS5 earlier, but seems you worked out a nice fix.
> 
> There's a couple minor issues left (including one nitpick I just added) but seems, unless @jraddaoui spots anything, like it should be good to go after that:
> 
> * https://github.com/artefactual/atom/pull/1636/files#r1272766735
> * https://github.com/artefactual/atom/pull/1636/files#r126313014

File: apps/qubit/modules/settings/actions/diacriticsAction.class.php [hunk 19]
```diff
@@ -0,0 +1,88 @@
...
+
+        $this->settingDefaults = [
+            'diacritics' => 0,
+        ];
+    }
+
+    public function execute($request)
+    {
+        parent::execute($request);
+    }
```
jraddaoui: A bit rusty on PHP, but this method may not be needed in the child class if it's only calling the parent class method.

apps/qubit/modules/settings/actions/diacriticsAction.class.php
```diff
@@ -0,0 +1,88 @@
...
+
+    public function uploadDiacritics($request)
+    {
+        $file = $request->getFiles('mappings');
+        $diacriticsMappingPath =
+            sfConfig::get('sf_plugins_dir') . DIRECTORY_SEPARATOR .
+            'arElasticSearchPlugin' . DIRECTORY_SEPARATOR .
+            'config' . DIRECTORY_SEPARATOR .
+            'diacritics_mapping.yml';
```
jraddaoui: We usually put config files overwriting the defaults in the main config folder, as it's easier to keep track of those configuration changes. And the way the file is read below, the main config folder will take precedence already:

https://github.com/artefactual/atom/pull/1636/files#diff-5f879544d637140233df54dbb8f025b8a59f4b150fd5bdb38d89e9493d12f0bfR142-R145

apps/qubit/modules/settings/actions/diacriticsAction.class.php
```diff
@@ -0,0 +1,88 @@
...
+
+        if (1 === (int) $request->getPostParameters()['diacritics']) {
+            if ('application/x-yaml' === $request->getFiles()['mappings']['type']) {
```
jraddaoui: I think you could set the mine type in the sfValidatorFile below. For example: https://github.com/artefactual/atom/blob/qa/2.x/apps/qubit/modules/informationobject/actions/uploadFindingAidAction.class.php#L54-L64.

apps/qubit/modules/settings/actions/diacriticsAction.class.php
jraddaoui: It should also return after setting the flash, right?

apps/qubit/modules/settings/actions/diacriticsAction.class.php
```diff
@@ -0,0 +1,88 @@
...
+                try {
+                    $file = Qubit::moveUploadFile($file);
+                } catch (sfException $e) {
+                    $this->getUser()->setFlash('error', $e->getMessage());
+                }
+
+                // Replace /config/diacritics_mapping.yml with uploaded file
+                $uploadMappingPath = $file['tmp_name'];
+                $diacriticsMappings = file_get_contents($uploadMappingPath);
+                file_put_contents($diacriticsMappingPath, $diacriticsMappings);
```
jraddaoui: There is a helper for this: https://github.com/artefactual/atom/blob/qa/2.x/apps/qubit/modules/informationobject/actions/uploadFindingAidAction.class.php#L84

apps/qubit/modules/settings/actions/diacriticsAction.class.php
jraddaoui: > Might be good to put an else in this logic to flash an error at the user if file isn't "application/x-yaml".

If setting the mine type in the sfValidatorFile, it will work like a normal form error.

apps/qubit/modules/settings/actions/diacriticsAction.class.php
```diff
@@ -0,0 +1,88 @@
...
+    }
+
+    protected function addField($name)
+    {
+        switch ($name) {
+            case 'diacritics':
+                $this->form->setDefault($name, $this->settingDefaults[$name]);
+                $this->form->setWidget($name, new sfWidgetFormSelectRadio(['choices' => [0 => $this->i18n->__('Disabled'), 1 => $this->i18n->__('Enabled')]], ['class' => 'radio']));
+                $this->form->setValidator($name, new sfValidatorChoice(['choices' => [1, 0]]));
+                $this->form->setWidget('mappings', new sfWidgetFormInputFile());
```
jraddaoui: It would be nice to set the accept value: https://github.com/artefactual/atom/blob/qa/2.x/apps/qubit/modules/informationobject/actions/uploadFindingAidAction.class.php#L57

apps/qubit/modules/settings/actions/diacriticsAction.class.php
```diff
@@ -0,0 +1,88 @@
...
+                $this->form->setValidator('mappings', new sfValidatorFile());
```
jraddaoui: And validate the mine type directly here, as mentioned above.

apps/qubit/modules/settings/actions/editAction.class.php
```diff
@@ -39,6 +39,10 @@ public function execute($request)
                     $this->getUser()->setFlash('notice', $this->updateMessage);
                 }
 
+                if (null !== $request->getFiles() && 'SettingsDiacriticsAction' === get_called_class()) {
```
jraddaoui: Why is the `get_called_class` check needed?

jraddaoui (changes requested):
Sorry, a bit late to the review of the form part and I still couldn't check the templates, but I have added a few comments.

apps/qubit/modules/settings/actions/editAction.class.php
melaniekung: I set this in case we add other settings in the future that requires calling getFiles as well. Should I leave that out for now since it's the only setting requiring getFiles?

jraddaoui: Oh, I see it now. This should be moved into the child class `processForm()` method then.

apps/qubit/modules/settings/actions/diacriticsAction.class.php
mcantelon: Yeah, I think you're right.

apps/qubit/modules/settings/actions/diacriticsAction.class.php
mcantelon: Good call!

apps/qubit/modules/settings/actions/diacriticsAction.class.php
melaniekung: this didn't work for me, at least not the sfValidatorFile part... if i add that, it skips over processForm and uploadDiacritics. I've come up with a slightly different way of validation, though!

1 thing i wasn't sure how to do:
- how to not display the update message if an error message is being thrown... currently we get both :upside_down_face:

apps/qubit/modules/settings/actions/diacriticsAction.class.php
```diff
@@ -0,0 +1,97 @@
...
+                        'Unable to upload diacritics mapping yaml file.'
+                    );
+
+                    return;
+                }
+            } catch (Exception $e) {
+                // Reset diacritics settings when uploading yaml fails
+                QubitSetting::findAndSave('diacritics', 0, ['sourceCulture' => true]);
+                unlink($diacriticsMappingPath);
+                $this->getUser()->setFlash('error', $this->context->i18n->__('Error: Unable to upload diacritics mapping.'));
```
mcantelon: If you add a line `unset($this->updateMessage);` after the `setFlash` line then you should be good!

mcantelon (changes requested):
Almost there!

mcantelon (approved):
Looks good!

mcantelon:
@jraddaoui How's this looking to you? I've approved but needs your okay to merge.

apps/qubit/modules/settings/templates/diacriticsSuccess.php
```diff
@@ -0,0 +1,80 @@
...
+        </td>
+        <td>
+          <span>
+            <?php echo __('YAML file containing char_filter mappings.'); ?>
+          </span></br>
+          <?php echo $form->mappings; ?><br />
+        </td>
+      </tr>
+    </tbody>
+  </table>
```
jraddaoui: To display error messages like the media type validation error, the fields need to be rendered with:

```php
<?php echo $form->mappings->label(__('Mappings YAML'))->renderRow(); ?>
```

You may be able to use the help message for the span text and I'd suggest to not use the table display and use something similar to the `csvValidatorSuccess.php` template.

apps/qubit/modules/settings/actions/diacriticsAction.class.php
```diff
@@ -0,0 +1,101 @@
...
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with Access to Memory (AtoM).  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+class SettingsDiacriticsAction extends SettingsEditAction
+{
+    // Arrays not allowed in class constants
+    public static $NAMES = [
+        'diacritics',
```
jraddaoui, line 23: I'd add the mappings field here and process it like others, instead of having a custom method.

apps/qubit/modules/settings/actions/diacriticsAction.class.php
```diff
@@ -0,0 +1,101 @@
...
+            case 'diacritics':
+                $this->form->setDefault($name, $this->settingDefaults[$name]);
+                $this->form->setWidget($name, new sfWidgetFormSelectRadio(['choices' => [0 => $this->i18n->__('Disabled'), 1 => $this->i18n->__('Enabled')]], ['class' => 'radio']));
+                $this->form->setValidator($name, new sfValidatorChoice(['choices' => [1, 0]]));
+                $this->form->setWidget('mappings', new sfWidgetFormInputFile([], ['accept' => '.yml,.yaml']));
+                $this->form->setValidator('mappings', new sfValidatorFile());
+
+                break;
+        }
+    }
```
jraddaoui: Something like this:

```php
    public function processForm()
    {
        foreach ($this->form as $field) {
            $this->processField($field);
        }
    }

    protected function addField($name)
    {
        switch ($name) {
            case 'diacritics':
                $this->form->setDefault($name, $this->settingDefaults[$name]);
                $this->form->setWidget($name, new sfWidgetFormSelectRadio(['choices' => [0 => $this->i18n->__('Disabled'), 1 => $this->i18n->__('Enabled')]], ['class' => 'radio']));
                $this->form->setValidator($name, new sfValidatorChoice(['choices' => [1, 0]]));
                
                break;

            case 'mappings':
                $this->form->setWidget('mappings', new sfWidgetFormInputFile([], ['accept' => '.yml,.yaml']));
                $this->form->setValidator('mappings', new sfValidatorFile(['mime_types' => ['text/plain']]));

                break;
        }
    }

    protected function processField($field)
    {
        switch ($name = $field->getName()) {
            case 'diacritics':
                parent::processField($field);
                
                break;
            
            case 'mappings':
                $file = $this->form->getValue('mappings');

                $diacriticsMappingPath =
                    sfConfig::get('sf_config_dir').DIRECTORY_SEPARATOR.
                    'diacritics_mapping.yml';

                if (null !== $file) {
                    try {
                        sfYaml::load($file->getTempName());

                        if (!move_uploaded_file($file->getTempName(), $diacriticsMappingPath)) {
                            $this->getUser()->setFlash('error', $this->context->i18n->__(
                                'Unable to upload diacritics mapping yaml file.'
                            ));
                            unset($this->updateMessage);

                            return;
                        }
                    } catch (Exception $e) {
                        // Reset diacritics settings when uploading yaml fails
                        QubitSetting::findAndSave('diacritics', 0, ['sourceCulture' => true]);
                        unlink($diacriticsMappingPath);
                        $this->getUser()->setFlash('error', $this->context->i18n->__(
                            'Unable to upload diacritics mapping yaml file.'
                        ));
                        unset($this->updateMessage);

                        return;
                    }
                } else {
                    // Reset diacritics settings when uploading yaml fails
                    QubitSetting::findAndSave('diacritics', 0, ['sourceCulture' => true]);
                    unlink($diacriticsMappingPath);
                }

                break;
        }
    }
```

apps/qubit/modules/settings/actions/diacriticsAction.class.php
jraddaoui: After changing the template I was able to see the media type validation error and it looks like YAML is being detected as `text/plain`, so this would work:

```php
$this->form->setValidator('mappings', new sfValidatorFile(['mime_types' => ['text/plain']]));
```

This will fail in many unwanted cases and for other plain text files it will fail in the try-catch parsing the contents.

plugins/arElasticSearchPlugin/lib/arElasticSearchPlugin.class.php
```diff
@@ -468,6 +484,12 @@ public static function modelClassFromQubitObjectClass($className)
      */
     protected function initialize()
     {
+        if (sfConfig::get('app_diacritics')) {
+            $this->config['index']['configuration']['analysis']['char_filter']['diacritics_lowercase'] = $this->loadDiacriticsMappings();
+        }
+
+        $this->loadMappings();
```
jraddaoui: Why do we need this extra mapping load?

jraddaoui (changes requested):
Getting close! Some tweaks for the validation below, still looking at the B5 template.

plugins/arElasticSearchPlugin/lib/arElasticSearchPlugin.class.php
melaniekung: this second mapping is for the mapping.yml, separate function than loadDiacriticsMapping, https://github.com/artefactual/atom/blob/35d1a58ed3a05c4aba4362ada9304809faaed497/plugins/arElasticSearchPlugin/lib/arElasticSearchPlugin.class.php#L118

plugins/arDominionB5Plugin/modules/settings/templates/diacriticsSuccess.php
```diff
@@ -0,0 +1,57 @@
...
+          <tr>
+            <td><?php echo __('Mappings YAML'); ?></td>
+            <td>
+              <span><?php echo __('YAML file containing char_filter mappings.'); ?></span></br>
+              <?php echo $form->mappings; ?><br/>
+            </td>
+          </tr>
+        </tbody>
+      </table>
+    </div>
```
jraddaoui: We are not using the table display on any B5 settings page anymore. Like for B2, you could follow the template from the CSV Validator settings.

---

For the file input, there is a way to style them better on B5, check any of the import pages or the finding aid upload form:

https://github.com/artefactual/atom/blob/qa/2.x/plugins/arDominionB5Plugin/modules/informationobject/templates/uploadFindingAidSuccess.php#L38

I'm not sure if that would be enough to display the validation errors, but give a try uploading a not accepted media type.

jraddaoui (commented):
For the B5 template:

plugins/arElasticSearchPlugin/lib/arElasticSearchPlugin.class.php
jraddaoui: But why do we need to call it here? It's not assigning the returned value to anything and it will be called below through `$this->loadAndNormalizeMappings();` before the mappings are really needed.

plugins/arDominionB5Plugin/modules/settings/templates/diacriticsSuccess.php
melaniekung: hi @jraddaoui , i've pushed some changes! thank you :)

apps/qubit/modules/settings/actions/diacriticsAction.class.php
```diff
@@ -0,0 +1,103 @@
...
+
+                            return;
+                        }
+                    } catch (Exception $e) {
+                        QubitSetting::findAndSave('diacritics', 0, ['sourceCulture' => true]);
+                        unlink($diacriticsMappingPath);
+                        $this->getUser()->setFlash('error', $this->context->i18n->__('Unable to upload diacritis mapping yaml file.'));
+                        unset($this->updateMessage);
+
+                        return;
```
jraddaoui: Nitpick: this return is no longer needed.

plugins/arDominionB5Plugin/modules/settings/templates/diacriticsSuccess.php
```diff
@@ -0,0 +1,78 @@
...
+        aria-expanded="true" aria-controls="diacritics-collapse">
+        <?php echo __('Diacritics Settings'); ?>
+      </button>
+    </h2>
+    <div id="diacritics-collapse" class="accordion-collapse collapse show" aria-labelledby="diacritics-heading">
+      <div class="accordion-body">
+        <?php echo render_field(
+          $form->diacritics
+              ->label(__('Diacritics'))
+        ); ?>
```
jraddaoui: Nitpick: it could fit in one line.

plugins/arDominionB5Plugin/modules/settings/templates/diacriticsSuccess.php
```diff
@@ -0,0 +1,78 @@
...
+      <div class="alert alert-info m-3 mb-0">
+        <p>
+          <?php echo __('Example CSV:'); ?>
+        </p>
+        <pre>type: mapping<br/>mappings:<br/>  - À => A<br/>  - Á => A</pre>
+      </div>
+
+      <div class="accordion-body">
+        <?php echo render_field($form->mappings
+            ->label(__('Mappings YAML'))); ?>
```
jraddaoui: Nitpick: it could fit in one line.

plugins/arDominionB5Plugin/modules/settings/templates/diacriticsSuccess.php
```diff
@@ -0,0 +1,78 @@
...
+  </div>
+</div>
+
+<section class="actions">
+  <input class="btn atom-btn-outline-success" type="submit" value="<?php echo __('Save'); ?>" />
+</section>
+
+</form>
+
+<?php end_slot(); ?>
```
jraddaoui: Nitpick: missing empty line at the end.

jraddaoui (commented):
Thanks @melaniekung! Only the comment about `$this->loadMappings();` left, and added a few nitpicks.

plugins/arDominionB5Plugin/modules/settings/templates/diacriticsSuccess.php
melaniekung: @jraddaoui changes made :)

jraddaoui (approved):
Thanks @melaniekung, nice work!

Code looks good to me but now that I check I don't see a ticket/issue related to it. Make sure the @artefactual/atom-maintainers know about this feature before merging, as these changes will require documentation and deployment considerations (for the `config/diacritics_mapping.yml` file). As we chatted, a next iteration of this feature could involve moving the actual mapping into the DB (now that it even has a settings page), to avoid the use of the filesystem and configuration files.

PR #1640, opened 2023-07-19
Add error message for XLIFF file load (#1627)

### LINKED ISSUE #1627: Investigate missing french translations in qtAccessionPlugin
Linked to this [Redmine issue](https://projects.artefactual.com/issues/13660). French translations for this plugin seem to be dropped when running a pull from weblate, and trying to import the existing translations (currently present in the 2.7.3 release) into the translations repo seems to have no effect.

It's important to note that unless this issue is fixed, french translations for this file will always be dropped when running a translation pull.

anvit:
Throwing an exception when Symfony runs into issues when loading XLIFF files. This will help with catching broken XML files due to translation errors. Without any error handling, this would lead Symfony to use an empty array as the array of translations when it is unable to read XML files, and would make it delete all existing translations for such a file.

### DIFF
--- a/vendor/symfony/lib/i18n/sfMessageSource_XLIFF.class.php
+++ b/vendor/symfony/lib/i18n/sfMessageSource_XLIFF.class.php
@@ -54,6 +54,16 @@ class sfMessageSource_XLIFF extends sfMessageSource_File
     if (!$xml = simplexml_load_file($filename))
     {
       $error = false;
+      $xmlErrors = libxml_get_errors();
+      $errorMessage = '';
+      foreach ($xmlErrors as $error) {
+        $errorMessage .= sprintf("%s[File]: %s at line %s, column %s\n",
+          $error->message, $error->file, $error->line, $error->column);
+      }
+      if ($errorMessage) {
+        throw new sfException(sprintf("Could not load XML file:\n\n%s",
+          $errorMessage));
+      }
 
       return $error;
     }


anvit:
This is one potential way to safeguard against this problem. Another could be to add a warning/exception in `vendor/symfony/lib/i18n/sfMessageSource.class.php` for scenarios where `loadData` doesn't return an array.
 https://github.com/artefactual/atom/blob/d996e1c4a0ab802c9900ab7b789f04d3960ad2f9/vendor/symfony/lib/i18n/sfMessageSource.class.php#L193-L201
 
Decided to go with adding it here since generating the exception right when we're having an issue with loading the XLIFF file seemed like the more logical behaviour, but let me know if there's a better way, or a reason not to do this.

vendor/symfony/lib/i18n/sfMessageSource_XLIFF.class.php
```diff
@@ -54,6 +54,15 @@ public function &loadData($filename)
     if (!$xml = simplexml_load_file($filename))
     {
       $error = false;
+      $xmlErrors = libxml_get_errors();
+      $errorMessage = '';
+      foreach ($xmlErrors as $error) {
+        $errorMessage .= $error->message . '[File]: ' . $error->file .
```
mcantelon: Using sprinf to assemble the error message might be cleaner, but otherwise looks solid.

mcantelon (changes requested):
One suggestion but looks good otherwise!

vendor/symfony/lib/i18n/sfMessageSource_XLIFF.class.php
anvit: Done!

mcantelon (approved):
Looks good to me!

PR #1641, opened 2023-07-25
Update cookie security settings

anvit:
Use secure and same-site flags for atom cookies.

### DIFF
--- a/apps/qubit/config/qubitConfiguration.class.php
+++ b/apps/qubit/config/qubitConfiguration.class.php
@@ -24,7 +24,7 @@ class qubitConfiguration extends sfApplicationConfiguration
 
     public function listenToChangeCultureEvent(sfEvent $event)
     {
-        setcookie('atom_culture', $event['culture'], ['path' => '/']);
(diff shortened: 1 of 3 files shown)


jraddaoui:
Nice!

We should make sure this works with Varnish and the example configuration from: https://github.com/artefactual/atom/blob/qa/2.x/docker/etc/varnish/default.vcl

jraddaoui:
Also, what about the `symfony` cookie?

anvit:
> Also, what about the `symfony` cookie?

It's working as expected now! My VM's production config had `session_cookie_secure` set to `false` for some reason but I noticed that that's not the case for the default config. The only thing that's not updated at this point is that the `symfony` cookie doesn't set the same-site flag (likely because symfony 1.4 just didn't have that flag).

sbreker (approved):
Looks great @anvit! Thanks for adding this!

I was wondering about adding `httponly` to atom_authenticated but I see that this cookie _is_ accessed from AtoM's fullwidthtreeview.js so that setting would not work.

PR #1649, opened 2023-08-31
Add deprecation notice for BS2 theme (#1642)

### LINKED ISSUE #1642: Feature: show a Bootstrap 2 deprecation notice in the web UI
Display a notice in the AtoM web user interface to notify administrators of the future deprecation of the Bootstrap 2 themes (Dominion and ArchivesCanada) and derivative custom themes and the need to switch to a Bootstrap 5 theme.  Displaying a notice in the UI would inform administrators that have not read about the deprecation in the user forum or release notes.

Design considerations:
- Where should the notice be displayed? It could be displayed at the top of the page like the new release notice and GDPR notice, but this could result in 3 notices being shown at the same time.
- Who should see the notice? I don't see any benefit to notifying public (non-authenticated) users, but I'm not sure if only administrators should see it, or if editors, translators, etc. should see it as well.  As the regular users of the software, an editor or translator will likely see the notice before an administrator.
- When should the notice be displayed? Displaying the notice every time a user logs will probably be annoying, so maybe it should only be shown once per user?
- What information should be put in the notice? Linking to a guide on upgrading a theme to Bootstrap 5 would probably be helpful for many administrators.

anvit:
Add deprecation notice that informs users that Bootstrap 2 themes are deprecated and will not be supported in future releases.

### DIFF
--- a/apps/qubit/modules/default/actions/bs2DeprecationMessageComponent.class.php
+++ b/apps/qubit/modules/default/actions/bs2DeprecationMessageComponent.class.php
@@ -1,0 +1,41 @@
+<?php
+
+/*
+ * This file is part of the Access to Memory (AtoM) software.
+ *
+ * Access to Memory (AtoM) is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * Access to Memory (AtoM) is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with Access to Memory (AtoM).  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+/**
+ * Display bootstrap 2 theme deprecation message component.
+ *
+ * @author Anvit Srivastav <asrivastav@artefactual.com>
+ */
+class DefaultBS2DeprecationMessageComponent extends sfComponent
+{
+    public function execute($request)
+    {
+        $hasTranslateOrEditAccess = $this->context->user->isAdministrator()
+            || $this->getUser()->hasGroup(QubitAclGroup::TRANSLATOR_ID)
+            || $this->getUser()->hasGroup(QubitAclGroup::EDITOR_ID);
+
+        // Only display this banner to editors, translators, and admins and
+        // do not display if the theme uses BS5 if it has previously been dismissed
+        if (!$hasTranslateOrEditAccess || sfConfig::get('app_b5_theme', false)
+            || null !== $this->context->user->getAttribute('bs2_deprecation_message_dismissed')
+        ) {
+            return sfView::NONE;
+        }
+    }
+}
--- a/apps/qubit/modules/default/templates/_bs2DeprecationMessage.php
+++ b/apps/qubit/modules/default/templates/_bs2DeprecationMessage.php
@@ -1,0 +1,10 @@
+<?php echo javascript_include_tag('bs2DeprecationMessage'); ?>
+
+<div class="animateNicely" id="bs2-deprecation-message">
+    <div class="alert alert-danger alert-banner">
+        <div id="bs2-deprecation-message-content">
+            <?php echo __('Bootstrap 2 themes have been deprecated and will be removed in a future release. Please consider switching to a Bootstrap 5 theme. %1%More info.%2%', ['%1%' => '<a href="https://www.accesstomemory.org/en/docs/latest/admin-manual/customization/theming/#bs2-update" target="_blank">', '%2%' => '</a>']); ?>
+        </div>
+        <button type="button" class="close">&times;</button>
+    </div>
+</div>
--- a/apps/qubit/templates/_header.php
+++ b/apps/qubit/templates/_header.php
@@ -1,3 +1,5 @@
+<?php echo get_component('default', 'bs2DeprecationMessage'); ?>
+
 <?php echo get_component('default', 'updateCheck'); ?>
 
 <?php echo get_component('default', 'privacyMessage'); ?>
--- a/js/bs2DeprecationMessage.js
+++ b/js/bs2DeprecationMessage.js
@@ -1,0 +1,31 @@
+(function ($) {
+
+  "use strict";
+
+  class Bs2DeprecationMessage {
+    constructor(element) {
+      this.$block = $(element);
+      this.$button = this.$block.find('button');
+      this.listen();
+    }
+
+    listen() {
+      this.$button.on('click', $.proxy(this.onBs2DeprecationMessageButton, this));
+    }
+
+    onBs2DeprecationMessageButton() {
+      this.$block.slideUp(100);
+      $.get('/default/bs2DeprecationMessageDismiss');
+    }
+  }
+
+  $(function ()
+  {
+    let $node = $('#bs2-deprecation-message');
+    if (0 < $node.length)
+    {
+      new Bs2DeprecationMessage($node.get(0));
+    }
+  });
+
+})(window.jQuery);
--- a/plugins/arDominionPlugin/css/less/scaffolding.less
+++ b/plugins/arDominionPlugin/css/less/scaffolding.less
@@ -1668,6 +1668,51 @@ body.login #content {
   }
 }
 
+// Bootstrap 2 theme deprecation notice
+@media (max-width: 767px) {
+  #bs2-deprecation-message {
+    margin-left: -20px;
+    margin-right: -20px;
+  }
+}
+
+#bs2-deprecation-message {
+  text-align: center;
+  position: sticky;
+  top: 0;
+  z-index: 1;
+
+  .alert-banner {
+    margin: 0 auto;
+    display: flex;
+    justify-content: space-between;
+    border: 0;
+    border-radius: 0;
+  }
+
+  button {
+    font-size: 1.75rem;
+    right: 0;
+    top: 0;
+    transform: translateX(-100%);
+  }
+
+  #bs2-deprecation-message-content {
+    padding: 8px 0;
+    color: @red;
+    position: relative;
+    left: 50%;
+    transform: translateX(-50%);
+    margin-right: 65px;
+    font-size: 0.85rem;
+
+    a:not(.btn) {
+      text-shadow: none;
+      text-decoration: underline;
+    }
+  }
+}
+
 // GDPR Privacy Message banner
 @media (max-width: 767px) {
   #privacy-message {
--- a/apps/qubit/modules/default/actions/bs2DeprecationMessageDismissAction.class.php
+++ b/apps/qubit/modules/default/actions/bs2DeprecationMessageDismissAction.class.php
@@ -1,0 +1,28 @@
+<?php
+
+/*
+ * This file is part of the Access to Memory (AtoM) software.
+ *
+ * Access to Memory (AtoM) is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * Access to Memory (AtoM) is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with Access to Memory (AtoM).  If not, see <http://www.gnu.org/licenses/>.
+ */
+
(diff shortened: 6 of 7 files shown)


anvit:
@melaniekung, @jraddaoui Still need to link this to an actual docs page, but I wanted your feedback on this, especially about if [apps/qubit/templates/_header.php](https://github.com/artefactual/atom/blob/c088704e63506d6bca5d9ea526e04fe59f12c5f9/apps/qubit/templates/_header.php#L1) is a good place to add this or if there's a better place to put it that might make client upgrades easier.

More details about what this looks like, and who will see this on the [issue page](https://github.com/artefactual/atom/issues/1642).

apps/qubit/modules/default/actions/bs2DeprecationMessageComponent.class.php
```diff
@@ -0,0 +1,40 @@
...
+/**
+ * Display bootstrap 2 theme deprecation message component.
+ *
+ * @author Anvit Srivastav <asrivastav@artefactual.com>
+ */
+class DefaultBS2DeprecationMessageComponent extends sfComponent
+{
+    public function execute($request)
+    {
+        if (!$this->context->user->isAuthenticated() || sfConfig::get('app_b5_theme', false)) {
```
jraddaoui: I'd only show this to admins, like the update check component: https://github.com/artefactual/atom/blob/qa/2.x/apps/qubit/modules/default/actions/updateCheckComponent.class.php#L29.

apps/qubit/modules/default/actions/bs2DeprecationMessageComponent.class.php
```diff
@@ -0,0 +1,40 @@
...
+            return sfView::NONE;
+        }
+
+        // Do not display the banner if it has previously been dismissed or is deactivated.
+        if (
+            null !== $this->context->user->getAttribute('bs2_deprecation_message_dismissed')
+        ) {
+            return sfView::NONE;
+        }
```
jraddaoui: Nitpick: all conditions could go in the same if.

apps/qubit/modules/default/templates/_bs2DeprecationMessage.php
```diff
@@ -0,0 +1,12 @@
+<?php $notificationMessage = __('Bootstrap 2 themes have been deprecated and will be removed in a future release. Please consider switching to a Bootstrap 5 theme. %1%More info.%2%', ['%1%' => '<a href="https://www.accesstomemory.org/download/" target="_blank">', '%2%' => '</a>']); ?>
+
+<?php echo javascript_include_tag('bs2DeprecationMessage'); ?>
+
+<div class="animateNicely" id="bs2-deprecation-message">
+    <div class="alert alert-danger alert-banner">
+        <div id="bs2-deprecation-message-content">
+            <?php echo __($notificationMessage); ?>
```
jraddaoui: Is this adding double translation? Is that an issue? I'd avoid the variable and use the string directly here.

js/bs2DeprecationMessage.js
```diff
@@ -0,0 +1,32 @@
+(function ($) {
+
+  "use strict";
+
+  class Bs2DeprecationMessage {
+    constructor(element) {
+      this.$element = $(element);
+      this.$Bs2DeprecationMessageBlock = this.$element.find('div[id="bs2-deprecation-message"]');
```
jraddaoui: Nitpick: `this.$element.find('#bs2-deprecation-message');`

js/bs2DeprecationMessage.js
```diff
@@ -0,0 +1,32 @@
...
+    }
+  }
+
+  $(function ()
+  {
+    var $node = $('body');
+    if (0 < $node.length)
+    {
+      new Bs2DeprecationMessage($node.get(0));
+    }
```
jraddaoui: I never liked this class/prototype initialization from the body, you could fin the element here directly, `var $node = $('#bs2-deprecation-message');`. This will avoid the initialization when it's not there and the extra find in the constructor.

In general, I think a class/prototype is not necessary in these cases but I know that's what we use to do. I'm not sure if we could use Bootstrap events in this version, but check the privacy message JS in B5: https://github.com/artefactual/atom/blob/qa/2.x/plugins/arDominionB5Plugin/js/privacyMessage.js.

js/bs2DeprecationMessage.js
```diff
@@ -0,0 +1,32 @@
+(function ($) {
+
+  "use strict";
+
+  class Bs2DeprecationMessage {
+    constructor(element) {
+      this.$element = $(element);
+      this.$Bs2DeprecationMessageBlock = this.$element.find('div[id="bs2-deprecation-message"]');
+      this.$Bs2DeprecationMessageButton = this.$element.find('button.bs2-deprecation-message-button');
```
jraddaoui: Nitpick: `bs2-deprecation-message-button` looks like it should be an id instead of a class.

jraddaoui: Nitpick: quite verbose variable names in a component already called the same, I'd name them `this.$block` and `this.$button`.

plugins/arDominionPlugin/css/less/scaffolding.less
```diff
@@ -1668,6 +1668,51 @@ body.login #content {
...
+        transform: translateX(-50%);
+        margin-right: 65px;
+        font-size: 0.85rem;
+
+        a:not(.btn) {
+            text-shadow: none;
+            text-decoration: underline;
+        }
+    }
+}
```
jraddaoui: Is all this style really needed using a Bootstrap 2 alert? Also, I think we use two spaces indentation in CSS.

apps/qubit/templates/_header.php
```diff
@@ -1,3 +1,5 @@
+<?php echo get_component('default', 'bs2DeprecationMessage'); ?>
+
```
jraddaoui, line 2: We should add this in arArchivesCanadaPlugin too: https://github.com/artefactual/atom/blob/qa/2.x/plugins/arArchivesCanadaPlugin/templates/_header.php#L2.

jraddaoui (changes requested):
Looking good @anvit!

I added some comments below for the current implementation but, as you noticed, what happens with users using a custom theme/header? My first thought is that we should inform them differently, maybe in the AtoM downloads page and of course in the release notes/announcement.

The reasoning behind that thought is that we could try to target some of them using a flash alert instead of an extra component/template, but we changed how those alerts are rendered in 2.7 and, if they are facing this issue, they are probably missing those flash alerts already.

So this will only affect users with a pretty basic theme or using the default B2 themes and it may be redundant if they already read the release notes/announcement, but it'll be a good reminder over time for them.

plugins/arDominionPlugin/css/less/scaffolding.less
anvit: Some of this styling is just centering the content and aligning the close icon with the buttons from the privacy and update check notifications to be consistent, and some of it is to visually distinguish it from those two since there is a scenario where a user might see all three, and it made sense to make this appear more prominent than the other two. I could potentially combine some of the styles for all three of these under the same classes though and clean some of the duplicated code up (there's a little bit that's duplicated between the other two alerts already).

Good catch on the indentation though, it seems like my editor's indentation default was set to 4 instead of 2 spaces for CSS!

apps/qubit/modules/default/actions/bs2DeprecationMessageComponent.class.php
anvit: We'd talked about who should see it on the issue [here](https://github.com/artefactual/atom/issues/1642) and the thought was that regular authenticated users, especially editors or translators, would be likely to come across this before an admin does and could bring it to their attention.

anvit:
Thanks so so much for the great feedback @jraddaoui (and for catching a bunch of issues!). Will add a commit fixing these. 

I just had another thought. Would it make things easier if I added this to the menu instead of the header (while having it still visually display where it is right now, as a sticky notification at the top of the header)? We're still going to do release announcements, so most users will hear about it that way, but I was wondering if that makes it automatically visible for folks using BS2 themes with custom headers?

apps/qubit/modules/default/actions/bs2DeprecationMessageComponent.class.php
jraddaoui: Added a comment there ;)

jraddaoui:
Hard to tell @anvit, not sure what menu are you referring to, but I guess it could have been customized too. Another option following your idea would be to modify the response directly from a config/filter class adding the HTML code. Something similar to what we do [here](https://github.com/artefactual/atom/blob/qa/2.x/plugins/arDominionPlugin/config/arDominionPluginConfiguration.class.php#L32) but modifying the content and doing it from `apps/qubit/config/qubitConfiguration.class.php` or from a filter.

I think we are worrying too much, I don't like that idea a lot and a static alert can get messy to style properly. Give it a try if you want and think it's good solution, otherwise I'd leave it as it is. Nice work!

anvit:
Addressed feedback, and updated the URL to the page where the details about Bootstrap 2 deprecation and instructions for changing to the Bootstrap 5 Dominion theme are going to be. Also decided to keep the notice in the header since that'll be the easiest to manage and maintain in the future.

apps/qubit/modules/default/actions/bs2DeprecationMessageComponent.class.php
```diff
@@ -26,13 +26,14 @@ class DefaultBS2DeprecationMessageComponent extends sfComponent
 {
     public function execute($request)
     {
-        if (!$this->context->user->isAuthenticated() || sfConfig::get('app_b5_theme', false)) {
-            return sfView::NONE;
-        }
+        $hasTranslateOrEditAccess = QubitAcl::check($this->resource, 'update')
```
jraddaoui: I'm not sure if we should check against `$this->resource`. This component may be called from all pages, so I'm not sure if a resource will exist in all cases.

js/bs2DeprecationMessage.js
```diff
@@ -4,25 +4,24 @@
...
-      this.$Bs2DeprecationMessageBlock.slideUp(100);
+      this.$block.slideUp(100);
       $.get('/default/bs2DeprecationMessageDismiss');
     }
   }
 
   $(function ()
   {
-    var $node = $('body');
+    let $node = $('#bs2-deprecation-message');
```
jraddaoui, line 24: Nice.

plugins/arArchivesCanadaPlugin/css/less/main.less
```diff
@@ -319,3 +319,48 @@ body.login #content {
...
+    transform: translateX(-50%);
+    margin-right: 65px;
+    font-size: 0.85rem;
+
+    a:not(.btn) {
+      text-shadow: none;
+      text-decoration: underline;
+    }
+  }
+}
```
jraddaoui: arArchivesCanadaPlugin is extending arDominionPlugin, if the style is the same this shouldn't be needed.

jraddaoui (changes requested):
Nicely done @anvit, thanks!

js/bs2DeprecationMessage.js
jraddaoui: I think this was left behind ;)

plugins/arArchivesCanadaPlugin/css/less/main.less
anvit: Fixed!

apps/qubit/modules/default/actions/bs2DeprecationMessageComponent.class.php
anvit: Ah, you're right! I'm updating it to simply test against `QubitInformationObject` instead, which is something I'd seen in a couple other places [such as in the sfIsdiahPlugin](https://github.com/artefactual/atom/blob/c088704e63506d6bca5d9ea526e04fe59f12c5f9/plugins/arDominionB5Plugin/modules/sfIsdiahPlugin/templates/indexSuccess.php#L280). It seems to work, but let me know if you have a better suggestion!

js/bs2DeprecationMessage.js
anvit: I've updated this to an ID, but I'm still using the class for styling. Let me know if our practice is to get rid of the class altogether in cases like this?

jraddaoui: I think a class adds no value if it's only used by a single element and this one seems to specific to be reused. Thinking a bit more, I'd not even use an id in this case, being the only button inside `#bs2-deprecation-message`.

apps/qubit/modules/default/actions/bs2DeprecationMessageComponent.class.php
jraddaoui: Not sure how that would work, it kind of makes sense for creation. but not for update. I'd just remove it and trust on the groups (not sure if admin check will be needed to).

anvit: Removing this breaks the check for admin accounts, so I'm adding the admin check in.

js/bs2DeprecationMessage.js
anvit: Fixed!

jraddaoui: Just nitpicking, I'd remove the class entirely and use the button in the LESS code too, nested under `#bs2-deprecation-message`.

jraddaoui (approved):
Thanks @anvit!!

PR #1658, opened 2023-09-18
Add tab for viewing failed jobs (#1613)

### LINKED ISSUE #1613: Feature: A tab for viewing failed jobs
Users would find it helpful to be able to filter and see only failed jobs in AtoM. 

In the Jobs page (after logging in, visit `Manage > Jobs` from the menu), we currently have an "All jobs" filter/tab, and a "Active jobs" filter/tab, but there's no easy way to filter and and see  failed jobs.

Another tab should be created in the browse page (`apps/qubit/modules/jobs/templates/browseSuccess.php` https://github.com/artefactual/atom/blob/d996e1c4a0ab802c9900ab7b789f04d3960ad2f9/apps/qubit/modules/jobs/templates/browseSuccess.php#L4-L7), similar to how we have a tab for active jobs, and the corresponding filter should be handled in the browse action (`apps/qubit/modules/jobs/actions/browseAction.class.php` https://github.com/artefactual/atom/blob/d996e1c4a0ab802c9900ab7b789f04d3960ad2f9/apps/qubit/modules/jobs/actions/browseAction.class.php#L53-L60).

melaniekung:
Add a "Failed Jobs" tab in the Jobs page (Manage > Jobs) to easily filter and and see failed jobs.

### DIFF
--- a/apps/qubit/modules/jobs/actions/browseAction.class.php
+++ b/apps/qubit/modules/jobs/actions/browseAction.class.php
@@ -57,6 +57,8 @@ class JobsBrowseAction extends DefaultBrowseAction
 
         if ('active' === $this->filter) {
             $criteria->add(QubitJob::STATUS_ID, QubitTerm::JOB_STATUS_IN_PROGRESS_ID);
+        } else if ('failed' === $this->filter) {
+            $criteria->add(QubitJob::STATUS_ID, QubitTerm::JOB_STATUS_ERROR_ID);
         }
 
         $criteria->addJoin(QubitJob::ID, QubitObject::ID);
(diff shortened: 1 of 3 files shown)


apps/qubit/modules/jobs/actions/browseAction.class.php
```diff
@@ -59,6 +59,10 @@ public function execute($request)
             $criteria->add(QubitJob::STATUS_ID, QubitTerm::JOB_STATUS_IN_PROGRESS_ID);
         }
 
+        if ('failed' === $this->filter) {
```
anvit: We could use an `else if` here instead to skip this check if the filter is already set to active.

anvit (approved):
Added a small suggestion, but the rest looks good to me!

PR #1659, opened 2023-09-19
Adhere to coding standards. #1613

### LINKED ISSUE #1613: Feature: A tab for viewing failed jobs

melaniekung:


anvit (approved):
Looks good!

PR #1664, opened 2023-09-23
Fix update function for 'Language and script notes'. (#13657)

melaniekung:
Bug was due to the 'Language and script notes' field being specific to the ISAD, DACS, and RAD templates only, so they were handled differently and was not checking to see if the string/field was updated. 

Added a function to check the stored memory to see if notes were updated when duplicating an archival description. This was created specifically for the 'Language and script notes' field, defined by the `QubitTerm::LANGUAGE_NOTE_ID`.

### DIFF
--- a/lib/model/QubitObject.php
+++ b/lib/model/QubitObject.php
@@ -305,6 +305,33 @@ class QubitObject extends BaseObject implements Zend_Acl_Resource_Interface
         return QubitStatus::getOne($criteria);
     }
 
+    /**
+     * Check to see if note is saved in memory (field was updated).
+     *
+     * Used by ISAD, DACS, and RAD templates when duplicating a description,
+     * specifcally for updating the 'Language and script notes' field.
+     *
+     * @return array $notes
+     */
+    public function getMemoryNotesByType(array $options = [])
+    {
+        // Check memory for notes
+        if (!count($this->notes) < 0) {
+            return $this->getNotesByType($options);
+        }
+
+        $noteTypeId = $options['noteTypeId'];
+        $notes = new ArrayObject();
+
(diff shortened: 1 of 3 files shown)


lib/model/QubitObject.php
```diff
@@ -305,6 +305,35 @@ public function getStatus($options = [])
...
+        $notes = new ArrayObject();
+
+        foreach ($this->notes as $note) {
+            if ($note->typeId == $noteTypeId) {
+                $notes->append($note);
+            }
+        }
+
+        if (count($notes) > 0) {
+            return $notes;
```
anvit: It seems like there is a scenario where we end up with 0 `$notes`. Would it be better to skip the if condition and just return `$notes` instead? Right now it'll end up returning `NULL`, which doesn't play with with array methods.

lib/model/QubitObject.php
```diff
@@ -305,6 +305,35 @@ public function getStatus($options = [])
...
+        // Check memory for notes
+        if (!count($this->notes) < 0) {
+            return $this->getNotesByType($options);
+        }
+
+        $noteTypeId = $options['noteTypeId'];
+        $notes = new ArrayObject();
+
+        foreach ($this->notes as $note) {
+            if ($note->typeId == $noteTypeId) {
```
anvit, line 327: I'm guessing these will always be string? If so we should prefer `===` since we're sure they're the same data type.

melaniekung: this is actually an integar (`$note->typeId`) and a string (`$noteTypeId`) so i kept it as `==`

anvit (approved):

PR #1665, opened 2023-09-25
Fix npm install for Docker. (#1662)

### LINKED ISSUE #1662: Problem: Docker gives an error when it is trying to install a package from npm
Hi People, 
I guess theres a typ0 in the Dockerfile in the following line: https://github.com/artefactual/atom/blob/9fd22ace13c16132d1b73f639c08f445ade02bd7/Dockerfile#L54

In order to work, we need to remove the second `npm` in the line.

melaniekung:
Fix for installing npm on Docker.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -51,7 +51,7 @@ RUN set -xe \
       bash \
       gnu-libiconv \
       fcgi \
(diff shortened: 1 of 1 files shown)


anvit (approved):
Looks good! 👍🏼

PR #1669, opened 2023-09-25
Add error handling for tools:rename-slug. (#1666)

### LINKED ISSUE #1666: Problem: tools:rename-slug: error handling
### Current Behavior

When running the rename-slug tool, if the destination slug (newSlug) already exists, the tool exits with an error:

Image

```
Unable to execute UPDATE statement. [wrapped: SQLSTATE[23000]: Integrity constraint violation: 1062 Duplicate entry 'wragge-sydney' for key 'slug.slug\_U\_2']
```

This is because the slug `wragge-sydney` already exists.

### Expected Behavior

Script should continue when an error is encountered and print a list of failed renames in the console output STDOUT.

melaniekung:
Add error handling for `tools:rename-slug` so that if the new slug already exists, the tool will take note and move on, instead of terminating. 

Once all the slugs have been updated, it will ouput the following into the command line STDOUT:
- list each slug that **_failed_** to rename
- provide a log file path listing all **_successful_** slug renames

### DIFF
--- a/lib/task/tools/renameSlugTask.class.php
+++ b/lib/task/tools/renameSlugTask.class.php
@@ -19,6 +19,10 @@
 
 class renameSlugTask extends arBaseTask
 {
+    protected $failedSlugs = [];
+
+    protected $logFile;
+
     protected function configure()
     {
         $this->addArguments([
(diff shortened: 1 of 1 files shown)


lib/task/tools/renameSlugTask.class.php
```diff
@@ -19,6 +19,8 @@
 
 class renameSlugTask extends arBaseTask
 {
+    protected static $globalErrors = [];
```
anvit: Small nitpick but maybe consider naming this array `failedSlugs` or something similar so that it describe what it contains more clearly.

lib/task/tools/renameSlugTask.class.php
```diff
@@ -61,6 +63,13 @@ protected function execute($arguments = [], $options = [])
         } else {
             throw new Exception('Either provide old and new slug values, or use the CSV option and supply a CSV file containing those values.');
         }
+
+        if (!empty($this->globalErrors)) {
+            $this->logSection('rename-slug', 'The following slugs were not updated:', null, 'ERROR');
+            foreach ($this->globalErrors as $err) {
+                $this->logSection('rename-slug', "{$err}", null, 'ERROR');
```
anvit: Since this is just displaying a single variable containing strings, we can skip the interpolation (double quotes and the curly braces) and just pass `$err` as the second parmeter.

anvit (approved):
Looks great! Nicely done!

PR #1676, opened 2023-10-05
Fix institutional search box options (#1635)

### LINKED ISSUE #1635: Regression: Global search institutional delimiters and advanced search link no longer working
### Current Behavior

1. Ensure you are using the default BS2 theme (notes on BS5 behavior below)
2. Navigate to Browse > Archival institutions
3. Click through to the view page of an institution
4. If desired, select a description from the holdings list and click through there as well to the view page
5. Now click into the global search box

**ISSUE 1:** No institutional delimiter radio button appears, allowing a user to limit searching to the most recent repository. The search is only global. 

6. Click on the "Advanced search" link in the global search drop-down menu: 

**ISSUE 2:** Page reloads, but the Advanced search menu is not expanded

### Expected Behavior

* The institutional search delimiter should appear when clicking into the global search box after visiting an archival institution
* Clicking on the Advanced search link in the global search drop-down should reload the page with the Advanced search section expanded - and possibly in focus as well (i.e. cursor in first adv search field)?

### Possible Solution

_No response_

### Context and Notes

First reported in the AtoM user forum, 2023-07-12: https://groups.google.com/g/ica-atom-users/c/VcfuqlXxC4M/m/yqqY_vL1AQAJ

Reproduced in the public demo site. The missing functionality is described further in the documentation here: 

* https://www.accesstomemory.org/docs/2.7/user-manual/access-content/search-atom/#search-box-delimiters

**BS5 behavior**

The institutional delimiter IS present in the BS5 theme. However, this might be an opportunity to revisit the UX on this. With the change introduced in BS5, the delimiter options are now hidden in the gear menu next to the global search, and not immediately visible in the drop-down when clicking into the search field. This could mean that defaulting to the institutional delimiter (i.e. automatically scoping the next search to a single institution rather than globally) after a user visits an archival institution view page or the view page of a linked description could be confusing, and result in users not finding expected results. 

At minimum I would suggest going forward that with this design choice, the search should  ALWAYS remain global by default, UNLESS a user clicks open the drop-down and sets it themselves. We should not be scoping user searches without visual feedback in advance. The other option  is to make some UI changes to make the delimiters more immediately visible, rather than hidden in a separate drop-down menu.

The advanced search link in this new global search drop-down menu also does not work as expected in BS5 - the advanced search menu is not expanded when the link is clicked and the page reloads. 

### Version used

AtoM 2.7.2 - 192

Ubuntu 20.04

en

### PHP version

PHP 7.4

### Contact details

_No response_

anvit:
- Fixed institutional search options not showing up in BS2 theme 
- Fixed advanced search options not auto-expanding on the browse page when using the showAdvanced url parameter.
- Changed the search box to have global search as the default option, and updated the BS5 search box to be closer to the BS2 search box in behaviour.

**Issue 1**
The institutional search options not showing up in BS2 themes is due to a quirk in how the quick links menu works. The quick links menu calls a method in `QubitObject` to find the route which matches a given URL, which ends up changing the value of `$this->resource` in `QubitMetadataRoute`: https://github.com/artefactual/atom/blob/9fd22ace13c16132d1b73f639c08f445ade02bd7/lib/routing/QubitMetadataRoute.class.php#L80

The reason this works for the BS5 theme is because the search bar is placed before the quick links menu in the template. Since it makes logical sense for the search bar to be before the quick links menu, opting for that since the alternative would be more complicated. Also, no style changes seem to be required because of how we've styled the BS2 search box.

**Issue 2**
The advanced search options not auto-expanding occurred due to the template using the variable name `show` whereas the associated action used a variable called `showAdvanced` (this works in other templates since they explicitly pass this info as show, for instance here: https://github.com/artefactual/atom/blob/7e25350ac8613c3c9d8555e1bc84603af62069bf/apps/qubit/modules/actor/templates/browseSuccess.php#L116-L122)

**Other UX/UI changes**
Updating the default search option to be global instead of institutional since this may be confusing/unexpected for users to suddenly have their searches limited by institute. Also updating the BS5 search bar dropdown to show search options instead of it being a dedicated button, similar to how the BS2 theme behaves. This would make it easier to find and access those options when needed.

More details about these issues/changes on the GitHub issue here: https://github.com/artefactual/atom/issues/1635

### DIFF
--- a/apps/qubit/templates/_header.php
+++ b/apps/qubit/templates/_header.php
@@ -22,6 +22,14 @@
     </h1>
   <?php } ?>
 
+  <div id="search-bar">
+
+    <?php echo get_component('menu', 'browseMenu', ['sf_cache_key' => $sf_user->getCulture().$sf_user->getUserID()]); ?>
+
+    <?php echo get_component('search', 'box'); ?>
+
+  </div>
+
   <nav>
 
     <?php echo get_component('menu', 'userMenu'); ?>
@@ -37,14 +45,6 @@
     <?php echo get_component('menu', 'mainMenu', ['sf_cache_key' => $sf_user->getCulture().$sf_user->getUserID()]); ?>
 
   </nav>
-
-  <div id="search-bar">
-
-    <?php echo get_component('menu', 'browseMenu', ['sf_cache_key' => $sf_user->getCulture().$sf_user->getUserID()]); ?>
-
-    <?php echo get_component('search', 'box'); ?>
-
-  </div>
 
   <?php echo get_component_slot('header'); ?>
 
--- a/plugins/arDominionB5Plugin/modules/search/templates/_box.php
+++ b/plugins/arDominionB5Plugin/modules/search/templates/_box.php
@@ -26,7 +26,7 @@
               type="radio"
               name="repos"
               id="search-realm-global"
-              <?php echo !isset($repository) ? 'checked' : ''; ?>
+              checked
               value>
             <label class="form-check-label" for="search-realm-global">
               <?php echo __('Global search'); ?>
@@ -39,7 +39,6 @@
                 type="radio"
                 name="repos"
                 id="search-realm-repo"
-                checked
                 value="<?php echo $repository->id; ?>">
               <label class="form-check-label" for="search-realm-repo">
                 <?php echo __('Search <span>%1%</span>', ['%1%' => render_title($repository)]); ?>
--- a/apps/qubit/modules/informationobject/templates/_advancedSearch.php
+++ b/apps/qubit/modules/informationobject/templates/_advancedSearch.php
@@ -6,15 +6,15 @@
 
   <a
     href="#"
-    class="advanced-search-toggle <?php echo $show ? 'open' : ''; ?>"
-    aria-expanded="<?php echo $show ? 'true' : 'false'; ?>"
+    class="advanced-search-toggle <?php echo $showAdvanced ? 'open' : ''; ?>"
+    aria-expanded="<?php echo $showAdvanced ? 'true' : 'false'; ?>"
     data-cy="advanced-search-toggle"
   >
     <?php echo __('Advanced search options'); ?>
   </a>
 
   <div class="advanced-search animateNicely"
-    <?php echo !$show ? 'style="display: none;"' : ''; ?>>
+    <?php echo !$showAdvanced ? 'style="display: none;"' : ''; ?>>
 
     <?php echo $form->renderFormTag(
       url_for(['module' => 'informationobject', 'action' => 'browse']),
--- a/apps/qubit/modules/search/templates/_box.php
+++ b/apps/qubit/modules/search/templates/_box.php
@@ -7,11 +7,7 @@
     <input type="hidden" name="topLod" value="0"/>
     <input type="hidden" name="sort" value="relevance"/>
 
-    <?php if (isset($repository) && !sfConfig::get('app_enable_institutional_scoping')) { ?>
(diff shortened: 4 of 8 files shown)


apps/qubit/templates/_header.php
```diff
@@ -38,14 +46,6 @@
...
   </nav>
 
-  <div id="search-bar">
-
-    <?php echo get_component('menu', 'browseMenu', ['sf_cache_key' => $sf_user->getCulture().$sf_user->getUserID()]); ?>
-
-    <?php echo get_component('search', 'box'); ?>
-
-  </div>
-
```
jraddaoui, line 48: Most custom themes won't be able to benefit from this solution. While B2 is being deprecated and themes should be updated, it would be great if we could find another solution. Don't worry too much if there is no "easy" alternative.

plugins/arDominionB5Plugin/modules/search/templates/_box.php
```diff
@@ -7,70 +62,6 @@ class="d-flex flex-grow-1 my-2"
...
-      <?php } ?>
-      <a class="dropdown-item" href="<?php echo url_for([
-          'module' => 'informationobject',
-          'action' => 'browse',
-          'showAdvanced' => true,
-          'topLod' => false,
-      ]); ?>">
-        <?php echo __('Advanced search'); ?>
-      </a>
-    </div>
```
jraddaoui: Moving this drop-down outside the search-box input was done for accessibility reasons. We should test these changes in that regard.

plugins/arDominionB5Plugin/scss/_layout.scss
```diff
@@ -96,6 +96,19 @@ header i.fa-2x {
...
+  margin-bottom: 0;
+  li {
+    &:last-child {
+      border-bottom: 1px solid $gray-400;
+      padding-bottom: 0.5rem;
+      margin-bottom: 0.5rem;
+    }
+  }
+}
+
```
jraddaoui: Most of these could be done with B5 classes.

jraddaoui (approved):
Nice work @anvit, thanks for the detailed description.

I'll add the accessibility comment to the issue too ;)

plugins/arDominionB5Plugin/modules/search/templates/_box.php
```diff
@@ -1,3 +1,58 @@
...
+    <a class="dropdown-item" href="<?php echo url_for([
+        'module' => 'informationobject',
+        'action' => 'browse',
+        'showAdvanced' => true,
+        'topLod' => false,
+    ]); ?>">
+      <?php echo __('Advanced search'); ?>
+    </a>
+  </div>
+</template>
```
jraddaoui: Why the use of a template and moving it with JS? If it's really needed, should it be visible hidden to avoid its temporary display.

plugins/arDominionB5Plugin/modules/search/templates/_box.php
anvit: Reverting this change due to accessibility concerns, and sticking with just making global search the default option.

apps/qubit/templates/_header.php
jraddaoui: Also, check arArchivesCanadaPlugin.

anvit: @jraddaoui just updated this. Unfortunately there isn't an easy and clean alternative. The arArchiveCanadaPlugin ended up needing a small styling change for this to work, which also seems a little awkward given that the change is for fixing the underlying logic of how quicklinks work.

jraddaoui (approved):
Thanks @anvit!

PR #1679, opened 2023-10-13
Fix GA4 tracking for fullwidth treeview. (#1606)

### LINKED ISSUE #1606: Problem: G4 analytics does not track pageviews from clicks in full-width treeview
### Current Behavior

First reported in the user forum, 2023-06-06: https://groups.google.com/g/ica-atom-users/c/8c2QkRNitsU/m/2NFkJgVdBAAJ

Lacking the right tools, I have not reproduced this locally - though the logic of the report makes sense and it's worth investigating what options there are to potentially address it. 

The full-width treeview uses AJAX to perform partial page loads, so that the current treeview position can be maintained while the body of the record view page can be updated based on the current node selection in the treeview. This is very different from how the older sidebar treeview works, where each node click performs a full request/response and page reload.

One consequence of this difference, as reported in the forum, is that the G4 Google Analytics module does not capture different node clicks in the full-width treeview as individual page clicks, so no analytics information is captured for any descriptions viewed within the same hierarchy. 

1. Log into AtoM as an administrator
2. Navigate to Admin > Settings > Treeview and set the treeview type to sidebar
3. Save and log out
4. Get a G4A tracking code via Google
5. Add it to the `config/app.yml` file in the appropriate spot, as described in the [documentation](https://www.accesstomemory.org/docs/latest/admin-manual/maintenance/web-analytics/#google-analytics)
6. Clear the cache and restart PHP-FPM
7. Wait ~24hrs or so for the analytics to begin reporting realtime data
8. Open your analytics page so you can monitor results as they happen
9. Open AtoM in a web browser as a public user and navigate to an archival unit with descendants
10. Use the sidebar treeview to navigate - confirm you can see page view data in the Analytics dashboard as you navigate around
11. Log into AtoM as an Admin
12. Via the settings, change the treeview type to "Full-width", save, and log out again
13. Repeat steps 8-10 but this time using the full-width treeview to navigate the descendants

### Expected Behavior

* Individual description page views should ideally be captured by web analytics tools like G4A when using the full-width treeview to navigate the descendants in an archival unit

### Possible Solution

_No response_

### Context and Notes

_No response_

### Version used

AtoM 2.7.2

_No response_

_No response_

### PHP version

PHP 7.4

### Contact details

_No response_

melaniekung:
With BS5's new fullwidth treeview, GA4 is not tracking any page views triggered via the treeview. We want to add a gtag event to send a page view to GA so that when a user views records using the fullwidth treeview, it can be tracked.

### DIFF
--- a/js/fullWidthTreeView.js
+++ b/js/fullWidthTreeView.js
@@ -187,6 +187,12 @@
         // Insert new content into page
         $('#main-column h1').first().replaceWith($(response.find('#main-column h1').first()));
 
+        // Get new document title for G4 Tracking
+        var doctitle = $('#main-column h1').first();
+
+        // Extract the text content without HTML tags and set title
+        document.title = doctitle[0].textContent;
+
         // Add empty breadcrumb section if current page has none, but response does
         if (!$('#main-column .breadcrumb').length && $(response.find('#main-column .breadcrumb').length))
         {
@@ -269,7 +275,7 @@
       }
     };
 
-    // On node open: remove tooltip after a node is selected, the 
+    // On node open: remove tooltip after a node is selected, the
     // node is reloaded and the first tooltip is never removed
     var openNodeListener = function (e, data)
     {
--- a/plugins/arDominionB5Plugin/js/fullWidthTreeView.js
+++ b/plugins/arDominionB5Plugin/js/fullWidthTreeView.js
@@ -288,6 +288,12 @@
           .first()
           .replaceWith($(response.find("#main-column h1").first()));
 
+        // Get new document title for G4 Tracking
+        var doctitle = $("#main-column h1").first();
+
+        // Extract the text content without HTML tags and set title
+        document.title = doctitle[0].textContent;
+
         // Add empty breadcrumb section if current page has none, but response does
         if (
           !$("#breadcrumb").length &&


js/fullWidthTreeView.js
```diff
@@ -224,6 +224,22 @@
...
+
+        // Get new document title for G4 Tracking
+        var doctitle = $('#main-column h1').first();
+
+        // Create a temporary element to parse the HTML
+        var tempElement = document.createElement('div');
+        tempElement.innerHTML = doctitle[0].outerHTML;
+
+        // Extract the text content without HTML tags and set title
+        document.title = tempElement.textContent;
```
anvit: You should be able to get the document title without creating an extra div element. `doctitle[0].textContent` should give you the title I believe.

plugins/arDominionB5Plugin/js/fullWidthTreeView.js
```diff
@@ -334,6 +334,22 @@
...
+
+        // Get new document title for G4 Tracking
+        var doctitle = $("#main-column h1").first();
+
+        // Create a temporary element to parse the HTML
+        var tempElement = document.createElement("div");
+        tempElement.innerHTML = doctitle[0].outerHTML;
+
+        // Extract the text content without HTML tags and set title
+        document.title = tempElement.textContent;
```
anvit: Same as the BS2 version, we shouldn't need the extra div element here and should be able to get the title without it.

anvit (approved):
Looks great! Glad we finally figured this one out.

PR #1681, opened 2023-10-16
Provide informative csrf attack message. (#1678)

### LINKED ISSUE #1678: CSRF Attack message
### Current Behavior

Currently, if a user stays logged into AtoM for a long period of time, their crsf token expires so if they try to re-login without refreshing the browser, they get a scary message saying "CSRF attack detected". 

### Solution

This isn't a result of a bug or any issue within the codebase so the fix is to provide a more informative/useful message like "Please refresh the page and log in again".

melaniekung:
Provide a more informative message for resolving Symfony's csrf token message. 

Currently, if a user stays logged into AtoM for a long period of time, their csrf token expires. If they try to re-login without refreshing the browser, they get a scary message saying "CSRF attack detected". The token is usually not actually attacked, but expired, so providing a more informative message on how to resolve the issue will address any concerns related to the message.

### DIFF
--- a/vendor/symfony/lib/validator/sfValidatorCSRFToken.class.php
+++ b/vendor/symfony/lib/validator/sfValidatorCSRFToken.class.php
@@ -27,7 +27,7 @@ class sfValidatorCSRFToken extends sfValidatorBase
 
     $this->setOption('required', true);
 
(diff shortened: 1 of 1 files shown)


anvit (approved):
Looks good!

PR #1682, opened 2023-10-17
Fix repeated build-nested-set warning (#1634)

### LINKED ISSUE #1634: Problem: Running the build-nested-set task outputs the search re-index warning repeatedly
### Current Behavior

1. Run `php symfony propel:build-nested-set`
2. Notice that the warning appears many times
image

### Expected Behavior

The warning should appear only once.

### Possible Solution

This is likely due to the changes introduced for [#13640](https://github.com/artefactual/atom/pull/1566) where the log command ends up running in a loop instead of being run just once.

### Context and Notes

Reported by Efthimios Mavrikas on Google groups [here](https://groups.google.com/g/ica-atom-users/c/ipHAQ7gqA48/m/yU_K4IutBAAJ).

### Version used
AtoM 2.7.2

phalene-bytes:
Removes "Note: you will need to rebuild your search index for updates to show up properly in search results." out of recursivelyUpdateTree() function

### DIFF
--- a/lib/task/propel/propelBuildNestedSetTask.class.php
+++ b/lib/task/propel/propelBuildNestedSetTask.class.php
@@ -95,6 +95,13 @@ class propelBuildNestedSetTask extends arBaseTask
             $this->conn->commit();
         }
 
+        if (!$options['index']) {
+            $this->logSection(
+                'propel',
+                'Note: you will need to rebuild your search index for updates to show up properly in search results.'
+            );
+        }
+
         $this->logSection('propel', 'Done!');
     }
 
@@ -157,11 +164,6 @@ EOF;
             if ($node['id'] != $classname::ROOT_ID) {
                 $this->reindexLft($classname, $node['id'], $node['lft']);
             }
-        } else {
-            $this->logSection(
-                'propel',
-                'Note: you will need to rebuild your search index for updates to show up properly in search results.'
-            );
         }
 
         return $width;


lib/task/propel/propelBuildNestedSetTask.class.php
```diff
@@ -119,8 +123,8 @@ protected function configure()
         $this->briefDescription = 'Build all nested set values.';
 
         $this->detailedDescription = <<<'EOF'
-Build nested set values. Optionally excluding tables (information_object, term, menu).
-EOF;
+        Build nested set values. Optionally excluding tables (information_object, term, menu).
```
anvit: I think these lines weren't indented because any extra spaces would show up on the command line when this is printed.

lib/task/propel/propelBuildNestedSetTask.class.php
```diff
@@ -173,4 +172,4 @@ protected function reindexLft(string $classname, int $id, int $lft)
             $classname, $id, ['lft' => $lft]
         );
     }
-}
+}
```
anvit: PHP code style checker might be failing the test because the file doesn't end with a new line character like it was before.

lib/task/propel/propelBuildNestedSetTask.class.php
```diff
@@ -95,6 +95,10 @@ public function execute($arguments = [], $options = [])
             $this->conn->commit();
         }
 
+        $this->logSection(
```
anvit: This message should be displayed only when the `$options['index']` is falsy.

anvit (changes requested):
I left some comments on a couple of minor issues, but otherwise this works as expected! Thanks @phalene-bytes!

anvit (commented):
Sorry, another minor nitpick! — the commit message header should be 50 characters max, and we usually have the issue number in the brackets at the end (this helps with GitHub's issue linking). I'd suggest amending the commit message to be less than 50 characters, and using the commit message body (72 characters per line) if you need more than 50 characters. 

These are a couple of examples: https://github.com/artefactual/atom/pull/1676/commits/1490c23139ba9749c02f1ae35bf63609e6a252b2, https://github.com/artefactual/atom/pull/1679/commits/c050c0b988b9aaf4f4e861a5d7013917f85c8d00

lib/task/propel/propelBuildNestedSetTask.class.php
phalene-bytes: ok so this one doesn't matter then? I saw some other commits with this failing so wondering when it matters and when it doesn't?

anvit (approved):
Looks good!

PR #1686, opened 2023-10-25
Standardize BS5 required form elements. (#61)

### LINKED ISSUE #1687: BS5 Theme: review/update required form elements
Form elements for BS5 that are noted as required (*) can create the * span element using the 
- lib/form/arB5WidgetFormSchemaFormatter.class.php

BS2 used:
- plugins/sfDrupalPlugin/lib/sfDrupalWidgetFormSchemaFormatter.class.php

The asterisk should show up as orange in BS5 because the primary-text class is added, but is showing up as gray in some form elements.
Example:
Screenshot 2023-10-24 at 9 32 27 PM

For this example, the * span element has directly added the BS2 format:
image

This difference is happening in several places, so need to review and update.

melaniekung:
The asterisk used to represent required fields in BS5 forms should consistently be orange.

### DIFF
--- a/plugins/arDominionB5Plugin/scss/_forms.scss
+++ b/plugins/arDominionB5Plugin/scss/_forms.scss
@@ -90,3 +90,7 @@ ul.multiInput,
 .accordion-body > .mb-3:last-child {
   margin-bottom: 0 !important;
 }
+
+.form-required {
+  color: $primary;
+}


jraddaoui (approved):

melaniekung:
@phalene-bytes is investigating a solution to the base issue rather than a patch up fix

Issue #1688, opened 2023-10-25
BS5 Theme: change BS5 navbar-toggler-icon to Font Awesome icon


phalene-bytes:
Benefits from implementing this change:
- Consistency: aligns with approach to using Font Awesome icons/classes throughout AtoM
- Flexibility and Control: grants more control over the icon appearance, making it easier to create custom themes
- Automation: updates navbar icon's appearance automatically and in sync with the other navbar icons, with minimal effort

PR #1689, opened 2023-10-26
Update navbar toggler icon (#1688)

phalene-bytes:
Updates navbar toggler icon from BS5 icon/class to Font Awesome icon for consistency and easier client BS5 theme customization (in sync with other navbar icons)

### DIFF
--- a/plugins/arDominionB5Plugin/templates/_header.php
+++ b/plugins/arDominionB5Plugin/templates/_header.php
@@ -26,8 +26,14 @@
         <?php } ?>
       </a>
     <?php } ?>
-    <button class="navbar-toggler atom-btn-secondary my-2 me-1" type="button" data-bs-toggle="collapse" data-bs-target="#navbar-content" aria-controls="navbar-content" aria-expanded="false">
-      <span class="navbar-toggler-icon"></span>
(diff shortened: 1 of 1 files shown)


plugins/arDominionB5Plugin/templates/_header.php
```diff
@@ -27,7 +27,7 @@
       </a>
     <?php } ?>
     <button class="navbar-toggler atom-btn-secondary my-2 me-1" type="button" data-bs-toggle="collapse" data-bs-target="#navbar-content" aria-controls="navbar-content" aria-expanded="false">
-      <span class="navbar-toggler-icon"></span>
+      <i class="fas fa-2x fa-fw fa-bars"></i>
```
phalene-bytes: not sure if there are any classes and/or ARIA or other attributes I should add

jraddaoui: I'd add `aria-hidden="true"` and maybe `title="<?php echo __('Toggle navigation'); ?>"`.

https://fontawesome.com/v4/accessibility/#:~:text=The%20simplest%20way%20to%20provide,it%20accessible%20to%20assistive%20technologies.

I have my conflicts about the title, which will work as a default tooltip, because we use BS5 tooltips in the menu icons. I don't think we need it for this button/icon but if you want to add it, check the clipboard menu button/icon:

https://github.com/artefactual/atom/blob/qa/2.x/plugins/arDominionB5Plugin/modules/menu/templates/_clipboardMenu.php#L20-L27

anvit (approved):
Looks good!

PR #1691, opened 2023-10-31
WIP: Update BS5 required form elements (#1687)

phalene-bytes:
WIP

### DIFF
--- a/apps/qubit/modules/actor/actions/editAction.class.php
+++ b/apps/qubit/modules/actor/actions/editAction.class.php
@@ -91,7 +91,7 @@ class ActorEditAction extends DefaultEditAction
         switch ($name) {
             case 'entityType':
                 $this->form->setDefault('entityType', $this->context->routing->generate(null, [$this->resource->entityType, 'module' => 'term']));
-                $this->form->setValidator('entityType', new sfValidatorString());
+                $this->form->setValidator('entityType', new sfValidatorString(['required' => true]));
 
                 $choices = [];
                 $choices[null] = null;
@@ -104,9 +104,15 @@ class ActorEditAction extends DefaultEditAction
                 break;
 
             case 'authorizedFormOfName':
-            case 'corporateBodyIdentifiers':
             case 'datesOfExistence':
             case 'descriptionIdentifier':
+                $this->form->setDefault($name, $this->resource[$name]);
+                $this->form->setValidator($name, new sfValidatorString(['required' => true]));
+                $this->form->setWidget($name, new sfWidgetFormInput());
+
+                break;
+
+            case 'corporateBodyIdentifiers':
             case 'institutionResponsibleIdentifier':
                 $this->form->setDefault($name, $this->resource[$name]);
                 $this->form->setValidator($name, new sfValidatorString());
--- a/apps/qubit/modules/repository/actions/editAction.class.php
+++ b/apps/qubit/modules/repository/actions/editAction.class.php
@@ -189,6 +189,12 @@ class RepositoryEditAction extends DefaultEditAction
 
             case 'identifier':
             case 'authorizedFormOfName':
+                $this->form->setDefault($name, $this->resource[$name]);
+                $this->form->setValidator($name, new sfValidatorString(['required' => true]));
+                $this->form->setWidget($name, new sfWidgetFormInput());
+
+                break;
+
             case 'descIdentifier':
             case 'descInstitutionIdentifier':
                 $this->form->setDefault($name, $this->resource[$name]);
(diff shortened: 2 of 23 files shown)


jraddaoui:
Right, I remember now why we left those "required" fields with the asterisk in black and not in orange. The actual required fields from each form will have the orange one, while the fields required by the standard will have a black one.

This fields required by the standard are not actually required in the form submission, they will produce a red alert warning on the view page for users with enough permissions but they shouldn't invalidate the form submission AFAIK.

@phalene-bytes, check with @fiver-watson before moving forward with this changes (if you have not done that before and we really need to change this).

fiver-watson:
Interesting... @jraddaoui does that mean that the BS5 templates have submission requirements that the BS2 ones did not? Because you could click Add New and then immediately click Create in BS2 and generate a totally blank record - so there were no fields required for submission. Is that different now?

In any case: if there are 2 different requirements here, then I like the idea of using the different colors to indicate standards requirement (but not enforced by AtoM) vs form requirements (which are). Otherwise, making them all orange is better than making them all black, I think?

jraddaoui:
> Interesting... @jraddaoui does that mean that the BS5 templates have submission requirements that the BS2 ones did not? Because you could click Add New and then immediately click Create in BS2 and generate a totally blank record - so there were no fields required for submission. Is that different now?

No, you are right, that should still be the same. We may not have that distinction in the main entity forms and they may be all fake required fields so it would be nice to have some consistency in the asterisk color. But this PR is making some of then really required at the validation level, which will introduce that change and what I'm trying to avoid.

> In any case: if there are 2 different requirements here, then I like the idea of using the different colors to indicate standards requirement (but not enforced by AtoM) vs form requirements (which are). Otherwise, making them all orange is better than making them all black, I think?

Yeah, I kind of remember making that choice of having two colors for that reason but I may be wrong, we do have some really required fields in other forms (import/export), but if they are not in the same form I'm not sure if we should have the two colors. If you ask me, I'd remove the asterisk if they are not really required fields and use the orange for all the others, but I understand we have that hint based on the archival standard.

fiver-watson:
I'd prefer that we keep SOME kind of indication about what fields the standards require - AtoM's big selling point is supposed to be standards-based description, so if we are not ENFORCING what the standards expect, we need to at least offer visual cues. 

Given that i don't think this PR is meant to turn into a big redesign, I'd suggest that just using the same orange for all is better than removing the asterisk altogether from the standards-based ones.

phalene-bytes:
@fiver-watson @jraddaoui Oh yeah, I noticed that the other day and was going to ask about that. I noticed it because I was checking between my docker (wouldn't go to view page after change) against the demo site and also because of the alert stub in `plugins/arDominionB5Plugin/modules/sfIsaarPlugin/templates/editSuccess.php`. So I changed it so the alert looks more like what's in indexSuccess.php:

UI before (after changing to validator required):
image

UI after:
image


I think I could make a Helper regex function or something so it works better, and maybe could figure out how to change the error message so it's more like what's in indexSuccess.php but maybe that's not what y'all want. And makes sense to me, pretty convenient and flexible not needing to fill out all the form-required stuff before having a record.

code before:
```
  <?php if (!$form->isValid() && $form->hasErrors()) { ?>
      <div class="alert alert-danger" role="alert">
        <ul class="<?php echo render_b5_show_list_css_classes(); ?>">
          <?php foreach ($form->getErrorSchema() as $error) { ?>
            <?php $error = sfOutputEscaper::unescape($error); ?>
            <li><?php echo $error->getMessage(); ?></li>
          <?php } ?>
        </ul>
      </div>
    <?php } ?>
```

code after:
```
    <?php if (!$form->isValid() && $form->hasErrors()) { ?>
      <div class="alert alert-danger" role="alert">
        <ul class="<?php echo render_b5_show_list_css_classes(); ?>">
          <?php foreach ($form as $name => $field) { ?>
            <?php if ($field->hasError()) { ?>
              <li>
                <?php 
                  $label = strip_tags($form[$name]->renderLabel());
                  $asteriskPos= strpos($label, '*');
                  $label = substr($label, 0, $asteriskPos);
                  echo $label . ' - ' . $field->getError(); 
                ?>
              </li>
            <?php } ?>
          <?php } ?>
        </ul>
      </div>
    <?php } ?>
```

phalene-bytes:
@fiver-watson I was thinking about this some more, if we want a visual cue, but there's some difference between AtoM required vs standards required, maybe make the visual cue `*` and `**`, because using a color cue was too subtle for me. Though maybe that's just me.

Granted, there's only 1 instance I've seen so far where both types of required were in the same form and the Accession Number gets populated automatically. It'd only be if you deleted that part, that it'd stay on the edit page (but I don't there's an alert banner for it)

PR #1693, opened 2023-11-01
Assign CSP header default to 'report-only'

sbreker:
Update default CSP 'repsonse_header' setting to be 'Content-Security-Policy-Report-Only'.


anvit (approved):
Looks good. Thanks Steve!

PR #1695, opened 2023-11-02
Fix missing Bootstrap imports. (#1694)

### LINKED ISSUE #1694: Problem: Fix missing Bootstrap import
When using a newer version of node (ex. installing from tarball which doesn't include package-lock.json), build fails due to some changes made by Bootstrap which requires new imports.

melaniekung:


anvit (approved):
Looks good!

PR #1699, opened 2023-11-02
Fix BS5 typography issues (#1698)

### LINKED ISSUE #1698: Problem: BS5 theme has typography issues and inconsistencies
The BS5 theme has some missing styles due to using `reboot.scss` from Bootstrap, which exists to remove the default styles that browsers add to HTML tags. This can be desirable while building a new theme from the ground up, but in our case has lead to the `blockquote`, `code`, `pre` tags missing some essential styling.

A few examples of this can be found when using markdown. Here is what the BS2 version looks like
BS2-markdown

Compare this to the BS5 version
BS5-markdown-current

Another issue that comes from `reboot.scss` is that the default sans serif typeface is `system-ui`. This leads to different users having widely different experiences, for instance: most current Windows versions default to `Segoe UI` typeface which doesn't ship with a `500` weight font, and as a result they won't see any sections with the `<strong>` tag with having any kind of emphasis applied:
BS5-home-current-windows

Here is what it looks like on Ubuntu 22.04 for comparison:
BS5-home-current-ubuntu22

anvit:
Added the Open Sans typeface (which is what arDominionPlugin uses) and set it as the default sans-serif typeface for the BS5 theme.

Also added the following style changes:
- Left border and padding for blockquote tags.
- Background color for code and pre tags.
- Padding for code and pre tags.
- Whitespace wrapping for code tags.
- Set the default font-weight for bold and strong tags to 700

Here's what it looks like with these styling changes applied:
BS5-markdown
BS5-home-new


anvit:
I went with the 700 font-weight for `<b>` and `<strong>` tags just because that's what the BS2 theme was using, but here's a comparison of what other weights look like, just in case this seems a too bold:

500 font-weight
600 font-weight
700 font-weight

melaniekung (approved):
Nice work, especially with the image in the description! :100: 

Only thought I had was the use of `@use` in _typography.scss. I know that Sass is trying to deprecate the use of `@import` but currently, all our .scss files uses `@import` so do we want to keep them consistent for now? 

Otherwise looks good to me!

plugins/arDominionB5Plugin/scss/_typography.scss
```diff
@@ -0,0 +1,32 @@
...
+@include faces($subsets: all, $weights: all, $styles: all);
+
+:root {
+  --bs-font-sans-serif: "Open Sans", sans-serif;
+}
+
+b,
+strong {
+  font-weight: 700;
+}
```
jraddaoui: Can this be changed with these vars? https://github.com/artefactual/atom/blob/qa/2.x/plugins/arDominionB5Plugin/scss/_variables.scss#L44-L47

In fact, removing those vars may do what you are doing directly.

plugins/arDominionB5Plugin/scss/_typography.scss
```diff
@@ -0,0 +1,32 @@
...
+
+blockquote {
+  border-left: 5px solid $gray-400;
+  padding-left: 1rem;
```
jraddaoui: I'd use `$spacer`.

plugins/arDominionB5Plugin/scss/_typography.scss
```diff
@@ -0,0 +1,32 @@
...
+}
+
+code {
+  background-color: $gray-200;
+  white-space: pre-wrap;
+  padding: 0.25rem;
```
jraddaoui: Same, `$spacer * 0.25`.

plugins/arDominionB5Plugin/scss/_typography.scss
```diff
@@ -0,0 +1,32 @@
...
+}
+
+pre {
+  background-color: $gray-200;
+  padding: 0.5rem;
```
jraddaoui: Same.

jraddaoui (approved):
Thanks @anvit!

PR #1702, opened 2023-11-03
Update Clipboard send to use form submit (#1701)

### LINKED ISSUE #1701: Problem: The clipboard send feature no longer sends the request data as a form submission
When the clipboard was moved to browser storage in [#11556](https://github.com/artefactual/atom/pull/1166) this functionality was changed to send information through an ajax request. Since the idea behind this feature was that it could be used to have the clipboard submit form info to external site and send the user there, that functionality is now broken.

anvit:
Remove the ajax request, and udate the clipboard send feature to send the information through a form submit instead. This allows the Send button to redirect to the intended URL like the documentation suggests.

### DIFF
--- a/js/clipboard.js
+++ b/js/clipboard.js
@@ -182,39 +182,56 @@
         return;
       }
 
+      let $form = $("<form />", {
+        "id": "sendForm",
+        "action": $sendButton.data("url"),
+        "method": $sendButton.data("method"),
+      });
+
       // Generate clipboard send data
-      var data = {base_url: $sendButton.data('site-base-url')};
+      let $baseUrl = $("<input />", {
+        "type": "hidden",
+        "name": "base_url",
+        "value": $sendButton.data('site-base-url')
+      });
+
+      $form.append($baseUrl);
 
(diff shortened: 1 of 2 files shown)


js/clipboard.js
```diff
@@ -182,39 +182,39 @@
...
       }
 
+      let $form = $("<form />", {
+        "id": "sendForm",
+        "action": $sendButton.data("url"),
+        "method": $sendButton.data("method"),
+      });
+
       // Generate clipboard send data
       var data = {base_url: $sendButton.data('site-base-url')};
```
jraddaoui: I thing you could get rid of this `data` variable and add the values to the form inputs directly.

js/clipboard.js
```diff
@@ -182,39 +182,39 @@
...
+      $form.append(`<input name='base_url' type='hidden' value='${data.base_url}'/>`);
```
jraddaoui: Nitpick: I like the way you created the form element, I think it's cleaner than a long string and I'd do the same for the inputs.

jraddaoui (approved):
Nice!

PR #1704, opened 2023-11-06
Fix multi-line i18n strings

anvit:
Join i18n strings that have been split into multiple lines to have them be recognized as a single unit for translations.


jraddaoui (approved):
Thanks @anvit! This had to be a pain, sorry I didn't realize when we introduced the code formatter. It also went under the radar for a long time!

anvit:
@jraddaoui No problem! To be honest it wasn't as bad as I thought it was going to be, because when I'd ran a pattern-match search, I found over 700 results 😅, but a lot of those turned out to be SQL statements that were split up, CLI exceptions, and other such results that don't affect the user.

PR #1705, opened 2023-11-06
Add validation for static page slugs. (#1703)

### LINKED ISSUE #1703: Problem: Invalid static pages are created with invalid slugs
### Current Behavior

1. Create a static page with an invalid url (ex. "test.123")
2. Notice that the page is created but cannot be used or deleted.

### Expected Behavior

Throw an error if slug is invalid and do not create record.

melaniekung:


anvit (changes requested):
This might need more complex validation than I'd first thought. This fails for valid slugs that don't follow a valid url pattern, which is a lot of them. For instance the strings `help` or `privacy-policy` are valid, but since they don't look like a valid URL pattern, this fails.

It will probably need to have some custom validation to check if it contains anything other than alphanumeric characters, dashes, and underscores.

lib/helper/QubitHelper.php
```diff
@@ -37,6 +37,10 @@ function render_field($field, $resource = null, array $options = [])
     $div = null;
     $culture = sfContext::getInstance()->user->getCulture();
 
+    if (isset($options['type']) && 'url' === $options['type']) {
+        $options['type'] = 'url';
+    }
```
jraddaoui: `$options['type']` already has that value based on the if condition.

jraddaoui (approved):
Nice!

anvit (approved):
Looks good to me!

Small nitpick that we might want the slugs to start with an alphanumeric character and not permit hyphen or underscore as the starting character, but as far as AtoM is concerned, these should still be valid and wouldn't lead to errors.

lib/helper/QubitHelper.php
```diff
@@ -37,6 +37,10 @@ function render_field($field, $resource = null, array $options = [])
     $div = null;
     $culture = sfContext::getInstance()->user->getCulture();
 
+    if (isset($options['pattern'])) {
+        $options['pattern'] = '^[a-zA-Z][a-zA-Z0-9\-_]*$';
```
jraddaoui, line 41: @melaniekung @anvit, this could be an issue later on if someone uses a different pattern before calling the helper. It shouldn't be needed if it's already set and it may modify its value.

melaniekung: thank you radda! addressed in new PR: https://github.com/artefactual/atom/pull/1707

PR #1706, opened 2023-11-07
Add GA sources to CSP directive

sbreker:
Add Google Analytics sources to CSP directive.

Address CSP issue with GA inline script in _footer template by adding template to the BS5 theme and inserting CSP nonce value. CSP is not supported in BS2 themes.

### DIFF
--- a/apps/qubit/modules/default/templates/_tagManager.php
+++ b/apps/qubit/modules/default/templates/_tagManager.php
@@ -1,4 +1,7 @@
 <?php if ('script' === $code) { ?>
+
+<?php if ('' === sfConfig::get('csp_nonce', '')) { ?>
+
 <!-- Google Tag Manager -->
 <script>(function(w,d,s,l,i){w[l]=w[l]||[];w[l].push({'gtm.start':
 new Date().getTime(),event:'gtm.js'});var f=d.getElementsByTagName(s)[0],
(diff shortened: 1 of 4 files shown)


apps/qubit/modules/default/templates/_tagManager.php
```diff
@@ -1,11 +1,28 @@
...
+<script <?php echo __(sfConfig::get('csp_nonce', '')); ?>>(function(w,d,s,l,i){w[l]=w[l]||[];w[l].push({'gtm.start':
+new Date().getTime(),event:'gtm.js'});var f=d.getElementsByTagName(s)[0],
+j=d.createElement(s),dl=l!='dataLayer'?'&l='+l:'';j.async=true;j.src=
+'https://www.googletagmanager.com/gtm.js?id='+i+dl;var n=d.querySelector('[nonce]');
+n&&j.setAttribute('nonce',n.nonce||n.getAttribute('nonce'));f.parentNode.insertBefore(j,f);
+})(window,document,'script','dataLayer','<?php echo $containerId; ?>');</script>
+<!-- End Google Tag Manager -->
+
+<?php } ?>
+
```
jraddaoui, line 25: Would this work with an empty nonce to avoid the duplication? Not so sure about the setAttribute part.

plugins/arDominionB5Plugin/templates/_footer.php
```diff
@@ -0,0 +1,32 @@
...
+      data-read-less-text="<?php echo __('Read less'); ?>">
+    </div>
+  </div>
+
+</footer>
+
+<?php $gaKey = sfConfig::get('app_google_analytics_api_key', ''); ?>
+<?php if (!empty($gaKey)) { ?>
+  <script <?php echo __(sfConfig::get('csp_nonce', '')); ?> async src="https://www.googletagmanager.com/gtag/js?id=<?php echo $gaKey; ?>"></script>
+  <script <?php echo __(sfConfig::get('csp_nonce', '')); ?>>
```
jraddaoui: Do we need to duplicate the footer in B5 just to add the nonce here? Would it work adding it directly to the existing footer?

jraddaoui (approved):
Nice!

plugins/arDominionB5Plugin/templates/_footer.php
sbreker: Getting the CSP nonce should not happen in BS2 as it is unsupported. It would have no effect to retrieve the nonce as it would evaluate to '', but I felt it was confusing to have that code in templates that could be run under BS2. I am open to changing this.

jraddaoui: Not sure if it happened in the other PR, but we are doing it in this one in the tag manager template. I'd prefer not duplicating the templates if possible, but not a big deal.

apps/qubit/modules/default/templates/_tagManager.php
sbreker: I am not confident that it would the way it tries to add the nonce attribute.

plugins/arDominionB5Plugin/templates/_footer.php
sbreker: I see - will revert this bit and add the nonce in the other footer file

PR #1707, opened 2023-11-08
Don't modify pattern option on render_field helper. (#1703)

### LINKED ISSUE #1703: Problem: Invalid static pages are created with invalid slugs

melaniekung:
Don't modify pattern option on render_field helper.


jraddaoui (approved):
Thanks @melaniekung!

I think `Don't modify pattern option on render_field helper` (or something along those lines) could be a better commit message.

PR #1716, opened 2023-11-16
Make Job logs more readable. (#1715)

### LINKED ISSUE #1715: Problem: Diffcult to read job logs
### Current Behavior

1. Go to a Failed Jobs report
2. There's a background color that makes the text difficult to read

### Expected Behavior

Should be easy to read logs

### Possible Solution

https://github.com/artefactual/atom/blob/9f59fcaa512110d5e8c697bea5d1f616961c2c89/plugins/arDominionB5Plugin/scss/_typography.scss#L22

^ the above background color should not be used in this scenario - removing the background will make the text more readable.

melaniekung:


anvit (approved):
Looks good! Thanks!

PR #1717, opened 2023-11-16
Add Google Maps urls to the CSP directives

anvit:
Update list of directives to include those supplied by Google to prevent CSP warnings when using Google Maps.
Ensure that the CSP nonce is available in the page so that Google Maps JS can use it. This eliminates CSP errors that occur when Google Maps are included on a page, but the page does not have any other elements that include the nonce. This change ensures the nonce appears at least once on the page before the map component is displayed.


melaniekung (approved):
Good job team! LGTM

sbreker (approved):
Thanks @anvit! :+1:

PR #1731, opened 2023-12-27
Exclude hidden elements from exports. (#1728)

### LINKED ISSUE #1728: Enhancement: Exclude hidden fields from CSVs exported by public users
### Current behaviour ###
- As an administrator: In the visible elements module, uncheck one or more archival description fields to make them hidden from the public.
- As a public user:
  - Add one or more archival descriptions to the clipboard.
  - Go to the clipboard and export a CSV.
  - The exported CSV includes the field(s) that have been unchecked in the Visible elements module.

### Expected behaviour ###
Exported CSVs should only contain fields that are visible to the user who exported the CSV.

### Possible solution ###
Ideally, exclude hidden fields from the CSV export if the user is not logged in. Another option that would work for us would be to hide the export button entirely for public users.

### Context ###
Hidden fields may contain sensitive information that should not be shared freely with the public (e.g. immediate source of acquisition where the donor prefers to be anonymous). Having this information included in publicly available CSV exports then becomes an information privacy/security risk.

Note that we would still want the full export functionality to be available to staff (authenticated users), just not to the public.

melaniekung:
Elements marked as hidden in Visible Elements settings are removed from XML and CSV exports by default. Administrators  have the option to include these hidden elements in exports.

### DIFF
--- a/lib/flatfile/QubitFlatfileExport.class.php
+++ b/lib/flatfile/QubitFlatfileExport.class.php
@@ -40,6 +40,8 @@ class QubitFlatfileExport
     protected $rowsPerFile = 1000;           // how many rows until creating new export file
 
     protected $separatorChar = '|';          // character to use when imploding arrays to a single value
+    protected $params;
+    protected $nonVisibleElementsIncluded;
 
     /*
      * Constructor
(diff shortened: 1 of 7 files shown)


lib/flatfile/QubitFlatfileExport.class.php
```diff
@@ -296,24 +308,82 @@ public function prepareRowFromResource()
...
         $this->modifyRowBeforeExport();
     }
 
+    /**
+     * Get parameters to determine hidden elements.
+     *
+     * @param array parameters for CSV export
+     * @param mixed $params
+     */
+    public function passParams($params)
```
anvit: I would suggest naming this function `setParams` since this behaves as a setter function

anvit (approved):
Looks good!

PR #1740, opened 2024-01-26
Remove memprof from dockerfile (#1739)

### LINKED ISSUE #1739: Problem: memprof Dockerfile typo
### Current Behavior

1. If you run `docker compose build --no-cache`, build fails at php8-pecl-memprof
image

https://github.com/artefactual/atom/blob/9fd22ace13c16132d1b73f639c08f445ade02bd7/Dockerfile#L40

Should be php81-pecl-memprof
https://dl-3.alpinelinux.org/alpine/edge/testing/x86_64/
image

phalene-bytes:
Docker images should now build successfully

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -36,8 +36,6 @@ RUN set -xe \
       libxslt \
       libmemcached-libs \
       libzip \
-    && apk add --update-cache --repository http://dl-3.alpinelinux.org/alpine/edge/testing/ --allow-untrusted \
-      php8-pecl-memprof \
     && apk del .phpext-builddeps \
     && pecl clear-cache \
     && apk add --no-cache --virtual .atom-deps \


melaniekung (changes requested):
hi eleanor, this fix looks good to me! but before we merge, can you provide a little more details on this fix in the commit message? was it really a typo or a dependency req for an image we're using in docker? also, a little background on what this pkg is used for, that would also be great. thank you!

phalene-bytes:
@melaniekung Mmm what do you think I should put?  After the fix, the rest of the Dockerfile commands are able to run (like building the B2 & B5 themes). Should I put:

Fix: Typo in Dockerfile (`php8-pecl-memprof` should be `php81-pecl-memprof`)
This dependency is a memory profiler for PHP, used to identify memory leaks in PHP scripts. The typo caused a failure in the Docker image build process. The fix now enables the remaining commands in the Dockerfile to run successfully.

For more information on this dependency, see: https://github.com/artefactual/atom/pull/1452

melaniekung:
i couldn't confirm if it was a typo or if the difference is due to support for php 8 vs php 8.1 since docker used to build fine. if my assumption is correct, then i'd say something like "Update pecl-memprof dependency, a memory profiler for PHP used to identify memory leaks in PHP scripts". otherwise "Fix typo for pecl-memprof dependency, a memory profiler for PHP used to identify memory leaks in PHP scripts" will do.

melaniekung (approved):

PR #1741, opened 2024-01-26
Add OIDC authentication to AtoM

sbreker:
This commit adds support for OIDC (OpenID Connect) user authentication in AtoM, implemented in a new arOidcPlugin.

The plugin also provides an optional mechanism for dynamically setting AclUserGroup membership on each login, based on the presence or absence of expected values in tokens returned from the OIDC endpoint during authentication.

Configuration of the plugin is handled manually by editing configuration files rather than through the AtoM user interface to avoid a "chicken-egg" problem of needing to log into AtoM as an administrator in order to set up the authentication mechanism. The steps required to enable the plugin are as follows, with all paths relative to the AtoM root directory:

- Activate the plugin by adding it to the $plugins array in config/ProjectConfiguration.class.php
- Configure the OIDC settings in plugins/arOidcPlugin/config/app.yml
- Change the default login module from user to OIDC in apps/qubit/config/settings.yml
- Change the user class to oidcUser in apps/qubit/config/factories.yml

Tasks such as DIP upload will use `authenticateWithBasicAuth` method in AtoM's base `myUser` user class to allow Archivematica to authenticate for DIP Upload using basic auth even when a single sign-on method such as OIDC is enabled in AtoM.


plugins/arOidcPlugin/config/arOidcPluginConfiguration.class.php
```diff
@@ -0,0 +1,38 @@
...
+ * You should have received a copy of the GNU General Public License
+ * along with Access to Memory (AtoM).  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+/**
+ * arOidcPlugin configuration.
+ */
+class arOidcPluginConfiguration extends sfPluginConfiguration
+{
+    public static $summary = 'OIDC authn authz plugin.';
```
anvit: Should this say "OIDC authenticaion authorization plugin."?

sbreker: I can change this if you like. Are these are [common abbreviations](https://www.google.com/search?q=authn+authz&oq=authn+a&gs_lcrp=EgZjaHJvbWUqBwgAEAAYgAQyBwgAEAAYgAQyBggBEEUYOTIHCAIQABiABDIHCAMQABiABDIHCAQQABiABDIHCAUQABiABDIHCAYQABiABDIHCAcQABiABDIICAgQABgWGB4yCAgJEAAYFhgeqAIAsAIA&sourceid=chrome&ie=UTF-8#ip=1)?

anvit: @sbreker Ah I see! I'd just noticed that we don't abbreviate them in arCasPlugin, but I think it's fine if they're common and used elsewhere.

plugins/arOidcPlugin/lib/oidcUser.class.php
```diff
@@ -0,0 +1,359 @@
...
+    {
+        $idToken = $this->getAttribute('oidc-token', null);
+        $this->unsetAttributes();
+        $this->signOut();
+
+        if (true == sfConfig::get('app_oidc_send_oidc_logout', false) && !empty($idToken)) {
+            $logoutRedirectUrl = sfConfig::get('app_oidc_logout_redirect_url', '');
+            if (empty($logoutRedirectUrl)) {
+                $logoutRedirectUrl = null;
+                $this->logger->err('Setting "app_oidc_logout_redirect_url" invalid. Unable to redirect on signout.');
```
melaniekung: small typo: "Unable to redirect on sign out."?

plugins/arOidcPlugin/config/app.yml
melaniekung, line 1: is it possible to use config variables to generate `$redirect_url` and `$logout_redirect_url` to avoid the need for users/sysadmins to manually edit this file to configure OIDC authentication?

melaniekung (commented):
nice job, steve! minor typo and question, but otherwise looks good to me!

plugins/arOidcPlugin/config/app.yml
sbreker: I think it would be undesirable to have these values be configurable in the WebUI since these values should never really change once set up, plus they should only be set by a sysadmin who knows what the implications are. Any updates to these values are likely to break the log in or log out flow.

The CAS plugin similarly contains these type of redirect values in its `app.yml` file.

plugins/arOidcPlugin/lib/oidcUser.class.php
sbreker: Updated.

anvit (approved):
There are a few files that use `elseif` instead of `else if` that's probably causing the tests to fail, but other than that this looks good to me!

sbreker:
> There are a few files that use `elseif` instead of `else if` that's probably causing the tests to fail, but other than that this looks good to me!

Thanks @anvit and @melaniekung! I have made the last change to satisfy PHP_cs_fixer. I will rebase, squash and merge.

plugins/arOidcPlugin/config/arOidcPluginConfiguration.class.php
sbreker: @anvit - I have replaced these abbreviations with 'authentication'

PR #1754, opened 2024-02-13
Fix GUI CSV Actor export for translations. (#1752)

### LINKED ISSUE #1752: Problem: GUI CSV export for actors only outputs single culture
### Current Behavior

1. Run CSV export for a actors available in multiple languages.
2. Resulting CSV only exports in 1 language.

### Expected Behavior

CSV export should export actors in all available languages.

melaniekung:


lib/job/arActorExportJob.class.php
```diff
@@ -71,8 +71,33 @@ protected function doExport($path)
                 return;
             }
 
-            $this->exportResource($resource, $path);
+            $this->csvActorExport($path, $resource);
             $this->logExportProgress();
         }
     }
+
+    protected function csvActorExport($path, $resource)
```
anvit: I would suggest changing the name of this function to avoid confusion with the `csvActionExport` class

lib/job/arActorExportJob.class.php
```diff
@@ -71,8 +71,33 @@ protected function doExport($path)
...
+        $cultures = array_keys(DefaultTranslationLinksComponent::getOtherCulturesAvailable($resource->actorI18ns, 'authorizedFormOfName', $resource->getAuthorizedFormOfName(['sourceCulture' => true])));
+
+        // Write row to file and initialize row
+        foreach ($cultures as $culture) {
+            $actor = QubitActor::getById($resource->id);
+            $this->context->getUser()->setCulture($culture);
+
+            $writer->exportResource($actor);
+
+            ++$itemsExported;
```
anvit: Are we using this count or logging this somewhere? If not, we can skip this variable entirely.

anvit (changes requested):
I have a couple of small nitpicks but looks good otherwise!

lib/job/arActorExportJob.class.php
melaniekung: `$itemsExported` is used in logs to show how many records were exported. although, with my new implementation, it's reporting number of rows exported, so i could move it out of the loop to accurrately represent items exported? (but descriptions are also reporting 2 records)

anvit (approved):

PR #1770, opened 2024-03-05
Update OIDC login action flow

sbreker:
Update logic to protect authenticate() if the user is not authenticated, and is redirected back to oidc/login.

Add logic to support redirecting the user back to their requested page when authentication succeeds.


anvit (approved):
LGTM!

PR #1779, opened 2024-03-19
Fix untranslated settings yes/no choices (#1772)

### LINKED ISSUE #1772: Problem: untranslated yes/no choice in some settings tabs
### Current Behavior

Sorry if anything isn't done correctly, this is my first time using git.
Currently in several settings tabs the yes/no choice is untranslated. A check on Weblate didn't yield any missing string in that regard. This concern the tabs:
* DIP upload settings
* Global
* Markdown
* Permissions
* Security

image
Other tabs are fine; on e.g. Identifiers, the option is translated.

image

The behaviour is observed with every language other than english.

### Expected Behavior

The yes/no choice should be translated.

### Possible Solution

I'm not a developer, so I can only suggest that the hardcoding of 'yes' and 'no' in lib/form/SettingsGlobalForm.class.php might have to do with it. On the other hand apps/qubit/modules/settings/templates/identifierSuccess.php doesn't have any of those.

### Context and Notes

_No response_

### Version used

2.8.0 - 193

_No response_

fr

### PHP version

_No response_

### Contact details

_No response_

melaniekung:
The following settings tabs have been fixed:

- DIP upload settings
- Global
- Markdown
- Permissions
- Security

### DIFF
--- a/apps/qubit/modules/settings/actions/dipUploadAction.class.php
+++ b/apps/qubit/modules/settings/actions/dipUploadAction.class.php
@@ -39,7 +39,8 @@ class SettingsDipUploadAction extends SettingsEditAction
     {
         switch ($name) {
             case 'stripExtensions':
-                $this->form->setWidget($name, new sfWidgetFormSelectRadio(['choices' => [1 => 'yes', 0 => 'no']], ['class' => 'radio']));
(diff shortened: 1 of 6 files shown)


anvit (approved):
Looks great! Thanks @melaniekung

PR #1780, opened 2024-03-20
Fix tooltips for BS5 (#1777)

### LINKED ISSUE #1777: Problem: settings to enable or disable tooltips do not work in BS5 templates
### Current Behavior

Noticed during a client training workshop yesterday.

1. Ensure that you are using the BS5 templates
2. Navigate to Admin > Settings > Global > Presentation, set "Show Tooltips" to `no`, and save
3. Navigate to Add > Archival description and inspect the resulting empty edit template

**Error encountered**

Because tooltips no longer appear in a JS popup beside the main body column of the edit page (in the BS5 templates, they appear directly under the edit field), the current setting to hide them does nothing. The tooltips are still visible in the edit page after changing the setting. 

### Expected Behavior

When the setting to show tootlips is set to "no," tooltips are not shown.

### Possible Solution

If the changes to the template layout make it hard or impossible to fix the expected setting behavior, one alternative would be to simply remove the setting. 

### Context and Notes

_No response_

### Version used

2.8.1 - 193

Docker env 

en

### PHP version

PHP 7.4

### Contact details

_No response_

melaniekung:


plugins/arDominionB5Plugin/templates/_layout_start_webpack.php
```diff
@@ -4,6 +4,7 @@
     <meta charset="utf-8">
     <meta name="viewport" content="width=device-width, initial-scale=1">
     <?php include_title(); ?>
+    <?php include_stylesheets(); ?>
```
jraddaoui: You don't want to include all the B2 CSS in B5.

css/hide_tooltips.css
```diff
@@ -1,3 +1,9 @@
+/* Hide tooltips for BS2 */
 .description {
   display: none;
 }
+
+/* Hide tooltips for BS5 */
+[id*='\-help'] {
+  display: none;
+}
```
jraddaoui: Some notes:

- This should go in the B5 theme SCSS files.
- How are they going to be shown based on the setting? This won't work with the existing B2 JS (which I believe is not even included in the B5 theme).
- Based on the discussion with @fiver-watson in Slack, we should only hide/show the help text for "standards-based guidance in the edit forms", not all input's help text.

And suggestions:

- Because they are no longer tooltips, instead of using JS to show/hide them, I'd add a class to the body based on the setting value so they are only hidden when the setting is set to not show tooltips. And add a new class to the help text that needs to be shown/hidden.
- I'd also change the setting name to something more accordingly with the current implementation, like "Show standards-based help text" instead of "Show Tooltips", but check with @fiver-watson first ;)

jraddaoui (changes requested):

css/hide_tooltips.css
melaniekung: hi @jraddaoui ! i looked into doing it only in BS5 but that would be a lot more work - which is fine but i want to first argue my approach XD

first off, these `*-help` ids are only used in the `arDominionB5Plugin` directory (checked by running `grep -ri --include='*.php' 'id.*\-help'`) and these changes doesn't affect how they work in BS2.

it is triggered by the [QubitDescription.class.php](https://github.com/artefactual/atom/blob/qa/2.x/lib/QubitDescription.class.php), so the setting will toggle these on/off, and only those for in the forms in the `add` tab, not all `*-help` classes (so those in settings, csv, etc. are all excluded from this change). because of this, i had to use the `<?php include stylesheets(); ?>` in the `_layout_start_webpack.php` for this change to take affect.

if you still think this approach is not best practice and should be addressed only in `arDominionB5Plugin`, i can definitely try that too!

jraddaoui: I see, thanks @melaniekung!

I still think it should go in the B5 theme, it will be easy to forget this when the B2 themes are fully removed. Also, `<?php include stylesheets(); ?>` may be including other styles that we don't want in B5.

cypress/e2e/descriptions.cy.js
```diff
@@ -7,7 +7,7 @@ describe('Descriptions', () => {
     cy.visit('/informationobject/add')
     cy.contains('Identity area').click()
     cy.get('input#identifier').type('123')
-    cy.get('input#title').type('Test description A')
+    cy.get('input#title').type('Test description A', {force: true});
```
jraddaoui: Why is this needed?

lib/QubitDescription.class.php
```diff
@@ -24,7 +24,24 @@ public static function addAssets(sfWebResponse $response)
...
+                                    tooltip.removeAttribute('data-bs-toggle');
+                                    tooltip.removeAttribute('data-bs-original-title');
+                                });
+                            });
+                        });
+                    });
+                </script>
+            <?php } else {
+                $response->addStylesheet('hide_tooltips', 'last');
+            }
```
jraddaoui: One of the goals from the B5 project was to have a better separation of concerns in the code. I think we don't need this as I'll explain below, but if we do, I'd love to see it in a JS file, so we can benefit from the proper tools for it (syntax checks, dep management, bundling, etc).

plugins/arDominionB5Plugin/js/tooltip.js
```diff
@@ -1,5 +1,17 @@
-import Tooltip from "bootstrap/js/dist/tooltip";
```
jraddaoui, line 1: I'd keep using the import, we may stop importing the entire Bootstrap bundle or we could stop exposing it as a global variable at some point.

plugins/arDominionB5Plugin/js/tooltip.js
```diff
@@ -1,5 +1,17 @@
...
-
-document
-  .querySelectorAll("[data-bs-toggle=tooltip]")
-  .forEach((element) => new Tooltip(element));
```
jraddaoui, line 5: Removing this will break all other tooltips.

plugins/arDominionB5Plugin/js/tooltip.js
```diff
@@ -1,5 +1,17 @@
...
+document.addEventListener("DOMContentLoaded", function () {
```
jraddaoui: I think we don't need an event listener for this case.

plugins/arDominionB5Plugin/js/tooltip.js
```diff
@@ -1,5 +1,17 @@
...
+  if (
+    window.location.pathname.endsWith("/add") ||
+    window.location.pathname.endsWith("/edit")
```
jraddaoui: Instead of checking the window location path I'd check the body classes. The add and edit pages should have an `edit` class in there.

plugins/arDominionB5Plugin/js/tooltip.js
```diff
@@ -1,5 +1,17 @@
...
+  ) {
+    var elements = document.querySelectorAll('[id$="-help"]');
+    elements.forEach(function (idElement) {
+      idElement.style.display = "none";
```
jraddaoui: Because this will always be needed, I'd do it in the SCSS code. Ideally, we would do it in the templates, but that's going to require a lot of changes in them.

plugins/arDominionB5Plugin/js/tooltip.js
```diff
@@ -1,5 +1,17 @@
...
+      parentElement = idElement.previousElementSibling;
```
jraddaoui: This won't work for autocomplete fields.

plugins/arDominionB5Plugin/js/tooltip.js
```diff
@@ -1,5 +1,17 @@
...
+      if (parentElement != null) {
+        parentElement.setAttribute("data-bs-toggle", "tooltip");
+        parentElement.setAttribute("title", idElement.textContent.trim());
+        var tooltip = new bootstrap.Tooltip(parentElement);
```
jraddaoui: Instead of enabling the tooltips here always and disabling them in the PHP code based on the setting value, we should only enable them here based on the setting value. To do so, we could use a class in the body tag, like I mentioned we have for the add/edit pages.

jraddaoui: For example:

```diff
diff --git a/plugins/arDominionB5Plugin/templates/_layout_start_webpack.php b/plugins/arDominionB5Plugin/templates/_layout_start_webpack.php
index 547251694..6756d6791 100644
--- a/plugins/arDominionB5Plugin/templates/_layout_start_webpack.php
+++ b/plugins/arDominionB5Plugin/templates/_layout_start_webpack.php
@@ -9,7 +9,7 @@
     <%= htmlWebpackPlugin.tags.headTags %>
     <?php echo get_component_slot('css'); ?>
   </head>
-  <body class="d-flex flex-column min-vh-100 <?php echo $sf_context->getModuleName(); ?> <?php echo $sf_context->getActionName(); ?>">
+  <body class="d-flex flex-column min-vh-100 <?php echo $sf_context->getModuleName(); ?> <?php echo $sf_context->getActionName(); ?><?php echo sfConfig::get('app_show_tooltips') ? ' show-edit-tooltips' : ''; ?>">
     <?php echo get_component('default', 'tagManager', ['code' => 'noscript']); ?>
     <?php echo get_partial('header'); ?>
     <?php include_slot('pre'); ?>
```

plugins/arDominionB5Plugin/js/tooltip.js
jraddaoui: For example:

```diff
diff --git a/plugins/arDominionB5Plugin/scss/_forms.scss b/plugins/arDominionB5Plugin/scss/_forms.scss
index f19711f28..5f940cff9 100644
--- a/plugins/arDominionB5Plugin/scss/_forms.scss
+++ b/plugins/arDominionB5Plugin/scss/_forms.scss
@@ -94,3 +94,7 @@ ul.multiInput,
 .form-required {
   color: $primary;
 }
+
+body.edit [id$=-help] {
+  display: none;
+}
```

You could make sure they are inside `.accordion-body` but that's probably not needed.

plugins/arDominionB5Plugin/js/tooltip.js
```diff
@@ -1,5 +1,17 @@
...
+      }
+    });
+  }
+});
```
jraddaoui: I gave it a try and this is the `tooltip.js` file I ended with, using jQuery:

```js
import Tooltip from "bootstrap/js/dist/tooltip";

(($) => {
  "use strict";

  $(() => {
    $("body.edit.show-edit-tooltips [id$=-help]").each((_, el) => {
      $(el)
        .prevAll(":not([type=hidden]):first")
        .attr({ "data-bs-toggle": "tooltip", title: el.textContent.trim() });
    });

    $("[data-bs-toggle=tooltip]").each((_, el) => new Tooltip(el));
  });
})(jQuery);
```

Some notes:
- The first function (`(($) => {`) isolates this code from the global scope.
- The second one (`$(() => {`) makes sure the code is executed when the DOM has been loaded.
- The selector `body.edit.show-edit-tooltips [id$=-help]` checks that we are on an edit page with the show tooltips setting enabled.
- `prevAll(":not([type=hidden]):first")` gets the first non hidden previous sibling (for the autocompletes issue).
- `attr()` sets multiple attributes at the same time.
- Finally, we initialize all tooltips on the page.

jraddaoui (changes requested):
Thanks @melaniekung, nice that you gave it a try to use proper tooltips in the end!

plugins/arDominionB5Plugin/js/tooltip.js
```diff
@@ -1,5 +1,16 @@
...
-  .forEach((element) => new Tooltip(element));
+(($) => {
+  "use strict";
+
+  $(() => {
+    $("body.edit.show-edit-tooltips [id$=-help]").each((_, el) => {
+      $(el)
+        .prevAll(":not([type=hidden]):first")
+        .attr({ "data-bs-toggle": "tooltip", title: el.textContent.trim() })
+        .attr({ "data-bs-placement": "right" });
```
jraddaoui: All three could go in one object and one `attr()` call.

plugins/arDominionB5Plugin/templates/_layout_start_webpack.php
```diff
@@ -9,7 +9,7 @@
     <%= htmlWebpackPlugin.tags.headTags %>
     <?php echo get_component_slot('css'); ?>
   </head>
-  <body class="d-flex flex-column min-vh-100 <?php echo $sf_context->getModuleName(); ?> <?php echo $sf_context->getActionName(); ?>">
+  <body class="d-flex flex-column min-vh-100 <?php echo $sf_context->getModuleName(); ?> <?php echo $sf_context->getActionName(); ?><?php echo sfConfig::get('app_show_tooltips') ? ' show-edit-tooltips' : ''; ?>">
```
jraddaoui, line 12: Thinking about this again, while I still think is the best solution for us, we should take a note for custom theme upgrades, as it will be needed in all B5 themes _layout_start_webpack.php file.

jraddaoui (approved):
Thanks @anvit!

anvit:
@melaniekung this looks good to go!

@jraddaoui we made a couple of small additional changes after we got some feedback from folks. We're setting these tooltips to trigger only on focus, and not both focus and hover (which is also consistent with how BS2's tooltips work).

Also Firefox tests seem to be failing, but that's likely a recent problem with Cypress because branches that were passing a couple days ago that have nothing to do with tooltips are failing the same tests now.

jraddaoui:
Cool, thanks for the update @anvit, looks good to me!

PR #1784, opened 2024-03-26
Fix clipboard for empty browser cache (#1783)

### LINKED ISSUE #1783: Problem: cannot add items to clipboard
### Current Behavior

1. Find any entry
2. Click Add to Clipboard
3. Nothing is added to the clipboard and we can see this error in the console:
  ```
   Uncaught TypeError: this.items is null
   toggle http://localhost:63001/js/clipboard.js:351
   ```

### Expected Behavior

The item should be added to the clipboard...

### Possible Solution

The problem seems to have appeared in this commit: `Fix clipboard storage behaviour (#1749)` (9cf5ac4ea4c31c2b8179d72a812bbfcf634e530d)

### Context and Notes

_No response_

### Version used

AtoM 2.8.1

Ubuntu 22.04

en

### PHP version

PHP 7.4

### Contact details

_No response_

anvit:
Set browser local storage values for clipboard if their corresponding variables are null on load, which happens if the browser storage is cleared.

### DIFF
--- a/js/clipboard.js
+++ b/js/clipboard.js
@@ -12,19 +12,8 @@
     this.storage = localStorage;
     this.types = ['informationObject', 'actor', 'repository'];
     this.initialItems = {'informationObject': [], 'actor': [], 'repository': []};
-    this.items = JSON.parse(this.storage.getItem('clipboard'));
-    this.exportTokens = JSON.parse(this.storage.getItem('exportTokens'));
-
-    if (!this.items)
-    {
-      this.items = this.initialItems;
-    }
-
-    if (!this.exportTokens)
-    {
-      this.exportTokens = [];
-    }
-
+    this.items = JSON.parse(this.storage.getItem('clipboard')) || this.initialItems;
+    this.exportTokens = JSON.parse(this.storage.getItem('exportTokens')) || [];
     this.init();
   };
 
@@ -344,7 +333,7 @@
 
       // Load items from local storage in case activity
       // in another tab has changed the content
-      this.items = JSON.parse(this.storage.getItem("clipboard"));
+      this.items = JSON.parse(this.storage.getItem("clipboard")) || this.initialItems;
 
       var type = $button.data('clipboard-type');
       var slug = $button.data('clipboard-slug');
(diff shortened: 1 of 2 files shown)


js/clipboard.js
```diff
@@ -18,11 +18,13 @@
...
     {
       this.items = this.initialItems;
+      this.storage.setItem('clipboard', JSON.stringify(this.initialItems));
     }
 
     if (!this.exportTokens)
     {
       this.exportTokens = [];
+      this.storage.setItem('exportTokens', JSON.stringify([]));
     }
```
jraddaoui: I'd change the following block:

```js
    this.items = JSON.parse(this.storage.getItem('clipboard'));
    this.exportTokens = JSON.parse(this.storage.getItem('exportTokens'));

    if (!this.items)
    {
      this.items = this.initialItems;
      this.storage.setItem('clipboard', JSON.stringify(this.initialItems));
    }

    if (!this.exportTokens)
    {
      this.exportTokens = [];
      this.storage.setItem('exportTokens', JSON.stringify([]));
    }
```

To this:

```js
    this.items = JSON.parse(this.storage.getItem('clipboard')) || this.initialItems;
    this.exportTokens = JSON.parse(this.storage.getItem('exportTokens')) || [];
```

I know that's just an optimization that won't fix the issue, but then you can do the same inside the toggle functions:

```js
    this.items = JSON.parse(this.storage.getItem('clipboard')) || this.initialItems;
```

I think that's better than writing the default/empty values into localstorage.

jraddaoui (changes requested):
Thanks @anvit!

js/clipboard.js
anvit: @jraddaoui Makes sense, this makes the code much cleaner! I've updated it, thanks!

plugins/arDominionB5Plugin/js/clipboard.js
```diff
@@ -11,21 +11,13 @@ import Tooltip from "bootstrap/js/dist/tooltip";
 
       this.storage = localStorage;
       this.types = ["informationObject", "actor", "repository"];
-      this.initialItems = JSON.stringify({
+      this.initialItems = {
```
jraddaoui: Nitpick: it could go in one line now.

jraddaoui (approved):
Thanks!

jraddaoui:
Make sure you make Prettier happy ;)

PR #1788, opened 2024-03-27
Automatically orient scaled images (#13606)

mcantelon:
Added option to ImageMagick convert tool to automatically orient scaled images.

### DIFF
--- a/plugins/sfThumbnailPlugin/lib/sfImageMagickAdapter.class.php
+++ b/plugins/sfThumbnailPlugin/lib/sfImageMagickAdapter.class.php
@@ -333,7 +333,7 @@ class sfImageMagickAdapter
                 break;
         } // end switch
 
-        $command .= ' -thumbnail ';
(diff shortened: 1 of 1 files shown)


mcantelon:
My commit references an issue on the old issue server. Do I need to make an issue on GitHub then change the number to that?

anvit (approved):
Looks good! @mcantelon I don't think there's a need to make a new GitHub issue for that, but maybe change the commit to say ref#13606 (without parenthesis, like the old commit format) to avoid GitHub from trying to connect it to an issue here?

fiver-watson:
FYI @mcantelon it is the reference image that is most often affected by this issue. Will this address possible rotation issues for both, or just the thumbnail?

mcantelon:
@fiver-watson It should be for all scaling.

mcantelon:
@anvit Thanks Anvit... good catch... I'll amend the commit message before merging!

PR #1795, opened 2024-04-03
Add PDF page D.O. support to regen. task (#1794)

### LINKED ISSUE #1794: Problem: Images rendered from PDF pages get removed, but not regenerated, when digital objects regeneration task is run
### Current Behavior

Here are the steps needs to enable the setting and upload a PDF:

1. Navigate to settings/uploads.
2. Change the "Upload multi-page files as multiple descriptions" setting to "Yes".
3. Click the "Save" button.
4. Navigate to informationobject/addand add an information object.
5. Once you've created the information object then, for the "More" select menu, select "Link Digital Object".
6. Click "Choose file", select a multi-page PDF file, then click the "Create" button.
7. Child information objects should now have been added for each page of the PDF.

Note the slug of the PDF's information object for later.

Then, from the command-line in the AtoM root directory, run the digital objects regeneration task:

```
php symfony digitalobject:regen-derivatives --force --index
```

Now navigate to the slug noted earlier or refresh the page if already on it. Child information objects for each PDF page should still exist, but the images for them will now be gone.

### Expected Behavior

PDF page images should be regenerated rather than just deleted.

### Possible Solution

I've made a fix PR that I'll submit.

### Context and Notes

AtoM has a setting that allows PDFs, uploaded as digital objects to information objects, to be "exploded" into child information objects, for each of the PDF's pages, with each information object having a digital object attached that's an image rendered from the PDF page.

### Version used

2.8.2 - 193

Ubuntu 20.04

en

### PHP version

PHP 7.4

### Contact details

mike@artefactual.com

mcantelon:
Fixed issue with the digital objects regeneration task (digitalobject:regen-derivatives) deleting, but not regenerating, digital objects representing PDF pages.

Removed unneeded and unused digital object class method.

### DIFF
--- a/lib/model/QubitDigitalObject.php
+++ b/lib/model/QubitDigitalObject.php
@@ -1662,18 +1662,6 @@ class QubitDigitalObject extends BaseDigitalObject
     }
 
     /**
-     * Return true if this is a compound digital object.
-     *
-     * @return bool
-     */
-    public function isCompoundObject()
-    {
-        $isCompoundObjectProp = QubitProperty::getOneByObjectIdAndName($this->id, 'is_compound_object');
-
-        return null !== $isCompoundObjectProp && '1' == $isCompoundObjectProp->getValue(['sourceCulture' => true]);
-    }
-
-    /**
      * Derive file path for a digital object asset.
      *
      * All digital object paths are keyed by object id that is the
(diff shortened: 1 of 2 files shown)


lib/model/QubitDigitalObject.php
```diff
@@ -1661,18 +1661,6 @@ public function isImage()
         return self::isImageFile($this->getName());
     }
 
-    /**
-     * Return true if this is a compound digital object.
-     *
-     * @return bool
-     */
-    public function isCompoundObject()
```
anvit, line 1669: Is this function no longer being used anywhere?

mcantelon: I grepped and could find any use of it, oddly! If you could double check, just in case I missed something, that'd be awesome.

anvit (approved):
Just double checked and there isn't another usage of `isCompoundObject `. Thanks @mcantelon

mcantelon:
Thanks @anvit !

PR #1796, opened 2024-04-04
Generate AtoM SBOM in CI

sbreker:
Add ability to generate and upload an AtoM SBOM to Dependency Track in CI.

This CI workflow will autocreate a new Dependency Track project for AtoM for any qa or stable branch this workflow is picked to. The projectName will be set to "atom" and the project version will be the branch name.

Because the parentName is set to 'AtoM', this SBOM will be grouped under an 'AtoM' project group which must be present in advance.

### DIFF
--- a/.github/workflows/generate-sbom.yml
+++ b/.github/workflows/generate-sbom.yml
@@ -1,0 +1,59 @@
+name: Generate and upload SBOM
+
+on:
+  push:
+    branches:
+    - qa/**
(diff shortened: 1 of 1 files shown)


sbreker:
I have added a second commit containing the `--include-dev-deps` option. Do you think this is something that should be monitored for this project?

melaniekung (approved):
LGTM, thanks steve!

PR #1800, opened 2024-04-16
Fix Trivy caching in CI

sbreker:
Update Trivy SBOM generation to invalidate the Trivy cache when package-lock.json or composer.lock are modified.


anvit (approved):
Looks good!

PR #1806, opened 2024-04-30
Fix translationLinksComponent dropdown (#1805)

### LINKED ISSUE #1805: Problem: Other language dropdown is empty
### Current Behavior

1. Go to an archival description that has translations available in a second language.
2. Click the `Other languages available` dropdown button. 

Notice that the dropdown results are empty.

### Expected Behavior

Clickable results for other available languages should pop up.

### Possible Solution

This bug seems to have been introduced in to #1751 

### Version used

AtoM 2.8.1

anvit:
The $translations variable in translationLinksComponent needs to be a property of $this in order to accessible by the template. Also adding in a currentCulture flag to the getOtherCulturesAvailable method so that the current culture can be excluded from the results.

### DIFF
--- a/apps/qubit/modules/default/actions/translationLinksComponent.class.php
+++ b/apps/qubit/modules/default/actions/translationLinksComponent.class.php
@@ -110,15 +110,19 @@ class DefaultTranslationLinksComponent extends sfComponent
             return sfView::NONE;
         }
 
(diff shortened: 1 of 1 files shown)


djjuhasz (approved):
LGTM :+1:

PR #1815, opened 2024-05-10
Add OIDC plugin create user setting

sbreker:
Add new setting to control the auto creation of AtoM user records. If set to 'true' AtoM user records will be automatically created on first login. If set to 'false' then a user record will not automatically be created - in this case the user must be manually created beforehand. Default value for this setting is 'true'.


melaniekung (approved):
LGTM, thanks @sbreker !

PR #1822, opened 2024-05-24
Allow manual user creation based on  OIDC setting

sbreker:
If OIDC Plugin setting auto_create_atom_user is set to false, allow manual creation of AtoM users in the AtoM UI. Passwords will not be able to be set since the authorization will come from the OIDC endpoint.


anvit (approved):
Looks good @sbreker !

PR #1824, opened 2024-06-11
Allow selection of secondary OIDC provider

sbreker:
Add the ability to define a secondary OIDC provider for authentication and select it using a query param on the request URL.

E.g. Use the 'sample_provider' provider by modifying the AtoM URL before pressing "Log in with SSO":

http://127.0.0.1:63001/index.php?secondary=sample_provider

### DIFF
--- a/plugins/arOidcPlugin/lib/oidcUser.class.php
+++ b/plugins/arOidcPlugin/lib/oidcUser.class.php
@@ -43,15 +43,24 @@ class oidcUser extends myUser implements Zend_Acl_Role_Interface
      *
      * @param null|mixed $username
      * @param null|mixed $password
-     *
-     * @return bool
      */
-    public function authenticate($username = null, $password = null)
+    public function authenticate($username = null, $password = null): bool
     {
         $authenticated = false;
         $user = null;
         $authenticateResult = false;
         $email = null;
+
+        // Get provider ID from session storage as it may have been set elsewhere.
+        $providerId = $this->getSessionProviderId();
+        // Validate and set provider ID in session.
+        if (null !== $providerId = $this->validateProviderId($providerId, true)) {
+            // Set Provider details in OIDC client.
+            $result = $this->setOidcProviderDetails($providerId);
+        }
+        if (null === $providerId || !isset($result) || false === $result) {
+            return $authenticated;
+        }
 
         if (isset($_REQUEST['code'])) {
             $this->logger->info('OIDC request "code" is set.');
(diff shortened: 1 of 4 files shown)


plugins/arOidcPlugin/lib/oidcUser.class.php
```diff
@@ -43,16 +43,27 @@ public function initialize(sfEventDispatcher $dispatcher, sfStorage $storage, $o
...
      */
-    public function authenticate($username = null, $password = null)
+    public function authenticate($username = null, $password = null): bool
     {
         $authenticated = false;
         $user = null;
         $authenticateResult = false;
         $email = null;
 
+        $this->logger->err(sprintf('%s', json_encode(sfConfig::getAll())));
```
anvit: Was this a debug message?

plugins/arOidcPlugin/lib/oidcUser.class.php
```diff
@@ -262,13 +275,135 @@ public function logout()
...
-                $this->oidcClient->signOut($idToken, $logoutRedirectUrl);
+                // Get saved session provider id.
+                $providerId = $this->getSessionProviderId();
+                // Set provider details in the OIDC client using provider id.
+                if (true === $result = $this->setOidcProviderDetails($providerId)) {
+                    $this->oidcClient->signOut($idToken, $logoutRedirectUrl);
+                }
+                $this->setProviderId();
             } catch (Exception $e) {
+                $this->setProviderId();
```
anvit: Is this repeated function call inside the catch a typo?

anvit (changes requested):
I think I found a couple minor lines of code that were accidentally left in, but everything else looks great!

plugins/arOidcPlugin/lib/oidcUser.class.php
sbreker: Yes it is! :laughing: :facepalm: 

Thanks for catching this! :+1:

plugins/arOidcPlugin/lib/oidcUser.class.php
sbreker: I was trying to ensure that setProviderId() would be run even if there was an exception generated in signOut().

I am wondering now if I could just move getSessionProviderID() and setProviderId() to before the try{}. I will have a closer look.

plugins/arOidcPlugin/lib/oidcUser.class.php
sbreker: Removed - see latest commit

plugins/arOidcPlugin/lib/oidcUser.class.php
sbreker: Updated and tested. See latest commit.

anvit (approved):
Looks good! Thanks @sbreker

PR #1833, opened 2024-06-21
Small spelling fix

misilot:
Adds a space between between "a" and "period"

### DIFF
--- a/apps/qubit/modules/clipboard/actions/exportAction.class.php
+++ b/apps/qubit/modules/clipboard/actions/exportAction.class.php
@@ -293,7 +293,7 @@ class ClipboardExportAction extends DefaultEditAction
 
         $responseData['success'] .= '</p><p>';
         $responseData['success'] .= $this->context->i18n->__(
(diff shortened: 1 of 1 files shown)


anvit:
Hi @misilot ! Thanks for finding and fixing this typo! We'll include this fix in the next AtoM release.

There are a couple things we'd need to do before we can accept this change. The update to `exportAction.class.php` looks good but the message xml files are generated when we run our script for pulling translations, so I would suggest reverting those since those will be fixed automatically (with the appropriate timestamp changes to the translation files) when we run the translations pull from weblate for the next release. 

Also, when you get a chance, would you be able to sign our Contributor Agreement? Here's a link with details about the agreement — https://wiki.accesstomemory.org/Development/Contribute_code#Copyright_and_license

misilot:
Is this a different CLA than already on file with Artefactual?

anvit:
@misilot received, thanks! The only other signed CLA I see is for Archivematica, and we have a separate one for AtoM.

anvit (approved):
Looks great, thanks!

PR #1835, opened 2024-06-25
Clear template cache when authenticating OIDC usr

sbreker:
Address OIDC authentication issue where cached menu templates are not refreshed on login.

When a user's groups are updated in the identity provider, this fix ensures that the main menu icons are updated to show all admin functions.


anvit (approved):
Looks good

PR #1836, opened 2024-07-09
Allow independent OIDC provider config

sbreker:
Allow provider specific config values to be set independently for each
provider instead of these settings applying to every configured OIDC
provider.


anvit (approved):
These look good to me! Thanks @sbreker !

misilot:
@sbreker are there any tips for implementing `getProviderConfigValue` with CAS? As this function doesn't exist and was added in 2.9. (See #1983)

Thanks!

Issue #1509, opened 2022-05-02
Accessibility


doana:
Hi folks, 

I wanted to bring the issue of web accessibility to your attention for this project as there are a few low-hanging fruit that could be addressed in order to create a much better experience for users of assistive technology like screen readers. 

Our local legislation references an international accessibility standard - The [Web Content Accessibility Guidelines (WCAG) version 2.1](https://www.w3.org/TR/WCAG21/). This document would be a great resource for the implementation of accessibility features in AtoM.  

In particular, the most prevalent issues that I saw in my quick checking were: 
1. Low [contrast](https://www.w3.org/WAI/WCAG21/Understanding/contrast-minimum.html) between text and its background in the Dominion theme. I haven't checked the Archives Canada theme for this issue. Examples of this issue include the link color, orange h2's, light grey h3's.
2. A lack of [labels for form fields](https://www.w3.org/WAI/WCAG21/Techniques/general/G131.html) sporadically throughout the application including the search and advanced search forms

If these issues could be addressed, it would go a long way toward creating a better experience for all users of AtoM.

fiver-watson:
Hi @doana , 

Thanks for raising this issue. It's one we are aware of, and we do have some very relevant work coming soon, described further below. First, I just wanted to mention that thus far, the greatest challenge for us in doing work like this to enhance AtoM has been our current business model. We have always strived for openness, and release not only AtoM under an open / libre license, but all supporting materials as well like our documentation, presentations and slides, wiki content, forum support, and more. Consequently, while our team invests a lot of time and effort in addressing bugs, maintenance upgrades, and performance improvements with each release, we have relied on our community to support feature enhancements and major dependency upgrades, either in the form of code contributions or via community development sponsorship. You can read about the history and development philosophy of the AtoM project thus far on our wiki, here: 

* https://wiki.accesstomemory.org/Development/Philosophy

Artefactual is in the process of trying to evolve our business model so we can more easily maintain and develop our applications (and create new ones), without compromising on our values and commitment to openness. This is a slow process that will take time to implement - so in the meantime, if you'd like to submit a pull request addressing some of these elements, we have some guidance [on our wiki](https://wiki.accesstomemory.org/Development#Development_resources); or if your institution might be interested in sponsoring accessibility development in AtoM, feel free to contact our team at info@artefactual.com to discuss options. 

That said: 

[...]

PR #1840, opened 2024-07-16
Improve the contrast

eviau-artefactual:
As part of https://github.com/artefactual/atom/issues/1509

I chose the colour by using a darker shade of orange: #c44e00 instead of #f60.

We can pick another colour - I only tested on Firefox using the integrated accessibility tools.

If my understanding of SCSS is correct, this should build correctly, given that I am unable to build and access this version to test it locally with Docker, following the instructions in https://www.accesstomemory.org/fr/docs/2.8/dev-manual/env/compose/ ... do these docs needs a review ?

### DIFF
--- a/plugins/arDominionB5Plugin/scss/_variables.scss
+++ b/plugins/arDominionB5Plugin/scss/_variables.scss
@@ -13,7 +13,7 @@ $gray-700: #495057 !default;
 $gray-800: #343a40 !default;
 $gray-900: #212529 !default;
 $black: #000 !default;
-$orange: #f60 !default;
+$orange: #c44e00 !default;
 $blue: #0a58ca !default;
 $primary: $orange !default;
 $secondary: $gray-700 !default;


anvit:
The change looks good to me, but tagging @fiver-watson in case there were historical branding or UX reasons for sticking to that specific shade of orange that I might not be aware of.

fiver-watson:
Looks okay to me, and seems to meet WCG AA requirements: 

atom-orange-contrast-check-black

Note: it would be good to double-check, but looking through the demo site, I couldn't actually find places where we use small black text on an orange background, so I'm not too concerned about this result. In general, we use orange text on white backgrounds, or the opposite: 

atom-orange-contrast-check-white
atom-orange-contrast-check-white-bg

(sidenote, I don't think we will ever achieve AAA compliance in the base theme. Instead, I have always thought that instead of the ArchivesCanada theme as a secondary theme, we should have an accessibility-focused second theme, that includes high contrast color changes, built-in buttons to resize page text, etc... but that's a whole separate thing). 

Additionally, AFAIK we already are not using the same orange as Artefactual has used elsewhere in its branding. The Circle-A logo appears to use #F56400; the AtoM logo used in the header of AtoM itself uses #FF6600; the Artefactual logo used in the header of the company website looks like the A is using #F36F2B; the Arty logo showing in place of Donald's face in the Teams section of the website uses #F46400, etc....

You could always check with BUD but so far, we have definitely NOT been consistent in our use of orange in our branding, so personally I think that adjusting this color to better support contrast for accessibility is totally fine!

anvit (approved):

PR #1841, opened 2024-07-16
Update Bootstrap 5 dependency to current version

anvit:
Unpinned Bootstrap 5 and updated it to the current version. Due to breaking changes introduced by Bootstrap in v5.2.0 and v5.3.0, and additional import for maps was needed, updating links using the heading class to use the default link colour was needed to be set, and updating the default background colour variable for cards and list group items needed to be set.


melaniekung (approved):
LGTM, thanks @anvit !

PR #1849, opened 2024-07-25
Update pcov unit test coverage config

sbreker:
Exclude AtoM's docker directory from pcov unit test coverage report. Add pcov option to ensure all folders are scanned for unit test coverage (plugin folder tests were not being discovered).


anvit (approved):
Looks good!

PR #1851, opened 2024-07-31
Update app.yml CSP Directives Structure

sawyerb-ksu:
Piping in directives on new lines allows for easier diff viewing when applying new changes.

### DIFF
--- a/config/app.yml
+++ b/config/app.yml
@@ -67,7 +67,15 @@ all:
       # 'Content-Security-Policy-Report-Only' or 'Content-Security-Policy'
       response_header: Content-Security-Policy-Report-Only
       # Configure CSP response directives.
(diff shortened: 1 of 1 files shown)


sbreker:
Hi Sawyer - thanks for this contribution to improve the readability of the CSP directive! This is a good add to AtoM and does make it easier to read and modify in the app.yml config file.

Unfortunately using the pipe "|" character in the value of a YAML key value pair causes the newline characters to be preserved when the value is read from the file. (Summary of this behaviour when parsing YAML can be found [here](https://stackoverflow.com/questions/3790454/how-do-i-break-a-string-in-yaml-over-multiple-lines) and [here](https://symfony.com/doc/current/reference/formats/yaml.html#strings)).

Since newlines are used in the HTML 1.1 spec as separators between the sections of an HTML response, values returned in the header must not contain newlines. See section 3 of [RFC 7230](https://www.rfc-editor.org/rfc/rfc7230) HTTP/1.1 Message Syntax and Routing for more info on newlines as separators.

The net result is that when the app.yml contains the "|" and the values are separated with newlines, the CSP header will be omitted from the response and so deactivates CSP headers - if the response is viewed in the developer tools of your browser before and after adding this modification to the CSP directive you will see that the header goes missing.

(When testing, be sure that your AtoM instance is not using a cached copy of your app.yml file. Search your AtoM instance and make sure there is not more than one copy! You may find one in `atom/app/qubit/config` which gets created at install time.)

Good news is that I think this can be corrected by using a different symbol in the app.yml file. Since line feeds cannot be contained in a header value, to achieve the result your are aiming for you could update this PR so that instead of using the pipe "|" in app.yml, use the greaterthan symbol ">" to omit any embedded newlines from the parsed value. I encourage you to update this enhancement and verify this does the trick, and then re-submit this PR. This would be a great improvement to AtoM's CSP config.

Additionally I will be looking at making a change to QubitCSPFilter.php to strip any newlines out of the directive when it is read from config.

sbreker (changes requested):
Great idea but unfortunately this will not work. See my comment - try changing the `|` to a `>` and retesting!

sawyerb-ksu:
@sbreker Thank you for the response and guidance/resources to review the yml requirements. I also appreciate not roasting me too much for missing the cached data prior to submitting the PR! I have re-tested with the recommended changes and found better results. I have updated the PR with the corrected change. for review.

Thanks again!

sbreker (approved):
Looks great! Thanks for this enhancement!

PR #1853, opened 2024-08-01
Fix accordion button icons for BS5 theme

anvit:
Add variables set for accordion to accordion-button since these are not present by default in the new BS5 version, and this breaks the `>` icons for section headers in the bootstrap 5 theme. Also updated bootstrap scss imports to follow the recommendation for v5.3


melaniekung (approved):
lgtm!

PR #1854, opened 2024-08-02
Add test for arSolrExistsQuery

sbreker:


test/phpunit/arSolrPlugin/lib/query/ArSolrExistsQueryTest.php
```diff
@@ -0,0 +1,68 @@
...
+ *
+ * @covers \arSolrExistsQuery
+ */
+class ArSolrExistsQueryTest extends TestCase
+{
+    public function createSolrExistsQueryProvider(): array
+    {
+        return [
+            'New arSolrExistsQuery with blank field' => ['field' => ''],
+            'New arSolrExistsQuery with null field' => ['field' => null],
```
sbreker: @anvit and @melaniekung: thinking about this - should '' or null trigger a validation error in the constructor? What will happen if a query is executed against Solr with these values?

melaniekung: hi steve, they both return parser errors

PR #1857, opened 2024-08-07
Ignore newlines in CSP directive in app.yml

sbreker:
This commit adds logic to QubitCspFilter to ignore newlines added when defining the CSP Directive in app.yml.

Refactor QubitCspFilter to move config reading logic to separate methods in the QubitCspFilter class.

Add tests to ensure that the config vars are read correctly from app.yml when they are formatted in different ways.


anvit (approved):
Looks great, thanks @sbreker !

PR #1858, opened 2024-08-08
Update cypress and fix flaky tests

anvit:
Update cypress version, and add retries for flaky tests. Also add optional chaining to autocomplete to avoid generating js error messages when no autocomplete test results are present.


melaniekung (approved):
lgtm, thanks anvit!

PR #1860, opened 2024-08-12
Fix identifier matching errors, refs #13582

PeteVox:
During command line imports with the _match-and-update_ flag set, it is possible for AtoM to false-match to an existing record based on **Title only** if Identifiers contain special characters, and Identifier patterns are 'similar' in the Repository. This causes metadata in matched record to be overwritten, parent id reassigned, and breaks the nested set.  

This seems to be related to the way that Elastic tokenizes special characters.

Problem first noted in Nov 2021, reported by Matthew Addis @ Arkivum (UK). https://projects.artefactual.com/issues/13582 

The proposed code-fix has been used locally in multiple AtoMs and across multiple versions since 2.5+. The modified field/index _identifier.untouched_ is already used in several places in code: https://github.com/search?q=repo%3Aartefactual%2Fatom%20identifier.untouched&type=code

### DIFF
--- a/lib/model/QubitInformationObject.php
+++ b/lib/model/QubitInformationObject.php
@@ -2171,7 +2171,7 @@ class QubitInformationObject extends BaseInformationObject
 
             // Use match query for exact matches.
             $queryText = new \Elastica\Query\Match();
-            $queryBool->addMust($queryText->setFieldQuery('identifier', $identifier));
+            $queryBool->addMust($queryText->setFieldQuery('identifier.untouched', $identifier));
 
             $queryText = new \Elastica\Query\Match();
             $queryBool->addMust($queryText->setFieldQuery(sprintf('i18n.%s.title.untouched', $currentCulture), $title));


anvit:
Hi @PeteVox ! Thank you so much for looking into this!
This change looks good to me, and applying this should resolve the reported issue.

That said, we're looking at [replacing ElasticSearch with a different search engine](https://groups.google.com/g/ica-atom-users/c/ceUkxL3rdS4/m/7gOb_tjJAAAJ) in the next release at the moment, due to licensing changes introduced for v7.11+, as well as the current version of Elastica that we use being incompatible with PHP8.0+. The alternative we're looking at right now is Solr, which has a few differences when it comes to its index. Since we would have to rewrite this section of code to work with Solr, we are closing this PR for now.

anvit:
Hi @PeteVox! As you might have gathered from the [Release Notes for AtoM 2.9.0](https://github.com/artefactual/atom/releases/tag/v2.9.0), we [decided to keep ElasticSearch](https://groups.google.com/g/ica-atom-users/c/sTb8io3bP1Y/m/VrHL0tFWAQAJ), so we're able to re-open this issue.

We would be happy to accept this change! Let us know if you'd have the capacity to rebase/apply your changes to the latest `qa/2.x`, and we can review and accept this PR.

PeteVox:
Hi @anvit. Thanks for taking another look at this, but I think the PR can stay closed. The previously problematic ElasticSearch 'Match' has been replaced with a new [direct database query](https://github.com/IMAGIZ/atom/blob/877fd479f7bc57bfd5953ddcef6841cbbfb47565/lib/model/QubitInformationObject.php#L2173), which removes the issue. 

Thanks again for considering the PR - much appreciated.

anvit:
Ah right! Thanks again @PeteVox !

PR #1880, opened 2024-10-09
Update ElasticSearch and  Elastica to 6.x

### LINKED ISSUE #1708: Problem:  ElasticSearch 5.x dependency
### Current Behavior

ElasticSearch 5.0 was released in 2016, and has been End of Life since many years.

### Expected Behavior

Support for Non-EOL products like ES7 and ES8

### Possible Solution

_No response_

### Context and Notes

_No response_

### Version used

_No response_

_No response_

_No response_

### PHP version

_No response_

### Contact details

_No response_

anvit:
Update ElasticSearch and Elastica and address issues introduced by breaking changes in ElasticSearch 6.x.

The most notable changes are:
- [Dropped support for multiple mapping types](https://www.elastic.co/guide/en/elasticsearch/reference/6.8/breaking-changes-6.0.html#_indices_created_before_6_0): ES 6.x no longer allows using multiple indices, They suggest handling this by creating multiple indices, each with its own type.
- [Mapping changes](https://www.elastic.co/guide/en/elasticsearch/reference/6.8/breaking-changes-6.0.html#breaking_60_mappings_changes): The `include_in_all` mapping parameter is disallowed and the `_all` field is disabled by default. 

Changes in Elastica:
- Elastica only has a addDocuments, deleteDocuments, and updateDocuments methods in 6.x and there are no longer addDocument, deleteDocument, updateDocument methods for handling a singular documents.


apps/qubit/modules/search/actions/autocompleteAction.class.php
```diff
@@ -76,7 +76,10 @@ public function execute($request)
 
         foreach ($items as $item) {
             $search = new \Elastica\Search($client);
-            $search->addIndex($index)->addType($index->getType($item['type']));
+            foreach ($indices as $type => $index) {
+                $itemType = QubitSearch::getInstance()->index->getIndexTypeName($item['type']);
+                $search->addIndex($index)->addType($index->getType($itemType));
```
jraddaoui: Instead of using a different type on each index, I'd suggest to use the dummy type name `_doc`, as it will be used in 7.x:

https://www.elastic.co/blog/removal-of-mapping-types-elasticsearch#schedule

Using the same type for all indices should simplify many places modified in this PR.

lib/QubitLftSyncer.class.php
```diff
@@ -92,8 +92,9 @@ private function repairEsChildrenLftValues()
         $results = QubitPdo::fetchAll($sql, $params, ['fetchMode' => PDO::FETCH_ASSOC]);
 
         $bulk = new Elastica\Bulk(QubitSearch::getInstance()->client);
-        $bulk->setIndex(QubitSearch::getInstance()->index->getName());
-        $bulk->setType('QubitInformationObject');
+        $type = 'QubitInformationObject';
+        $bulk->setIndex(QubitSearch::getInstance()->index->getType($type)->getName());
+        $bulk->setType(QubitSearch::getInstance()->index->getIndexTypeName($type));
```
jraddaoui: This would be just `$bulk->setType("_doc");`

plugins/arElasticSearchPlugin/lib/arElasticSearchIndexDecorator.class.php
```diff
@@ -17,67 +17,66 @@
...
 {
     protected $_instance;
 
-    public function __construct(Elastica\Index $instance)
+    protected $_indexPrefix;
+
+    public function __construct($prefix)
     {
-        $this->_instance = $instance;
+        $this->_instance = [];
```
jraddaoui: It feels strange to use an index decorator to manage multiple indexes. I'd remove this class and move the logic altogether to `arElasticSearchPlugin.class.php`.

plugins/arElasticSearchPlugin/lib/arElasticSearchPluginUtil.class.php
```diff
@@ -482,8 +482,8 @@ protected static function getAllObjectStringFields(
                 // Get string fields included in _all
                 elseif (
                     (
-                        !isset($propertyProperties['include_in_all'])
-                        || $propertyProperties['include_in_all']
+                        !isset($propertyProperties['copy_to'])
+                        || ('_all' == $propertyProperties['copy_to'])
```
jraddaoui: This condition will now copy to `_all` all the mapping fields where you removed `include_in_all: false`. It may be better to add `copy_to: _all` to the fields we really want to include and change this condition to `isset($propertyProperties['copy_to']) && ('_all' == $propertyProperties['copy_to'])`.

jraddaoui (changes requested):
Looking great @anvit!

I have not been able to look at everything in detail yet, but I have added a few comments that I think are important and may require some work already. Let me know your thoughts ;)

plugins/arElasticSearchPlugin/lib/arElasticSearchPluginUtil.class.php
anvit: Updated the mappings.yml file to copy all relevant mappings to _all

plugins/arElasticSearchPlugin/lib/arElasticSearchIndexDecorator.class.php
anvit: I looked at removing this at first but realized that one reason we'd kept this file was that since we introduced multiple indices, we needed to use the ES index name set in the AtoM config as the prefix for all index names, and we would be repeating a lot of code for adding in the prefix if we didn't include a function for doing this.

I've renamed this class and refactored some code to clarify what this class is for.

plugins/arElasticSearchPlugin/lib/arElasticSearchPlugin.class.php
```diff
@@ -44,6 +52,13 @@ class arElasticSearchPlugin extends QubitSearchEngine
      */
     public $index;
```
jraddaoui, line 53: It's not longer an Elastic_Index and it could be a good opportunity to name it `indicesWrapper` or something like that, since you are already changing many of the places where it's used.

plugins/arElasticSearchPlugin/lib/arElasticSearchPlugin.class.php
```diff
@@ -44,6 +52,13 @@ class arElasticSearchPlugin extends QubitSearchEngine
...
 
+    /**
+     * Current batch type, used for batch flush.
+     *
+     * @var mixed defaults to null
+     */
+    protected $currentBatchType;
```
jraddaoui: Maybe `$currentBatchIndex`, `$currentBatchIndexName` or `$currentBatchName` is better?

plugins/arElasticSearchPlugin/lib/arElasticSearchPlugin.class.php
```diff
@@ -90,9 +105,7 @@ public function __construct(array $options = [])
         // Verify the version running in the server
         $this->checkVersion();
 
-        // TODO: arElasticSearchIndexDecorator can be removed when Elastica >= 6.x. It is present to
-        // provide Elastica/Index::updateByQuery().
-        $this->index = new arElasticSearchIndexDecorator($this->client->getIndex($this->config['index']['name']));
+        $this->index = new arElasticSearchMultiIndexWrapper($this->config['index']['name']);
```
jraddaoui: I think you don't need the prefix in the wrapper ...

plugins/arElasticSearchPlugin/lib/arElasticSearchPlugin.class.php
```diff
@@ -493,57 +537,86 @@ protected function initialize()
...
+        $this->loadAndNormalizeMappings();
 
-                        if (sfConfig::get('app_diacritics')) {
-                            $filters = array_merge($filters, ['diacritics_lowercase']);
-                        }
+        // Iterate over types (actor, informationobject, ...)
+        foreach ($this->mappings as $typeName => $typeProperties) {
+            $typeName = 'Qubit'.sfInflector::camelize($typeName);
+            $this->index->createIndex($typeName,
+                $this->client->getIndex($this->index->getIndexName($typeName))
```
jraddaoui: ... it may only be needed here. Removing it from the wrapper will simplify things a bit, just using the default index names internally.

plugins/arElasticSearchPlugin/lib/arElasticSearchPlugin.class.php
```diff
@@ -493,57 +537,86 @@ protected function initialize()
...
+                    );
                 }
 
-                $this->index->create(
-                    $this->config['index']['configuration'],
-                    ['recreate' => true]
-                );
-            }
+                // Load and normalize mappings
+                $this->loadAndNormalizeMappings();
```
jraddaoui: Why calling this again inside the loop?

plugins/arElasticSearchPlugin/lib/arElasticSearchMultiIndexWrapper.class.php
```diff
@@ -0,0 +1,89 @@
...
+ */
+
+/**
+ * arElasticSearchMultiIndexWrapper facilitates handling ElasticSearch indices so that
+ * all indices for an AtoM installation can share a common prefix without having to
+ * explicitly specify it in common indexing and search functions.
+ */
+class arElasticSearchMultiIndexWrapper
+{
+    protected $_instance;
```
jraddaoui: I'd call this `indices`. Also, why the underscore?

plugins/arElasticSearchPlugin/lib/arElasticSearchMultiIndexWrapper.class.php
```diff
@@ -0,0 +1,89 @@
...
+
+    protected $_indexPrefix;
```
jraddaoui: As I said, you could leave the prefix outside of the wrapper. Use the default names internally and append the prefix only when calling ES. Removing the need of `getIndexName`.

plugins/arElasticSearchPlugin/lib/arElasticSearchMultiIndexWrapper.class.php
```diff
@@ -0,0 +1,89 @@
...
+
+    public function __construct($prefix)
+    {
+        $this->_instance = [];
+
+        $this->_indexPrefix = $prefix;
+    }
+
+    public function createIndex($name, Elastica\Index $index)
```
jraddaoui: I'd call this function `addIndex`.

plugins/arElasticSearchPlugin/lib/arElasticSearchMultiIndexWrapper.class.php
```diff
@@ -0,0 +1,89 @@
...
+    // Return the index element from the array of indices
+    // that matches the qualified index name
+    public function getIndex($name)
+    {
+        $name = $this->getIndexName($name);
+
+        return $this->_instance[$name];
+    }
+
+    public function getInstance()
```
jraddaoui: I'd call it `getIndices`.

plugins/arElasticSearchPlugin/lib/arElasticSearchMultiIndexWrapper.class.php
```diff
@@ -0,0 +1,89 @@
...
+    {
+        return $this->_instance;
+    }
+}
```
jraddaoui, line 88: Considering all already said, I think you could live without this class and move the remaining pieces to arElasticSearchPlugin. Having `indices` there directly instead of the wrapper.

plugins/arElasticSearchPlugin/lib/arElasticSearchPlugin.class.php
```diff
@@ -30,6 +30,14 @@ class arElasticSearchPlugin extends QubitSearchEngine
      */
     public const MIN_VERSION = '1.3.0';
```
jraddaoui: Should this be updated?

plugins/arElasticSearchPlugin/lib/arElasticSearchPlugin.class.php
```diff
@@ -343,10 +356,23 @@ public function addDocument($data, $type)
...
+        // Setting a dummy type since it is required in ES 6.x
+        // but it can be removed in 7.x when it becomes optional
+        $document->setType(self::ES_TYPE);
+
+        if (!$this->currentBatchType) {
+            $this->currentBatchType = $type;
+        }
 
         if ($this->batchMode) {
             // Add this document to the batch add queue
```
jraddaoui: This comment is misplaced now.

plugins/arElasticSearchPlugin/lib/arElasticSearchPlugin.class.php
```diff
@@ -343,10 +356,23 @@ public function addDocument($data, $type)
...
+            if ($this->currentBatchType != $type) {
+                $this->flushBatch();
+                $this->currentBatchType = $type;
+                $this->index->refresh();
```
jraddaoui: It would be nice to be able to pass an index name to flush only that one, now that we are flushing when we change the index. Defaults to refresh them all.

If you remove the wrapper and have access to the `indices` here, you could even refresh it directly (`$this->indices[$name]->refresh();`) and only do the full refresh over all in the `__destruct` function.

plugins/arElasticSearchPlugin/lib/arElasticSearchPlugin.class.php
```diff
@@ -427,12 +459,24 @@ public function delete($object)
...
             // the document to be deleted and add this document object to the batch delete
             // queue.
             $document = new \Elastica\Document($object->id);
-            $document->setType(get_class($object));
+            $document->setType(self::ES_TYPE);
+
+            if ($this->currentBatchType != $type) {
+                $this->flushBatch();
+                $this->currentBatchType = $type;
+                $this->index->refresh();
```
jraddaoui: Same here.

plugins/arElasticSearchPlugin/lib/arElasticSearchPlugin.class.php
```diff
@@ -493,57 +537,86 @@ protected function initialize()
...
 
-            // Load and normalize mappings
-            $this->loadAndNormalizeMappings();
+                // Iterate over types (actor, informationobject, ...)
+                foreach ($this->mappings as $typeName => $typeProperties) {
+                    $typeName = 'Qubit'.sfInflector::camelize($typeName);
+
+                    if ($indexType != $this->index->getIndexName($typeName)) {
+                        continue;
+                    }
```
jraddaoui: This foreach and if seem unnecessary, you should be able to access the normalized mapping related to the index by using the index name directly `$this->mappings[$indexName]`.

plugins/arElasticSearchPlugin/lib/arElasticSearchPlugin.class.php
```diff
@@ -493,57 +537,86 @@ protected function initialize()
...
+                    // Define mapping in elasticsearch
+                    $mapping = new \Elastica\Type\Mapping();
 
-                // Define mapping in elasticsearch
-                $mapping = new \Elastica\Type\Mapping();
-                $mapping->setType($this->index->getType($typeName));
-                $mapping->setProperties($typeProperties['properties']);
+                    // Setting a dummy type since it is required in ES 6.x
+                    // but it can be removed in 7.x when it becomes optional
+                    $mapping->setType($index->getType(self::ES_TYPE));
```
jraddaoui: It should be enough with `$mapping->setType(self::ES_TYPE);`.

apps/qubit/modules/search/actions/autocompleteAction.class.php
```diff
@@ -76,7 +76,12 @@ public function execute($request)
 
         foreach ($items as $item) {
             $search = new \Elastica\Search($client);
-            $search->addIndex($index)->addType($index->getType($item['type']));
+            foreach ($indices as $type => $index) {
+                $itemType = QubitSearch::getInstance()::ES_TYPE;
+
+                // This can be updated in ES 7.x when type params are optional
+                $search->addIndex($index)->addType($index->getType($itemType));
```
jraddaoui: Same, we should be able to use the constant directly, right? `$search->addIndex($index)->addType(QubitSearch::getInstance()::ES_TYPE);`

lib/QubitLftSyncer.class.php
```diff
@@ -92,8 +92,9 @@ private function repairEsChildrenLftValues()
         $results = QubitPdo::fetchAll($sql, $params, ['fetchMode' => PDO::FETCH_ASSOC]);
 
         $bulk = new Elastica\Bulk(QubitSearch::getInstance()->client);
-        $bulk->setIndex(QubitSearch::getInstance()->index->getName());
-        $bulk->setType('QubitInformationObject');
+        $type = 'QubitInformationObject';
+        $bulk->setIndex(QubitSearch::getInstance()->index->getIndex($type)->getName());
+        $bulk->setType(QubitSearch::getInstance()->index->getIndexName($type));
```
jraddaoui: Shouldn't this be constant? `$bulk->setType(QubitSearch::getInstance()::ES_TYPE);`

lib/QubitLftSyncer.class.php
```diff
@@ -92,8 +92,9 @@ private function repairEsChildrenLftValues()
         $results = QubitPdo::fetchAll($sql, $params, ['fetchMode' => PDO::FETCH_ASSOC]);
 
         $bulk = new Elastica\Bulk(QubitSearch::getInstance()->client);
-        $bulk->setIndex(QubitSearch::getInstance()->index->getName());
-        $bulk->setType('QubitInformationObject');
+        $type = 'QubitInformationObject';
+        $bulk->setIndex(QubitSearch::getInstance()->index->getIndex($type)->getName());
```
jraddaoui: We may need the prefix here too. Right now, it's getting the index name from ES directly but you could do something like `$bulk->setIndex(QubitSearch::getInstance()->prefix + 'QubitInformationObject');`, if you add the prefix to arElasticSearchPlugin and make it public.

apps/qubit/modules/search/actions/autocompleteAction.class.php
```diff
@@ -76,7 +76,12 @@ public function execute($request)
 
         foreach ($items as $item) {
             $search = new \Elastica\Search($client);
-            $search->addIndex($index)->addType($index->getType($item['type']));
+            foreach ($indices as $type => $index) {
```
jraddaoui: I'd try to avoid the use of `$type`, I'd use $name in this case.

lib/QubitLftSyncer.class.php
```diff
@@ -92,8 +92,9 @@ private function repairEsChildrenLftValues()
         $results = QubitPdo::fetchAll($sql, $params, ['fetchMode' => PDO::FETCH_ASSOC]);
 
         $bulk = new Elastica\Bulk(QubitSearch::getInstance()->client);
-        $bulk->setIndex(QubitSearch::getInstance()->index->getName());
-        $bulk->setType('QubitInformationObject');
+        $type = 'QubitInformationObject';
```
jraddaoui: No need for a variable, if you still want one I'd call it `$name`.

lib/job/arUpdatePublicationStatusJob.class.php
```diff
@@ -91,7 +91,8 @@ public function runJob($parameters)
             'conflicts' => 'proceed',
         ];
 
-        $response = QubitSearch::getInstance()->index->updateByQuery($query, $queryScript, $options)->getData();
+        $type = 'QubitInformationObject';
```
jraddaoui: Same, no need for a var, if you want it I think it's better called `$name`.

lib/task/tools/updatePublicationStatusTask.class.php
```diff
@@ -166,14 +166,15 @@ protected function updatePublicationStatusDescendants($resource, $publicationSta
...
-                'inline' => 'ctx._source.publicationStatusId = '.$publicationStatus->id,
+                'source' => 'ctx._source.publicationStatusId = '.$publicationStatus->id,
                 'lang' => 'painless',
             ],
         ]);
 
         $options = ['conflicts' => 'proceed'];
 
-        $response = QubitSearch::getInstance()->index->updateByQuery($query, $queryScript, $options)->getData();
+        $type = 'QubitInformationObject';
```
jraddaoui: Same.

plugins/arElasticSearchPlugin/lib/arElasticSearchPlugin.class.php
```diff
@@ -343,10 +356,23 @@ public function addDocument($data, $type)
...
 
         $document = new \Elastica\Document($id, $data);
-        $document->setType($type);
+
+        // Setting a dummy type since it is required in ES 6.x
+        // but it can be removed in 7.x when it becomes optional
+        $document->setType(self::ES_TYPE);
+
+        if (!$this->currentBatchType) {
+            $this->currentBatchType = $type;
```
jraddaoui: We may even want to change the param name to `$indexName` instead of `$type`.

jraddaoui: Also, unrelated to these changes, shouldn't this function be checking if it's enabled? Like all the others updates and delete.

plugins/arElasticSearchPlugin/lib/arElasticSearchPlugin.class.php
```diff
@@ -379,10 +405,16 @@ public function partialUpdate($object, $data)
             return;
         }
 
+        $type = get_class($object);
```
jraddaoui: Call it `$name` or `$indexName`.

plugins/arElasticSearchPlugin/lib/arElasticSearchPlugin.class.php
```diff
@@ -427,12 +459,24 @@ public function delete($object)
         }
 
         if ($this->batchMode) {
+            $type = get_class($object);
```
jraddaoui: Same.

plugins/arElasticSearchPlugin/lib/arElasticSearchPlugin.class.php
```diff
@@ -493,57 +537,86 @@ protected function initialize()
...
-                        }
+        // Load and normalize mappings
+        $this->loadAndNormalizeMappings();
 
-                        if (sfConfig::get('app_diacritics')) {
-                            $filters = array_merge($filters, ['diacritics_lowercase']);
-                        }
+        // Iterate over types (actor, informationobject, ...)
+        foreach ($this->mappings as $typeName => $typeProperties) {
+            $typeName = 'Qubit'.sfInflector::camelize($typeName);
```
jraddaoui: `$indexName`

plugins/arElasticSearchPlugin/lib/arElasticSearchPlugin.class.php
```diff
@@ -493,57 +537,86 @@ protected function initialize()
...
+            $this->index->createIndex($typeName,
+                $this->client->getIndex($this->index->getIndexName($typeName))
+            );
+        }
 
-                        $this->config['index']['configuration']['analysis']['analyzer'][$key]['char_filter'] = $filters;
+        foreach ($this->index->getInstance() as $indexType => $index) {
```
jraddaoui: `$indexName`

plugins/arElasticSearchPlugin/lib/arElasticSearchPlugin.class.php
```diff
@@ -493,57 +537,86 @@ protected function initialize()
...
-                    ['recreate' => true]
-                );
-            }
+                // Load and normalize mappings
+                $this->loadAndNormalizeMappings();
 
-            // Load and normalize mappings
-            $this->loadAndNormalizeMappings();
+                // Iterate over types (actor, informationobject, ...)
+                foreach ($this->mappings as $typeName => $typeProperties) {
```
jraddaoui: `$indexName`, `$indexProperties`

plugins/arElasticSearchPlugin/config/mapping.yml
```diff
@@ -61,22 +61,22 @@ mapping:
...
     type: nested
     properties:
-      start_date: { type: date, include_in_all: false }
-      end_date: { type: date, include_in_all: false }
-      start_date_string: { type: keyword }
-      end_date_string: { type: keyword }
-      type_id: { type: integer, include_in_all: false }
-      actor_id: { type: integer, include_in_all: false }
+      start_date: { type: date, copy_to: _all }
+      end_date: { type: date, copy_to: _all }
```
jraddaoui: I think these two are wrong and they should not have `copy_to: _all`.

plugins/arElasticSearchPlugin/config/mapping.yml
```diff
@@ -91,8 +91,8 @@ mapping:
         dynamic: strict
     dynamic: strict
     properties:
-      date: { type: date, include_in_all: false }
-      date_string: { type: keyword }
+      date: { type: date, copy_to: _all }
```
jraddaoui: Same.

plugins/arElasticSearchPlugin/config/mapping.yml
```diff
@@ -102,111 +102,111 @@ mapping:
...
-      rights_holder: { type: text, include_in_all: false }
-      rights_note: { type: text, include_in_all: false }
-      license_terms: { type: text, include_in_all: false }
+      basis: { type: text, copy_to: _all }
+      start_date: { type: date, copy_to: _all }
+      end_date: { type: date, copy_to: _all }
+      copyright_status: { type: text, copy_to: _all }
+      rights_holder: { type: text, copy_to: _all }
+      rights_note: { type: text, copy_to: _all }
+      license_terms: { type: text, copy_to: _all }
```
jraddaoui: Same.

plugins/arElasticSearchPlugin/config/mapping.yml
```diff
@@ -102,111 +102,111 @@ mapping:
...
     dynamic: strict
     properties:
-      act: { type: text, include_in_all: false }
-      restriction: { type: text, include_in_all: false }
-      start_date: { type: date, include_in_all: false }
-      end_date: { type: date, include_in_all: false }
+      act: { type: text, copy_to: _all }
+      restriction: { type: text, copy_to: _all }
+      start_date: { type: date, copy_to: _all }
+      end_date: { type: date, copy_to: _all }
```
jraddaoui: Same.

plugins/arElasticSearchPlugin/config/mapping.yml
```diff
@@ -102,111 +102,111 @@ mapping:
...
+      data_size: { type: long, copy_to: _all }
+      footer_size: { type: long, copy_to: _all }
+      language: { type: keyword, copy_to: _all }
+      color_primaries: { type: keyword, copy_to: _all }
+      transfer_characteristics: { type: keyword, copy_to: _all }
+      matrix_coefficients: { type: keyword, copy_to: _all }
+      is_streamable: { type: boolean, copy_to: _all }
+      writing_application: { type: keyword, copy_to: _all }
+      file_last_modification_date: { type: date, copy_to: _all }
+      file_last_modification_date_local: { type: date, copy_to: _all }
```
jraddaoui: Same.

plugins/arElasticSearchPlugin/config/mapping.yml
```diff
@@ -223,84 +223,84 @@ mapping:
...
-      last_modified: { type: date, include_in_all: false }
-      date_ingested: { type: date, include_in_all: false }
-      size: { type: long, include_in_all: false }
-      mime_type: { type: keyword, include_in_all: false }
+      puid: { type: keyword, copy_to: _all }
+      filename: { type: keyword, copy_to: _all }
+      last_modified: { type: date, copy_to: _all }
+      date_ingested: { type: date, copy_to: _all }
+      size: { type: long, copy_to: _all }
+      mime_type: { type: keyword, copy_to: _all }
```
jraddaoui: Same.

plugins/arElasticSearchPlugin/config/mapping.yml
```diff
@@ -630,5 +630,5 @@ mapping:
             search_analyzer: standard
             term_vector: with_positions_offsets
       # Not nested date fields for sorting
-      start_date_sort: { type: date, include_in_all: false }
-      end_date_sort: { type: date, include_in_all: false }
+      start_date_sort: { type: date, copy_to: _all }
+      end_date_sort: { type: date, copy_to: _all }
```
jraddaoui: Same for a lot of fields, there is a lot. If the field had `include_in_all: false` it should not have `copy_to: _all`. And, looking at the condition, we only care about `text` types, so we only need to add `copy_to: _all` to those when needed.

jraddaoui (commented):
Thanks @anvit! This looks a lot clearer to me now.

plugins/arElasticSearchPlugin/lib/arElasticSearchPlugin.class.php
anvit: We're planning on retaining the old name for backwards compatibility of custom themes for now. Will probably update this in the future when we refactor away ElasticSearch's use in AtoM that's outside the Search/ElasticSearch plugin.

plugins/arElasticSearchPlugin/lib/arElasticSearchPlugin.class.php
anvit: `mapping->setType` expects an `Elastica\Type`, which is why we're using `getType`:


https://github.com/ruflin/Elastica/blob/f95cf646f884a8e7b73bd12eda3e5753994b24d3/lib/Elastica/Index.php#L70-L80

apps/qubit/modules/search/actions/autocompleteAction.class.php
```diff
@@ -76,7 +76,12 @@ public function execute($request)
...
+                $elasticSearchTypeName = QubitSearch::getInstance()::ES_TYPE;
+
+                // This can be updated in ES 7.x when type params are optional
+                $search->addIndex($index)->addType($index->getType($elasticSearchTypeName));
+            }
```
jraddaoui: This works the same for me:

```php
foreach ($indices as $index) {
    $search->addIndex($index)->addType(QubitSearch::getInstance()::ES_TYPE);
}
```

And I think the type is already optional here, this also works for me:

```php
foreach ($indices as $index) {
    $search->addIndex($index);
}
```

PS. Somehow, actors are duplicated and shown as repositories in the results:

image

lib/QubitLftSyncer.class.php
```diff
@@ -92,8 +92,8 @@ private function repairEsChildrenLftValues()
         $results = QubitPdo::fetchAll($sql, $params, ['fetchMode' => PDO::FETCH_ASSOC]);
 
         $bulk = new Elastica\Bulk(QubitSearch::getInstance()->client);
-        $bulk->setIndex(QubitSearch::getInstance()->index->getName());
-        $bulk->setType('QubitInformationObject');
+        $bulk->setIndex(QubitSearch::getInstance()->index->getIndex('QubitInformationObject')->getName());
```
jraddaoui: You could pass the index directly, without calling getName:

https://github.com/ruflin/Elastica/blob/6.x/lib/Elastica/Bulk.php#L50-L64

lib/QubitLftSyncer.class.php
```diff
@@ -92,8 +92,8 @@ private function repairEsChildrenLftValues()
...
+        $bulk->setType(QubitSearch::getInstance()::ES_TYPE);
```
jraddaoui, line 96: Looking at the usage of `_type` in https://github.com/ruflin/Elastica/blob/6.x/lib/Elastica/Bulk.php I bet we don't need to set it here either, but I have not tested it.

plugins/arElasticSearchPlugin/lib/arElasticSearchMultiIndexWrapper.class.php
```diff
@@ -0,0 +1,83 @@
...
+ * You should have received a copy of the GNU General Public License
+ * along with Access to Memory (AtoM).  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+/**
+ * arElasticSearchMultiIndexWrapper facilitates handling ElasticSearch indices
+ * and has methods that match signatures of pre ES 6.x methods that used a
+ * single index with multiple types instead of multiple index with a single
+ * type or no type.
+ */
```
jraddaoui: Maybe add a note that we have `indices` here in a separate class instead of moving it to arElasticSearchPlugin class to maintain backwards compatibility in client themes, similar to the getType function, but as a whole for the entire class.

plugins/arElasticSearchPlugin/lib/arElasticSearchPlugin.class.php
```diff
@@ -355,7 +385,7 @@ public function addDocument($data, $type)
                 $this->index->refresh();
             }
         } else {
-            $this->index->getType($type)->addDocument($document);
+            $this->index->getIndex($indexName)->addDocuments([$document]);
```
jraddaoui: Why getting the index instead of using the wrapper directly like in flushBatch?

```php
$this->index->addDocuments($indexName, [$document]);
```

plugins/arElasticSearchPlugin/lib/arElasticSearchPlugin.class.php
```diff
@@ -379,10 +409,16 @@ public function partialUpdate($object, $data)
...
+
         $document = new \Elastica\Document($object->id, $data);
 
+        // Setting a dummy type since it is required in ES 6.x
+        // but it can be removed in 7.x when it becomes optional
+        $document->setType(self::ES_TYPE);
+
         try {
-            $this->index->getType(get_class($object))->updateDocument($document);
+            $this->index->getIndex($indexName)->updateDocuments([$document]);
```
jraddaoui: Going with the wrapper in the end, we may want to add methods for the update and the partial updates in there too. Not a big deal for sure.

plugins/arElasticSearchPlugin/lib/arElasticSearchPlugin.class.php
jraddaoui: I'll follow up about this in another comment.

plugins/arElasticSearchPlugin/lib/arElasticSearchPlugin.class.php
jraddaoui: I see, thanks!

apps/qubit/modules/search/actions/autocompleteAction.class.php
jraddaoui: Addressed in another comment.

lib/QubitLftSyncer.class.php
jraddaoui: Addressed in a following comment.

plugins/arElasticSearchPlugin/lib/arElasticSearchPlugin.class.php
```diff
@@ -493,57 +541,84 @@ protected function initialize()
...
-                $mapping->send();
+                    $this->log(sprintf('Defining mapping %s...', $indexName));
+
+                    // In ES 7.x this should be changed to:
+                    // $mapping->send($index, [ 'include_type_name' => false ])
+                    // which can be removed in 8.x since that is the default behaviour
+                    // and will have be removed by 9.x when it is discontinued
+                    $mapping->send();
+                }
             }
```
jraddaoui: There are two unnecessary loops that make this a lot harder to read. If you make the ES index a variable in the first loop, there is no need to loop over getIndices nor the mapping again:

```php
foreach ($this->mappings as $indexName => $indexProperties) {
    $indexName = 'Qubit'.sfInflector::camelize($indexName);
    $prefixedIndexName = $this->config['index']['name'].'_'.strtolower($indexName);
    $index = $this->client->getIndex($prefixedIndexName);
    $this->index->addIndex($indexName, $index);

    try {
        $index->open();
    } catch (Exception $e) {
        // If the index has not been initialized, create it
        if ($e instanceof \Elastica\Exception\ResponseException) {
            // Removed for the review
        }

        // Define mapping in elasticsearch
        $mapping = new \Elastica\Type\Mapping();

        // Setting a dummy type since it is required in ES 6.x
        // but it can be removed in 7.x when it becomes optional
        $mapping->setType($index->getType(self::ES_TYPE));
        $mapping->setProperties($indexProperties['properties']);

        // Parse other parameters
        unset($this->mapping[$indexName]->indexProperties['properties']);
        foreach ($this->mapping[$indexName]->indexProperties as $key => $value) {
            $mapping->setParam($key, $value);
        }

        $this->log(sprintf('Defining mapping %s...', $indexName));

        // In ES 7.x this should be changed to:
        // $mapping->send($index, [ 'include_type_name' => false ])
        // which can be removed in 8.x since that is the default behaviour
        // and will have be removed by 9.x when it is discontinued
        $mapping->send();
    }
}
```

jraddaoui: Also:

- It feels rare that we define the mapping outside the `if`.
- We may want to use the prefixed index name in the log:

  ```php
  $this->log(sprintf('Defining mapping for index %s...', $prefixedIndexName));
  ```

plugins/arElasticSearchPlugin/lib/arElasticSearchPlugin.class.php
```diff
@@ -605,9 +680,9 @@ private function displayTypesToIndex($excludeTypes)
...
 
-        foreach ($this->mappings as $typeName => $typeProperties) {
-            if (!in_array(strtolower($typeName), $excludeTypes)) {
-                $this->log(' - '.$typeName);
+        foreach ($this->mappings as $indexName => $indexProperties) {
+            if (!in_array(strtolower($indexName), $excludeTypes)) {
+                $this->log(' - '.$indexName);
                 ++$typeCount;
             }
         }
```
jraddaoui: We may want to change this too, "Indices that will be updated." and use the prefixed index name.

plugins/arElasticSearchPlugin/config/mapping.yml
```diff
@@ -32,11 +32,11 @@ mapping:
...
     properties:
-      contact_person: { type: keyword }
+      contact_person: { type: keyword, copy_to: _all }
       street_address: { type: text, index: false }
-      postal_code: { type: text, include_in_all: false }
-      country_code: { type: keyword, include_in_all: false }
-      location: { type: geo_point }
+      postal_code: { type: text }
+      country_code: { type: keyword }
+      location: { type: geo_point, copy_to: _all }
```
jraddaoui: The condition checks the type too, making sure it's a `text` type field. It doesn't matter if you add it here, but it won't be copied to _all in the end and adds some confusion.

plugins/arElasticSearchPlugin/config/mapping.yml
```diff
@@ -61,22 +61,22 @@ mapping:
...
-      start_date: { type: date, include_in_all: false }
-      end_date: { type: date, include_in_all: false }
-      start_date_string: { type: keyword }
-      end_date_string: { type: keyword }
-      type_id: { type: integer, include_in_all: false }
-      actor_id: { type: integer, include_in_all: false }
+      start_date: { type: date }
+      end_date: { type: date }
+      start_date_string: { type: keyword, copy_to: _all }
+      end_date_string: { type: keyword, copy_to: _all }
```
jraddaoui: Same.

plugins/arElasticSearchPlugin/config/mapping.yml
```diff
@@ -91,8 +91,8 @@ mapping:
         dynamic: strict
     dynamic: strict
     properties:
-      date: { type: date, include_in_all: false }
-      date_string: { type: keyword }
+      date: { type: date }
+      date_string: { type: keyword, copy_to: _all }
```
jraddaoui: Same.

plugins/arElasticSearchPlugin/config/mapping.yml
```diff
@@ -102,111 +102,111 @@ mapping:
     _foreign_types: { contact_informations: contact_information }
     dynamic: strict
     properties:
-      slug: { type: keyword }
+      slug: { type: keyword, copy_to: _all }
```
jraddaoui: I think you get the point ;)

jraddaoui (commented):
Thanks @anvit! Getting there, just a few minor tweaks!

apps/qubit/modules/search/actions/autocompleteAction.class.php
jraddaoui: I found the issue, this is now adding all indices to all searches. We should get the wrapper before the loop, instead of getting all indices:

```php
$client = QubitSearch::getInstance()->client;
$indexWrapper = QubitSearch::getInstance()->index;
```

And only add the expected index to each search inside the loop:

```php
$search = new \Elastica\Search($client);
$search->addIndex($indexWrapper->getIndex($item['type']));
```

jraddaoui: And a couple of extra details after that:

- Changing this and the for loops in the initialize function makes `getIndices()` unnecessary in the index wrapper class.
- This should be noted as something that could break custom themes extending the autocomplete.

lib/QubitLftSyncer.class.php
jraddaoui: Ignore this, I just tested it and it looks like it's still needed.

plugins/arElasticSearchPlugin/lib/arElasticSearchPlugin.class.php
jraddaoui: Sorry, I meant refresh when I said flush. This is my suggestion to avoid refreshing all indices when we only flush the batch in one of them:

* Looking at the calls of `$this->index->refresh();` they all happen after `$this->flushBatch();`. So I'd move that call inside the flushBatch method and remove it from all other places.
* Now that you are sure it only happens there, you also know that we only flush the batch for one index, so you could change the call to include the index name: `$this->index->refresh($this->currentBatchIndexName);`.
* And the refresh function in the wrapper could require a `$name` parameter and only refresh that index:
  
  ```php
  public function refresh($name)
  {
      $this->indices[$name]->refresh();
  }
  ```

Additionally, similar to the comment I made about the update methods, up to you if you want to include it in the wrapper or not, you could do it outside like you are doing now for the updates: `$this->index->getIndex($this->currentBatchIndexName)->refresh();`. I think it's important to be consistent so, now that we are going with the wrapper, I think it makes more sense to put everything there.

plugins/arElasticSearchPlugin/lib/arElasticSearchPlugin.class.php
```diff
@@ -427,12 +463,24 @@ public function delete($object)
...
 
         if ($this->batchMode) {
+            $indexName = get_class($object);
```
jraddaoui: I just noticed, this needs to happen before the `if`, as it's also used in the `else`.

plugins/arElasticSearchPlugin/lib/arElasticSearchPlugin.class.php
jraddaoui: Okay, maybe not here. These need to match the `exclude-types` argument from the search:populate task.

Now that I think about that, we should probably revisit how is that done with multiple indexes instead of types.

plugins/arElasticSearchPlugin/lib/arElasticSearchMultiIndexWrapper.class.php
```diff
@@ -0,0 +1,83 @@
...
+    {
+        $this->indices[$name] = $index;
+    }
+
+    public function delete()
+    {
+        foreach ($this->indices as $index) {
+            $index->delete();
+        }
+    }
```
jraddaoui: Given all the feedback above, this would be the only function without a `$name` parameter in the wrapper (apart from the constructor). And, when we revisit the `exclude-types` option I think we'll add that parameter.

It makes a lot of sense to me that all methods from this class require a name parameter now that it's a multi index wrapper instead of an index decorator.

It would be even clearer if `$index` was called `$indicesWrapper` in `arElasticSearchPlugin`, but preventing issues in custom code is a stronger point than clarity.

Just a note.

plugins/arElasticSearchPlugin/lib/arElasticSearchPlugin.class.php
anvit: @jraddaoui it does feel odd to define the mappings outside the `if`. I let it be because I wasn't sure if there was some historical reason or if there was a scenario where the index existed but could not be opened because it didn't have mappings declared. In my testing changing it to:
```
try {                                                             
  $index->open();                                      
} catch (\Elastica\Exception\ResponseException $e) {
// If the index has not been initialized, create it
```
didn't seem to cause any problems (both in the scenario when the index existed and when it didn't). Let me know if you want me to make this change

jraddaoui: Yeah, changing the catch and removing the if makes sense to me.

apps/qubit/modules/search/actions/autocompleteAction.class.php
```diff
@@ -76,7 +75,8 @@ public function execute($request)
 
         foreach ($items as $item) {
             $search = new \Elastica\Search($client);
-            $search->addIndex($index)->addType($index->getType($item['type']));
+            $indexWrapper = QubitSearch::getInstance()->index;
```
jraddaoui: Better if we get the wrapper only once before the loop.

plugins/arElasticSearchPlugin/lib/arElasticSearchPlugin.class.php
```diff
@@ -493,61 +540,84 @@ protected function initialize()
...
+                // Setting a dummy type since it is required in ES 6.x
+                // but it can be removed in 7.x when it becomes optional
+                $mapping->setType($index->getType(self::ES_TYPE));
+                $mapping->setProperties($indexProperties['properties']);
 
                 // Parse other parameters
-                unset($typeProperties['properties']);
-                foreach ($typeProperties as $key => $value) {
+                unset($this->mapping[$indexName]->indexProperties['properties']);
+                foreach ($this->mapping[$indexName]->indexProperties as $key => $value) {
```
jraddaoui: Can't you do this directly over the `$indexProperties` variable? Why accessing the mapping again?

plugins/arElasticSearchPlugin/lib/arElasticSearchPlugin.class.php
```diff
@@ -493,61 +540,84 @@ protected function initialize()
...
+                }
+
+                if (sfConfig::get('app_diacritics')) {
+                    $filters = array_merge($filters, ['diacritics_lowercase']);
+                }
+
+                $this->config['index']['configuration']['analysis']['analyzer'][$key]['char_filter'] = $filters;
+            }
+        }
+    }
```
jraddaoui, line 608: Nice move!

plugins/arElasticSearchPlugin/config/mapping.yml
```diff
@@ -630,5 +630,5 @@ mapping:
             search_analyzer: standard
             term_vector: with_positions_offsets
       # Not nested date fields for sorting
-      start_date_sort: { type: date, include_in_all: false }
-      end_date_sort: { type: date, include_in_all: false }
+      start_date_sort: { type: date }
+      end_date_sort: { type: date }
```
jraddaoui, line 634: Neat! A lot clearer compared to the previous version.

jraddaoui (commented):
Nice, thanks @anvit!

plugins/arElasticSearchPlugin/lib/arElasticSearchMultiIndexWrapper.class.php
```diff
@@ -0,0 +1,96 @@
...
+    // been removed or refactored.
+    public function getType($name)
+    {
+        return $this->getIndex($name);
+    }
+
+    public function getIndices()
+    {
+        return $this->indices;
+    }
```
jraddaoui: This function is no longer used.

plugins/arElasticSearchPlugin/lib/arElasticSearchPlugin.class.php
jraddaoui: I think what I suggested above:

> Looking at the calls of $this->index->refresh(); they all happen after $this->flushBatch();. So I'd move that call inside the flushBatch method and remove it from all other places.

It would have simplified the latest changes and the code in general, only one call to refresh inside flushBatch, and always using `$this->currentBatchIndexName`, no need to make the index name optional since it always happens per index.

Sorry if that wasn't clear, and feel free to leave it as it is ;)

jraddaoui (approved):

PR #1881, opened 2024-10-10
Update composer deps and update php-cs-fixer conf

anvit:
Update composer dependencies:
- jasig/phpcas is no longer maintained. apereo/phpcas is the suggested replacement.
- Updating to php-cs-fixer 3.64.0 introduces some new rules which would lead to changes in over ~400 files. Selectively changing or turning off some of these rules to avoid noise and to maintain the current standard in AtoM


jraddaoui:
Nice @anvit!

At the time, we decided to follow the rules giving by php-cs-fixer and the community as much as possible and avoid our own opinions/discussions. I'd try to follow those new rules when possible too. I'd suggest you continue this PR and enable them one by one to see what they do and, unless you have a really strong argument against them, follow the default as much as possible. Check [the comment in the config](https://github.com/artefactual/atom/blob/qa/2.x/.php-cs-fixer.dist.php#L13-L21) to know why we changed only one rule. Maybe with the new version that rule could be restored to the default too.

lib/CsvImportAuditer.class.php
```diff
@@ -116,7 +116,7 @@ public function validateFilename($filename): string
         return $filename;
     }
 
-    public function setOptions(array $options = null): void
+    public function setOptions(?array $options = null): void
```
jraddaoui, line 119: TIL about nullable type syntactic sugar.

plugins/qtAccessionPlugin/modules/accession/templates/indexSuccess.php
```diff
@@ -199,7 +199,7 @@
   <section class="actions">
     <ul>
 
-      <?php if (QubitAcl::check($resource, 'update') || (QubitAcl::check($resource, 'translate'))) { ?>
+      <?php if (QubitAcl::check($resource, 'update') || QubitAcl::check($resource, 'translate')) { ?>
```
jraddaoui, line 202: Amazing how many unneeded parenthesis we had.

jraddaoui (approved):
Thanks @anvit! 

I think they are mostly nice changes, always tricky in templates, but it's nice.

PR #1882, opened 2024-10-21
Fix OIDC plugin getOidcInstance() logic

sbreker:
Fixed issue with arOidc::getOidcInstance() where if it were called a second time getOidcInstance() would return null instead of an instance of OpenIDConnectClient.


melaniekung (approved):
LGTM, thanks @sbreker !

PR #1884, opened 2024-10-29
Add oidcUserTest

sbreker:
Test oidcUser::authenticate() method with mock OpenIDConnectClient.

Modify phpunit.xml to set phpunit option 'stderr' to "true". This prevents error messages when tests call the setcookie() function.


melaniekung (approved):
LGTM, thanks @sbreker

PR #1886, opened 2024-10-30
Add OIDC plugin activation control

sbreker:
Add the ability for an admin to activate the OIDC plugin in a way that does not require updating the ProjectConfiguration class file, or using the plugin admin interface.

If the empty file 'activate-oidc-plugin' is present in the root of the AtoM installation, the OIDC plugin will be enabled.

At this point in Symfony's startup, in ProjectConfiguration::setup() config files like app.yml are not yet loaded which is why the presence of an OS file is being used.


melaniekung (approved):
👍

PR #1895, opened 2024-11-25
Add Include non-visible elements option for BS2 clipboard exp. (#1894)

### LINKED ISSUE #1894: Problem: Clipboard exports missing Include non-visible elements option in BS2 theme.
### Current Behavior

1. Use Dominion BS2 theme
2. Add item to clipboard
3. Go to clipboard export
4. Note that there's no option to Include non-visible elements

### Expected Behavior

All available options should be present in both BS2 and BS5 themes.

melaniekung:


anvit (approved):
LGTM!

PR #1897, opened 2024-11-25
Fix clipboard exports with hidden visible element headers. (#1896)

### LINKED ISSUE #1896: Problem: Clipboard exports missing fields related to hidden Visible Element Headers
### Current Behavior

1. Go to Visible Elements in Admin settings and hide headers for your template
2. Add item to clipboard
3. Export clipboard item, without selecting 'Include non-visible elements' option
4. Note that the CSV are missing fields related to hideen headers

### Expected Behavior

Hiding a header title should not affect the records in any way or form.

### LINKED ISSUE #1925: Problem: CSV export has data in wrong columns when elements are hidden
### Current Behavior

1. Go to Admin > Visible Elements.
2. Under ISAD template - area headings, turn off `Description control area`.
3. Try exporting a description that has at least 1 descendant, and make sure the `Include descendants` option is checked while exporting.  

### Expected Behavior

Notice that all of the columns towards the end, starting from the 2nd entry, contain data that does not match their headers. 

### Possible Solution

The output csv should have data that lines up with the headers for all columns when certain columns are set to not be visible any more.

melaniekung:
Hiding visible element headers should not affect description exports in any way. Exclude visible element headers from `getHiddenVisibleElementCsvHeaders()` by excluding settings values ending in `_area`.

### DIFF
--- a/lib/flatfile/QubitFlatfileExport.class.php
+++ b/lib/flatfile/QubitFlatfileExport.class.php
@@ -357,6 +357,7 @@ class QubitFlatfileExport
                 if (
                     (false !== strpos($setting, 'app_element_visibility_'.$template))
                     && (!strpos($setting, '__source'))
+                    && str_ends_with('_area', $setting) // Ignore hidden element headers
                     && (0 == sfConfig::get($setting))
                 ) {
                     array_push($nonVisibleElements, $setting);


anvit (approved):
Looks good!

PR #1902, opened 2024-12-14
Remove duplicate sfPropelPlugin phing code

melaniekung:
Set Composer to use Composer's vendor phing directory and remove sfPropelPlugin's copy of phing.


sbreker (approved):
Thanks @melaniekung! :+1:

PR #1907, opened 2025-01-09
ES 6.x Updates

melaniekung:


jraddaoui (approved):
LGTM!

plugins/arElasticSearchPlugin/lib/arElasticSearchPlugin.class.php
```diff
@@ -502,6 +502,11 @@ public static function modelClassFromQubitObjectClass($className)
         return str_replace('Qubit', 'arElasticSearch', $className);
     }
 
+    public function getEsType()
+    {
+        return self::ES_TYPE;
+    }
+
```
jraddaoui: This got lost in the review somehow ... Why the get function if it's a public constant?

melaniekung: oh like accessing them like this instead?
```
--- a/lib/task/search/arDocumentTask.class.php
+++ b/lib/task/search/arDocumentTask.class.php
@@ -38,7 +38,7 @@ class arSearchDocumentTask extends arBaseTask
             $this->log(sprintf("Fetching data for %s ID %d...\n", $slugObject->className, $slugObject->id));
 
             // Dummy type is needed for ES 6, this should be updated in ES 7.x when type is not required for getDocument()
-            $esType = QubitSearch::getInstance()->getEsType();
+            $esType = arElasticSearchPlugin::ES_TYPE;
```

jraddaoui: Or `QubitSearch::getInstance()::ES_TYPE;`, like we do in https://github.com/artefactual/atom/blob/qa/2.x/lib/QubitLftSyncer.class.php#L96.

melaniekung: updated, thanks @jraddaoui !

PR #1908, opened 2025-01-14
Update default CSP directive

sbreker:
Add form-action directive to default CSP header. Limit form target to 'self'.


anvit (approved):
:+1:  thanks @sbreker

PR #1909, opened 2025-01-22
Apply updates for PHP 8.x

### LINKED ISSUE #1709: Problem: PHP 7.4 dependency
### Current Behavior

PHP 7.4 is EOL since November last year.

### Expected Behavior

Support for newer and supported PHP-versions

### Possible Solution

_No response_

### Context and Notes

_No response_

### Version used

_No response_

_No response_

_No response_

### PHP version

_No response_

### Contact details

_No response_

anvit:
Update Atom and its associated dependencies (Symfony, net_gearman, sfPropel) to work with PHP 8.x. 

- For Symfony related changes, we've referred to [Friends of Symfony](https://github.com/FriendsOfSymfony1/symfony1).
- Net Gearman was updated to 2.2.3 from their [repo](https://github.com/brianlmoon/net_gearman) while retaining an AtoM specific customization to `Job.php`
- Added #[\ReturnTypeWillChange] attributes to functions that were extending certain vendor classes (in line with changes that Friends of Symfony applied)

Other changes that address PHP 8.x related deprecation warnings and errors —
- Dynamic property creation is deprecated in PHP 8.x so we've added declarations for properties created dynamically.
- Optional params are expected to the last params in a function declaration, so updated functions and their calls where this was not the case.
- Some comparisons (notable string to number comparison) has changed since PHP 8 and is a breaking change, so conditionals that were broken because of this have been updated.
-  There have been [changes to what are partially supported callables](https://wiki.php.net/rfc/deprecate_partially_supported_callables) in PHP 8.2, so call_user_func_array calls that use $this have been updated.
- `implode` no longer supports having the array first and separator second in the function call.
- `preg_replace` expects the second parameter to be either string or array so the places where we were using `NULL` have been updated to use empty string instead.


lib/job/arGearman.class.php
```diff
@@ -98,7 +98,7 @@ public static function getAbilities($options = [])
                 $abilities = array_merge($abilities, $config['worker_types'][$type]);
             }
         } else {
-            $abilities = call_user_func_array('array_merge', $config['worker_types']);
+            $abilities = call_user_func_array('array_merge', array_values($config['worker_types']));
```
djjuhasz: I've never been a big fan of the `call_user_func_array()` and similar functions because if the named function doesn't exist you don't get an error until run time.  With PHP 8+ I believe the following code is functionality equivalent, but allows static analysis of the `array_merge()` call:

```
array_merge(...array_values($config['worker_types']));
``` 

Test case: https://php-play.dev/?c=DwfgDgFmBQAkCGACAvIg2gcnhlA%2BdW2eiGAjKRgDQkBGNOy%2BZATBgLrUb3GbxGMlmFTgGMRDJs1YcS4nhgDMFNmwDc0OADNSKRCPgAbAwH0ArgGcApgCdjm0wDsRx%2BNevwAngAosbz8YBbGwBzSypEV3cPYwA3Q1NLcy8EAEoU9VhNZl1I-yDrUK8AOhLc6LiDBKTU9I0wawBLBwAXY2tk7Vr6pubEACIAHQdgGmtEAHpcPvVulraO5nSgA&v=8.4&f=html

vendor/symfony/lib/plugins/sfPropelPlugin/lib/vendor/propel/util/PropelConfiguration.php
```diff
@@ -57,6 +57,7 @@ public function __construct(array $parameters = array())
 	/**
 	 * @see        http://www.php.net/ArrayAccess
 	 */
+  #[\ReturnTypeWillChange]
```
djjuhasz: Why aren't these directives aligned with the comment and function blocks?

lib/model/QubitActor.php
```diff
@@ -70,7 +70,7 @@ public function __get($name)
                 return [];
         }
 
-        return call_user_func_array([$this, 'BaseActor::__get'], $args);
+        return call_user_func_array('BaseActor::__get', $args);
```
djjuhasz, line 73: Again, in cases where the first parameter of `call_user_func_array()` is a constant, I'd prefer to call the function using the "..." unpack operator for the parameter array to allow static analysis of the function call, e.g. 

```PHP
BaseActor::__get(...$args);
```

I realize that would mean a lot of changes to this commit, and may be impractical in this PR, but it would be a nice change in the future I think.

lib/task/import/csvAccessionImportTask.class.php
```diff
@@ -225,7 +225,9 @@ public function execute($arguments = [], $options = [])
                     if (!empty($identifiers) || !empty($identifierNotes)) {
                         $identifierTypes = $self->rowStatusVars['alternativeIdentifierTypes'];
 
-                        for ($index = 0; $index < max(count($identifiers), count($identifierNotes)); ++$index) {
```
djjuhasz, line 228: Wow, did PHP8 error  on the original `max()` call even though it's inside a conditional block that has  already confirmed that `$identifiers` and `$identifierNotes` are not empty?  It seems redundant to have to check for emptiness again. :open_mouth:

vendor/net_gearman/Net/Gearman/Job.php
```diff
@@ -65,29 +63,49 @@ abstract class Net_Gearman_Job
      */
     static public function factory($job, $conn, $handle, $initParams=array())
     {
-        $file = NET_GEARMAN_JOB_PATH . '/' . $job . '.php';
+        if (empty($initParams['path'])) {
```
djjuhasz, line 66: I think it's too much work for this PR, but it would be great to remove  `vendor/net_gearman` in favour of the composer package in the future.  Of course we'd have to figure out how to either wrap `Net_German_Job` or handle these customizations in the caller somehow.

plugins/sfRadPlugin/lib/sfRadPlugin.class.php
```diff
@@ -105,7 +105,7 @@ public function setProperty($name, $value)
 
             case 'languageNotes':
                 $note = $this->resource->getMemoryNotesByType(['noteTypeId' => QubitTerm::LANGUAGE_NOTE_ID])->offsetGet(0);
-                $missingNote = 0 === count($note);
+                $missingNote = 0 === strlen($note);
```
djjuhasz: I think `$missingNote = empty($note);` would achieve the same result and be easier to read.

plugins/sfRadPlugin/modules/sfRadPlugin/templates/indexSuccess.php
```diff
@@ -86,7 +86,7 @@
     <div class="titleNotes">
       <ul>
         <?php foreach ($resource->getNotesByTaxonomy(['taxonomyId' => QubitTaxonomy::RAD_TITLE_NOTE_ID]) as $item) { ?>
-          <?php if (0 != count($item->getContent(['cultureFallback' => true]))) { ?>
+          <?php if (0 != strlen($item->getContent(['cultureFallback' => true]))) { ?>
```
djjuhasz: Likewise, this could be replaced with the equivalent:

```PHP
if (!empty($item->getContent(['cultureFallback' => true])) {
```

Which is easier to read I think.

lib/validator/QubitValidatorCountable.class.php
```diff
@@ -36,6 +36,6 @@ protected function doClean($value)
      */
     protected function isEmpty($value)
     {
-        return 1 > count($value);
+        return 1 > count((array) $value);
```
djjuhasz: I think 

```PHP
return empty($value);
```

is equivalent and easier to read.  

I wonder why `sfValidatorBase` has an `isEmpty()` method at all instead of just using the built-in `empty()` function?  Maybe `empty()` has slightly different handling of "falsey" values?

djjuhasz (approved):
This looks really good to me @anvit and @melaniekung.  I made a handful of comments for possible improvements, but it's mostly just minor style changes and I didn't see any blockers.

vendor/symfony/lib/plugins/sfPropelPlugin/lib/vendor/propel/util/PropelConfiguration.php
anvit: Fixed! The fix look like a big change, but basically this entire file was using tabs instead of spaced for indents, except for the newly added lines which were using spaces.

lib/job/arGearman.class.php
anvit: Fixed! There are several other places where we use `call_user_func_array` and like you pointed out, that would be a big refactor, but we'll get to it in the future, since stuff like this is what made static analysis to find all of the things that would break with PHP 8.x challenging in the first place.

lib/task/import/csvAccessionImportTask.class.php
anvit: There is a scenario when one of the two is not empty, in which case it would enter the code block. The fix was to address `count` being called on a non countable value which was permissible pre PHP 8.0, but now throws an error. So in the scenario where `$identifierNotes` is empty for instance but `$identifiers` is not, we wouldn't want to attempt running `count($identifierNotes)`;

vendor/net_gearman/Net/Gearman/Job.php
anvit: Yes! This is on our radar for the future. Overriding one file did not sound trivial given how Net Gearman is written, so we saved it for a future task.

lib/task/import/csvAccessionImportTask.class.php
djjuhasz: Ahh, yeah that makes sense.

vendor/net_gearman/Net/Gearman/Job.php
djjuhasz: Nice!

vendor/symfony/lib/plugins/sfPropelPlugin/lib/vendor/propel/util/PropelConfiguration.php
djjuhasz: Great. :)

PR #1911, opened 2025-01-24
Add OIDC 'unauthorized' failure message

sbreker:
If an error occurs in AtoM after successful authentication with the OIDC endpoint, end the AtoM session and redirect to the standard AtoM 'Unauthorized' message template.

Previously, errors that occurred in AtoM after successful authentication with the OIDC endpoint would fail silently redirecting the user to the AtoM home page in an unauthenticated state.


anvit (approved):
Looks good! Thanks @sbreker

PR #1916, opened 2025-01-31
Fix DO images in the edit DO page (#1915)

### LINKED ISSUE #1915: Problem: Edit Digital Object page is missing images
### Current Behavior

### Expected Behavior

anvit:
Using a div with appropriate styling classes instead of a section for the image in the edit DO page header since blank.js filters/removes any sections that do not have content meant for sections.

Also updating the resource sent to the QubitAcl check in DO's `editAction` to point to the parents' object instead of the resources' object (similar to [viewAction](https://github.com/artefactual/atom/blob/qa/2.x/apps/qubit/modules/digitalobject/actions/viewAction.class.php#L97-L125)) in the scenario when that object is empty.  In PHP 7.4 the Acl check succeeds despite `$this->resource->object` being empty since it isn't strictly `null` (QubitAcl tests `null !== $resource` but doesn't test if it is empty. `$this->resource->object` is empty in PHP 7.4 and null in PHP 8.x), which ends up being a problem in PHP 8.x and the Acl check failing.

### DIFF
--- a/apps/qubit/modules/digitalobject/actions/showComponent.class.php
+++ b/apps/qubit/modules/digitalobject/actions/showComponent.class.php
@@ -74,12 +74,20 @@ class DigitalObjectShowComponent extends sfComponent
      */
     private function checkShowGenericIcon()
     {
+        $resourceObject = $this->resource->object;
+
+        // Get the parent resouce (IO or Actor) for Acl check
+        // if resource->object is empty and parent exists
+        if (!$resourceObject && $this->resouce->parent) {
+            $resourceObject = $this->resource->parent->object;
+        }
+
         switch ($this->usageType) {
             case QubitTerm::REFERENCE_ID:
-                return !QubitAcl::check($this->resource->object, 'readReference');
+                return !QubitAcl::check($resourceObject, 'readReference');
 
             case QubitTerm::THUMBNAIL_ID:
-                return !QubitAcl::check($this->resource->object, 'readThumbnail');
+                return !QubitAcl::check($resourceObject, 'readThumbnail');
         }
     }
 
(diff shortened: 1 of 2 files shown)


apps/qubit/modules/digitalobject/actions/showComponent.class.php
```diff
@@ -74,12 +74,19 @@ public function execute($request)
      */
     private function checkShowGenericIcon()
     {
+        $resourceObject = $this->resource->object;
+
+        // For digital objects, get the parent resouce (IO or Actor) for Acl check
+        if (!$resourceObject && ('QubitDigitalObject' === get_class($this->resource))) {
+            $resourceObject = $this->resource->parent->object;
+        }
```
jraddaoui: I think it has to be a `QubitDigitalObject` to be in this module as the resource, I'd change that to check that parent is not null, that way you kind of check the usage type of the DO.

apps/qubit/modules/digitalobject/actions/showComponent.class.php
```diff
@@ -74,12 +74,19 @@ public function execute($request)
...
+
         switch ($this->usageType) {
             case QubitTerm::REFERENCE_ID:
-                return !QubitAcl::check($this->resource->object, 'readReference');
+                return !QubitAcl::check($resourceObject, 'readReference');
 
             case QubitTerm::THUMBNAIL_ID:
-                return !QubitAcl::check($this->resource->object, 'readThumbnail');
+                return !QubitAcl::check($resourceObject, 'readThumbnail');
         }
```
jraddaoui, line 91: Based on this switch it will always be the resource parent, unless `$this->usageType` doesn't match `$this->resource->usageId`. Only the master will be related to the IO/Actor.

jraddaoui (approved):
:ok_hand:

PR #1919, opened 2025-02-04
Bump gh actions/cache to v4

anvit:
Upgrading actions/cache to v4 since v2 is deprecated.


melaniekung (approved):
lgtm, thanks @anvit !

PR #1926, opened 2025-02-12
Re-encode AV Files Only when Necessary

### LINKED ISSUE #1912: Enhancement: Create Audio and Video Derivatives with FFmpeg Only if Necessary
### Current Behavior

The `createAudioDerivative` and `createVideoDerivative` methods in the `QubitDigitalObject` class process audio and video files with FFmpeg even if the master files are in the correct format. This is an issue when ingesting large amounts of video and audio, since FFmpeg can take a long time to re-encode files (especially large video files).

For example, video reference derivatives are created in this format:
- .mp4 container
- h264 video
- yuv420p pixel format
- aac audio
- 44.1 kHz audio sample rate

But, if the master video file is *already in this format*, then FFmpeg is still used to re-encode all streams, which is inefficient.

### Expected Behavior

AtoM should skip re-encoding audio/video streams with FFmpeg if the streams are already encoded correctly.

### Possible Solution

My organization (National Centre for Truth and Reconciliation) has a working fix in our custom version of AtoM which I am willing to adapt to merge into `qa/2.x`.

Basically, it does this:

- Check if `ffprobe` is available.
- Use `ffprobe` to analyze streams.
- If the video is already in the correct format, **copy it**.
- If the video is partially in the correct format, construct a minimal FFmpeg command that encodes only the streams that need re-encoding.
- If the video does not match any of the correct parameters, then do a full re-encode.

These changes make it much quicker to get video and audio into AtoM when they are in the proper format already.

### Context and Notes

An example command for when only the audio is in the wrong format is this:

`ffmpeg -y -i <INPUT> -c:v copy -c:a aac -ar 44100 <OUTPUT> 2>&1`

The `-c:v copy` instruction is used to tell FFmpeg to copy the video. This is *much* faster than re-encoding with `libx264`.

### Version used

AtoM 2.8.2

_No response_

_No response_

### PHP version

PHP 7.4

### Contact details

_No response_

danloveg:
Addresses https://github.com/artefactual/atom/issues/1912

These changes concern the creation of digital object reference derivatives for audio and video (AV) files. The changes here are authored by myself and @yenaing-oo.

We have optimized the `ffmpeg` commands used to generate audio and video derivatives by analyzing the master files with `ffprobe` before encoding. Based on a file's parameters, a minimal FFmpeg command is generated to make only those changes that are required to create the reference file.

For example, if only the audio needs to be re-encoded in a video file, then the video stream is copied and the audio is re-encoded to bring it in line with the expected audio encoding and sample rate.

The purpose of this change is to make importing audio and video objects quicker, especially when files are already in the correct format. If an AV file is in the right format already, it is simply `copy()`-ed, rather than processed with FFmpeg.

---

We created a test package of audio and video files that this change can be tested against:  [issue-1912-test-files-import.zip](https://github.com/user-attachments/files/18774430/issue-1912-test-files-import.zip)

Assuming AtoM is being run in Docker, extract them to the root of the repository, and run the following to import them:

```shell
php symfony csv:import issue-1912-test-files-import.csv --index
```

We saw an improvement in import time using the changes included in this PR:

- With the new changes, importing the test package took on average 11.39s (over three runs)
- *Without* these changes, importing the test package took on average 15.05s (over three runs)

To run the timing tests, we used this command to avoid including nested-set build time and search indexing time:

```shell
php symfony cc && time php symfony csv:import issue-1912-test-files-import.csv --skip-nested-set-build
```

This is a modest timing improvement, but since a lot of the test files do require re-encoding we'd expect to see a greater timing improvement for files that don't require any re-encoding.

### DIFF
--- a/lib/model/QubitDigitalObject.php
+++ b/lib/model/QubitDigitalObject.php
@@ -2721,23 +2721,19 @@ class QubitDigitalObject extends BaseDigitalObject
 
         foreach($ffProbeInfo['streams'] as $stream) {
             // Read only the first video and audio streams
-            if (!empty($format['video']) && !empty($format['audio']))
-            {
+            if (!empty($format['video']) && !empty($format['audio'])) {
                 break;
             }
 
             $codecType = strtolower($stream['codec_type']);
 
-            if ($codecType === 'video' && empty($format['video']))
-            {
+            if ($codecType === 'video' && empty($format['video'])) {
                 $format['video'] = [];
                 $format['video']['codec_name'] = $stream['codec_name'];
                 $format['video']['pix_fmt']    = $stream['pix_fmt'] ?? null;
                 $format['video']['width']      = (int) $stream['width'];
                 $format['video']['height']     = (int) $stream['height'];
-            }
-            else if ($codecType === 'audio' && empty($format['audio']))
-            {
+            } else if ($codecType === 'audio' && empty($format['audio'])) {
                 $format['audio'] = [];
                 $format['audio']['codec_name'] = $stream['codec_name'];
                 $format['audio']['sample_rate'] = (int) $stream['sample_rate'];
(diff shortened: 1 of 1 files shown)


anvit (approved):
Thanks @danloveg , @yenaing-oo ! This looks great!

PR #1927, opened 2025-02-13
Render markdown for BS5 values.

melaniekung:


anvit (approved):
This makes sense. Thanks mel!

PR #1928, opened 2025-02-13
Fix object for Acl check in showComponent (#1915)

### LINKED ISSUE #1915: Problem: Edit Digital Object page is missing images

anvit:
Fix typo and update showComponent to always use parent object for Acl check.

### DIFF
--- a/apps/qubit/modules/digitalobject/actions/showComponent.class.php
+++ b/apps/qubit/modules/digitalobject/actions/showComponent.class.php
@@ -74,20 +74,14 @@ class DigitalObjectShowComponent extends sfComponent
      */
     private function checkShowGenericIcon()
     {
-        $resourceObject = $this->resource->object;
-
-        // Get the parent resouce (IO or Actor) for Acl check
-        // if resource->object is empty and parent exists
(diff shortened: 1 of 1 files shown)


apps/qubit/modules/digitalobject/actions/showComponent.class.php
```diff
@@ -74,13 +74,8 @@ public function execute($request)
...
     private function checkShowGenericIcon()
     {
-        $resourceObject = $this->resource->object;
-
         // Get the parent resouce (IO or Actor) for Acl check
-        // if resource->object is empty and parent exists
-        if (!$resourceObject && $this->resouce->parent) {
-            $resourceObject = $this->resource->parent->object;
-        }
+        $resourceObject = $this->resource->parent->object;
```
jraddaoui: Do you really need `$resourceObject` as a var? I'd do it directly inside the switch on each call. That way you prevent the possibility of somebody adding master usage type to the switch and using the parent when it shouldn't.

jraddaoui (approved):
:ok_hand:

PR #1929, opened 2025-02-13
Fix Acl check for QubitContactInformation.

melaniekung:
Address `null` `$resource` in `QubitAcl` for `QubitContactInformation`.

### DIFF
--- a/apps/qubit/modules/contactinformation/actions/indexAction.class.php
+++ b/apps/qubit/modules/contactinformation/actions/indexAction.class.php
@@ -21,14 +21,11 @@ class ContactInformationIndexAction extends sfAction
 {
     public function execute($request)
     {
-        $hasEditAccess = QubitAcl::check($this->resource, 'update')
-            || $this->getUser()->hasGroup(QubitAclGroup::EDITOR_ID);
-        // Check user authorization
-        if (!($this->getUser()->isAuthenticated() && $hasEditAccess)) {
+        $this->resource = QubitContactInformation::getById($request->id);
+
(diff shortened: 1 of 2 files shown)


apps/qubit/modules/contactinformation/actions/indexAction.class.php
```diff
@@ -21,15 +21,15 @@ class ContactInformationIndexAction extends sfAction
...
     public function execute($request)
     {
+        $this->resource = QubitContactInformation::getById($request->id);
+
         $hasEditAccess = QubitAcl::check($this->resource, 'update')
             || $this->getUser()->hasGroup(QubitAclGroup::EDITOR_ID);
         // Check user authorization
         if (!($this->getUser()->isAuthenticated() && $hasEditAccess)) {
             QubitAcl::forwardUnauthorized();
         }
```
jraddaoui: If I understand correctly, `QubitAcl::check()` will now check if the user is authenticated AND if the user is an admin or an editor. So we could simplify this entire block to:

```php
if (!QubitAcl::check($this->resource, 'update')) {
    QubitAcl::forwardUnauthorized();
}
```

Also, `update` in an index action?

melaniekung: ah, right. thanks, @jraddaoui !

jraddaoui (approved):
Thanks @melaniekung! Still feels rare to use `update` from an index action, but not sure if this is used as a component in the edit pages from other entities only, so I'll leave that up to you if you want to investigate ;)

PR #1930, opened 2025-02-14
Revert fix for showComponent and fix typo (#1915)

### LINKED ISSUE #1915: Problem: Edit Digital Object page is missing images

anvit:
Revert the change to showComponent that was changing the showComponent to always use the parent object for reference and thumbnail since this broke description page with attatched digital objects. Updating the last fix to showComponent with the typo fixed resolves this bug since it uses the parent object only when it is not null.

### DIFF
--- a/apps/qubit/modules/digitalobject/actions/showComponent.class.php
+++ b/apps/qubit/modules/digitalobject/actions/showComponent.class.php
@@ -74,20 +74,14 @@ class DigitalObjectShowComponent extends sfComponent
      */
     private function checkShowGenericIcon()
     {
-        $resourceObject = $this->resource->object;
-
-        // Get the parent resouce (IO or Actor) for Acl check
-        // if resource->object is empty and parent exists
-        if (!$resourceObject && $this->resource->parent) {
-            $resourceObject = $this->resource->parent->object;
-        }
-
         switch ($this->usageType) {
             case QubitTerm::REFERENCE_ID:
-                return !QubitAcl::check($resourceObject, 'readReference');
+                // Use the parent resource (IO or Actor) for Acl check
+                return !QubitAcl::check($this->resource->object, 'readReference');
 
             case QubitTerm::THUMBNAIL_ID:
-                return !QubitAcl::check($resourceObject, 'readThumbnail');
+                // Use the parent resource (IO or Actor) for Acl check
+                return !QubitAcl::check($this->resource->object, 'readThumbnail');
         }
     }
 
--- a/apps/qubit/modules/digitalobject/actions/showComponent.class.php
+++ b/apps/qubit/modules/digitalobject/actions/showComponent.class.php
@@ -74,14 +74,20 @@ class DigitalObjectShowComponent extends sfComponent
      */
     private function checkShowGenericIcon()
     {
+        $resourceObject = $this->resource->object;
+
+        // Get the parent resouce (IO or Actor) for Acl check
+        // if resource->object is empty and parent exists
+        if (!$resourceObject && $this->resource->parent) {
+            $resourceObject = $this->resource->parent->object;
+        }
+
         switch ($this->usageType) {
             case QubitTerm::REFERENCE_ID:
-                // Use the parent resource (IO or Actor) for Acl check
-                return !QubitAcl::check($this->resource->parent->object, 'readReference');
+                return !QubitAcl::check($resourceObject, 'readReference');
 
             case QubitTerm::THUMBNAIL_ID:
-                // Use the parent resource (IO or Actor) for Acl check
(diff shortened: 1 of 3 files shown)


apps/qubit/modules/digitalobject/actions/showComponent.class.php
```diff
@@ -74,14 +74,20 @@ public function execute($request)
...
     {
+        $resourceObject = $this->resource->object;
+
+        // Get the parent resouce (IO or Actor) for Acl check
+        // if resource->object is empty and parent exists
+        if (!$resourceObject && $this->resource->parent) {
+            $resourceObject = $this->resource->parent->object;
+        }
+
         switch ($this->usageType) {
```
jraddaoui, line 77: I see, `$this->usageType` doesn't have to be the usage type from `$this->resource`. Sorry for the confusion, I should have looked at this deeper.

The show DO component is meant to be called with the master DO as resource, but that doesn't seem to be the case in the edit representation component. The edit representation component is called **only** from the DO edit page, passing the resource (master DO) and the representation:

https://github.com/artefactual/atom/blob/qa/2.x/apps/qubit/modules/digitalobject/templates/editSuccess.php#L63
https://github.com/artefactual/atom/blob/qa/2.x/apps/qubit/modules/digitalobject/templates/editSuccess.php#L99

Then, as I said, this component calls the show component using the representation as resource instead of the master:

https://github.com/artefactual/atom/blob/qa/2.x/apps/qubit/modules/digitalobject/templates/_editRepresentation.php#L12

I believe the actual fix for the initial issue is just changing `$representation` by `$resource` in that call to the show DO component.

---

Just looking at this component, I can see more usages of `$this->resource->object` (lines 44, 98, 103, 118 and 142 in `qa/2.x`). I can also see some of the child components using it and using properties that may only be set in the master DO (`mimeType`, `mediaTypeId`):

https://github.com/artefactual/atom/blob/qa/2.x/apps/qubit/modules/digitalobject/actions/showCompoundComponent.class.php#L36
https://github.com/artefactual/atom/blob/qa/2.x/apps/qubit/modules/digitalobject/actions/showDownloadComponent.class.php#L62
https://github.com/artefactual/atom/blob/qa/2.x/apps/qubit/modules/digitalobject/actions/showGenericIconComponent.class.php#L40

While other child components do consider the case of the resource being a representation, and also have a parent look-up:

https://github.com/artefactual/atom/blob/qa/2.x/apps/qubit/modules/digitalobject/actions/showAudioComponent.class.php#L34-L40
https://github.com/artefactual/atom/blob/qa/2.x/apps/qubit/modules/digitalobject/actions/showImageComponent.class.php#L34-L40
https://github.com/artefactual/atom/blob/qa/2.x/apps/qubit/modules/digitalobject/actions/showImageComponent.class.php#L34-L40
https://github.com/artefactual/atom/blob/qa/2.x/apps/qubit/modules/digitalobject/actions/showTextComponent.class.php#L34-L40
https://github.com/artefactual/atom/blob/qa/2.x/apps/qubit/modules/digitalobject/actions/showVideoComponent.class.php#L34-L40

---

If you ask me:

- I'd make sure the edit representation component doesn't use the representation for a reason I'm not seeing.
- I'd fix the call from the edit representation component, using the master DO instead of the representation as resource.
- I'd make sure that's the only place using it with a representation (don't trust me).
- Optionally: I'd make sure this component and its children components are only called with a master DO and return an error (or error view) to prevent that in the future.
- Optionally: I'd remove the parent look-ups from this component and its children.

jraddaoui (changes requested):
Thanks @anvit! This goes deeper than I thought.

apps/qubit/modules/digitalobject/actions/showComponent.class.php
anvit: @jraddaoui thanks so much for looking into this! 
I couldn't find logical reason for why we'd be using representation instead of the appropriate resource, and I've searched and I couldn't find any other instances where we send the representation instead of the resource.
I did find a few instances where we use the associated digital object associated by id (for instance rad, mods, and dacs templates — https://github.com/artefactual/atom/blob/36393ea8f4e9e9ffc20e2919bd20ca02f2108247/plugins/sfRadPlugin/modules/sfRadPlugin/templates/indexSuccess.php#L55) but that seems correct.

jraddaoui (approved):
:ok_hand:

apps/qubit/modules/digitalobject/actions/showComponent.class.php
jraddaoui: Yeah, that will always be the master DO.

PR #1931, opened 2025-02-19
Refactor getByTitleIdentifierAndRepo()

sbreker:
Refactor getByTitleIdentifierAndRepo() to use a SQL query instead of looking for a description match in the ES index.

This should make description matching during csv or xml import more successful since it is possible for the AtoM DB and the ES index to get out of sync which can cause matching to fail.


anvit (approved):
LGTM! Thanks!

PR #1933, opened 2025-02-21
Ensure check for `pdftotext` works on all environments.

melaniekung:
Some environments (especially restricted hosting environments, such as Rocky9) do not allow `which` to be executed inside `exec()`.


anvit (approved):
:+1: Thanks for figuring this out @melaniekung

PR #1936, opened 2025-02-25
Remove tests asserting PDO class type

sbreker:
Removed two tests that made assertions that the class type should be type 'DebugPDO'. This is not a good assertion to make as it prevents this class type from changing.

Also removed a commented out line.


anvit (approved):
Looks good! Thanks @sbreker

PR #1939, opened 2025-03-04
Bump composer dependencies

anvit:


melaniekung (approved):
lgtm, thanks @anvit

anvit:
Marking this as a WIP. PHP-CS Fixer has updated defaults for `operator_linebreak`, `php_unit_data_provider_method_order` that will need to be reviewed, and a few composer packages have updated their supported PHP version to be `>8.1.2` instead of `>7.4` so this is blocked by #1942 until gh unit test related changes have been addressed.

PR #1941, opened 2025-03-04
Fix exports for PHP 8.x (#1940)

### LINKED ISSUE #1940: Problem: CSV export failing in PHP 8.x
### Current Behavior

1. Attempt to export a CSV
2. Note that the CSV export job fails with this error:
```
Exception: count(): Argument #1 ($value) must be of type Countable|array, int given
File: /atom/src/lib/job/arExportJob.class.php
Line: 58
```

### Expected Behavior

CSV export should succeed.

anvit:
Fix condition for counting number of exported items to just use the variable instead of count. This is always an integer that is incremented, not a countable value, and PHP 8.0 only allows countable values as arguments for count.

### DIFF
--- a/lib/job/arExportJob.class.php
+++ b/lib/job/arExportJob.class.php
@@ -55,7 +55,7 @@ class arExportJob extends arBaseJob
 
         $this->doExport($tempPath);
 
-        if (count($this->itemsExported) > 0) {
+        if ($this->itemsExported > 0) {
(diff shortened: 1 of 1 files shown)


melaniekung (commented):
LGTM , thanks @anvit

melaniekung (approved):
LGTM , thanks @anvit

PR #1948, opened 2025-03-07
Fix add function page for PHP 8.x (#1947)

### LINKED ISSUE #1947: Problem: Unable to add functions in PHP 8.x
### Current Behavior

1. Log in
2. Attempt to add a function
3. There is a 500 error, with this in the error log:
```
WARNING: [pool atom] child 15 said into stderr: "NOTICE: PHP message: PHP Fatal error:  Uncaught TypeError: get_class(): Argument #1 ($object) must be of type object, null given in /atom/src/plugins
/qbAclPlugin/lib/QubitAcl.class.php:767"                                                                                                                                                                                               
WARNING: [pool atom] child 15 said into stderr: "Stack trace:"
WARNING: [pool atom] child 15 said into stderr: "#0 /atom/src/plugins/qbAclPlugin/lib/QubitAcl.class.php(132): QubitAcl::checkAccessByClass(NULL, Object(myUser), 'create', Array)"
WARNING: [pool atom] child 15 said into stderr: "#1 /atom/src/apps/qubit/modules/function/actions/editAction.class.php(64): QubitAcl::check(NULL, Array)"
WARNING: [pool atom] child 15 said into stderr: "#2 /atom/src/plugins/sfIsdfPlugin/modules/sfIsdfPlugin/actions/editAction.class.php(47): FunctionEditAction->earlyExecute()"
WARNING: [pool atom] child 15 said into stderr: "#3 /atom/src/apps/qubit/modules/default/actions/editAction.class.php(33): sfIsdfPluginEditAction->earlyExecute()"
WARNING: [pool atom] child 15 said into stderr: "#4 /atom/src/apps/qubit/modules/function/actions/editAction.class.php(29): DefaultEditAction->execute(Object(sfWebRequest))"
WARNING: [pool atom] child 15 said into stderr: "#5 /atom/src/cache/qubit/prod/config/config_core_compile.yml.php(975): FunctionEditAction->execute(Object(sfWebRequest))"
WARNING: [pool atom] child 15 said into stderr: "#6 /atom/src/cache/qubit/prod/config/config_core_compile.yml.php(970): sfExecutionFilter->executeActio..."
```
4. Also attempt to visit <atom-site-url>/function/add
5. Note the same error

### Expected Behavior

Logged out users should get a message telling them they're not authorized to access the page if they attempt to access the add functions page, not a 500 error. Logged in users should be able to add functions as normal

anvit:
Fix editAction for function to use resource instead of $this->parent which is always null. $resource needs to be valid in QubitAcl since get_class will not accept null values in PHP 8.x and expects a valid object.

### DIFF
--- a/apps/qubit/modules/function/actions/editAction.class.php
+++ b/apps/qubit/modules/function/actions/editAction.class.php
@@ -61,7 +61,7 @@ class FunctionEditAction extends DefaultEditAction
             $this->form->setWidget('serialNumber', new sfWidgetFormInputHidden());
         } else {
             // Check authorization
-            if (!QubitAcl::check($this->parent, 'create')) {
+            if (!QubitAcl::check($this->resource, 'create')) {
                 QubitAcl::forwardUnauthorized();
             }
         }


sbreker:
Should a check be added in QubitAcl::check() or QubitAcl::checkAccessByClass() to catch null values being passed in?

anvit:
> Should a check be added in QubitAcl::check() or QubitAcl::checkAccessByClass() to catch null values being passed in?

@sbreker I've considered this, but wasn't sure what the correct response would be. We could throw an exception if we wanted, and we'd get the exception instead of the error with a stack trace in the logs, but the root causes would still need to be addressed.

apps/qubit/modules/function/actions/editAction.class.php
```diff
@@ -61,7 +61,7 @@ protected function earlyExecute()
             $this->form->setWidget('serialNumber', new sfWidgetFormInputHidden());
         } else {
             // Check authorization
-            if (!QubitAcl::check($this->parent, 'create')) {
+            if (!QubitAcl::check($this->resource, 'create')) {
```
djjuhasz, line 64: I think this is fine.  If `QubitFunctionObject` allowed hierarchies, then it would make sense to check permissions against the parent because the resource we are creating doesn't exist yet (which is why we check against the parent for actors, information objects, etc.).  `QubitAcl` has hard-coded roles for editing/creating a function resource anyway (https://github.com/artefactual/atom/blob/qa/2.x/plugins/qbAclPlugin/lib/QubitAcl.class.php#L772), so any `QubitFunctionObject` (including an empty one) will produce the same results here.

djjuhasz (approved):
LGTM :+1:

PR #1949, opened 2025-03-10
Fix integration test repos

anvit:
Update integration test workflow to run apt update first before attempting to install packages.


melaniekung (approved):
LGTM, thanks @anvit

PR #1956, opened 2025-03-13
Update gh unit test workflow in AtoM to use ubuntu 24.04

### LINKED ISSUE #1942: Update gh unit test workflow in AtoM to use ubuntu 24.04
GitHub is discontinuing Ubuntu 20.04, so we would need to update the unit test workflow to use 24.04.

PHP 8.x also breaks unit tests, since they're dependent on DebugPDO which has not been updated for PHP 8.0. The themselves tests also need updates since they contain a lot of PHP 8.x related deprecation warnings.

### LINKED ISSUE #1965: Replace DebugPDO with PropelPDO for Tests
DebugPDO class is not compatible with modern PHP versions - replacing DebugPDO with PropelPDO ensures compatibility with current and future PHP versions.

Currently strictly used for PHPUnit tests.

melaniekung:


lib/model/om/BaseTaxonomy.php
```diff
@@ -80,7 +80,7 @@ public function __isset($name)
 
     try
     {
-      return call_user_func_array(array($this, 'QubitObject::__isset'), $args);
+      return call_user_func_array('QubitObject::__isset', $args);
```
anvit, line 83: I would try `QubitObject::__isset(...$args);` and see if that works here. We made similar fixes to QubitActor as well to allow static analysis tools to potentially catch issues with these functions in the future.

lib/model/om/BaseTaxonomy.php
```diff
@@ -129,7 +129,7 @@ public function __get($name)
 
     try
     {
-      return call_user_func_array(array($this, 'QubitObject::__get'), $args);
+      return call_user_func_array( 'QubitObject::__get', $args);
```
anvit, line 132: Same as above, try `QubitObject::__get(...$args);`

lib/model/om/BaseTaxonomy.php
```diff
@@ -212,7 +212,7 @@ public function __set($name, $value)
       $options = $args[2];
     }
 
-    call_user_func_array(array($this, 'QubitObject::__set'), $args);
+    call_user_func_array('QubitObject::__set', $args);
```
anvit, line 215: Same as above, try `QubitObject::__set(...$args);`

lib/task/import/validate/csvValidatorResultCollection.php
```diff
@@ -129,13 +134,13 @@ public function renderResultsAsText(bool $verbose = false): string
         $outputString = "CSV Results:\n";
 
         foreach ($this->results as $result) {
+            $filename = $result->getFilename();
             if ($filename !== $result->getFilename()) {
```
anvit, line 138: I'm not sure if this fix would work as intended. This condition will now never be false, wouldn't it?

lib/task/tools/settingsTask.class.php
```diff
@@ -121,7 +121,12 @@ public function getSetting($name, $options = [])
             $criteria->add($this->ormClasses['setting']::SCOPE, $options['scope']);
         }
 
-        return $this->ormClasses['setting']::getOne($criteria);
+        try {
+            return $this->ormClasses['setting']::getOne($criteria);
+        } catch (Exception $e) {
+            throw new Exception('Setting does not exist.');
```
anvit, line 132: Could we include a setting name in the exception message so that it's easier to figure out which setting is causing a problem?

test/phpunit/PhysicalobjectCsvImporterTest.php
```diff
@@ -629,7 +637,7 @@ public function testProgressUpdateFreqOne()
     {
         $importer = new PhysicalObjectCsvImporter($this->context, $this->vdbcon);
         $importer->setOption('progressFrequency', 1);
-        $data = $importer->processRow(['name' => 'foo', 'culture' => 'en']);
+        $data = $importer->processRow(['name' => 'foo', 'culture' => 'en', 'type' => '', 'legacyId' => '', 'location' => '', 'descriptionSlugs' => '']);
```
anvit: Are these required fields? If yes, do we have a test to check that it fails as expected when the required fields are not included?

test/phpunit/QubitFindingAidTest.php
```diff
@@ -23,7 +24,7 @@ public function setUp(): void
     public function testSetResourceFromConstructor()
     {
         $resource = new QubitInformationObject();
-        $findingAid = new QubitFindingAid($resource);
+        $findingAid = new QubitFindingAid($resource, ['homeDir' => '/']);
```
anvit: Instead of updating the homeDir for all the tests, maybe changing the home dir permissions for the github unit test yaml would be better, similar to how we updated integration tests?

vendor/symfony/lib/plugins/sfPropelPlugin/lib/vendor/propel/util/DebugPDO.php
```diff
@@ -321,13 +322,13 @@ public function exec($sql)
...
-	 * 
+	 *
 	 * Overridden for query counting and logging.
-	 * 
+	 *
 	 * @see        http://php.net/manual/en/pdo.query.php for a description of the possible parameters.
 	 * @return     PDOStatement
 	 */
-	public function query()
+	public function query(string $query, ?int $fetchMode = null, mixed ...$fetchModeArgs): PDOStatement|false
```
anvit: Is DebugPDO still in use after changes to the tests?

anvit (changes requested):

lib/model/om/BaseTaxonomy.php
melaniekung: will address these in a future refactor

vendor/symfony/lib/plugins/sfPropelPlugin/lib/vendor/propel/Propel.php
```diff
@@ -663,7 +663,8 @@ private static function processDriverOptions($source, &$write_to)
 			}
 			$key = constant($key);
 
-			$value = $optiondata['value'];
+			isset($optiondata['value']) ? $value = $optiondata['value'] : $value = null;
```
anvit: I would consider re-writing this as `$value = isset($optiondata['value']) ?  $optiondata['value'] :  null;` to keep things DRY.

anvit (approved):
Added a minor suggestion but the rest looks good

PR #1961, opened 2025-03-19
Fix PHP 8.x deprecation warnings (#1960)

### LINKED ISSUE #1960: Problem: PHP 8.x Deprecation warnings in AtoM logs
Tasks like generating finding aid are generating deprecation warnings. These are not visible in the UI but they still show up in AtoM logs and should be addressed.

anvit:
- Add ReturnTypeWillChange attribute to all offset Get/Set/Unset/Exists methods
- Fix deprecation warnings for call_user_func_array calls
- Fix instances of strlen and strtolower in sfEadPlugin, BaseActor, and
BaseProperty that could run with null values and generate deprecation
warnings, since PHP 8 expects strlen and strtolower to only accept
strings.
- Add missing property to sfEadPlugin.class.php since dynamically creating
properties is deprecated in PHP 8.

### DIFF
--- a/lib/model/QubitDigitalObject.php
+++ b/lib/model/QubitDigitalObject.php
@@ -1117,7 +1117,7 @@ class QubitDigitalObject extends BaseDigitalObject
                 return $this->values['reference'];
         }
 
-        return call_user_func_array([$this, 'BaseDigitalObject::__get'], $args);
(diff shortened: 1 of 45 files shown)


melaniekung (approved):
LGTM , thanks @anvit

PR #1962, opened 2025-03-24
Add js challenge for human verification

sbreker:
This commit introduces a new js challenge feature aimed at verifying that a visitor is a real human. The implementation includes:

- A dedicated challenge page that displays a friendly message with countdown timer before automatic redirection.
- Logic to set cookies indicating success of js browser tests.

### DIFF
--- a/docker/bootstrap.php
+++ b/docker/bootstrap.php
@@ -116,6 +116,18 @@ all:
       worker-src 'self' blob:;
       connect-src 'self' https://*.google-analytics.com https://*.analytics.google.com https://*.googletagmanager.com https://*.googleapis.com *.google.com https://*.gstatic.com  data: blob:;
       frame-ancestors 'self';
+  user_challenge:
+    activated: false
+    test_headless: true
+    cookiename_visited: atom_visited
+    cookiename_headless: atom_headless
+    cookiename_js: atom_js
+    salt: 'REPLACE_WITH_SECRET'
+    delay_seconds: 5
+    cookie_days: 3
+    cidr_exceptions:
+      - '192.168.1.0/24'
+      - '10.0.0.0/8'
 
 EOT;
 
--- a/lib/QubitGeoIpHelper.class.php
+++ b/lib/QubitGeoIpHelper.class.php
@@ -1,0 +1,218 @@
+<?php
+
+/*
+ * This file is part of the Access to Memory (AtoM) software.
+ *
+ * Access to Memory (AtoM) is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * Access to Memory (AtoM) is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with Access to Memory (AtoM).  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+use GeoIp2\Database\Reader;
+
+class QubitGeoIpHelper
+{
+    protected $asnDbPath;
+    protected $cityDbPath;
+
+    public function __construct($asnDbPath = null, $cityDbPath = null)
+    {
+        $this->setAsnDbPath($asnDbPath ?: sfConfig::get('app_geoip_asn_db_path', ''));
+        $this->setCityDbPath($cityDbPath ?: sfConfig::get('app_geoip_city_db_path', ''));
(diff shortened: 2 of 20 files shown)


lib/QubitGeoIpHelper.class.php
```diff
@@ -0,0 +1,135 @@
...
+     *
+     * @param string $ip
+     *
+     * @return bool true if the IP is private/reserved
+     */
+    public function isPrivateIP($ip)
+    {
+        $long = ip2long($ip);
+
+        return ($long >= ip2long('10.0.0.0') && $long <= ip2long('10.255.255.255'))
```
djjuhasz: Are [IPv6 private addresses](https://en.wikipedia.org/wiki/Private_network#Private_IPv6_addresses) relevant here?

lib/QubitGeoIpHelper.class.php
```diff
@@ -0,0 +1,135 @@
...
+
+    /**
+     * Check if an IP address is within a given CIDR range.
+     *
+     * @param string $ip   the IP address to check
+     * @param string $cidr The CIDR range, e.g., "192.168.1.0/24".
+     *
+     * @return bool true if $ip is in the CIDR range, false otherwise
+     */
+    public function ipInCidr($ip, $cidr)
```
djjuhasz, line 108: I'm not sure what this will return if `$ip` is an IPv6 address. :thinking:

lib/filter/QubitUserChallengeFilter.class.php
```diff
@@ -0,0 +1,224 @@
...
+
+        $result = $this->shouldBypassChallenge($country, $clientAsn, $geoIpHelper, $userAgent);
+        if (true === $result) {
+            $filterChain->execute();
+
+            return;
+        }
+
+        // Convert days to seconds since epoch.
+        $cookielifespan = time() + 60 * 60 * 24 * sfConfig::get('app_user_challenge_cookie_days', 90);
```
djjuhasz: Nitpick: should be `$cookieLifespan`

djjuhasz (commented):
This is very cool @sbreker! From reading the code I think it will challenge IPv6 remote addresses properly, but it might be worth adding some IPv6 unit tests to confirm.

lib/filter/QubitUserChallengeFilter.class.php
sbreker: Thanks! Fixed...

lib/QubitGeoIpHelper.class.php
sbreker: I have added support for this. See latest commits.

lib/QubitGeoIpHelper.class.php
sbreker: I have added support for this. See latest commits.

docker/bootstrap.php
```diff
@@ -116,6 +116,18 @@ function get_host_and_port($value, $default_port)
       worker-src 'self' blob:;
       connect-src 'self' https://*.google-analytics.com https://*.analytics.google.com https://*.googletagmanager.com https://*.googleapis.com *.google.com https://*.gstatic.com  data: blob:;
       frame-ancestors 'self';
+  user_challenge:
+    activated: true
```
anvit: Wasn't the user challenge supposed to be disabled by default? Is the plan to leave it on for docker by default?

lib/filter/QubitUserChallengeFilter.class.php
```diff
@@ -0,0 +1,224 @@
...
+     * @param mixed $country
+     * @param mixed $clientAsn
+     * @param mixed $geoIpHelpers
+     * @param mixed $geoIpHelper
+     * @param mixed $userAgent
+     */
+    public function shouldBypassChallenge($country, $clientAsn, $geoIpHelper, $userAgent)
+    {
+        // Exceptions Check: Country code.
+        if (null !== $country) {
```
anvit: Minor nitpick so feel free to ignore this, but reversing these conditions to make the code less deeply nested would improve readability:
```
if (null === $country) {
  return false;
}

$countryExceptions = sfConfig::get('app_user_challenge_country_exceptions', []);
...
```

anvit (approved):
I had a couple small comments but it looks good otherwise!

docker/bootstrap.php
sbreker: I first thought perhaps this might be good to have activated when running the docker env, but am now having second thoughts as well. I will set this to FALSE.

lib/filter/QubitUserChallengeFilter.class.php
sbreker: Hi @anvit - I have added a commit that refactors `shouldBypassChallenge()`.

docker/bootstrap.php
sbreker: I have pushed a commit that sets this to false.

anvit (approved):
Thanks @sbreker ! Looks great!

PR #1969, opened 2025-04-02
Fix JS challenge unit test for PHP 8.3

anvit:
Fix unit tests for js challenge which were returning PHP 8.x related deprecation warnings and errors. 
- Update QubitUserChallenge and QubitUserChallengeFilter to explicitly declare dynamic properties.
- Also update getRemoteAddress in the QubitUserChallengeFilter to return null if request->getPathInfoArray returns null.


sbreker (approved):
Thanks! :+1:

PR #1973, opened 2025-04-07
Address high memory usage in tools:rename-slug

### LINKED ISSUE #1971: Problem: tools:rename-slug is memory inefficient
### Current Behavior

1.  Run `tools:rename-slug` in an AtoM instance with a large number of slugs. My organization's AtoM instance has 3.5 million slugs, for example.

The memory footprint of this command will be very large since all slugs are kept in memory.

### Expected Behavior

All slugs should not be kept in memory. There may be, at minimum, two slugs loaded in memory for each slug rename operation. Namely: the old slug, and the new slug. All other slugs are irrelevant to processing.

### Possible Solution

- In `lib/task/tools/renameSlugTask.class.php`, load only the current slug and the new slug (if either exist) instead of loading all slugs in `getAllSlugs()`.

In other words, instead of doing this:

```sql
SELECT slug FROM slug ORDER BY slug;
```

Essentially, do this instead:

```sql
SELECT slug FROM slug WHERE slug=? OR slug=?;
```

Where ?, ? are the current and new slug, respectively.

I have a working fix with these changes I can open a pull request for.

### Context and Notes

I'm seeing a consistent 4GB of memory being used in my organization's AtoM instance when running this tool with the `--csv` option. It also runs very slowly, since every slug is retrieved from the database for each row of the CSV.

### Version used

AtoM 2.8.2+

_No response_

_No response_

### PHP version

_No response_

### Contact details

_No response_

danloveg:
Addresses #1971 

The function `getAllSlugs()` has been removed in favour of fetching only the current (i.e., old) and new slug from the database.

This change improves performance by reducing the amount of memory this tool uses, especially when there are a large number of slugs in the database. Additionally, the database is leveraged to search for the new slug, instead of using the PHP `in_array` function which should further help improve performance.

### DIFF
--- a/lib/task/tools/renameSlugTask.class.php
+++ b/lib/task/tools/renameSlugTask.class.php
@@ -106,19 +106,31 @@ EOF;
 
     protected function renameSlug($oldSlug, $newSlug)
     {
+        // Search for both the new and old slugs
         $criteria = new Criteria();
-        $criteria->add(QubitSlug::SLUG, $oldSlug);
-        $slug = QubitSlug::getOne($criteria);
+        $c1 = $criteria->getNewCriterion(QubitSlug::SLUG, $oldSlug);
+        $c2 = $criteria->getNewCriterion(QubitSlug::SLUG, $newSlug);
+        $c1->addOr($c2);
+        $criteria->add($c1);
 
-        $existingSlugs = $this->getAllSlugs();
+        $oldSlugObject = null;
+        $newSlugObject = null;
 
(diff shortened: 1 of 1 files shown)


anvit (approved):
Thanks! This looks great! Should be a significant improvement for db's with a large number of slugs!

danloveg:
> Thanks! This looks great! Should be a significant improvement for db's with a large number of slugs!

Thanks for reviewing!

PR #1980, opened 2025-04-10
Copy reference_code to _all field for search.

### LINKED ISSUE #1977: Problem: Identifiers are not searchable
### Current Behavior

1. Search for an identifier/reference code in the general search bar
2. Results are not conclusive

### Expected Behavior

All descriptions containing the identifer/reference code searched should be returned.

### Notes

Using the advanced search and specifying that we want to search by identifier works as expected.

### Version used

AtoM 2.9.0

melaniekung:


anvit (approved):
Looks good!

PR #1985, opened 2025-04-13
Clipboard fixes

mInnes-archives:
PR fixes a bug that breaks clipboard save functionality caused by changes in PHP 8.

In PHP 8, the [count()](https://www.php.net/manual/en/function.count.php) function now throws a fatal error if the array doesn't exist (such as when any of the 3 clipboards are empty), rather than just returning 0 as it did with previous versions. This PR just adds a check for each key and only counts them if they exist.

PR also fixes a bug where the clipboard button type was incorrect for repository pages for the BS5 theme. This was causing these clipboard items to be counted as Archival descriptions and not show up properly in the clipboard.


anvit (approved):
Thanks @mInnes-archives ! And good catch finding and fixing the incorrect type!

Issue #1991, opened 2025-04-16
Problem: Empty form values mask submitted values in advanced search


danloveg:
### Current Behavior

This is an issue with the advanced search for information objects. Under certain conditions it's possible for applied filters to be completely removed from the advanced search, as if the user had never entered those filters. Causing the date range entered to be completely removed can be reproduced reliably.

**Steps to reproduce the behavior**

1. From the main page, go directly to `/index.php/informationobject/browse?showAdvanced=1&topLod=0`
2. Under **Find results with:**, type **Sample** (or any other search term)
3. Click **Search**
4. When the results appear, expand the advanced search dropdown again
5. Enter **2000-01-01** in the start date range
6. Enter **2000-12-31** in the end date range
7. Click **Search**

The date range will completely disappear, i.e., none of the results are filtered by the date, and if you open up the advanced search dropdown again, the Start and End date are not filled.

If you add the date range *yet again* and make a third search, the date range will be present in the advanced search filters.

### Expected Behavior

The start and end date are kept in the advanced search query when the second search is made.

### Possible Solution

This appears to be associated with the fact that certain fields are coming back as empty strings from the form if they aren't filled. When the advanced search form is first submitted, fields like `startDate` and `endDate` are coming back as empty strings. Then when the advanced search form is submitted again, those fields that came back as empty strings are completely removed from the fields submitted by the form.

I'm not entirely sure *why* a form field coming back as empty would mask the re-submission of that same field, but filtering the empty values out when binding values to the form seems to do the trick.

### Context and Notes

This issue is disruptive because it causes users to have to enter their advanced search criteria twice to be able to make a search with their filters.

### Version used

AtoM 2.9 (latest qa/2.x)

### Operating System and version

Docker

### PHP version

PHP 8.3

danloveg:
Closed by #1992

PR #1992, opened 2025-04-16
Filter empty fields from bound form values

danloveg:
Addresses #1991 

Before this change, the values bound to the advanced search form look like this:

- First request, searching "Sample":
  - `sf0="Sample"` is bound to the form
  - `startDate=""` is bound to the form
  - `endDate=""` is bound to the form
- Second request, adding date range "2000-01-01" to "2000-12-31"
  - `sf0="Sample"` is bound to the form
  - `startDate` is not set, hence it is not bound to the form
  - `endDate` is not set, hence it is not bound to the form

This has the effect of completely removing the searched date range, as if the user had never entered the date range. If a *third* request is made with a date range, the startDate and endDate are both set. I'm thinking this is probably because there was no value bound to the form for startDate and endDate in the second search request.

With the changes included in this pull request, the values bound to the form look like this:

- First request, searching "Sample":
  - `sf0="Sample"` is bound to the form
  - `startDate` is not set, hence it is not bound to the form
  - `endDate` is not set, hence it is not bound to the form
- Second request, adding date range "2000-01-01" to "2000-12-31"
  - `sf0="Sample"` is bound to the form
  - `startDate="2000-01-01"` is bound to the form
  - `endDate="2000-12-31"` is bound to the form

This prevents the startDate and endDate from being erased from the results.

### DIFF
--- a/apps/qubit/modules/informationobject/actions/browseAction.class.php
+++ b/apps/qubit/modules/informationobject/actions/browseAction.class.php
@@ -196,9 +196,14 @@ class InformationObjectBrowseAction extends DefaultBrowseAction
         // Add advanced form filter to the query
         $this->search->addAdvancedSearchFilters($this::$NAMES, $this->getParameters, $this->template);
 
+        // Bind filled values to form (skip null and empty fields)
+        $filteredRequestParams = array_filter($request->getRequestParameters(), fn ($item) => !empty($item));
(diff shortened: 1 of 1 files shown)


melaniekung (approved):
Thanks @danloveg for the PR, it looks good to me!

PR #1999, opened 2025-04-21
Update DebugPDO.php for PHP8

misilot:
From the ica-atom-users listserv Alberto Pereira recommended this change to make debug mode work.

https://groups.google.com/g/ica-atom-users/c/jBw4fyD5ZvE/m/L8B1ECsKBwAJ

### DIFF
--- a/vendor/symfony/lib/plugins/sfPropelPlugin/lib/vendor/propel/util/DebugPDO.php
+++ b/vendor/symfony/lib/plugins/sfPropelPlugin/lib/vendor/propel/util/DebugPDO.php
@@ -328,7 +328,7 @@ class DebugPDO extends PropelPDO
 	 * @see        http://php.net/manual/en/pdo.query.php for a description of the possible parameters.
 	 * @return     PDOStatement
 	 */
-	public function query()
+	public function query(string $query, ?int $fetchMode = null, mixed ...$fetchModeArgs): PDOStatement|false
 	{
 		$debug	= $this->getDebugSnapshot();
 		$args	= func_get_args();


anvit:
Hi @misilot! Thank you for submitting this PR. 

This should be fixed in qa/2.x due to changes made in #1942 which essentially removes DebugPDO from being used by AtoM, so we're closing this PR as the changes in PR will have no effect.

The reason behind this is that [propel2 updated DebugPDO](https://github.com/propelorm/Propel2/blob/master/src/Propel/Runtime/Connection/DebugPDO.php) to simply be a wrapper around the PDO rather than being its own class, and it performs the same as PDO with a few additional logging messages.

I also did want to note that at the moment `qubit_dev` is not functional for Bootstrap 5. We plan to address both propel and `qubit_dev` in a future release.

melaniekung:
FYI, if you encounter the following error in AtoM 2.9.1+ :
```
PHP Fatal error:  Declaration of DebugPDO::query() must be compatible with PDO::query(string $query, ?int $fetchMode = null, mixed ...$fetchModeArgs): PDOStatement|false in /usr/share/nginx/atom/vendor/symfony/lib/plugins/sfPropelPlugin/lib/vendor/propel/util/DebugPDO.php on line 331
```
Please update `config/config.php` (listed in `.gitignore`) to replace `DebugPDO` with `PropelPDO`, similar to our changes for `docker/bootstrap.php` and `‎lib/arInstall.class.php` files in our [Update GH unit tests to use Ubuntu 24.04. (#1942)](https://github.com/artefactual/atom/commit/b9a6b8e5d2c298ffb8a32671de0d81326668b4c9).

PR #2000, opened 2025-04-21
Add missing function for CAS (#1983)

### LINKED ISSUE #1983: Problem: Cannot Access User Menu when using CAS Authentication
### Current Behavior

1. Enable CAS
2. Try to go to /user/list
3. Get Error page

```
atom-1  | [11-Apr-2025 14:19:31] WARNING: [pool atom] child 12 said into stderr: "NOTICE: PHP message: Call to undefined method casUser::getProviderConfigValue."
```
I'm wondering if related to this commit? 

### Expected Behavior

Able to view user list and make changes

### Possible Solution

_No response_

### Context and Notes

_No response_

### Version used

_No response_

_No response_

_No response_

### PHP version

_No response_

### Contact details

_No response_

misilot:
Fixes #1983 by returning the default value.

### DIFF
--- a/plugins/arDominionB5Plugin/modules/user/templates/_showActions.mod_ext_auth.php
+++ b/plugins/arDominionB5Plugin/modules/user/templates/_showActions.mod_ext_auth.php
@@ -9,7 +9,7 @@
       <li><?php echo link_to(__('Delete'), [$resource, 'module' => 'user', 'action' => 'delete'], ['class' => 'btn atom-btn-outline-danger']); ?></li>
     <?php } ?>
 
-    <?php if (false === sfContext::getinstance()->user->getProviderConfigValue('auto_create_atom_user', true)) { ?>
+    <?php if ($sf_context->getConfiguration()->isPluginEnabled('arOidcPlugin') && false === sfContext::getinstance()->user->getProviderConfigValue('auto_create_atom_user', true)) { ?>
       <?php if (QubitAcl::check($resource, 'create')) { ?>
         <li><?php echo link_to(__('Add new'), ['module' => 'user', 'action' => 'add'], ['class' => 'btn atom-btn-outline-light']); ?></li>
       <?php } ?>
--- a/plugins/arDominionB5Plugin/modules/user/templates/editSuccess.mod_ext_auth.php
+++ b/plugins/arDominionB5Plugin/modules/user/templates/editSuccess.mod_ext_auth.php
@@ -27,7 +27,7 @@
           <div id="basic-collapse" class="accordion-collapse collapse show" aria-labelledby="basic-heading">
             <div class="accordion-body">
 
-              <?php if (false === sfContext::getinstance()->user->getProviderConfigValue('auto_create_atom_user', true)) { ?>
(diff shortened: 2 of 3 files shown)


anvit:
Hi @misilot! Thanks for creating this PR and bringing this issue to our attention! We spent some time discussing this, and we think this issue needs a slight change address the root cause. We have two different approaches in mind:

1. Like you noted in the issue, this traces its way back to `lib/myUser.class.php`. One potential fix would be to have a method for getting the config value in that class instead of in `casUser` and since `oidcUser` has its own version, it would override the `myUser` method and behave normally. Instead of simply returning the default value, I would have it look up a value from `app.yml` similar to how `oidcUser` does it, and return the default if it is empty. I would also suggest changing the name of the method to `getConfigValue`  instead of `getProviderConfigValue` and updating those instances in the rest of the code base (they are [only present in 9 files](https://github.com/search?q=repo%3Aartefactual%2Fatom%20getProviderConfigValue&type=code), so it's still a small change) since `provider` doesn't make sense outside of the oidc context. This solution is the cleaner and more comprehensive one of the two.

2. Alternatively, updating the 3 instances where we call `getProviderConfigValue` outside of the oidc context to first check if the oidc plugin is in use (something like `$sf_context->getConfiguration()->isPluginEnabled('arOidcPlugin')`). I believe these are in [plugins/arDominionB5Plugin/modules/user/templates/_showActions.mod_ext_auth.php](https://github.com/artefactual/atom/blob/811e7600ba1c1767c99a1a1af08be9643715b8c5/plugins/arDominionB5Plugin/modules/user/templates/_showActions.mod_ext_auth.php#L12), [plugins/arDominionB5Plugin/modules/user/templates/listSuccess.mod_ext_auth.php](https://github.com/artefactual/atom/blob/811e7600ba1c1767c99a1a1af08be9643715b8c5/plugins/arDominionB5Plugin/modules/user/templates/listSuccess.mod_ext_auth.php#L83), and [plugins/arDominionB5Plugin/modules/user/templates/editSuccess.mod_ext_auth.php](https://github.com/artefactual/atom/blob/811e7600ba1c1767c99a1a1af08be9643715b8c5/plugins/arDominionB5Plugin/modules/user/templates/editSuccess.mod_ext_auth.php#L30).

Let us know if you'd have the capacity to update the PR to make this change. If not, we'd be happy to take this over and address this!

misilot:
@anvit I don't have time to look at this for probably another week, so if you have the time, that would be awesome. Thank you!

anvit:
@misilot That is totally fine if it takes you longer! We're in the process of finalizing the `2.9.1` release this week, so no rush since this will be going into the next release.

misilot:
@anvit ok, so you are leaving CAS broken in 2.9.1?

misilot:
@anvit I chose option 2 and wrapped plugin enabled check with each call. I don't have a way to test oidc, but I assume it should still work.

sarah-mason:
Hi @misilot - thank you so much for taking the time to update your code to one of the options suggested. We'll do the oidc tests, thanks for the heads up.

This is a good fix and we will get it into 2.9.1.  Upon further review of the two options discussed, we think that option 1 will be more robust over time -- but we will do this refactoring for a future release. Again, thank you and do let us know if you have any questions.

anvit (approved):
Looks good!

Issue #1901, opened 2024-12-06
Enhancement: Add Accessions CSV export feature


ravengermain:
### Current Behavior

AtoM users are able to import accessions using a CSV file but there is no existing functionality to export an accessions CSV. Accessions are one of the only major entity types without an export option, which can make migrating content difficult. 

### Expected Behavior

Add an Accessions CSV export feature to AtoM. 

### Possible Solution

This could be accomplished by adding a clipboard button to accession record pages so that those with the proper user permissions (e.g. administrators) can add records to the clipboard for export. We should also include an "export all" option on the "Browse accessions" page. 

### Version used

AtoM 2.8.2

ravengermain:
When this functionality is added, it should also include the ability to export search results from the accessions module page. This would be consistent with the ability to export search results for archival descriptions in AtoM as described here: https://accesstomemory.org/en/docs/2.9/user-manual/import-export/csv-export/#export-search-and-browse-results-in-csv-format.

danloveg:
@melaniekung I see we looked at this issue at similar times - I was just working on this functionality in a branch in my repo. I'll make a PR for it, maybe we can combine the work we've both done on it.

PR #2003, opened 2025-04-21
Add CLI accession export task

### LINKED ISSUE #1862: Feature: Ability to export accession records
AtoM doesn't have an ability to export accession records as CSVs.

danloveg:
Partially addresses #1862
Partially addresses #1901

This pull request adds an accession export task that can be run from the AtoM backend. It's triggered like this:

```shell
php symfony csv:accession-export /path/to/output.csv
```

In an effort to keep the size of this pull request manageable, this PR *does not* include a way to export accessions from the web front end. This is why I've mentioned this PR only partially addresses the issues I've mentioned above.

In addition to the CSV export feature, I'm suggesting changing the `creators` column in the accession CSV template to `eventActors`. The reason I'm making this suggestion is that events are being duplicated in the database because of the existence of the `creators` column. See commit 1421a78d546eb6ea7654c0c57ad2e8dba1c7e35a for more details on why this is happening.

Other changes that will be needed not included in this PR:

- Changing the name of the column would require an update to this page of the AtoM Wiki: https://wiki.accesstomemory.org/Resources/CSV_templates#Accessions
- I can write up a new section of the documentation in the atom-docs for this new exporter if this PR is accepted. It would probably go here: https://www.accesstomemory.org/en/docs/2.9/admin-manual/maintenance/cli-import-export/#export-csv-files-from-the-command-line


yenaing-oo:
I tested this task and noticed that the exported CSVs include "NULL" values in certain fields, which are then also displayed  as "NULL" for those fields on the front-end slug page for the corresponding accessions.

I'm working on fixing this bug. Just wanted to flag it here in case others encounter the same behaviour.

lib/task/export/csvExportAccessionsTask.class.php
```diff
@@ -0,0 +1,95 @@
...
+            $accessionRecord = QubitAccession::getById($row['id']);
+            $this->context->getUser()->setCulture($row['culture']);
+
+            $writer->exportResource($accessionRecord);
+
+            $this->indicateProgress($options['items-until-update']);
+            ++$itemsExported;
+        }
+
+        $this->log('');
```
anvit, line 72: Is this empty log a typo or leftover from debugging?

lib/flatfile/csvAccessionExport.class.php
```diff
@@ -0,0 +1,232 @@
...
+    }
+
+    /*
+     * Make modifications to the row before it is written to the output. These operations can't be
+     * handled with a simple mapping, and must be populated manually.
+     *
+     * @return void
+     */
+    protected function modifyRowBeforeExport()
+    {
```
anvit, line 74: I noticed that this doesn't call the `parent::modifyRowBeforeExport`, and was wondering if that was on purpose?

anvit (commented):
Hi @danloveg , just finished taking a look at this PR, and it looks good! Just have a couple of small flags that I wanted to clarify before approving and merging.

lib/flatfile/csvAccessionExport.class.php
danloveg: Good catch, I did indeed forget to add that. I've added the call to the parent class in [042810f](https://github.com/artefactual/atom/pull/2003/commits/042810f7ed2ad85eac1f5e99b7ea4fdb3d48f9b3)

lib/task/export/csvExportAccessionsTask.class.php
danloveg: I've added that to add a new line after all of the `.`s in the output. Without this empty log call, the output looks like this:

no_newline

Adding a `\n` to the message itself also doesn't work, because I think the log messages get trimmed/stripped before they're written to the output. I tried this and the output looks the same as in my screenshot:

```php
$this->logSection('csv', "\nExport complete!");
```

EDIT: Something similar is also done in other CSV exporters, like here: https://github.com/artefactual/atom/blob/d2f2848cbddb82174ab90deb6cea3151f523df90/lib/task/export/exportAuthorityRecordsTask.class.php#L65

anvit:
Thanks @danloveg and @yenaing-oo !

anvit (approved):

PR #2006, opened 2025-04-23
Hide non-visible elements in CSV exports

### LINKED ISSUE #2005: Problem: CSV exports not hiding visible elements
### Current Behavior

1. Go to Admin > Visible Elements
2. Set some elements to hidden
3. Export CSV as public user or without checking the 'Include non-visible elements option'
4. Note that non-visible elements are included in the CSV

### Expected Behavior

Non-visible elements should only be included in CSVs exported by authenticated users using the 'Include non-visible elements' option.

### Version Used
2.9.1

melaniekung:


anvit (approved):
Thanks @melaniekung !

PR #2008, opened 2025-04-24
BS5 uses alert-danger instead of alert-error.

### LINKED ISSUE #2007: Missing .alert-error class in arDominionBS5Theme
Replace .alert-error with .alert-danger

melaniekung:


anvit (approved):
Looks good! This change would be a breaking change for BS2, so let's remember to merge this along with the BS2 removal change.

PR #2016, opened 2025-04-30
Clarify Limit IP settings for authenticated users.

### LINKED ISSUE #2015: Clarify Limit IP settings affects all authenticated users

melaniekung:


anvit (changes requested):
I would recommend changing only the actual source string in `securityAction.class.php` as a part of this fix. The XMLs should be updated through the script automatically when we run through our translations pull process.

anvit (approved):

PR #2021, opened 2025-05-01
Fix broken string from translating html entities

### LINKED ISSUE #2020: Broken string in messages.xml from uploadLimit
`&nbsp;` gets translated into `&amp;nbsp` when `&nbsp;` is included in `echo __()` functions

melaniekung:


anvit (approved):
:+1:

PR #2022, opened 2025-05-04
Add clipboard retention in save success message

mInnes-archives:
PR adds additional feedback to clipboard save success message to let users know how many days until clipboards are eligible for deletion. This reflects the clipboard max age setting which is typically only viewable by users with admin privileges.

I've also made the success message sensative to the correct grammatical number for 'item/items' saved, as well as for 'day/days' the clipboard will be retained. I think I've done this in a way that should work with Weblate.

### DIFF
--- a/apps/qubit/modules/clipboard/actions/saveAction.class.php
+++ b/apps/qubit/modules/clipboard/actions/saveAction.class.php
@@ -49,20 +49,9 @@ class ClipboardSaveAction extends sfAction
 
             $loadUrl = $this->context->routing->generate(null, ['module' => 'clipboard', 'action' => 'load']);
 
-            if (1 == $itemsCount) {
-                $itemsUnit = $this->context->i18n->__('item');
-            } else {
-                $itemsUnit = $this->context->i18n->__('items');
-            }
-
             $retention = sfconfig::get('app_clipboard_save_max_age');
-            if (1 == $retention) {
-                $retentionUnit = $this->context->i18n->__('day');
-            } else {
-                $retentionUnit = $this->context->i18n->__('days');
-            }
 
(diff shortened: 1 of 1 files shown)


sarah-mason:
Thanks for this contributions @mInnes-archives! We'll discuss it at our next team meeting and get back to you soon.

anvit:
Hi @mInnes-archives! Just finished taking a look at this and having a date in the clipboard retention message is definitely valuable and makes sense!

I left a comment requesting a small change to make this easier to deal with for translators.

apps/qubit/modules/clipboard/actions/saveAction.class.php
```diff
@@ -48,8 +48,21 @@ public function execute($request)
             }
 
             $loadUrl = $this->context->routing->generate(null, ['module' => 'clipboard', 'action' => 'load']);
-            $message = $this->context->i18n->__('Clipboard saved with %1% items. Clipboard ID is <b>%2%</b>. Please write this number down. When you want to reload this clipboard in the future, open the Clipboard menu, select <a class="alert-link" href="%3%">Load clipboard</a>, and enter this number in the Clipboard ID field.', ['%1%' => $itemsCount, '%2%' => $password, '%3%' => $loadUrl]);
 
+            if (1 == $itemsCount) {
```
anvit: We generally try to avoid having snippets inserted into other strings since sometimes that makes it tricky to translate into languages that are structured differently as this forces specific words into specific places in a sentence. I would suggest simplifying this and maybe sticking with something like `item(s)` and `day(s)` to make it easier to translate.

anvit (changes requested):

anvit (approved):
Looks great! Thanks @mInnes-archives !

PR #1924, opened 2025-02-11
Fix digital object file check failure

miceno:
Solves #1923.

PR #2025, opened 2025-05-08
Check if 'pathToDigialObject' exists. (#2024)

### LINKED ISSUE #2024: Problem: CSV Validation fails for digitalObjectPath
### Current Behavior

1. Validate a CSV with the digitalObjectPath filled out
2. Validation always fails via GUI and will fail via CLI if `path-to-digital-objects` flag is not used

### Expected Behavior

CSV validation should not fail unless the path is truly incorrect

### Version used

AtoM 2.9.1

melaniekung:


miceno:
I feel it is related to #1924

anvit (approved):
Makes sense! Thanks @melaniekung

PR #2047, opened 2025-06-05
Add test-debug composer script

### LINKED ISSUE #2046: Add ability to debug PHPUnit tests
It would be helpful if there were any easy, documented way to debug PHPUnit tests to be able to see what's going on inside unit tests. Personally, this would be very helpful to be able to understand how the unit tests work so that I can write more tests.

Since we have `xdebug` installed in the Docker container now (see #1993), I've found a way to debug PHPUnit tests as they're running in Docker.

These are the steps I used to debug tests in VS Code:

1. Start the application in Docker with `compose.dev.yml`
2. Start listening for xdebug in VS Code.
3. Set breakpoints in tests.
4. Run this command (this assumes you've already run `export COMPOSE_FILE="$PWD/docker/docker-compose.dev.yml"`)

  ```shell
  docker compose exec atom bash -c "XDEBUG_SESSION=1 composer test"
  ```

Basically, there just needs to be an `XDEBUG_SESSION` environment variable set when the tests run. I can wrap this functionality inside a nicer composer script that doesn't require you to do `bash -c`.

I've found that I can't debug the tests locally because I don't have all the proper PHP packages and the proper PHP version installed locally. But theoretically it would be possible to debug them locally, too.

danloveg:
This PR adds a new composer command to run phpunit with xdebug enabled. This makes it so that you can debug tests in your IDE with xdebug.

Closes #2046

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -31,6 +31,10 @@
     },
     "scripts": {
         "test": "phpunit",
+        "test-debug": [
+            "@putenv XDEBUG_SESSION=1",
+            "phpunit"
+        ],
         "test-cov": "@php -dpcov.enabled=1 -dpcov.directory=. ./vendor/composer/bin/phpunit --coverage-html=.coverage/html",
         "test-cov-strict": "@php -dpcov.enabled=1 -dpcov.directory=. ./vendor/composer/bin/phpunit --coverage-html=.coverage/html --strict-coverage",
         "php-cs-fixer": "php-cs-fixer"


danloveg:
I'm currently trying to figure out how to get xdebug to stop complaining in tests that purposely throw exceptions. The debugger stops any time an exception is thrown in a test, even if the test is expecting an exception with `$this->expectException(MyException::class)`.

EDIT: Figured it out. It's because I had the "Errors" breakpoint option set in VS Code. With that option unset, the debugger isn't stopping at expected exceptions any more.

Screenshot 2025-06-05 102601

anvit (approved):
Looks good @danloveg :+1:

PR #2051, opened 2025-06-09
Fix i18n string in the browse page (#2050)

### LINKED ISSUE #2050: Problem: Browse page missing information in other languages
### Current Behavior

1. Change language to ex Spanish
2. Go to Archival Description browse page
3. Note in the 'Narrow your results by' sidebar that sections such as Creators, Name, and Archival Instituions are only displaying the number of occurrence but not the item (creator, name, and institution)

### Expected Behavior

When changing languages, all information is expected to be shown.

### Version used

AtoM 2.9.1

en/es

anvit:
Cast inputs to strlen as strings in BaseActor and BaseProperty instead of passing empty string if they're null.

(strlen expects strings, and will output a deprecation warning if null is passed to it since PHP 8.2)

### DIFF
--- a/lib/model/om/BaseActor.php
+++ b/lib/model/om/BaseActor.php
@@ -242,7 +242,7 @@ abstract class BaseActor extends QubitObject implements ArrayAccess
 
     try
     {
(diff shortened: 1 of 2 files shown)


melaniekung (approved):
thanks @anvit !

PR #2054, opened 2025-06-13
Execute JS challenge logic before Symfony startup

sbreker:
Extract the JS challenge from the Symfony filterchain, action, and template into
minimal PHP entry points. These new files run before Symfony’s autoloader is
invoked. Scripts are CSP‐compliant and use localStorage for state. Bypassing
the full framework bootstrap for bot and headless‐browser endpoints eliminates
unnecessary overhead and significantly improves request performance.


docker/bootstrap.php
```diff
@@ -292,7 +288,6 @@ class: sfMemcacheCache
 \n
 # Development-specific configuration
 expose_php = on
-opcache.validate_timestamps = on
```
anvit: Is this change related to this fix? I couldn't figure out why this was removed from the docker config.

anvit (commented):
I just have a question about a small change, everything else looks good!

docker/bootstrap.php
sbreker: Good catch! Accidental leftover from trying to squeeze performance out of PHP. I will restore this line.

sbreker: Fixed in most recent commit

anvit (approved):
Thanks @sbreker !

PR #2060, opened 2025-06-19
Exclude hidden elements from Finding Aids. (#2030)

### LINKED ISSUE #2030: Problem: Finding aids include information that is deselected in Visible Elements menu
### Current Behavior

1. Set the RAD description template as default
2. Deselect the Control Area from the Visible Elements menu
3. Add information to the Control Area of a published description and confirm that it is not visible from an unauthenticated view.
4. Generate a finding aid for that description
5. Confirm that under the "Other Notes" section all of the info added to the Control Area is present 
Note: the Control Area source is also pulled into the "Note [Sources]" section despite this being deselected 

### Expected Behavior

The Visible Elements feature is designed to allow administrators to hide certain information from public view. That hidden/sensitive information should therefore not be pulled into finding aids generated by the public. The ideal behaviour would be that if a certain section is deselected in the Visible Elements menu, that section will then be skipped when a Finding Aid is generated for a description. 

### Possible Solution

_No response_

### Context and Notes

_No response_

### Version used

_No response_

_No response_

_No response_

### PHP version

_No response_

### Contact details

_No response_

melaniekung:


lib/QubitFindingAidGenerator.class.php
```diff
@@ -268,14 +268,9 @@ public function generate()
...
-        // enabled
-        $eadFilePath = $this->getEadCacheFilePath();
-
-        // Generate an EAD file, if there is no cached EAD file
-        if (empty($eadFilePath)) {
-            $eadFilePath = $this->generateEadFile();
-        }
+        // Generate an EAD file
+        // Don't use cached file to exclude hidden elements
+        $eadFilePath = $this->generateEadFile();
```
anvit, line 272: If we're not using the getEadCacheFilePath function any more, we should remove it entirely. Also, if we do that, the comment about not using the cached file wouldn't be useful in this file; I think commit message itself would provide enough context that the comment here wouldn't be needed.

lib/QubitFindingAidGenerator.class.php
```diff
@@ -392,6 +387,7 @@ public function generateEadFile(): string
 
         $options = [
             'public' => ('public' === $this->getAuthLevel()),
+            'findingAid' => true,
```
anvit: Maybe calling it `findingAidVisibility` or something along those lines would make it more clear what this option is really for.

anvit (approved):
Added a couple of small flags, but the code looks good otherwise! Thanks!

PR #2062, opened 2025-06-20
Fix Lang & Script notes in ISAD. (#2061)

### LINKED ISSUE #2061: Problem: Error with adding Language and Script notes field in ISAD & DACS
### Current Behavior

1. Set default description template to ISAD or DACS (which also uses the sfIsadPlugin class)
2. Add/edit a description with the `Language and script note` filled out
3. Notice that you get a 500 error

### Expected Behavior

There should be no error when adding values to fields

### Context and Notes

"NOTICE: PHP message: PHP Fatal error:  Uncaught Error: Attempt to assign property "content" on null in /atom/src/plugins/sfIsadPlugin/lib/sfIsadPlugin.class.php:94"

### Version used

2.9

melaniekung:


anvit (approved):
This makes sense. Thanks @melaniekung

PR #2064, opened 2025-06-23
Add 'show' class to accordions with errors. (#2063)

### LINKED ISSUE #2063: AtoM Enhancement: Expand sections with errors
### Current Behavior

1. Create a new accession record with an existing Accession Number
2. Click Create
3. Notice that "nothing" happens
4. Expand the Basic Information section
5. Notice that you get an error for the Accession Number

### Expected Behavior

A section should be expanded by default if there's an error in the section.

### Context and Notes

Should be tested on other records.

### Version used

2.9.1

melaniekung:


anvit (approved):
This makes sense. Thanks @melaniekung !

PR #2069, opened 2025-06-24
Fix implode call for arGenerateReportJob (#2068)

### LINKED ISSUE #2068: Problem: Error on attempting to generate box label report
### Current Behavior

1. Go to a description with physical storage linked.
2. Click `Reports` and attempt to generate a box label report (CSV)
3. Notice this error on the jobs page —
```
Exception: implode(): Argument #2 ($array) must be of type ?array, string given

File: /atom/src/lib/job/arGenerateReportJob.class.php

Line: 240
``` 

### Expected Behavior

Report generation should be successful.

### Possible Solution

This is very likely a bug in `arGenerateReportJob.class.php` introduced with PHP 8.x.

### Version used

AtoM 2.9.1

anvit:
Update implode function call in arGenerateReportJob to use string parameter first and array second since the alternative function signature that allows the string to be second has been discontinued in PHP 8.0

### DIFF
--- a/lib/job/arGenerateReportJob.class.php
+++ b/lib/job/arGenerateReportJob.class.php
@@ -237,7 +237,7 @@ class arGenerateReportJob extends arBaseJob
                     'referenceCode' => $informationObject->referenceCode,
                     'physicalObjectName' => $item->__toString(),
                     'title' => $informationObject->__toString(),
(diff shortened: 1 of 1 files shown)


melaniekung (approved):
thanks, @anvit

PR #2080, opened 2025-07-09
Enable translations for clipboard 'count'. (#2078)

### LINKED ISSUE #2078: Problem: Clipboard 'count' is untranslatable when hardcoded in js
### Current Behavior

Clipboard menu has counts where `count` is hardcoded in `plugins/arDominionB5Plugin/js/clipboard.js` and therefore not passed to weblate for translations

### Expected Behavior

All strings should be translatable via weblate.

melaniekung:


anvit (approved):
Looks good! Thanks for the updates @melaniekung !

anvit (approved):
Thanks @melaniekung !

anvit (approved):
Thanks for the updates @melaniekung !

PR #2081, opened 2025-07-09
Add Web analytics settings page for GA. (#2079)

### LINKED ISSUE #2079: Add Google analytics to settings for users to update.
Add setting to allow users to update Google Analytics code

melaniekung:


apps/qubit/modules/settings/actions/analyticsAction.class.php
```diff
@@ -0,0 +1,98 @@
...
+                        '  google_analytics_api_key: '.$gaTrackingId,
+                        $contents
+                    );
+
+                    if (null === $contents) {
+                        throw new RuntimeException('Regex error while updating app.yml');
+                    }
+
+                    // Write the updated contents back
+                    $result = file_put_contents($configAppYml, $contents);
```
anvit: @sbreker wanted to get your opinion about this if you think this is a good approach or if there are any implications for writing into the app.yml file. And do we need to sanitize inputs or apply some sort of regex filter before we accept these?

melaniekung: talked to @sbreker about this today and we've decided that this should be a variable added to the database but at least for now, we will check `app.yml` for this value and override it with any values in the database.

sbreker: As we discussed earlier I would avoid writing from the UI into this server side file because:

- The app.yml file may be written by Ansible scripts. Changes to this file will be lost if the site is redeployed or upgraded.
- it looks like this could be prone to injection as the replacement yml could have more than just the `google_analytics_api_key` node in it.
- there may be permissions issues with writing to this file/folder.

If we need to update this setting via the WebUI, I would consider migrating the setting from the app.yml file to the settings table in the database and adding a migration.

apps/qubit/modules/default/templates/_googleAnalytics.php
```diff
@@ -1,4 +1,4 @@
-<?php $gaKey = sfConfig::get('app_google_analytics_api_key', ''); ?>
+<?php $gaKey = '' != QubitSetting::getByName('google_analytics') ? QubitSetting::getByName('google_analytics') : sfConfig::get('app_google_analytics_api_key', ''); ?>
```
sbreker: Avoid retrieving the setting twice:
```suggestion
<?php
$gaKey = QubitSetting::getByName('google_analytics');
if (empty($gaKey)) {
    $gaKey = sfConfig::get('app_google_analytics_api_key', '');
}
?>
```

apps/qubit/modules/settings/actions/analyticsAction.class.php
```diff
@@ -0,0 +1,74 @@
...
+ * You should have received a copy of the GNU General Public License
+ * along with Access to Memory (AtoM).  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+/**
+ * Security.
+ *
+ * @author     Peter Van Garderen <peter@artefactual.com>
+ * @author     Jack Bates <jack@nottheoilrig.com>
+ * @author     David Juhasz <david@artefactual.com>
```
sbreker: Update comment

sbreker (changes requested):
A couple nitpicks but otherwise looks great! Does anything display on the settings page if the key is not set in the DB but instead set in the app.yml file?

apps/qubit/modules/settings/actions/analyticsAction.class.php
melaniekung: good catch, thank you!

sbreker (approved):
Looks good @melaniekung! :+1:

PR #2082, opened 2025-07-09
Enable mouse scroll nav for carousel. (#2076)

### LINKED ISSUE #2076: Enhancement: mouse scroll navigation in BS5 carousel
The image carousel in Bootstrap 5 does not support using a mouse wheel to scroll through the images. 
Being able to do this would be an improvement in usability, and the lack of the feature is also a departure from the behaviour present in the BS2 theme, since that used a different carousel internally.

melaniekung:


plugins/arDominionB5Plugin/js/imageflow.js
```diff
@@ -48,6 +48,35 @@
...
+
+            // Decide how many slides to move based on scroll strength
+            let jump;
+            if (absDelta < 50) {
+              jump = 1;
+            } else if (absDelta < 100) {
+              jump = 2;
+            } else {
+              jump = 3;
+            }
```
anvit: This could be simplified like this:

```
let sensitivity = 50;
let maxSlidesScroll = 3;
// 
let jump = Math.min(Math.ceil((absDelta / sensitivity)), maxSlidesScroll);
```

Also I think declaring the sensitivity and number of slides scrolled as variables like this makes it easier to make sense of and change in the future.

plugins/arDominionB5Plugin/js/imageflow.js
```diff
@@ -48,6 +48,35 @@
...
+          // Mouse scroll navigation
+          $("#collapse-carousel").on("wheel", function (e) {
+            e.preventDefault();
+
+            const slider = $("#atom-slider-images");
+            const titleSlider = $("#atom-slider-title");
+            let currentSlide = slider.slick("slickCurrentSlide");
+            let delta = e.originalEvent.deltaY;
+
+            // Sensitivity factor
```
anvit: This is more of scroll amount rather than a sensitivity factor.

anvit (commented):
Couple of small suggestions but looks good otherwise!

anvit (approved):
Thanks @melaniekung !

Issue #1987, opened 2025-04-15
Problem: connection to physical object prevents display of archival description


Hans-Arno:
### Current Behavior

**Steps to reproduce the behavior**
1. Go to an existing archival description
2. Link to a physical storage
3. Try to show that description (any archival description with physical storage): "The server returned a 500 Internal Server Error. ", error log: `2025/04/15 11:21:37 [error] 1224#1224: *243 FastCGI sent in stderr: "PHP message: Unknown record property "levelOfDescription" on "QubitPhysicalObject"" while reading response header from upstream, client: 192.168.178.80, server: _, request: "GET /postkartenalbum HTTP/1.1", upstream: "fastcgi://unix:/run/php-fpm.atom.sock:", host: "192.168.178.63", referrer: "http://192.168.178.63/bilder-sammlung"
`

### Expected Behavior

Archival descriptions with assigned physical Storage shall display, no error 500

The export of all physical storage from menu Manage/Physical Storage button "Export storage report" has to start a job running without error.

### Possible Solution

There is a change from 2.9.0 (OK) to 2.9.1 (only ADs without storage are working)

### Context and Notes

As version 2.9.0 is still buggy, I check coming 2.9.1 as an production candidate based on data imported into version 2.8.2.

### Version used

AtoM 2.9.1-193 from Github qa/2.x

### Operating System and version

Ubuntu 24.4

### Default installation culture

de

### PHP version

PHP 8.3

### Contact details

hans-arno@mielsch.de

Hans-Arno:
@melaniekung : sorry, but it is now working for **authenticated** users only, if the user is not authenticated he still gets an error 500 instead of a display.

melaniekung:
Hi @Hans-Arno , thanks pointing this out, and we appreciate your thorough testing! We've opened a new issue for this issue: https://github.com/artefactual/atom/issues/1996

PR #1900, opened 2024-12-03
Related material improvements

mInnes-archives:
PR adds some small improvements to how related materials are listed in descriptions and in relevant edit forms:

1. If draft descriptions are linked, these will now only be visible to authenticated users. [bug fix]
2. The title of linked descriptions now include the level of description and reference code for descriptions with such values, as well as the draft status when relevant. Both BS2 & BS5 themes are covered.

With regards to the second point, given that titles can be fairly generic in an archival context, just having the title alone ends up being really unclear and unhelpful at times. At BC Archives we have been avoiding using the 'Related materials' field because of how it displays. We've just been using the 'Associated materials' field instead, which technically isn't what it's intended for. Hopefully this can help get us using the appropriate field in the future and benefit from having reciprocal links. I imagine there are others who might be avoiding it or finding it not as clear as it could be for users for similar reasons.

PR #2085, opened 2025-07-10
Restore new related material formatting and fix 500 error when linking containers

mInnes-archives:
PR reverts all previous fix attempts to the related material descriptions field back to what was introduced in #1900 and properly addresses the 500 error caused when storage containers were added to descriptions with related material links.

Basically since containers are also relations, the relations here should have just included information objects. #1900 didn't do that, so it crashed (#1987). The existing type check just needed to be moved earlier. The previous fix attempts using `property_exists` and `method_exists` were always returning `false`, meaning the error was being removed by way of never allowing the new elements added in #1900 to be included in the link. All intended functionality is now restored (as far as I can tell).

### DIFF
--- a/apps/qubit/modules/informationobject/templates/_relatedMaterialDescriptions.php
+++ b/apps/qubit/modules/informationobject/templates/_relatedMaterialDescriptions.php
@@ -5,55 +5,54 @@
   <?php } else { ?>
     <h3><?php echo __('Related descriptions'); ?></h3>
   <?php } ?>
-
   <div>
     <ul>
       <?php foreach ($resource->relationsRelatedBysubjectId as $item) { ?>
-        <?php if ($sf_user->isAuthenticated() || QubitTerm::PUBLICATION_STATUS_PUBLISHED_ID == $item->object->getPublicationStatus()->statusId) { ?>
-          <?php $itemTitle = $item->object->__toString(); ?>
+        <?php if (isset($item->type) && QubitTerm::RELATED_MATERIAL_DESCRIPTIONS_ID == $item->type->id) { ?>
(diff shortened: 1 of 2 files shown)


mInnes-archives:
I'd just like to add that this also addresses the following buggy behaviour currently being experienced with this field:

- related materials are not being displayed on both subject and object description pages
- this field is only visible to authenticated users, regardless of the publication status of the description

mInnes-archives:
@anvit The bugs reported in this forum thread are related to this issue: https://groups.google.com/g/ica-atom-users/c/mxILsWx_BxU

The entities reporting missing properties are all non-information object relations linked to descriptions in various ways. Johan noticed that the 500 errors can still be reproduced with non-B5 themes. I only addressed the B5 theme here, but will add a commit that covers the others.

mInnes-archives:
067b1ff tested and confirmed to fix the 500 errors Johan was able to reproduce with non-B5 themes.

anvit:
> @anvit The bugs reported in this forum thread are related to this issue: https://groups.google.com/g/ica-atom-users/c/mxILsWx_BxU
> 
> The entities reporting missing properties are all non-information object relations linked to descriptions in various ways. Johan noticed that the 500 errors can still be reproduced with non-B5 themes. I only addressed the B5 theme here, but will add a commit that covers the others.

@mInnes-archives ah I see! No wonder I wasn't able to reproduce this, I hadn't been testing with those. Thanks so much for the fixes! Will take a look tomorrow. 

I see that you've pushed the fixes for the BS2 version as well, but just wanted to give you a heads up that 2.10 will be removing BS2 support entirely, so while those changes might be helpful for anyone who wants to cherry-pick them from the qa branch, they would not actually ship in the new release and will be removed when we merge #2066

anvit (approved):
Thanks @mInnes-archives !

PR #2086, opened 2025-07-11
Fix CSV exports with hidden elements. (#2041)

### LINKED ISSUE #2041: Problem: Wrong CSV exported columns when invisible elements
### Current Behavior

1. Make sure all elements are visible on "Admin > Visible elements > ISAD template elements" and "Admin > Visible elements > ISAD template area headings"
2. Create a Item (item1) using ISAD template: fill all the elements and set the title to "item1". Fill a value on the field `Archival history`.
2. Create a Item (item2) using ISAD template: fill all the elements and set the title to "item2". Fill a value on the field `Archival history`.
3. Run a CSV export from the command line using:
```
php -d memory_limit=-1 symfony csv:export --public all-elements-output.csv
```
4. Visit "Admin > Visible elements > ISAD template elements"
5. Make field `Archival history` invisible.
6. Repeat the export from the command line, to a new output file:
```
php -d memory_limit=-1 symfony csv:export --public invisible-elements-output.csv
```
The `invisible-elements-output.csv` file is not properly formed: although the value of the `archivalHistory` field is not included in each row, there are missing some values like those of field `culture` at the end of every row. Use the `all-elements-output.csv` to identify the values.

### Expected Behavior

1. Make sure all elements are visible on "Admin > Visible elements > ISAD template elements" and "Admin > Visible elements > ISAD template area headings"
2. Create a Item (item1) using ISAD template: fill all the elements and set the title to "item1". Fill a value on the field `Archival history`.
2. Create a Item (item2) using ISAD template: fill all the elements and set the title to "item2". Fill a value on the field `Archival history`.
3. Run a CSV export from the command line using:
```
php -d memory_limit=-1 symfony csv:export --public all-elements-output.csv
```
4. Visit "Admin > Visible elements > ISAD template elements"
5. Make field `Archival history` invisible.
6. Repeat the export from the command line, to a new output file:
```
php -d memory_limit=-1 symfony csv:export --public invisible-elements-output.csv
```
9. review `invisible-elements-output.csv` file and check that headers and fields are properly aligned, that is, for each header element, the value of the row is the value of the field.

### Possible Solution

The root cause is the way the invisible fields on the `lib/flatfile/QubitFlatfileExport.class.php` file are reviewed.

A quick fix is making sure that the same process that is held for the first record is done any other record, by commenting the line at https://github.com/artefactual/atom/blob/698d595e0382c6d067c224e89ed85d143cb4e410/lib/flatfile/QubitFlatfileExport.class.php#L246

### Context and Notes

[invisible-elements-output.csv](https://github.com/user-attachments/files/20523410/invisible-elements-output.csv)
[all-elements-output.csv](https://github.com/user-attachments/files/20523411/all-elements-output.csv)

### Version used

AtoM 2.8.2 and 2.9.2

_No response_

_No response_

### PHP version

PHP 7.4, PHP 8.3

### Contact details

_No response_

melaniekung:
Use column indexes to ensure hidden elements are removed from child exports and aligned correctly.

### DIFF
--- a/lib/flatfile/QubitFlatfileExport.class.php
+++ b/lib/flatfile/QubitFlatfileExport.class.php
@@ -25,6 +25,7 @@
 class QubitFlatfileExport
 {
     public $columnNames = [];       // ordered header column names
+    public $maxColumnIndexes = 0;     // total column indexes used for removing hidden elements
     public $standardColumns = [];       // flatfile columns that are object properties
     public $columnMap = [];       // flatfile columns that map to object properties
     public $propertyMap = [];       // flatfile columns that map to Qubit properties
@@ -42,6 +43,7 @@ class QubitFlatfileExport
     protected $separatorChar = '|';          // character to use when imploding arrays to a single value
     protected $params;
     protected $nonVisibleElementsIncluded;
+    protected $nonVisibleElementsIndexes = [];
     protected $path;
     protected $standard;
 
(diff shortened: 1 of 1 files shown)


anvit (approved):
Looks good for now, but @melaniekung and I talked about visibility being a pain because we don't use associative arrays for indexing, which should be something we refactor in the future.

PR #2092, opened 2025-07-14
Reduce year input to 4 digits for fields with datepickers

mInnes-archives:
PR reduces the number of possible digits for the year to 4 when inputing start and end dates in advanced search for information objects.

Some browsers like Chrome/Edge allow 6 digits, which means that the input doesn't naturally move from the YYYY segment to the MM segment when using these browsers. For reference, Firefox just uses 4 digits.


melaniekung (approved):
Thanks @mInnes-archives , this looks good to me!

mInnes-archives:
Apologies for submitting additional commits after review, but these address all the same things, just in different modules.

PR #2096, opened 2025-07-14
Exclude accessionNumber from public CSV exports.

### LINKED ISSUE #2090: Problem: Public CSV exports should exclude accessionNumber field.
### Current Behavior

1. Create an Archival description linked to an Accession.
2. Export CSV for the description as an authenticated user.
3. Note that accessionNumber is exported

### Expected Behavior

Public exports should exclude the accessionNumber field and it should only be included in authenticated exports.

### Version used

AtoM 2.9.2

melaniekung:
This PR uses the logic from [Fix CSV exports with hidden elements. (#2086)](https://github.com/artefactual/atom/pull/2086).

### DIFF
--- a/lib/flatfile/QubitFlatfileExport.class.php
+++ b/lib/flatfile/QubitFlatfileExport.class.php
@@ -281,6 +281,18 @@ class QubitFlatfileExport
             }
         }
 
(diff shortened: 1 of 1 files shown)


anvit (approved):
:+1:

PR #2102, opened 2025-07-21
Enhance accessibility of PDF finding aids

timtomch:
This addresses a number of accessibility issues with PDF finding aids generated from EAD through XSLT/FO. Some of the identified issues (e.g. nonfunctional hyperlinks, runoff text in table cells) also enhance the general usability of the generated PDFs.

This work was developed on behalf of the University of Toronto in April 2024. After testing on a development instance, it has been [deployed to production](https://discoverarchives.library.utoronto.ca/) since February 2025 (AtoM stable/2.8.x). As these changes only impact XSLT transforms, it should be compatible with v. 2.9.x and later.

## Summary of changes
* Apache FOP is called with the `-a` flag to turn on PDF accessibility features
* AtoM's three-letter language codes are mapped to two-letter codes (ISO639/1) as expected by Acrobat for accurate language recognition. This required an update to the languages list in `helper-functions.xsl`
* Added XMP metadata to PDF output
* Added hierarchical title tags
* Fixed URLs to make them actionable by screen readers
* Tagged repeated elements (eg page headers, logos) as artifacts so they can be skipped by screen readers
* Fixed table hierarchies
* Fixed some general display issues (e.g. text overflowing table cells) 

This is my first PR on this project, so apologies if I missed a step!

@kellibee123 is my contact at UofT and can provide more details on the testing and remediation process.


sarah-mason:
Hi @timtomch - thank you for your contribution and for already signing the contributor agreements! I've run the tests, which have all passed. Next step will be that the AtoM Maintainers will take a look at it soon and we'll be in touch soon. But if you have any follow up questions or requests, please do ping us.

lib/task/pdf/ead-pdf-inventory-summary.xsl
```diff
@@ -259,20 +267,20 @@
...
+                <fo:table-cell>
+                    <fo:block>Title</fo:block>
+                </fo:table-cell>
+                <fo:table-cell>
+                    <fo:block>Dates</fo:block>
+                </fo:table-cell>
+                <fo:table-cell>
+                    <fo:block>Physical description</fo:block>
+                </fo:table-cell>
+            </fo:table-row>
```
anvit: There seems to be an extra indent on this block.

lib/task/pdf/helper-functions.xsl
```diff
@@ -280,14 +281,507 @@
...
+            <xsl:otherwise><xsl:value-of select="$node/@type"/></xsl:otherwise>
         </xsl:choose>
     </xsl:function>
-
-    <!--
-        Prints out full language name from abbreviation.
-        List based on the ISO 639-2b three-letter language codes (http://www.loc.gov/standards/iso639-2/php/code_list.php).
+    <!-- 
+        Language code look-up map
+        List based on the ISO 639-2 language codes variants (https://en.wikipedia.org/wiki/List_of_ISO_639-2_codes).
```
anvit: I would suggest linking to the original source rather thank Wikipedia. In this case, Wikipedia links back to the [Library of Congress link](http://www.loc.gov/standards/iso639-2/php/code_list.php) that was already in place. I did notice that the LOC page is missing ISO639-3 codes, but their page references their website links to the [ISO639-3 page](https://iso639-3.sil.org/code_tables/639/data) which would be my preferred source if we wanted to link back to a list for future reference.

anvit (changes requested):
Thanks for submitting this PR @timtomch ! We're super grateful for this change to make PDF finding aids more accessible! I have a couple of small suggestions but this otherwise looks great.

lib/task/pdf/helper-functions.xsl
timtomch: Good point, I've reverted to the LOC and ISO sources as requested.

anvit (approved):
The updates look good. Thanks @timtomch !

PR #2106, opened 2025-07-28
Fix PHP 8x error for plugin page

melaniekung:


anvit (approved):
Looks good! Another bug that probably never worked as intended, but managed to go undetected for so long!

PR #2109, opened 2025-07-29
Fix PHP 8x error for update fixtures task.

melaniekung:


anvit (approved):
Thanks @melaniekung ! I'm surprised this ever worked as intended, as it would have been just resulted in an empty string in PHP 7.x. But I guess the method this is in only cares about the translations themselves, not the folder, so it managed to slip under the radar.

PR #2114, opened 2025-08-01
Add clipboard count UI labels to AtoM fixtures

anvit:
Adding these to fixtures so that they're present by default in new AtoM installations.


melaniekung (approved):
Thanks @anvit !

PR #2118, opened 2025-08-05
Add user to export context

### LINKED ISSUE #2117: Problem: expected user object does not exist during accession export
### Current Behavior

1. Create at least one accession
2. Run this command from the backend: `php symfony csv:accession-export accessions.csv`

You will get the following error:

```
>> csv       Found 1 accession records to export. Starting export.
PHP Fatal error:  Uncaught Error: Call to a member function isAuthenticated() on null in /atom/src/lib/flatfile/QubitFlatfileExport.class.php:285
Stack trace:
#0 /atom/src/lib/task/export/csvExportAccessionsTask.class.php(66): QubitFlatfileExport->exportResource(Object(QubitAccession))
#1 /atom/src/vendor/symfony/lib/task/sfBaseTask.class.php(70): csvExportAccessionsTask->execute(Array, Array)
#2 /atom/src/vendor/symfony/lib/task/sfTask.class.php(97): sfBaseTask->doRun(Object(sfCommandManager), NULL)
#3 /atom/src/vendor/symfony/lib/command/sfSymfonyCommandApplication.class.php(76): sfTask->runFromCLI(Object(sfCommandManager), NULL)
#4 /atom/src/vendor/symfony/lib/command/cli.php(20): sfSymfonyCommandApplication->run()
#5 /atom/src/symfony(14): include('/atom/src/vendo...')
#6 {main}
  thrown in /atom/src/lib/flatfile/QubitFlatfileExport.class.php on line 285
```

This is because `this->user` is not set when this code runs.

### Expected Behavior

The accession export should not fail.

### Possible Solution

During a CLI export, there may not be a user set. This code should probably check if it's running in the CLI, and if so, bypass this check.

I don't see the user property being set in the `QubitFlatFileExport` class anywhere either, this might be a good thing to set if certain exporters need to know who is making the export.

I can make a PR to fix this bug.

### Context and Notes

_No response_

### Version used

AtoM 2.9.2 (Latest qa/2.x)

Docker (`php:8.3-fpm-alpine`)

en

### PHP version

PHP 8.3

### Contact details

_No response_

danloveg:
Closes #2117 

Sets the user and the environment in the `QubitFlatfileExport` class.

For the `accessionNumber` logic, the field is excluded if the exporter is not running in the cli/worker, and if the user is not set or if the user is unauthenticated.

### DIFF
--- a/lib/flatfile/QubitFlatfileExport.class.php
+++ b/lib/flatfile/QubitFlatfileExport.class.php
@@ -30,6 +30,7 @@ class QubitFlatfileExport
     public $columnMap = [];       // flatfile columns that map to object properties
     public $propertyMap = [];       // flatfile columns that map to Qubit properties
     public $user;          // user doing the export
+    public $environment;   // environment the export is being run in
     protected $configurationLoaded = false;  // has the configuuration been loaded?
 
     protected $resource;                     // current resource being exported
@@ -70,6 +71,11 @@ class QubitFlatfileExport
         if (false !== $rowsPerFile) {
             $this->rowsPerFile = $rowsPerFile;
         }
+
+        $context = sfContext::getInstance();
+
+        $this->user = $context->getUser();
+        $this->environment = $context->getConfiguration()->getEnvironment();
 
         include_once sfConfig::get('sf_root_dir').'/lib/helper/QubitHelper.php';
     }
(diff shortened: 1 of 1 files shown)


anvit (approved):
Thanks @danloveg

PR #2119, opened 2025-08-05
Bump cypress version

anvit:
Bump cypress to v14.5.3. This should resolve failing tests since Chrome Devtools Protocol (CDP) was discontinued in Firefox 141, and cypress 14.1.0 added support for WebDriver BiDi to resolve this. Details [here](https://github.com/cypress-io/cypress/issues/32148)


melaniekung (approved):
thanks @anvit !

PR #2123, opened 2025-08-06
Fix shifted columns in CSV exports

### LINKED ISSUE #2122: Problem: Column mismatches in QubitFlatfileExport
### Current Behavior

1. Run any CSV export, where *only one* column is excluded from the export. A good example is an information object export that excludes the `accessionNumber`

The hidden column is not removed from rows after the second row, so cells in row 2 onward are shifted one column right. The culture column, for example, will appear one column *after* the actual culture heading, like this:

| Row Number | culture |  |
| ---------- | ------- | -- |
| 1 | en | |
| 2 | | en |
| 3 | | en |
| 4 | | en |

`culture` is supposed to be the last column, but an extra "empty" column is created because the column that is supposed to get removed doesn't actually get removed.

This is effectively an off-by-one error. See here:

- One extra element is added:

https://github.com/artefactual/atom/blob/2a5966634063b8e98975ed4c9448ff518e4976c4/lib/flatfile/QubitFlatfileExport.class.php#L111-L112

- The total number of columns - the total number of hidden columns is compared with that number:

https://github.com/artefactual/atom/blob/2a5966634063b8e98975ed4c9448ff518e4976c4/lib/flatfile/QubitFlatfileExport.class.php#L315-L316

Only one column is being hidden, but the `+ 1` that was added to the max columns is "hiding" the fact that this column should be removed.

### Expected Behavior

None of the rows in the CSV should be shifted right.

### Possible Solution

Rather than check the length of the row, a solution is to check if indices of hidden columns are in the row, and remove them if they are.

I can open a PR for this.

### Context and Notes

_No response_

### Version used

AtoM 2.9.2 (Latest qa/2.x)

Docker (`php:8.3-fpm-alpine`)

en

### PHP version

PHP 8.3

### Contact details

_No response_

danloveg:
Closes #2122 

Fixes the off-by-one error causing columns to be shifted right in flatfile exports.

I've also written some unit tests to run some checks against the `exportResource` logic.

### DIFF
--- a/test/phpunit/QubitFlatfileExportTest.php
+++ b/test/phpunit/QubitFlatfileExportTest.php
@@ -184,7 +184,42 @@ class QubitFlatfileExportTest extends TestCase
         $this->assertEquals(['1|2|3', 'u|v|w', 'C'], $rowData);
     }
 
-    public function testExportSingleResourceHiddenCols(): void
+    public function testExportSingleResourceOneHiddenCol(): void
+    {
+        $csvFile = $this->vfs->url().'/output.csv';
+
+        // Hide colA
+        $mockExporter = $this->createMockExporter(
+            outputPath: $csvFile,
+            columnNames: ['colA', 'colB', 'colC', 'colD'],
+            hiddenColumns: ['colA'],
+        );
+
+        $mockResource = $this->createMockResource(properties: [
+            'colA' => 'A',
+            'colB' => 'B',
+            'colC' => 'C',
+            'colD' => 'D',
+        ]);
+
+        $mockExporter->exportResource($mockResource);
+
+        $this->assertTrue(file_exists($csvFile));
+
+        $csvContent = file_get_contents($csvFile);
+        $rows = str_getcsv($csvContent, "\n");
+
+        // Should have two rows (one header, one data row)
+        $this->assertCount(2, $rows);
+
+        $headerData = str_getcsv($rows[0]);
+        $rowData = str_getcsv($rows[1]);
+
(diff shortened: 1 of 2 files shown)


test/phpunit/QubitFlatfileExportTest.php
```diff
@@ -0,0 +1,302 @@
...
+        $headerData = str_getcsv($rows[0]);
+        $rowData1 = str_getcsv($rows[1]);
+        $rowData2 = str_getcsv($rows[2]);
+        $rowData3 = str_getcsv($rows[3]);
+
+        $this->assertEquals(['colB', 'colC'], $headerData);
+        $this->assertEquals(['B1', 'C1'], $rowData1);
+        $this->assertEquals(['B2', 'C2'], $rowData2);
+        $this->assertEquals(['B3', 'C3'], $rowData3);
+    }
```
anvit, line 336: It would be good to have a test for exporting multiple resources with multiple hidden columns as well (similar to the single resource one, with columns B and D hidden).

anvit (commented):
Thanks so much @danloveg ! I added a comment about one additional test, but other than that this looks good!

test/phpunit/QubitFlatfileExportTest.php
danloveg: I agree! I also added a test for one hidden column when exporting a single resource. See: https://github.com/artefactual/atom/pull/2123/commits/581d6d64a741822dbabd363552bd2e7822dcaa22

anvit: Thank you, this is perfect!

anvit (approved):

PR #2127, opened 2025-08-07
Run unit test that creates objects in a transaction

### LINKED ISSUE #2125: Problem: QubitInformationObjectTest::testGetByTitleIdentifierAndRepo pollutes database
### Current Behavior

1. Run atom in Docker using `docker-compose.dev.yml`
2. Run the unit tests with `docker compose exec atom composer test`
3. Count the number of "TestDescription" objects in the database by running the command: `docker compose exec percona mysql -u root -p atom --execute 'SELECT COUNT(*) FROM information_object_i18n WHERE title LIKE "TestDescription%"';`
  - *Note:* the password is in `docker/etc/environment`

Successive runs of step 2 will show an increasing count in step 3.

### Expected Behavior

This unit test should not create extra data in the database. It should either use mocked objects, or delete the objects it creates after the test is finished.

### Possible Solution

Wrap this unit test in a transaction so that the objects it creates are removed after the test exits.

### Context and Notes

I run the unit tests as I develop code in Docker. Since the unit tests require running against a database, running the tests in Docker against my dev environment is the easiest way of running these tests. Ideally, I don't have to reset the database after running unit tests.

### Version used

AtoM 2.9.2 (latest qa/2.x)

Docker (`php:8.3-fpm-alpine`)

en

### PHP version

PHP 8.3

### Contact details

_No response_

danloveg:
Closes #2125 

Wraps the `testGetByTitleIdentifierAndRepo` unit test within a transaction so that the objects it creates in the database are rolled back when the test exits or throws an exception.

### DIFF
--- a/test/phpunit/QubitInformationObjectTest.php
+++ b/test/phpunit/QubitInformationObjectTest.php
@@ -23,46 +23,48 @@ class QubitInformationObjectTest extends TestCase
      */
     public function testGetByTitleIdentifierAndRepo($identifier, $title, $repoName, $hasLinkedRepo, $expectedTitle)
     {
-        $randomString = rand(1000000, 9999999);
+        $this->withTransaction(function () use ($identifier, $title, $repoName, $hasLinkedRepo, $expectedTitle) {
+            $randomString = rand(1000000, 9999999);
 
-        $io = new QubitInformationObject();
-        $io->title = 'TestDescriptionTitle'.$randomString;
-        $io->identifier = 'TestDescriptionIdentifier'.$randomString;
+            $io = new QubitInformationObject();
+            $io->title = 'TestDescriptionTitle'.$randomString;
+            $io->identifier = 'TestDescriptionIdentifier'.$randomString;
 
-        // Set up a linked repository if needed for the test.
-        if (true === $hasLinkedRepo) {
-            $repository = new QubitRepository();
-            $repository->indexOnSave = false;
-            $repository->setAuthorizedFormOfName('TestRepository'.$randomString);
-            $repository->save();
-            $io->setRepositoryId($repository->id);
-        }
+            // Set up a linked repository if needed for the test.
+            if (true === $hasLinkedRepo) {
+                $repository = new QubitRepository();
+                $repository->indexOnSave = false;
+                $repository->setAuthorizedFormOfName('TestRepository'.$randomString);
+                $repository->save();
+                $io->setRepositoryId($repository->id);
+            }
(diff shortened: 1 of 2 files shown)


melaniekung:
hi @danloveg , just an official announcemnt as mentioned before: 
Announcement: BS2 removal in 2.10 and merger/closure of all open pull requests

With the release of version 2.10, we will be removing BS2 from AtoM. We've realized that it will require a large amount of work to rebase any existing pull requests once we merge this to qa/2.x. This commit will include moving multiple templates and javascripts from plugins/arDominionB5Plugin to apps/qubit or js/. 

Therefore, we will be merging or closing all existing PRs after September 1, 2025 in order to do the BS2 removal commit. If you’re planning to contribute to AtoM after this date, please be sure to pull qa/2.x once the BS2 removal changes are merged, and start any future branches from the latest qa in order to avoid having to do a large rebase.

We appreciate your patience as we work through this process. If you have any questions or want to flag any upcoming changes, please feel free to contact us at maintainers@artefactual.com.

Thank you to all of the AtoM contributors! Your work continues to strengthen the open source project.

test/phpunit/QubitInformationObjectTest.php
```diff
@@ -23,46 +23,48 @@ class QubitInformationObjectTest extends TestCase
      */
     public function testGetByTitleIdentifierAndRepo($identifier, $title, $repoName, $hasLinkedRepo, $expectedTitle)
     {
-        $randomString = rand(1000000, 9999999);
+        $this->withTransaction(function () use ($identifier, $title, $repoName, $hasLinkedRepo, $expectedTitle) {
```
anvit: If we're not using the `withTransaction` method anywhere else, would it make sense to skip adding a new method and just include the try catch block here itself?

danloveg: How about this: https://github.com/artefactual/atom/pull/2127/commits/3754e2292764d3962526413715219f1da544517c

I checked for other `phpunit` tests that use the database like this test case does and none of them do, but it's possible that we might want to write more tests in the future that do make use of writing to the database (especially integration tests).

I moved this function out to a new `TransactionTestCase` that tests can use instead of the default `TestCase`. This new class can be used in case someone wants to run their tests in transactions.

anvit: Thanks! I like this approach a lot more!

anvit (approved):
Thanks @danloveg !

PR #2128, opened 2025-08-11
Add ability to do bulk CSV exports. (#2057)

### LINKED ISSUE #2057: Feature: Ability to do bulk CSV exports
At the moment, AtoM supports the ability to bulk exports only in the XML format. It would be helpful to have the ability to bulk export in the CSV format as well.

melaniekung:


anvit (approved):
Thanks @melaniekung !

PR #2130, opened 2025-08-12
Fix header content security policy violation

### LINKED ISSUE #2129: Problem: style attribute in customizable header violates CSP
### Current Behavior

1. Run the app in Docker
2. Open the dev console in Chrome or Firefox (the browsers I tested)

You will see a warning about the style attribute not being applied due to the Content Security Policy. The header does not have any special style - changing the colour in the settings does not change the header.

### Expected Behavior

The header style should be applied based on the `app_header_background_colour` setting.

### Possible Solution

Use a nonce on a `<style>` element, and apply styling to the `#top-bar` element instead of using an inline style attribute.

### Context and Notes

Firefox:

Image

Chrome:

Image

### Version used

AtoM 2.9.2 (latest qa/2.x)

Docker (`php:8.3-fpm-alpine`)

en

### PHP version

PHP 8.3

### Contact details

_No response_

danloveg:
Closes #2129 

Moves the CSS from an inline style attribute to a `<style>` element with a nonce that obeys the site's content security policy.

### DIFF
--- a/plugins/arDominionB5Plugin/templates/_header.php
+++ b/plugins/arDominionB5Plugin/templates/_header.php
@@ -14,7 +14,15 @@
   </div>
 <?php } ?>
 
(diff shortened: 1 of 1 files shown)


melaniekung (approved):
Thanks for catching this @danloveg !

PR #2133, opened 2025-08-13
Add password confirmation before updating user.

### LINKED ISSUE #2134: Problem: AtoM allows password changes without requiring user authentication
### Expected behaviour

Users should be required to re-authenticate (e.g., by entering their current password) before they are allowed to change their account password.

### Current behaviour

AtoM allows users to change their password via the user edit form without requiring their existing password. As a result, anyone with access to an active session can change the account password without verification.

### Steps to reproduce

Log in as an administrator
Go to Admin > Users
Click on the Edit button of any existing user in the Users table
You can set a new password for the user without providing your current password

melaniekung:


sbreker (approved):
LGTM :+1:  Thanks for adding this!

PR #2136, opened 2025-08-18
Fix XML esc_specialchars for DO URLs (#2135)

### LINKED ISSUE #2135: Problem: XML not escaping for access copy URLs in EAD export
URLs containing ampersands (&) in access copy links are breaking EAD XML generation (and therefore breaking Finding Aid generation) with `start tag expected, '<' not found` error.

1. Create a new description/update an existing description and link an external DO where the URL includes an ampersand (&).
2. Attempt to export EAD XML or generating a Finding Aid.

melaniekung:
Added proper XML escaping using escape_dc(esc_specialchars()) for DAO href attributes to handle special characters in URLs.


anvit (approved):
Good catch @melaniekung !

PR #2138, opened 2025-08-19
Mark currentPassword as optional if user doesn't exist

### LINKED ISSUE #2137: Problem: Creating new users requires non-existent currentPassword field
### Current Behavior

1. Run the app in Docker with `docker-compose.dev.yml`.
2. Log in as the demo user, or another admin that can create users.
3. Fill out the form to create a new user at `/index.php/user/add`
4. There is a note that you need to enter your current password, but there is no field to do so.
5. When you click Create, **the user object won't be created**, and there is no feedback about why.

### Expected Behavior

The user object should be created and/or there should be feedback about why the user was not created.

### Possible Solution

The `currentPassword` field isn't being rendered when the user doesn't already exist. I believe this is why there is no feedback when clicking Create; the field isn't rendered so the errors on the field aren't rendered either.

I think that this field should not be marked as required when the user doesn't already exist. Also, the message about entering your current password shouldn't be shown if it's not required.

### Context and Notes

Image

### Version used

AtoM 2.9.2 (latest qa/2.x)

Docker (`php:8.3-fpm-alpine`)

en

### PHP version

PHP 8.3

### Contact details

_No response_

danloveg:
Closes #2137 

Makes `currentPassword` optional if the user doesn't exist yet. Conditionally displays the message about current password only if the user already exists.

### DIFF
--- a/apps/qubit/modules/user/actions/editAction.class.php
+++ b/apps/qubit/modules/user/actions/editAction.class.php
@@ -196,9 +196,9 @@ class UserEditAction extends DefaultEditAction
                 $this->form->setDefault('currentPassword', null);
 
                 $this->form->setValidator('currentPassword', new sfValidatorAnd([
-                    new sfValidatorString(['required' => !isset($this->getRoute()->resource)]),
+                    new sfValidatorString(['required' => isset($this->getRoute()->resource)]),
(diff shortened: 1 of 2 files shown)


melaniekung (approved):
thank you for catching this @danloveg!

PR #2140, opened 2025-08-21
arOidcPlugin: use `expires_in` for access token

### LINKED ISSUE #2139: Problem: OIDC plugin refreshes access token only once due to use of refresh_expires_in
### Current Behavior

When using AtoM with Keycloak OIDC, the access token is only refreshed once after login. After that, no further refresh requests are sent until the refresh token itself is near expiry. This can cause sessions to end prematurely, even though a valid refresh token is available.

1. Configure AtoM with Keycloak OIDC.

2. Set Access Token Lifespan in Keycloak to a short value (e.g. 1 minute).

3. Log in to AtoM and browse authenticated pages.

4. Observe with tcpdump or Keycloak logs that only one refresh request is sent.

### Expected Behavior

AtoM should refresh the access token periodically according to its configured lifetime (expires_in or the token’s exp claim).

### Possible Solution

** Proposed fix **
Change the code to update oidc-expiry using expires_in instead of refresh_expires_in.
A minimal patch that resolves the issue (plugins/arOidcPlugin/lib/oidcUser.class.php):

```
- if (isset($refreshResult->refresh_expires_in) && is_numeric($refreshResult->refresh_expires_in)) {
-     $newExpiryTime = $currentTime + (int) $refreshResult->refresh_expires_in;
+ if (isset($refreshResult->expires_in) && is_numeric($refreshResult->expires_in)) {
+     $newExpiryTime = $currentTime + (int) $refreshResult->expires_in;
      $this->logger->info(sprintf('$newExpiryTime: %s', $newExpiryTime));
```

### Context and Notes

In `plugins/arOidcPlugin/lib/oidcUser.class.php`, the plugin updates oidc-expiry using refresh_expires_in from the refresh response. This is the refresh token lifetime, not the access token lifetime. Consequently, AtoM does not refresh the access token again until the refresh token itself is expiring.

I tested this small change and it works.

I was checking:

* keycloak session to see when the "Last Access" value changed (only changin 1 time)
* tcpdump in AtoM server to see when AtoM connects to keycloak server

After the change in test VM:

* More than 3 hours connection when running "Session Alive" plugin in AtoM URL. (it is still connected but I think it is enough to create the issue/PR)
* Setting token lifespan=1min in keycloak's client settings to force the refresh as soon as possible.

### Version used

AtoM 2.8 and 2.9

Ubuntu 20, Ubuntu22 and Rocky9

en

### PHP version

php 7.4 and 8.1

### Contact details

_No response_

mamedin:
The OIDC plugin was previously updating `oidc-expiry` using `refresh_expires_in` from the token endpoint response. This value represents the *refresh token* lifetime, not the access token’s. As a result, AtoM only refreshed the access token once and then waited until the refresh token was close to expiry, causing user sessions to end unexpectedly.

This change fixes the logic to use `expires_in` to compute the new access token expiry. With this adjustment, AtoM refreshes tokens regularly according to the configured access token lifespan, improving session stability when using Keycloak and other OIDC providers.

Connects to https://github.com/artefactual/atom/issues/2139

### DIFF
--- a/plugins/arOidcPlugin/lib/oidcUser.class.php
+++ b/plugins/arOidcPlugin/lib/oidcUser.class.php
@@ -255,8 +255,8 @@ class oidcUser extends myUser implements Zend_Acl_Role_Interface
                     return false;
                 }
 
-                if (isset($refreshResult->refresh_expires_in) && is_numeric($refreshResult->refresh_expires_in)) {
-                    $newExpiryTime = $currentTime + (int) $refreshResult->refresh_expires_in;
+                if (isset($refreshResult->expires_in) && is_numeric($refreshResult->expires_in)) {
+                    $newExpiryTime = $currentTime + (int) $refreshResult->expires_in;
                     $this->logger->info(sprintf('$newExpiryTime: %s', $newExpiryTime));
                 } else {
                     $this->logger->err('Error calculating new token expiry. Session no longer active.');


sbreker (approved):
Thanks for this @mamedin! Amazing sleuthing! :+1:

PR #2141, opened 2025-08-26
Sync edit and index group permissions. (#2094)

### LINKED ISSUE #2094: Problem: Group permissions are not set/displayed correctly
### Current Behavior

1. Log in, and go to ⚙️ > Groups > Contributor > Archival institution permissions.
2. Change the `Update` permission to `Grant` and save.
3. Notice that the permissions has updated correctly.
4. Edit the permission again.
5. Notice that the edit table does not reflect the updated permissions and notice that everything shows `inherit`.
6. Change the `Update` permission to `Deny` and save.
7. Notice that it has now been set to `Inherit`

### Expected Behavior

Permissions should be accurately displayed while editing, and setting a permission to `Deny` should update it correctly and not set it to `Inherit`.

melaniekung:
- Populate permissions variable and pass to partial index template
- Distinguish between Deny (0) and Inherit (null)
- Remove extra column in group permissions settings


anvit (approved):
Thanks @melaniekung

PR #2143, opened 2025-08-26
Display draft DOs in carousel. (#2142)

### LINKED ISSUE #2142: Display draft DOs in carousel, if permissions permits.
For authenticated users, if they have permissions to view draft digital objects, they should be displayed in the carousel. 

This bug only applies in specific permissions scenario, to reproduce:
1. As an administrator, go to Settings > Groups > authenticated > Archival description permissions > Edit
2. Set Permissions by archival description for any description with draft DOs, granting the following:
    - View draft
    - Access thumbnail
3. Ensure that for `All archival description`, the same actions are set to `Inherit`
4. Log out as administrator
5. Log in as an authenticated user (not administrator)
6. View the description with draft DOs

Expected Behaviour:
It is expected that `Inherit` permissions from the `All archival description` should inherit the `grant` permissions from the archival description specific settings.

melaniekung:


anvit (approved):
Thanks @melaniekung

PR #2144, opened 2025-08-26
Add empty io check to data integrity and repair

anvit:
Add ability to search for and delete information objects with all columns set to NULL in the data integrity and repair task.


melaniekung (approved):
Thanks @anvit, lots of users will be happy with this addition !

melaniekung (approved):
Thanks @anvit , LGTM!

PR #2146, opened 2025-08-27
Clarify public finding aid setting. (#2145)

### LINKED ISSUE #2145: Unclear Finding aid setting
The `Generate Finding Aid as public user` setting sounds like this setting will allow public users to generate Finding aids, `When set to 'yes' generated Finding Aids will exclude Drafts`.

melaniekung:


anvit (approved):
Looks good!

PR #2155, opened 2025-09-05
Close cURL handles on DO downloads

anvit:
- Added `close()` method in sfCurlAdapter to explicitly release the cURL handle.
- Added `close()` passthrough in sfWebBrowser to call adapter cleanup safely.
- Updated QubitDigitalObject::downloadExternalObject() to call `$browser->close()` after successful download and before throwing exceptions.

This ensures that external digital object imports release sockets immediately, preventing buildup of CLOSE_WAIT connections during large CSV imports.

### DIFF
--- a/lib/arInstall.class.php
+++ b/lib/arInstall.class.php
@@ -299,6 +299,7 @@ class arInstall
 
         $curlHttpCode = curl_getinfo($curl, CURLINFO_HTTP_CODE);
         curl_close($curl);
+        $curl = null;
 
         if (200 !== $curlHttpCode) {
             throw new Exception("Elasticsearch error: {$curlHttpCode}");
--- a/lib/arInstall.class.php
+++ b/lib/arInstall.class.php
@@ -298,7 +298,7 @@ class arInstall
         }
 
         $curlHttpCode = curl_getinfo($curl, CURLINFO_HTTP_CODE);
-        curl_close($curl);
+        // set curl handle to null to close connection
         $curl = null;
 
         if (200 !== $curlHttpCode) {
--- a/plugins/arStorageServicePlugin/lib/arStorageServiceUtils.class.php
+++ b/plugins/arStorageServicePlugin/lib/arStorageServiceUtils.class.php
@@ -62,7 +62,7 @@ class arStorageServiceUtils
         curl_exec($ch);
         $status = curl_getinfo($ch, CURLINFO_RESPONSE_CODE);
 
-        curl_close($ch);
+        // set curl handle to null to close connection
         $ch = null;
 
         return $status;
--- a/plugins/arStorageServicePlugin/lib/arStorageServiceUtils.class.php
+++ b/plugins/arStorageServicePlugin/lib/arStorageServiceUtils.class.php
@@ -63,6 +63,7 @@ class arStorageServiceUtils
         $status = curl_getinfo($ch, CURLINFO_RESPONSE_CODE);
 
         curl_close($ch);
+        $ch = null;
 
         return $status;
     }
--- a/plugins/sfModsPlugin/lib/sfModsConvertor.class.php
+++ b/plugins/sfModsPlugin/lib/sfModsConvertor.class.php
@@ -339,6 +339,7 @@ class sfModsConvertor extends QubitSaxParser
         file_put_contents($tempFile, curl_exec($curlSession));
 
         curl_close($curlSession);
+        $curlSession = null;
 
         // Add temp file to digital object import queue
         $pathParts = pathinfo($this->data());
--- a/plugins/sfModsPlugin/lib/sfModsConvertor.class.php
+++ b/plugins/sfModsPlugin/lib/sfModsConvertor.class.php
@@ -338,7 +338,7 @@ class sfModsConvertor extends QubitSaxParser
         $tempFile = tempnam(sys_get_temp_dir(), 'atomFile');
         file_put_contents($tempFile, curl_exec($curlSession));
 
-        curl_close($curlSession);
+        // set curl handle to null to close connection
         $curlSession = null;
 
         // Add temp file to digital object import queue
--- a/plugins/sfWebBrowserPlugin/lib/sfCurlAdapter.class.php
+++ b/plugins/sfWebBrowserPlugin/lib/sfCurlAdapter.class.php
@@ -124,7 +124,7 @@ class sfCurlAdapter
 
     public function __destruct()
     {
-        curl_close($this->curl);
+        $this->close();
     }
 
     public function close()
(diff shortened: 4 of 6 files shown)


melaniekung (approved):
LGTM!

plugins/arStorageServicePlugin/lib/arStorageServiceUtils.class.php
```diff
@@ -63,6 +63,7 @@ public static function getFileFromStorageService($url)
         $status = curl_getinfo($ch, CURLINFO_RESPONSE_CODE);
 
         curl_close($ch);
```
sbreker: `curl_close()` can be removed as it no longer does anything. See PHP 8 breaking changes [here](https://www.php.net/manual/en/migration80.incompatible.php#migration80.incompatible.resource2object) - search for curl.

plugins/sfModsPlugin/lib/sfModsConvertor.class.php
```diff
@@ -339,6 +339,7 @@ protected function electronicLocatorTag()
         file_put_contents($tempFile, curl_exec($curlSession));
 
         curl_close($curlSession);
```
sbreker: `curl_close()` can be removed as it no longer does anything. See PHP 8 breaking changes.

plugins/sfWebBrowserPlugin/lib/sfCurlAdapter.class.php
```diff
@@ -124,7 +124,15 @@ public function __construct($options = [])
...
     public function __destruct()
     {
-        curl_close($this->curl);
+        $this->close();
+    }
+
+    public function close()
+    {
+        if ($this->curl) {     // works for resource and CurlHandle
+            curl_close($this->curl);
```
sbreker: This curl_close() can be removed too.

lib/arInstall.class.php
```diff
@@ -299,6 +299,7 @@ public static function checkSearchConnection($options)
 
         $curlHttpCode = curl_getinfo($curl, CURLINFO_HTTP_CODE);
         curl_close($curl);
```
sbreker: `curl_close()` can be removed as it no longer does anything. See PHP 8 breaking changes [here](https://www.php.net/manual/en/migration80.incompatible.php#migration80.incompatible.resource2object) - search for curl.

sbreker (changes requested):
I think all calls to `curl_close()` can be removed. Let me know what you think.

anvit:
@sbreker I'd left them in earlier because the method isn't deprecated or anything, but you're right, if it's a noop, it's better to remove it altogether and if a future PHP updated restores this functionality back to curl_close again, we can think about changing it back.

I've removed curl_close from the code, and added comments that spell out that we're setting the curl handle to null to close the connection.

sbreker (approved):
Thanks @anvit! Looks good!

lib/arInstall.class.php
```diff
@@ -298,7 +298,8 @@ public static function checkSearchConnection($options)
         }
 
         $curlHttpCode = curl_getinfo($curl, CURLINFO_HTTP_CODE);
-        curl_close($curl);
+        // set curl handle to null to close connection
+        $curl = null;
```
miceno, line 302: I am afraid that setting a variable to null is not the same as closing... This is a code-smell for a later problem of memory leaks and unused resources.

plugins/arStorageServicePlugin/lib/arStorageServiceUtils.class.php
```diff
@@ -62,7 +62,8 @@ public static function getFileFromStorageService($url)
         curl_exec($ch);
         $status = curl_getinfo($ch, CURLINFO_RESPONSE_CODE);
 
-        curl_close($ch);
+        // set curl handle to null to close connection
+        $ch = null;
```
miceno, line 66: I am afraid that setting a variable to null is not the same as closing... This is a code-smell for a later problem of memory leaks and unused resources.

plugins/sfModsPlugin/lib/sfModsConvertor.class.php
```diff
@@ -338,7 +338,8 @@ protected function electronicLocatorTag()
         $tempFile = tempnam(sys_get_temp_dir(), 'atomFile');
         file_put_contents($tempFile, curl_exec($curlSession));
 
-        curl_close($curlSession);
+        // set curl handle to null to close connection
+        $curlSession = null;
```
miceno, line 342: I am afraid that setting a variable to null is not the same as closing... This is a code-smell for a later problem of memory leaks and unused resources.

plugins/sfWebBrowserPlugin/lib/sfCurlAdapter.class.php
```diff
@@ -124,7 +124,15 @@ public function __construct($options = [])
...
     {
-        curl_close($this->curl);
+        $this->close();
+    }
+
+    public function close()
+    {
+        // If curl handle exists, set it to null to close connection
+        if ($this->curl) {
+            $this->curl = null;
```
miceno, line 134: I am afraid that setting a variable to null is not the same as closing... This is a code-smell for a later problem of memory leaks and unused resources.

lib/arInstall.class.php
miceno: As per checking with Copilot and other resources the appropiate solution is the one before this change:
```suggestion
        curl_close($curl);
```

plugins/sfWebBrowserPlugin/lib/sfCurlAdapter.class.php
miceno: The selected code is not properly releasing the cURL handle or its resources. Setting `$this->curl = null;` only removes the reference to the cURL handle, but does not actually close it. To properly release the resources, you should call `curl_close($this->curl)` before setting it to null.

anvit:
Hi @miceno ! I understand that this looks like a code smell — without having the context for this problem that certainly what it looks like, so I wanted to share some additional details about this PR.

We originally created it because we noticed a small memory leak on digital object downloads. When looking into it we found that the official PHP documentation says that [curl_close() does nothing in PHP 8.x](https://www.php.net/manual/en/function.curl-close.php), and in fact the correct way to free up memory is to set the handle to null so that it [no longer referenced, which frees up the memory](https://www.php.net/manual/en/migration80.incompatible.php#migration80.incompatible.resource2object).

Given all of this, for now we've decided to [remove the non-functional curl_close code](https://github.com/artefactual/atom/pull/2155#issuecomment-3272325505), and if a future version of PHP updates how this is supposed to be handled, we can revisit this then.

I also wanted to note for the future, that LLM tools like Copilot cannot be relied on for accuracy, and much less so on a code base like AtoM which is primarily in PHP, using a legacy framework, and having generally far smaller developer community discussions that would have made it into the common crawl and other major sources of training data for context that would be essential for being able to generate responses that would be accurate.

Thank you again for taking the time to share your input, and looking through the file changes for this pull request! We really appreciate that!

PR #2156, opened 2025-09-09
Update build package.xml with BS5 build files

### LINKED ISSUE #2150: Problem: 2.9.x release tarball is missing package.json?
### Current Behavior

The tarball mentioned in the installation instructions (see below) seems to be **missing [package.json which is present on github tree](https://github.com/artefactual/atom/blob/qa/2.x/package.json)?**

1. Go to the [official installation instructions for AtoM v2.9 on Ubuntu 24.04](https://www.accesstomemory.org/en/docs/2.9/admin-manual/installation/ubuntu/#installation-ubuntu)

2. Scroll down to the [Option 1: Download the Tarball](https://www.accesstomemory.org/en/docs/2.9/admin-manual/installation/ubuntu/#option-1-download-the-tarball)

3. Open the [Download page](https://www.accesstomemory.org/en/download/) and get any recent .tar.gz

4. Unpack the tar: Try to find any `package.json` files. There's only one, but that's not enough.

5. Trying to `npm install` (following steps from "[compile theme files](https://www.accesstomemory.org/en/docs/2.9/admin-manual/installation/ubuntu/#compile-theme-files)" setup instructions fails with "package.json not found" error.

Works fine, when following [Option 2: Check out from git repository](https://www.accesstomemory.org/en/docs/2.9/admin-manual/installation/ubuntu/#option-2-check-out-the-code-from-our-git-repository), instead of using the release tar.

### Expected Behavior

The tarball should contain all files necessary for the official installation instructions to work without errors.

### Possible Solution

Use "OPTION 2: install from git" - or try manually downloading package.json to the extracted tarball? (have NOT tried that, though)

### Context and Notes

_No response_

### Version used

_No response_

_No response_

_No response_

### PHP version

_No response_

### Contact details

_No response_

anvit:
Update build package.xml to include package.json, package-lock.json, webpack.config.js. Also update the developers section and point the email addresses to the maintainers email.

### DIFF
--- a/config/package.xml
+++ b/config/package.xml
@@ -9,16 +9,28 @@ Access to Memory (AtoM) is a fully web-based toolkit that will allow institution
 The goal is to provide an easy-to-use, flexible toolkit that complies with open standards (e.g. Dublin Core, ICA-ISAD, MODS, EAD, METS) and is developed using an open architecture that takes advantage of emerging web-based tools and practices.
   </description>
   <lead>
+    <name>Anvit Srivastav</name>
+    <user>anvit</user>
+    <email>maintainers@artefactual.com</email>
+    <active>yes</active>
+  </lead>
+  <lead>
+    <name>Melanie Kung</name>
+    <user>melaniekung</user>
+    <email>maintainers@artefactual.com</email>
+    <active>yes</active>
+  </lead>
+  <lead>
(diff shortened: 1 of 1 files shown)


melaniekung (approved):
LGTM, thanks @anvit

PR #2160, opened 2025-09-10
Update ES plugin to call forcemergeAll()

sbreker:
Elastica has deprecated the function `optimizeAll()` in favour of `forcemergeAll()`. This commit updates arElasticSearchPlugin::optimize() to use this new Elastica method.


anvit (approved):
Thanks @sbreker !

PR #2162, opened 2025-09-16
Improve DO load task URL handling

sbreker:
Replace call to @get_headers() with curl so URL validation will follow redirects successfully. get_headers() will not follow redirects and will fail URL validation causing the asset to not be downloaded.

Update the digital object load task to use `$this->` instead of `self::` when calling non-static digitalObjectLoadTask class methods.


melaniekung (approved):
LGTM, thanks @sbreker !

PR #2168, opened 2025-09-25
Fix Symfony debug mode in BS5 theme

### LINKED ISSUE #2031: Problem: Debug mode broken in BS5
### Current Behavior

1. Open the index.php file in your text editor of choice
2. Find the following line:
```
$configuration = ProjectConfiguration::getApplicationConfiguration('qubit',
'prod', false);
```
3. Update the “prod” environment to “dev” and “false” flag to “true”
```
$configuration = ProjectConfiguration::getApplicationConfiguration('qubit',
'dev', true);
```
4. Save the index.php file
5. Reload AtoM in your web browser
6. Notice the fatal error that breaks debug mode
```
Fatal error: Declaration of DebugPDO::query() must be compatible with PDO::query(string $query, ?int $fetchMode = null, mixed ...$fetchModeArgs): PDOStatement|false in /atom/src/vendor/symfony/lib/plugins/sfPropelPlugin/lib/vendor/propel/util/DebugPDO.php on line 331
```

### Expected Behavior

Debug mode should not be broken

### Possible Solution

Update sfPropelPlugin's DebugPDO::query() method to match the signature of PDO::query() like so:
```
 	#[\ReturnTypeWillChange]
	public function query(string $query, ?int $fetchMode = null, mixed ...$fetchModeArgs)
	{
		$debug	= $this->getDebugSnapshot();
		$args	= func_get_args();
		if (version_compare(PHP_VERSION, '5.3', '<')) {
			$return	= call_user_func_array(array($this, 'parent::query'), $args);
		} else {
			$return	= call_user_func_array('parent::query', $args);
		}
		
		$sql = $args[0];
		$this->log($sql, null, __METHOD__, $debug);
		$this->setLastExecutedQuery($sql);
		$this->incrementQueryCount();
		
		return parent::query($query, $fetchMode, ...$fetchModeArgs);
	}
```

danloveg:
Closes #2031 

I've been curious about using debug mode in Symfony so I took a look at this issue. Here is the debug mode bar working in the BS5 theme:

debug mode working

The main issue besides the one fixed in the `DebugPDO` class was the fact that **the Symfony debug bar is incompatible with the strict content security policy** that's enabled for BS5 themes. Rather than refactor how the debug bar gets rendered, I disabled the CSP when debug mode is on.

## Other issues addressed

One interesting thing with debug mode is that warnings and errors are rendered directly to the HTML displayed to the user. I wasn't able to fix all of these, but I fixed a number of the most obvious ones that get rendered to the homepage.

The "random" changes I made in this pull request address those errors. For example, adding the `$parsedown` variable to `QubitMarkdown.class.php` fixes this error from being rendered:

parsedown error

### DIFF
--- a/lib/routing/QubitRoute.class.php
+++ b/lib/routing/QubitRoute.class.php
@@ -146,7 +146,7 @@ class QubitRoute extends sfRoute
         foreach ($tokens as $token) {
             switch ($token[0]) {
                 case 'variable':
-                    if ((!$optional || !isset($this->defaults[$token[3]]) || $parameters[$token[3]] != $this->defaults[$token[3]]) && $parameters[$token[3]]) {
+                    if ((!$optional || !isset($this->defaults[$token[3]]) || $parameters[$token[3]] != $this->defaults[$token[3]]) && null !== $parameters[$token[3]]) {
                         $url[] = QubitRoute::urlencode3986($parameters[$token[3]]);
                         $optional = false;
                     }
--- a/lib/routing/QubitRoute.class.php
+++ b/lib/routing/QubitRoute.class.php
@@ -146,7 +146,7 @@ class QubitRoute extends sfRoute
         foreach ($tokens as $token) {
             switch ($token[0]) {
                 case 'variable':
-                    if (!$optional || !isset($this->defaults[$token[3]]) || $parameters[$token[3]] != $this->defaults[$token[3]]) {
+                    if ((!$optional || !isset($this->defaults[$token[3]]) || $parameters[$token[3]] != $this->defaults[$token[3]]) && $parameters[$token[3]]) {
                         $url[] = QubitRoute::urlencode3986($parameters[$token[3]]);
                         $optional = false;
                     }
--- a/lib/QubitMarkdown.class.php
+++ b/lib/QubitMarkdown.class.php
@@ -31,6 +31,8 @@ class QubitMarkdown
         'bold' => '__',
     ];
 
+    protected $parsedown;
+
     private $enabled;
 
     protected function __construct()
--- a/lib/filter/QubitCSPFilter.php
+++ b/lib/filter/QubitCSPFilter.php
@@ -21,8 +21,8 @@ class QubitCSP extends sfFilter
 {
     public function execute($filterChain)
     {
-        // Only use CSP headers if theme is b5.
-        if (!sfConfig::get('app_b5_theme', false)) {
+        // Only use CSP headers if theme is b5, or if not debugging
+        if (!sfConfig::get('app_b5_theme', false) || $this->getContext()->getConfiguration()->isDebug()) {
             $filterChain->execute();
 
             return;
--- a/lib/form/arB5WidgetFormSchemaFormatter.class.php
+++ b/lib/form/arB5WidgetFormSchemaFormatter.class.php
@@ -19,6 +19,7 @@
 
 class arB5WidgetFormSchemaFormatter extends sfWidgetFormSchemaFormatter
 {
+    public $form;
     protected $rowFormat = "<div class=\"mb-3\">\n  %label%\n  %field%\n"
         ."  %error%\n  %help%\n  %hidden_fields%\n</div>";
     protected $formErrorListFormat = '<div class="alert alert-danger"'
--- a/lib/model/QubitMenu.php
+++ b/lib/model/QubitMenu.php
@@ -470,6 +470,7 @@ class QubitMenu extends BaseMenu
 
         $module = $route['parameters']['module'];
         $action = $route['parameters']['action'];
+
         // Check access from security.yml rules
         return $context->getUser()->checkModuleActionAccess($module, $action);
     }
--- a/lib/model/QubitMenu.php
+++ b/lib/model/QubitMenu.php
@@ -462,9 +462,14 @@ class QubitMenu extends BaseMenu
         // Parse module and action from path
         $path = $this->getPath(['getUrl' => true, 'resolveAlias' => true]);
         $route = $context->getRouting()->findRoute($path);
+
+        // If the route does not exist, just return true
+        if (false == $route) {
+            return true;
+        }
+
         $module = $route['parameters']['module'];
         $action = $route['parameters']['action'];
-
         // Check access from security.yml rules
         return $context->getUser()->checkModuleActionAccess($module, $action);
     }
--- a/lib/model/om/BaseStaticPage.php
+++ b/lib/model/om/BaseStaticPage.php
@@ -76,7 +76,7 @@ abstract class BaseStaticPage extends QubitObject implements ArrayAccess
 
     try
     {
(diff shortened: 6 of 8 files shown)


anvit:
Adding @sbreker in case you have any thoughts on the debug mode bypassing CSP.

anvit:
I also did want to note that we also made a change to use PropelPDO instead of DebugPDO for tests in #1965 since Propel itself stopped maintaining DebugPDO. I would try to see if updating the fixtures to use PropelPDO works for debug mode, and if it does, making that change so that we're no longer trying to maintainer that file ourselves.

lib/routing/QubitRoute.class.php
```diff
@@ -146,7 +146,7 @@ protected function generateWithTokens($parameters)
         foreach ($tokens as $token) {
             switch ($token[0]) {
                 case 'variable':
-                    if (!$optional || !isset($this->defaults[$token[3]]) || $parameters[$token[3]] != $this->defaults[$token[3]]) {
+                    if ((!$optional || !isset($this->defaults[$token[3]]) || $parameters[$token[3]] != $this->defaults[$token[3]]) && $parameters[$token[3]]) {
```
sbreker: Should we use PHP's empty() to check if $parameters[$token[3]] has anything in it at the end of this condition? Is $token[3] a string?

sbreker:
> Adding @sbreker in case you have any thoughts on the debug mode bypassing CSP.

I think this is an ok strategy in this case to get debug working. Debug mode is a special case typically only used during application development in a private environment. 

A note should be added to the AtoM docs in the CSP section, and in the 'debug mode' section that CSP is deactivated regardless of the AtoM configuration settings when debug mode is activated.

lib/routing/QubitRoute.class.php
danloveg: `$token[3]` is a string, but that part of `$parameters` might be null. I'm so used to Python handling all "falsy" values the same way, understandable that we want to be more specific here. Fixed that here: https://github.com/artefactual/atom/pull/2168/commits/230caf09ede934e129fbead77b6f36f72dd43765

null-value

danloveg:
> I also did want to note that we also made a change to use PropelPDO instead of DebugPDO for tests in #1965 since Propel itself stopped maintaining DebugPDO. I would try to see if updating the fixtures to use PropelPDO works for debug mode, and if it does, making that change so that we're no longer trying to maintainer that file ourselves.

@anvit Could you point me to where this class is set? I can't seem to figure out how to change the class that gets used in dev/debug mode. I'm running AtoM in Docker. My best guess was that it's set in `docker/bootstrap.php` here: https://github.com/artefactual/atom/blob/1d0c2d0caed8bdea3668aefa679b377efe86b60d/docker/bootstrap.php#L217 

But that's already set to `PropelPDO` so I don't know why it's using `DebugPDO`.

anvit:
> > I also did want to note that we also made a change to use PropelPDO instead of DebugPDO for tests in #1965 since Propel itself stopped maintaining DebugPDO. I would try to see if updating the fixtures to use PropelPDO works for debug mode, and if it does, making that change so that we're no longer trying to maintainer that file ourselves.
> 
> @anvit Could you point me to where this class is set? I can't seem to figure out how to change the class that gets used in dev/debug mode. I'm running AtoM in Docker. My best guess was that it's set in `docker/bootstrap.php` here:
> 
> https://github.com/artefactual/atom/blob/1d0c2d0caed8bdea3668aefa679b377efe86b60d/docker/bootstrap.php#L217
> 
> But that's already set to `PropelPDO` so I don't know why it's using `DebugPDO`.

@danloveg I would first double check that your `config/config.php` correctly lists PropelPDO instead of DebugPDO. That file is generated by bootstrap.php and isn't tracked in Github so that could potentially be it. My second guess would be `vendor/symfony/lib/plugins/sfPropelPlugin/config/skeleton/config/databases.yml` since that's the only other place I see it being referenced in our codebase but this is just from a quick code search so I'm not entirely sure.

danloveg:
> > > I also did want to note that we also made a change to use PropelPDO instead of DebugPDO for tests in #1965 since Propel itself stopped maintaining DebugPDO. I would try to see if updating the fixtures to use PropelPDO works for debug mode, and if it does, making that change so that we're no longer trying to maintainer that file ourselves.
> > 
> > 
> > @anvit Could you point me to where this class is set? I can't seem to figure out how to change the class that gets used in dev/debug mode. I'm running AtoM in Docker. My best guess was that it's set in `docker/bootstrap.php` here:
> > https://github.com/artefactual/atom/blob/1d0c2d0caed8bdea3668aefa679b377efe86b60d/docker/bootstrap.php#L217
> > 
> > But that's already set to `PropelPDO` so I don't know why it's using `DebugPDO`.
> 
> @danloveg I would first double check that your `config/config.php` correctly lists PropelPDO instead of DebugPDO. That file is generated by bootstrap.php and isn't tracked in Github so that could potentially be it. My second guess would be `vendor/symfony/lib/plugins/sfPropelPlugin/config/skeleton/config/databases.yml` since that's the only other place I see it being referenced in our codebase but this is just from a quick code search so I'm not entirely sure.

Yeah the `config/config.php` file gets re-written every time the Docker container starts so it does have `PropelPDO` listed instead of `DebugPDO`.

I did some more digging and found that `PropelPDO` *is* being used - the only reason `DebugPDO` is coming up is because it's auto-loaded here:

https://github.com/artefactual/atom/blob/7aa33449aa37094d945059d92722296d10c313c6/vendor/symfony/lib/plugins/sfPropelPlugin/lib/vendor/propel/Propel.php#L216

If you run this script while AtoM is running (`php symfony tools:run which_pdo.php`) in debug mode:

```php
<?php
// which_pdo.php

$con = Propel::getConnection();
echo get_class($con) . "\n";
```

You'll see it outputs **PropelPDO**.

So DebugPDO is not being used, but it is still loaded. In this PR I've essentially just fixed it so that it can be loaded, but the code isn't being used.

Not sure what you want to do here - there is one other place where DebugPDO is being referenced, but that's only for a constant:

https://github.com/artefactual/atom/blob/7aa33449aa37094d945059d92722296d10c313c6/vendor/symfony/lib/plugins/sfPropelPlugin/lib/debug/sfWebDebugPanelPropel.class.php#L80

anvit (approved):

anvit:
@danloveg Thanks for this fix! @melaniekung and I chatted about what to do about DebugPDO, and decided to file a new issue #2223 just to keep that as a separate issue from this. We'll remove that from Propel's autoload in a separate PR.

Thanks!

danloveg:
> @danloveg Thanks for this fix! @melaniekung and I chatted about what to do about DebugPDO, and decided to file a new issue #2223 just to keep that as a separate issue from this. We'll remove that from Propel's autoload in a separate PR.
> 
> Thanks!

That makes sense. It feels silly fixing it just for it to never run but yeah removing it is a bigger job!

PR #2173, opened 2025-09-26
Add optional sidebar block “Newest additions”

alejandroarturom:
I have been testing a small improvement for the sidebar: adding an optional block that displays the five most recent archival descriptions.

The implementation uses QubitInformationObject with a simple criteria (ORDER BY created_at DESC LIMIT 5) and reuses existing helper functions like render_title() and url_for().

The goal is to provide end users with a quick overview of the latest material available in the system, in the same way the “Popular this week” block highlights frequently accessed descriptions.

The code is lightweight, requires only one query, and can be easily extended to support caching if desired.
Screenshot 2025-09-25 at 11 25 18 PM


sarah-mason:
@alejandroarturom - thank you so much for your contribution! I've run the tests on it and it is failing PHP CS fixer. If you have a chance, could you please have a look at that? We'll take a look at your PR in our team meeting next week. Thank you again for supporting AtoM!

PR #2184, opened 2025-10-01
Fix export bulk csv

### LINKED ISSUE #2179: Problem: Bad code on new CSV Bulk import CLI
### Current Behavior

1. review the code in the loop at the https://github.com/artefactual/atom/blob/db256d11177d81000d269da625e4fec162c426a3/lib/task/export/csvExportBulkTask.class.php#L90-L106
2. the code is setting a filepath in the loop, but the file is created before entering into the loop.
3. performance of import is impacted.

### Expected Behavior

The path check should be before entering the loop.

### Possible Solution

_No response_

### Context and Notes

_No response_

### Version used

AtoM 2.10

_No response_

_No response_

### PHP version

PHP 8.3

### Contact details

_No response_

danloveg:
Closes #2179 

- Make a new CSV writer for each row so that each description gets written to a separate file.
- Fix duplicate description count (`$itemsExported` was incremented twice each loop iteration)

Previously, the same CSV writer at the same path was being used every iteration, so it was essentially working like a regular `csv:export`. These changes make it so that each description gets written to a separate file which I believe is the proper way for this task to function.

### DIFF
--- a/lib/task/export/csvExportBulkTask.class.php
+++ b/lib/task/export/csvExportBulkTask.class.php
@@ -63,20 +63,7 @@ class csvExportBulkTask extends exportBulkBaseTask
 
         echo 'Exporting as '.strtoupper($options['standard']).".\n";
 
-        // Instantiate CSV writer
-        $writer = new csvInformationObjectExport(
-            $arguments['path'],
-            $options['standard'],
-            $options['rows-per-file']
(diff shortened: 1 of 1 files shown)


anvit (approved):
This looks good! Thanks @danloveg !

PR #2188, opened 2025-10-02
Fixed typo on "port" config in bootstrap.php

stephenfrank:


anvit (approved):
Good catch! Thanks @stephenfrank !

PR #2189, opened 2025-10-02
Fix clipboard print preview

### LINKED ISSUE #2171: Problem: Clipboard Print preview not working
### Current Behavior

1. Go to 'Clipboard' > 'Go to clipboard'
2. Select 'Print preview'
3. Note that the page looks the same

### Expected Behavior

[Documentation](https://accesstomemory.org/en/docs/2.9/user-manual/access-content/clipboard/#print-preview) states: 
Clicking this option will display a printer-optimized version of the page, with all buttons removed.

danloveg:
Closes #2171 

Bootstrap 5 has a `d-print-none` class that can be used to hide elements when printing. The `print-preview.css` file wasn't being included, so I added `include_stylesheets()` in the base template.

I don't know what the original print preview looked like but I reduced the thumbnail and font sizes to fit more results per printed page.

I've also not done a full scan of possible elements that need to be hidden but I think I've got most of them.


sarah-mason:
@danloveg - thanks for this PR and the others you've been working on lately! All checks have been passing, so no feedback there. Mel's on holiday at the moment, and we're still finishing off a few things in the 2.10 release cycle, so the maintainers will start reviewing them in a week or two. But we can also discuss further at the next dev meeting if you are able to attend.

danloveg:
> @danloveg - thanks for this PR and the others you've been working on lately! All checks have been passing, so no feedback there. Mel's on holiday at the moment, and we're still finishing off a few things in the 2.10 release cycle, so the maintainers will start reviewing them in a week or two. But we can also discuss further at the next dev meeting if you are able to attend.

Sounds great, no rush at all - I figured there was a lot to do after releasing a major version of AtoM! I will be there at the dev meeting next week, definitely.

css/print-preview.css
```diff
@@ -1,2 +1,32 @@
-/* Minified CSS version of arDominionPlugin print.less */
-@media all{#update-check,header,#site-slogan,#sidebar,#context-menu,section.actions,footer,#sfWebDebug,button.clipboard,a.clipboard-print,.browse-options,#sort-header{display:none;} .container,.container .row>.span9,#main-column,#main-column .row>.span7{width:100%;margin:0;padding:0;} .row{margin-left:0;} body{padding-left:0;padding-right:0;} #content{border:none;box-shadow:none;} #coverflow{padding-top:0 !important;}#coverflow #imageflow,#coverflow #imageflow_images{height:auto !important;} #coverflow img{position:relative !important;display:inline-block !important;visibility:visible !important;margin:5px !important;left:0 !important;top:0 !important;opacity:1 !important;height:auto !important;width:auto !important;} #coverflow #imageflow_caption,#coverflow #imageflow_navigation,#coverflow .result-count{display:none;} #fullwidth-treeview-row{height:auto !important;} #content .field>h3{padding:0;margin:0;margin-top:2px;}}
+@media all {
```
anvit: I had a couple thoughts about this approach. Instead of putting the print-preview code here, since this is no longer a minified css file generated from a less file, I think having this in a sass partial (like `plugins/arDominionB5Plugin/scss/_rtl.scss`) and then including that in `plugins/arDominionB5Plugin/scss/main.scss` would be my preference.

We would no longer need the `include_stylesheets` call in the webpack.php file, and having print specific styling within `@media print` would ensure that that doesn't get applied for screens so we can skip the addStylesheet call in viewAction as well.

danloveg: That's a much cleaner solution! I did not know you could conditionally include CSS like that with Sass, that's really useful. I made that change here: https://github.com/artefactual/atom/pull/2189/commits/855f2f38c22195bd24a85272c39a8ae154eb1841

anvit (approved):
@danloveg Thanks for addressing the feedback!

PR #2194, opened 2025-10-06
Add imagick extension support

### LINKED ISSUE #2163: Enhancement: Use Imagick PHP extension instead of calling ImageMagick directly
Update AtoM to use the [Imagick PHP extension](https://github.com/Imagick/imagick). Phase out calling ImageMagick with the PHP exec() function.

Changes recommended:

- Update AtoM installation process & docs, any ansible scripts that install PHP, and Dockerfile to install https://github.com/Imagick/imagick PHP extension.
- Update AtoM to use the Imagick PHP extension in `lib/model/QubitDigitalObject.php`
- Find any other potential calls to ImageMagick: e.g. `plugins/sfThumbnailPlugin/lib/sfImageMagickAdapter.class.php`
- Provide a secure policy.xml to block dangerous and unused file formats.

There are numerous other places in AtoM where exec() is used - these should be examined to see if these calls can similarly be replaced with a call to a library/API.

Copied from [this Tier 3 card](https://gitlab.artefactual.com/artefactual/tier3/-/issues/153).

danloveg:
Closes #2163 

- Installs the imagick extension in the application image.
- Adds a new `sfImagickAdapter` that can be used with the `sfThumbnail` class.
- ~~Adds deprecation warnings to methods using `pdfinfo`, `convert`, or `identify` with `sfImageMagickAdapter`. The old adapter still works, but the new adapter is used if `imagick` is installed.~~ (Removed sfImageMagickAdapter)
- Refactored exploding multi-page assets to use `imagick` if it's installed.

Things that are not complete:

- I have not created a policy.xml file, I wanted some feedback on this PR before looking into that.
- This PR does not update any of the deployment documentation or any Ansible scripts.
- ~~Unit tests.~~ (Tests added)
- ~~Copyright block at the top of the new adapter? I added a new file to a plugin that was not developed at Artefactual, not sure what sort of copyright case that falls under. I'd guess the regular AtoM copyright block but I'm not sure whether the licenses are compatible.~~ (Added the default AtoM copyright block based on CODE_OF_CONDUCT.md)

Also of note, I did not implement the fancy thumbnail method that `sfImageMagickAdapter` does since I couldn't see any place where those custom methods were being used.


lib/model/QubitDigitalObject.php
```diff
@@ -2012,18 +2030,48 @@ public function explodeMultiPageAsset()
...
+                $imagick->setResolution(300, 300);
 
-            if (1 == $status) {
-                throw new sfException('Encountered error'.(is_array($output) && count($output) > 0 ? ': '.implode('\n'.$output) : ' ').' while running convert (ImageMagick).');
-            }
+                foreach ($imagick as $index => $page) {
+                    $page->setImageAlphaChannel(Imagick::ALPHACHANNEL_REMOVE);
+                    $page->setImageBackgroundColor('white');
+                    $page->setImageFormat('jpeg');
+                    $page->setImageCompressionQuality(100);
```
sbreker: Nitpick: These three values should possibly be local public consts like THUMB_EXTENSION

lib/model/QubitDigitalObject.php
```diff
@@ -2012,18 +2030,48 @@ public function explodeMultiPageAsset()
...
+                    $page->writeImage($filename);
+
+                    $fileList[] = $filename;
+                }
 
-            // Build an array of the exploded file names
-            for ($i = 0; $i < $pageCount; ++$i) {
-                $fileList[] = $filenameMinusExtension.sprintf('_%02d.', $i).self::THUMB_EXTENSION;
+                $imagick->clear();
+            } else {
```
sbreker: In these fallback cases, we perhaps should still call `sfImageMagickAdapter::isImageMagickAvailable`

lib/model/QubitDigitalObject.php
```diff
@@ -1929,37 +1929,53 @@ public function createRepresentations($usageId, $connection = null)
...
+        $pages = null;
+
+        if ($this->canThumbnail()) {
+            if (extension_loaded('imagick')) {
+                // Only read metadata for performance
+                $im = new Imagick();
+                $im->pingImage($filename);
+                $pages = $im->count();
+                $im->clear();
+            } elseif ('sfImageMagickAdapter' === QubitDigitalObject::getThumbnailAdapter()) {
```
sbreker: In these fallback cases, we perhaps should still call `sfImageMagickAdapter::isImageMagickAvailable`

lib/model/QubitDigitalObject.php
danloveg: Agreed, that's fixed in 628d8a58008e83b78fb03027475b5c8e884b4c87

lib/model/QubitDigitalObject.php
danloveg: Fixed: 628d8a58008e83b78fb03027475b5c8e884b4c87

lib/model/QubitDigitalObject.php
danloveg: That makes sense to me. I added constants in dc5477b6bbc5f3ccf81947e8b0c29a029f2e5d65

Dockerfile
```diff
@@ -28,12 +29,8 @@ RUN set -xe \
       xsl \
       zip \
       ldap \
-    && pecl install apcu pcov xdebug \
-    && curl -Ls https://github.com/websupport-sk/pecl-memcache/archive/refs/tags/8.2.tar.gz | tar xz -C / \
```
sbreker, line 32: This memcache change seems unrelated to the ImageMagick change and as such I think this would be best contributed as a separate pull request to avoid introducing side effects with these changes.

lib/model/QubitDigitalObject.php
```diff
@@ -1929,39 +1935,57 @@ public function createRepresentations($usageId, $connection = null)
...
+                    'Using pdfinfo shell commands is deprecated in setPageCount. Install the Imagick PHP extension for better performance and security.',
+                    E_USER_DEPRECATED,
+                );
 
-            // If processing a PDF, attempt to use pdfinfo as it's faster
-            if ('pdf' == strtolower($extension) && sfImageMagickAdapter::pdfinfoToolAvailable()) {
                 $pages = sfImageMagickAdapter::getPdfinfoPageCount($filename);
-            } else {
-                $command = 'identify '.$filename;
+            } elseif (sfImageMagickAdapter::isImageMagickAvailable()) {
```
sbreker: I've been thinking about the logic to fall back to using ImageMagick in the event that the Imagick PHP extension is not present. I think that while this may provide some convenience with respect to AtoM upgrade requirements, I think it is more important to completely remove the original way of calling ImageMagick with exec() calls and rely solely on your new work to introduce Imagick. Removing the calls to exec() entirely will avoid triggering alerts in future security scans and remove the risk associated with these calls. In this case, if we remove all dependency on sfImageMagickAdapter.class.php, then we might be able to delete sfImageMagickAdapter.class.php entirely (I have not checked if there are other locations where it is referenced - but removing it would have implications for the PDF page count methods in there).

lib/model/QubitDigitalObject.php
```diff
@@ -1929,39 +1935,57 @@ public function createRepresentations($usageId, $connection = null)
...
-        if ($this->canThumbnail() && sfImageMagickAdapter::isImageMagickAvailable()) {
-            $filename = ($this->derivativesGeneratedFromExternalMaster($this->usageId)) ? $this->getLocalPath() : $this->getAbsolutePath();
-
-            $extension = pathinfo($filename, PATHINFO_EXTENSION);
+        $filename = ($this->derivativesGeneratedFromExternalMaster($this->usageId)) ? $this->getLocalPath() : $this->getAbsolutePath();
+        $extension = pathinfo($filename, PATHINFO_EXTENSION);
+        $pages = null;
+
+        if ($this->canThumbnail()) {
+            if (extension_loaded('imagick')) {
```
sbreker: This gets checked a number of times to ensure if imagick is available. Perhaps this should be removed to be a method in sfImagickAdapter like sfImageMagickAdapter does. I think Imagick also requires ImageMagick to be installed, so perhaps a a method like `sfImagickAdapter::isImagickAvailable()` could encapsulate the logic to check that both Imagick and ImageMagick are present. Let me know what you think.

sbreker (changes requested):
Hi Daniel - thank you so much for this code contribution! I have added some additional feedback regarding the memcache change, and the logic to fall back to using ImageMagick via exec(). Please let me know if there is anything I should be considering with respect to these requests. Interested to hear your thoughts.

lib/model/QubitDigitalObject.php
danloveg: That would certainly simplify things in the code. One issue I can see with that change is that we'll want to ensure the upgrade instructions are up to date for whichever version of AtoM this ends up in since [php8.3-imagick](https://launchpad.net/ubuntu/+source/php-imagick) will be a dependency requirement ~~, and [imagemagick](https://launchpad.net/ubuntu/+source/imagemagick) will no longer be required.~~

The imagick adapter handles PDF page counts without the use of Poppler (i.e., `pdfinfo`), having spent some time with this part of the code, I think completely removing `sfImageMagickAdapter` shouldn't be an issue at all.

lib/model/QubitDigitalObject.php
danloveg: The two places I have this are in those methods that fall back to using `sfImageMagickAdapter` when the extension isn't loaded - if we completely remove the old adapter, then there'll be no need to call `extension_loaded('imagick')`.

~~And Imagick *doesn't* require ImageMagick to be installed, the ImageMagick C code is directly compiled into the imagick extension when it's built. [See here](https://github.com/Imagick/imagick/blob/14880bd7ec2042fb3272772542992b16536522b8/php_imagick_defs.h#L25), the [MagickWand C API](https://imagemagick.org/script/magick-wand.php) is included and that's how the C gets compiled into the resulting PHP extension. The only thing that's needed is the `imagemagick-dev` package for building the extension. In Ubuntu, that won't be needed since there is already a pre-built package available: [php8.3-imagick](https://launchpad.net/ubuntu/+source/php-imagick).~~

EDIT: You do still need imagemagick installed, I misunderstood how it works! The extension fails to be loaded if imagemagick is not installed.

lib/model/QubitDigitalObject.php
danloveg: I made some commits for what it would look like without `sfImageMagickAdapter`, let me know what you think.

If needed, I can revert this commit range to bring it back:

d29c313d2497084e1e4520d6773bf5f7fc5b2185..5438fc53ed0c0b88277ac3593731a1dfdbe00bd0

Dockerfile
danloveg: Fixed in https://github.com/artefactual/atom/pull/2194/commits/2179fcf830ede4adbd32d4556e7a85aa7b1cbf45 (and https://github.com/artefactual/atom/pull/2194/commits/138b8f70c4fcd97f14f2bb7cf62402e8ac6d192a)

lib/model/om/BaseDigitalObject.php
```diff
@@ -118,7 +118,7 @@ public function __get($name)
 
     try
     {
-      return call_user_func_array('QubitObject::__get', $args);
+      return parent::__get(...$args);
```
sbreker: I think this file is generated by Symfony. Was this change the result of running a `php symfony` commands?

lib/model/QubitDigitalObject.php
```diff
@@ -2012,19 +2013,34 @@ public function explodeMultiPageAsset()
 
             $filenameMinusExtension = preg_replace('/\.[a-zA-Z]{2,3}$/', '', $path);
 
-            $command = 'convert -density 300 -alpha remove -quality 100 ';
-            $command .= $path;
-            $command .= ' '.$filenameMinusExtension.'_%02d.'.self::THUMB_EXTENSION;
-            exec($command, $output, $status);
+            $imagick = new Imagick();
```
sbreker, line 2020: We should check that `extension_loaded('imagick')` is true before and issue some kind of error if it is not.

lib/model/QubitDigitalObject.php
```diff
@@ -1929,38 +1935,31 @@ public function createRepresentations($usageId, $connection = null)
...
 
-            // If processing a PDF, attempt to use pdfinfo as it's faster
-            if ('pdf' == strtolower($extension) && sfImageMagickAdapter::pdfinfoToolAvailable()) {
-                $pages = sfImageMagickAdapter::getPdfinfoPageCount($filename);
-            } else {
-                $command = 'identify '.$filename;
-                exec($command, $output, $status);
-                $pages = count($output);
-            }
+        $im = new Imagick();
```
sbreker, line 1956: We should check that `extension_loaded('imagick')` is true before and issue some kind of error if it is not.

plugins/sfThumbnailPlugin/lib/sfThumbnail.class.php
```diff
@@ -56,8 +61,10 @@ public function __construct($maxWidth = null, $maxHeight = null, $scale = true,
         if (!$adapterClass) {
             if (extension_loaded('gd')) {
                 $adapterClass = 'sfGDAdapter';
+            } elseif (extension_loaded('imagick')) {
```
sbreker: Above in QubitDigitalObject.php we preferred Imagick over 'gd' but here we prefer 'gd' which could result in strange behaviour. It might be better to try loading first using Imagick here .

See:
https://github.com/artefactual/atom/pull/2194#diff-7e1e0787b6c198eada5e76743b0dda3e2761d8d6577584b93f4f930eadad7ce3R2415-R2418

lib/model/QubitDigitalObject.php
```diff
@@ -2012,19 +2013,34 @@ public function explodeMultiPageAsset()
 
             $filenameMinusExtension = preg_replace('/\.[a-zA-Z]{2,3}$/', '', $path);
```
sbreker: Unrelated to your change but this parsing of extensions seems brittle. We should consider changing to use `pathinfo($path, PATHINFO_DIRNAME|PATHINFO_FILENAME|etc)`.

plugins/sfThumbnailPlugin/lib/sfImagickAdapter.class.php
```diff
@@ -0,0 +1,392 @@
...
+            echo $this->source->getImageBlob();
+        } else {
+            $this->source->writeImage($thumbDest);
+        }
+    }
+
+    public function freeSource()
+    {
+        if (null !== $this->source) {
+            $this->source->clear();
```
sbreker, line 270: We should perhaps also `$this->source = null` after calling `clear()` to ensure the PHP clears the memory.

lib/model/om/BaseDigitalObject.php
danloveg: I can't remember why I made that change initially, but things are still working with that change reverted.

lib/model/QubitDigitalObject.php
danloveg: This function will only attempt to instantiate an `Imagick` instance now if the imagick extension is loaded: https://github.com/artefactual/atom/pull/2194/commits/4a92bf2a8b7bdd37d10d93d6e9dda5e0d796f5e0

lib/model/QubitDigitalObject.php
danloveg: This function will only attempt to instantiate an `Imagick` instance now if the imagick extension is loaded: https://github.com/artefactual/atom/pull/2194/commits/4a92bf2a8b7bdd37d10d93d6e9dda5e0d796f5e0

plugins/sfThumbnailPlugin/lib/sfThumbnail.class.php
danloveg: Fixed in https://github.com/artefactual/atom/pull/2194/commits/d6de81e86f4cd1b99793136ffac40daf9e528bc6

plugins/sfThumbnailPlugin/lib/sfImagickAdapter.class.php
danloveg: Fixed in https://github.com/artefactual/atom/pull/2194/commits/d6de81e86f4cd1b99793136ffac40daf9e528bc6

lib/model/QubitDigitalObject.php
danloveg: I changed this line to use `pathinfo`: https://github.com/artefactual/atom/pull/2194/commits/c721411409d0f04415b602027d7b0d67f50f33be

lib/model/om/BaseDigitalObject.php
danloveg: I take that back - I remember why I made that change now. It's because if you change it back to how it was, this is how the unit test output looks:

deprecated-warnings

apps/qubit/modules/settings/templates/digitalObjectDerivativesSuccess.php
```diff
@@ -39,7 +39,7 @@
               ); ?>
             <?php } else { ?>
               <div class="alert alert-danger" role="alert">
-                <?php echo __('The pdfinfo tool is required to use this functionality. Please contact your system administrator.'); ?>
+                <?php echo __('The imagick PHP extension is required to use this functionality. Please contact your system administrator.'); ?>
```
sbreker: Nitpick: 'imagick' should be capitalised here - I think the official repo uses "Imagick".

danloveg: Fixed in a8b06105a72b719789980707f36c5bea55cd3311

plugins/sfThumbnailPlugin/lib/sfImagickAdapter.class.php
```diff
@@ -0,0 +1,393 @@
...
+ *
+ * sfImagickAdapter converts images to thumbnails using the Imagick library.
+ *
+ * @see https://github.com/Imagick/imagick
+ *
+ * @author Daniel Lovegrove <d.lovegrove11@gmail.com>
+ */
+class sfImagickAdapter
+{
+    protected const string PDF_EXTRACT_BACKGROUND = 'white';
```
anvit: I would suggest omitting the type here since this would break in PHP 8.1 and PHP 8.2, which our documentation still lists as being supported in AtoM 2.10.x

lib/model/QubitDigitalObject.php
```diff
@@ -30,6 +30,16 @@ class QubitDigitalObject extends BaseDigitalObject
     public const THUMB_MIME_TYPE = 'image/jpeg';
     public const THUMB_EXTENSION = 'jpg';
 
+    // Constants for exploding multi-page assets
+    public const string MULTI_PAGE_ASSET_EXTRACT_BACKGROUND = 'white';
+    public const string MULTI_PAGE_ASSET_EXTRACT_FORMAT = 'jpeg';
+    public const int MULTI_PAGE_ASSET_EXTRACT_QUALITY = 100;
+    public const int MULTI_PAGE_ASSET_EXTRACT_RESOLUTION = 300;  // This controls the DPI
```
anvit: We should omit the types here as well since this would break in PHP 8.1 and PHP 8.2.

anvit (changes requested):

plugins/sfThumbnailPlugin/lib/sfImagickAdapter.class.php
danloveg: Good catch, I removed the types in https://github.com/artefactual/atom/pull/2194/commits/a36230a0c5569d080b558bf30c1ea4d174b41824

lib/model/QubitDigitalObject.php
danloveg: Removed in https://github.com/artefactual/atom/pull/2194/commits/a36230a0c5569d080b558bf30c1ea4d174b41824

anvit (approved):

sbreker (approved):
Hi @danloveg - Thanks for these changes! I think it would be best to merge this now, and the work to add a policy file can be added in a subsequent commit.

danloveg:
> Hi @danloveg - Thanks for these changes! I think it would be best to commit this now, and the work to add a policy file can be added in a subsequent commit.

Sounds great, glad to see this will be merged in soon!

PR #2195, opened 2025-10-07
Dependabot fixes: change package versions

anvit:
Updating `mediaelement` version based on dependabot alert.


melaniekung (approved):
thanks @anvit

PR #2197, opened 2025-10-12
Add german translation for table sorting

SergeoLacruz:


sarah-mason:
Thanks for your contribution @SergeoLacruz! All checks have passed, so there's nothing needed from you at the moment. We have an AtoM maintainer on holiday at the moment and we're finishing up a few things from the 2.10 release cycle, but the maintainers will get to reviewing the PRs in a week or two. Please let me know if you have any questions in the meantime. Again, thank you for your contribution!

anvit:
Hi @SergeoLacruz ! For translations related changes, we accept them through Weblate not GitHub, and these XML files always get always overwritten during our process when we pull translations from Weblate. You can read more about contributing translations to weblate [here](https://wiki.accesstomemory.org/Resources/Translation). 

I did notice that there were German translations that were already present for all of these terms but they didn't line up with the ones you'd submitted, and they're all waiting on review. I would recommend going and changing or reviewing those, and they'll make their way into our next release when we pull the translations for it.

SergeoLacruz:
Thanks for the hint. I will look into this. 
What about the files in atom/lib/task/pdf used for pdf generation. Are these also coming with Weblate?

anvit:
@SergeoLacruz those XSLT stylesheets for PDF generation don't have an i18n process unfortunately. They do not automatically switch with AtoM's language and they aren't tracked in Weblate, however people can edit/swap those with local versions if they would like, and we have community contributed [French language](https://wiki.accesstomemory.org/wiki/Community/Community_resources/Customization#French_XSLTs_for_finding_aid_generation) ones in our wiki.

We're happy to link to German language ones too if you'd like to contribute those! Please send those to contributors@artefactual.com and we'll add them to the wiki.

SergeoLacruz:
I just created a simple sed script that translates the Englich text to German. 
https://github.com/SergeoLacruz/atom-pdf-german-tanslator
Not the best solution but OK for me right now. It might have errors. I still need to tune it.
May be it is useful for others too.

sarah-mason:
Thanks for sharing with the community @SergeoLacruz! I will add it to our [community development resources page](https://wiki.accesstomemory.org/wiki/Community/Community_resources/Development) on the wiki and I recommend you share it on the [user forum](https://groups.google.com/u/1/g/ica-atom-users) as well.

PR #2201, opened 2025-10-16
Remove BS2, jQuery, and uppy from vendor/

### LINKED ISSUE #2182: Remove vendor code already being included via package.json
I noticed a few different libraries in `vendor/` that are already being included in the app by NPM packages in `package.json`.

This includes:

- `vendor/bootstrap`. This is Boostrap 2 code, and is not needed any more. Bootstrap 5 is included via `node_modules/`.
- `vendor/jquery*`. jQuery is included via `node_modules/`
- `vendor/jstree`. This is also being included via `node_modules/`
- `vendor/uppy`. Same as here.

I've experimented with removing all of these (and removing their references from `apps/config/qubit/view.yml`) and everything still works since Webpack isn't using the CSS or JS from any of those directories.

My question is, are these still being included for any reason (e.g., supporting legacy BS2 themes)? If there's no reason why they're being kept around, I can open a pull request to get rid of them.

Additionally, I noticed that the only `.less` files in this repository are those in `vendor/bootstrap`, so if those are removed, it should also be possible to remove the `npm install less@<4.0.0` in the Dockerfile.

danloveg:
Closes #2182 

Removes the code mentioned in the issue except for `jstree` which is still referenced in `apps/qubit/modules/informationobject/actions/indexAction.class.php`.

There is likely still some vendor code that could be removed but these were the lowest hanging fruit to remove.


anvit (approved):
I looked through the codebase and all of these seem to be safe to remove! Thanks s much @danloveg !

PR #2204, opened 2025-10-21
Allow YYYY for Acquisition date. (#2157)

### LINKED ISSUE #2157: Enhancement: Expand Acquisition date in accession and deaccession form to include YYYY only or free-form entries
Currently the Acquisition date in the accession form requires a YYYY-MM-DD entry. If only a year is entered, the change will not save and generates a ""YYYY" does not match the date format (YYYY-MM-DD)" message.

Use case: Legacy acquisitions do not always include a full acquisition date. In other cases, legacy records for which little administrative information is available are accessioned in the present in which a guesstimate year is used. 

One work around we've hit on is using January 1 as a stand in date (YYYY-01-01) for instances where we only know a year. 

Running version: AtoM 2.9.1

melaniekung:


plugins/qtAccessionPlugin/modules/accession/actions/editAction.class.php
```diff
@@ -226,17 +226,35 @@ protected function addField($name)
                 break;
 
             case 'date':
-                $this->form->setDefault('date', Qubit::renderDate($this->resource['date']));
-
                 if (!isset($this->resource->id)) {
                     $dt = new DateTime();
                     $this->form->setDefault('date', $dt->format('Y-m-d'));
+                } else {
```
anvit: I'm not sure if this change is required. Wouldn't it be the same in either scenario?

plugins/qtAccessionPlugin/modules/accession/templates/editSuccess.php
```diff
@@ -52,7 +52,7 @@
                     ->help(__('Accession date represents the date of receipt of the materials and is added during the donation process.'))
                     ->label(__('Acquisition date').' <span class="form-required" title="'.__('This field is marked as mandatory in the relevant descriptive standard.').'">*</span><span class="visually-hidden">'.__('This field is marked as mandatory in the relevant descriptive standard.').'</span>'),
                 null,
-                ['type' => 'date']
+                ['type' => 'string']
```
anvit, line 55: Should this also apply for deaccession (`plugins/arDominionB5Plugin/modules/deaccession/templates/editSuccess.php`), and also if YYYY is permitted for accession, would it make sense to have the same in deaccession?

anvit (approved):
Thanks @melaniekung

PR #2004, opened 2025-04-22
Allow searching accessions by linked terms

danloveg:
Closes #1986 

This pull request adds the four term types linked to accession records to the Elasticsearch mapping for accessions:

- Acquisition type
- Resource type
- Processing status
- Processing priority

I've also added these terms to the searchable criteria when searching for accession records.

PR #2207, opened 2025-10-23
Add facet filters to accession browse page

### LINKED ISSUE #2159: Enhancement: Introduction of facets to Browse accessions display
Facets based on the default accession fields (e.g. Date, Acquisition type, Resource type, Processing status, etc.) would allow for improved functionality and utility of accession management in AtoM. 

Use case(s): 

- Institutional archives that have received, and will continue to receive, record transfers from administrative units and want to view a list of all accessions received from a unit over time.
- Private archives that have received multiple donations from the same donor over time but not necessarily to the same fonds or collection. Facets would allow for seeing a list of all accessions, regardless of where the records ended up, associated with a particular donor. 
- Archive of any type wanting to view all de-accessioned accessions or all donations/transfers of a particular type at once.  

Currently a keyword search needs to be used to attempt to pull together a list by title, provided titles have been consistently applied or the unit name has not changed over the course of several decades. 

Running version: AtoM 2.9.1

danloveg:
Closes #2159

Adds filters on the accession browse page for the following accession fields:

- Acquisition type
- Resource type
- Processing status
- Processing priority
- Donor name
- Creator name

The first four filters are made possible by #2004, since facet filters use the elastic search index to filter the query.

I was able to remove the query setup from the `AccessionBrowseAction::execute` function by inheriting from `DefaultBrowseAction` which already does all of the same query setup.

These changes will also make it easier to implement an advanced search form for accessions in the future if that's a desired change. I'm passing an empty list to the `addAdvancedSearchFilters` function for now.

accession-facets


danloveg:
I also fixed the **Export CSV** button so that it works by exporting only the filtered accessions.

sromkey:
Hey @danloveg , thanks for this! We want to give it a close look over since it changes how accession searching works, but will definitely consider for 2.11.

danloveg:
> Hey @danloveg , thanks for this! We want to give it a close look over since it changes how accession searching works, but will definitely consider for 2.11.

Sounds good, thanks for the update

lib/job/arAccessionExportJob.class.php
```diff
@@ -38,24 +38,24 @@ class arAccessionExportJob extends arExportJob
...
+        } else {
+            $query->addAggFilters(AccessionBrowseAction::$AGGS, $parameters['params']);
+            $query->addAdvancedSearchFilters([], $parameters['params'], 'accession');
         }
 
         return QubitSearch::getInstance()
             ->index
             ->getIndex('QubitAccession')
-            ->createSearch($query->getQuery(false, false));
+            ->createSearch($query->getQuery(false));
```
anvit: This shouldn't make a difference functionally speaking since getQuery should by default be filtering drafts, even if `false` wasn't passed, but I'm curious if there was a reason this param was removed?

plugins/qtAccessionPlugin/modules/accession/actions/browseAction.class.php
```diff
@@ -17,133 +17,101 @@
...
-                'resourceType.i18n.%s.name' => 1,
-                'alternativeIdentifiers.i18n.%s.name' => 1,
-                'creators.i18n.%s.authorizedFormOfName' => 1,
-                'alternativeIdentifiers.i18n.%s.note' => 1,
-                'alternativeIdentifiers.type.i18n.%s.name' => 1,
-                'accessionEvents.i18n.%s.agent' => 1,
-                'accessionEvents.type.i18n.%s.name' => 1,
-                'accessionEvents.notes.i18n.%s.content' => 1,
-                'donors.contactInformations.contactPerson' => 1,
-                'accessionEvents.dateString' => 1,
```
anvit, line 115: I didn't see these weights being set anywhere and I don't think they'd be inherited from the browse action either. Doesn't removing them impact the search result order?

danloveg: Thanks for pointing that out, I didn't consider that.

I found that `addAdvancedSearchFilters()` makes a call to the `setBoostValues()` function, so I added boost values for the accession fields that were being boosted previously there. See here: https://github.com/artefactual/atom/pull/2207/commits/e372ba44521030c477ff588f4be6670764b91178

The ordering should be the same now that the fields are boosted properly.

Screenshot 2026-02-18 133658

lib/job/arAccessionExportJob.class.php
danloveg: I found it wasn't even necessary to call `getQuery()` since `addAdvancedSearchFilters()` creates a BoolQuery already. I'm using that here now: https://github.com/artefactual/atom/pull/2207/commits/e372ba44521030c477ff588f4be6670764b91178

anvit (approved):
Thanks @danloveg !

PR #2191, opened 2025-10-03
Add object details to physical storage page and report

danloveg:
Closes #2186 

- Duplicates the "related description" logic when browsing an information object for information objects linked to physical storage (added identifier, level of description, and an optional (Draft) part)
- Adds accession title to accessions linked to physical storage
- Adds the description's identifier to the physical storage report title

Physical storage page before:

phys_storage_before

Physical storage page now:

phys_storage_after

Physical storage report before:

phys_report_before

Physical storage report now:

phys_report_after

PR #2209, opened 2025-10-29
Add ability to add to clipboard from physical storage page

### LINKED ISSUE #2205: Enhancement: Add archival descriptions to clipboard from a Physical Storage view
Because in archival practice the physical order of material does not always match the intellectual/hierarchical it would be helpful to be able to view a physical container record (e.g. a Box) and add all of the content to the clipboard. This would be helpful for physical management processes such as digitization, conservation, etc.

A related request would be to add the results of a search to the clipboard without the need to click each search result (this has come up internally but is not yet in a public issue akaik)

danloveg:
Closes #2205

Adds a **More ^** menu to the physical storage page with two new options:

- "Add archival descriptions to clipboard"
- "Add accessions to clipboard"

If there are archival descriptions, you can click the former button to add them all to your clipboard. Similarly, if there are accessions, you can click the latter button to add them all to your clipboard.

The menu won't show up if there are no archival descriptions or accessions.

I created a new way to get records onto the clipboard, this could be useful for adding to the clipboard in other ways, too. See: `bulkAddItems()` in `clipboard.js`.

This PR will likely conflict with #2191 since I've edited the `indexSuccess.php` for the physical storage in that PR too.

phys-storage-add-to-clipboard


js/clipboard.js
```diff
@@ -362,6 +362,80 @@ import Tooltip from "bootstrap/js/dist/tooltip";
...
+     * Add multiple new items to the clipboard of a single given type
+     * @param {Array} slug The slug of the item to add
+     * @param {String} type The type of item the slug is for
+     */
+    bulkAddItems(
+      slugs,
+      type,
+      singleAddedMessage = "Added 1 item to the clipboard",
+      pluralAddedMessage = "Added %1% items to the clipboard",
+      alreadyAddedMessage = "All items are already on the clipboard"
```
anvit: We should avoid text messages in js since those don't get picked up by weblate for translation. Since we're already getting these from the php file, I think it's safe to remove the defaults here

danloveg: Sounds good, I removed those defaults

anvit (approved):
Thanks @danloveg !

PR #2213, opened 2025-11-06
Add local only option to extract-text task

### LINKED ISSUE #2214: Enhancement: Add a local only option to the text extraction PDF task
Having a local only option for the CLI extract-text task would be helpful for scenarios where we only want to run it for local files when one of or more remote PDF sources are known to be offline or broken currently.

anvit:
Add option to allow running extract-text task only for local pdfs


melaniekung (approved):
LGTM, thanks!

PR #2218, opened 2025-11-11
Fix missing divider in RTL layout (#2217)

### LINKED ISSUE #2217: Problem: Missing divider in RTL layout
### Current Behavior

1. Switch to a language that reads RTL.
2. Open any description.
3. Notice that there is no divider visible to the left of the title on each row. It is hard to notice but there is a divider on the right side which makes the border look thicker than usual.

### Expected Behavior

There should be a border present to the left of the title in an RTL layout.

### Possible Solution

_No response_

### Context and Notes

_No response_

### Version used

AtoM 2.10.0

_No response_

_No response_

### PHP version

_No response_

### Contact details

_No response_

anvit:
Add a border-left and set border-right to 0 for RTL layout

### DIFF
--- a/plugins/arDominionB5Plugin/scss/_rtl.scss
+++ b/plugins/arDominionB5Plugin/scss/_rtl.scss
@@ -191,4 +191,10 @@ html[dir="rtl"] {
   .alert-info pre {
     direction: ltr;
   }
+
+  .field .border-end {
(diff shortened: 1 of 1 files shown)


melaniekung (approved):
LGTM, thanks!

PR #2220, opened 2025-11-12
Enforce CSP by default, not report-only. (#2219)

### LINKED ISSUE #2219: AtoM CSP headers should not be 'report-only' by default
AtoM's CSP header feature currently defaults to 'Content-Security-Policy-Report-Only' which does not actually enforce these headers. This value should be changed to 'Content-Security-Policy' so that it is enforcing these policies by default for new installs.

melaniekung:


anvit (approved):
Looks good!

PR #2225, opened 2025-11-14
Remove DebugPDO and related

### LINKED ISSUE #2223: Remove DebugPDO from AtoM
DebugPDO is still being loaded by Propel https://github.com/artefactual/atom/blob/7aa33449aa37094d945059d92722296d10c313c6/vendor/symfony/lib/plugins/sfPropelPlugin/lib/vendor/propel/Propel.php#L216

Since this is no longer maintained, it should not be loaded by AtoM

danloveg:
Closes #2223 

Removes the following:

- `DebugPDO`
- `DebugPDOStatement`
- `MssqlDebugPDO` (a sub-class of DebugPDO)
- `sfWebDebugPanelPropel`
  - This is a panel for the debug bar that is intended to show SQL logs. DebugPDO was how those SQL logs were created, but since DebugPDO is now removed, there's no way for SQL queries to be logged and this panel will never appear. I removed this file because of this.


anvit (approved):
Thanks @danloveg

PR #2231, opened 2025-11-18
CSV import culture-agnostic matching

sbreker:
- Make getByTitleIdentifierAndRepo culture-agnostic:
  - Title: replace culture-bound `INNER JOIN` with `EXISTS on` information_object_i18n (no culture filter).
  - Repo name: match via actor_i18n without culture filter.
- Anchor recursive CTE to the candidate IO (io.id = :id) to find the nearest repository instead of generating hierarchy for all rows.
- Remove dependency on user/session culture (no sfContext user reads).
- Drop repository table join since actor_i18n suffices for name matching, and `effective_repo_id` originates from information_object.repository_id which guarantees the actor refers to a repository.
- Preserve behavior when repoName is null: return the first candidate by identifier + title
- Add PHPUnit tests covering missing-translation scenarios:
  - Title only in source/default culture.
  - Repo name only in source/default culture.
  - Inherited repo where the name exists only in source culture.

### DIFF
--- a/test/phpunit/QubitInformationObjectTest.php
+++ b/test/phpunit/QubitInformationObjectTest.php
@@ -53,13 +53,15 @@ class QubitInformationObjectTest extends TransactionTestCase
 
         if (null === $expectedTitle) {
             // No match expected.
-            $this->assertNull($result, 'Expected null result when no matching record exists.');
+            $this->assertIsArray($result);
+            $this->assertCount(0, $result, 'Expected empty result when no matching record exists.');
         } else {
             // A match is expected.
-            $this->assertNotNull($result, 'Expected a valid integer id when data should match.');
-            $this->assertIsInt($result, 'Expected the returned id to be an integer.');
+            $this->assertIsArray($result);
+            $this->assertCount(1, $result, 'Expected exactly one matching id.');
+            $this->assertIsInt($result[0], 'Expected the returned id to be an integer.');
 
-            $resultIo = QubitInformationObject::getById($result);
+            $resultIo = QubitInformationObject::getById($result[0]);
             $this->assertNotNull($resultIo, 'Expected a valid information object.');
             $this->assertEquals($expectedTitle.$randomString, $resultIo->title, 'The information object title does not match expected.');
         }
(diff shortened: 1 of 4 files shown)


test/phpunit/QubitInformationObjectTest.php
```diff
@@ -287,4 +287,157 @@ public function testGetByTitleIdentifierAndRepoWithNearestParentRepository()
...
+
+        $this->assertNotNull($result, 'Expected a match when repo is inherited and repo name only exists in source culture.');
+        $this->assertIsInt($result);
+
+        // Restore culture
+        sfPropel::setDefaultCulture($prevCulture);
+        if (sfContext::hasInstance()) {
+            sfContext::getInstance()->getUser()->setCulture($prevCulture);
+        }
+    }
```
anvit, line 451: I think it would be good to have a test cases for 3-4 levels of inheritance too.since Series > Sub-series > Fond > Item with repo set only on Series would not be an uncommon scenario. 

I would also consider adding test cases for no matches present for identifier, title, and/or repo.

test/phpunit/QubitInformationObjectTest.php
```diff
@@ -285,6 +290,163 @@ public function testGetByTitleIdentifierAndRepoWithNearestParentRepository()
             'TestGrandparentRepository'.$randomString
         );
 
-        $this->assertNull($resultWithWrongRepo, 'Expected null when searching with grandparent repository since nearest parent has different repository.');
+        $this->assertIsArray($resultWithWrongRepo, 'Expected an array of ids.');
+        $this->assertCount(0, $resultWithWrongRepo, 'Expected empty result when searching with grandparent repository since nearest parent has different repository.');
+    }
+
```
danloveg, line 296: Another test for when the number of candidates is > 1 would be useful since that's a new type of structure that `getByTitleIdentifierAndRepo` can return. For example, if you had two distinct information objects with the same title + identifier linking to the same repository, I'd expect the number of candidates to be 2.

test/phpunit/QubitInformationObjectTest.php
sbreker: Hi Anvit! I think the test "testGetByTitleIdentifierAndRepoWithNearestParentRepository()" is covering tracking the repo up through 3 levels of relationships.

I have cases covering 'no match' in:
- QubitInformationObjectTest::testGetByTitleIdentifierAndRepo
- QubitInformationObjectTest::testGetByTitleIdentifierAndRepoInherit
- QubitInformationObjectTest::testGetByTitleIdentifierAndRepoWithNearestParentRepository

test/phpunit/QubitInformationObjectTest.php
sbreker: Good idea! I have added two new tests:
- testGetByTitleIdentifierAndRepoMatchesWhenTitleOnlyInSourceCulture
- testGetByTitleIdentifierAndRepoMatchesWhenRepoNameOnlyInSourceCulture

sbreker:
Hi @anvit and @danloveg - thank you for your feedback so far! I have taken the WIP label off this PR now.

test/phpunit/QubitInformationObjectTest.php
anvit: Hi @sbreker ! I think `testGetByTitleIdentifierAndRepoWithNearestParentRepository` technically only tracks 2 levels of inheritance, and I still think it would be good to add another test case which checks `Grandparent (repo A) -> Parent (no repo) -> Child (no repo)` also has the child description being matched for repo A.

sbreker: Sorry @anvit - I pasted the wrong test - does this one match what you are asking: `testGetByTitleIdentifierAndRepoWithMultiLevelInheritance`

anvit: Yes that does! Thanks!

anvit (approved):
Thanks @sbreker !

PR #2233, opened 2025-11-19
Fix csp violation breaking “back to previous page” link

### LINKED ISSUE #2232: Problem: CSP Violation: Previous Page Link Not Functioning
### Current Behavior

1.Navigate to any invalid route For instance [http://localhost:63001/invalidwebsite](url)
2. Click on the` Back to previous page` link displayed on the error page.
Image
4. Observe that nothing happens; the previous page does not load.
5. Open the browser’s developer console and note the CSP (Content Security Policy) error indicating that the inline JavaScript used by this link has been blocked.
Image

### Expected Behavior

When a user clicks the `Back to previous page` link, the application should successfully navigate to the last visited page. The link should function consistently across all browsers without triggering any CSP violations, using an approved and CSP-compliant JavaScript implementation.

### Possible Solution

Move the inline JavaScript used by the `Back to previous page` link into an external JavaScript file. This will ensure the functionality complies with the site’s Content Security Policy, prevent CSP blocking, and restore proper back navigation behavior.

### Context and Notes

This issue prevents the Back to previous page link from functioning, which impacts navigation and interrupts the user experience on error pages. My goal is to ensure the link works reliably without triggering CSP violations.

To diagnose the problem, I reviewed the browser console and identified that the inline JavaScript was being blocked by the existing Content Security Policy. 

I tested a fix by moving the inline JavaScript into an external file, and this removed the CSP warnings and restored the expected behavior. I will be creating a pull request with this solution.

### Version used

AtoM 2.10

macOS Sonoma 14.5

en

### PHP version

PHP 8.3.1

### Contact details

Ash.Singhal@umanitoba.ca

Eashana0104:
Closes https://github.com/artefactual/atom/issues/2232

Changes Made -

- Extracted inline back-navigation JS into an external file for CSP compliance.
- Added a new back-navigation handler with proper history and fallback behavior.
- Updated all templates that previously relied on inline JavaScript for the Back to previous page link to use the new external script.

Clicking on `Back to previous page`
Screenshot 2025-11-19 at 11 50 18 AM

No more CSP compliance errors on clicking `Back to previous page`
Screenshot 2025-11-19 at 2 09 41 PM

Leads back to the previous page in history.
Screenshot 2025-11-19 at 2 11 25 PM


sarah-mason:
Hi @Eashana0104! Thank you so much for your contribution and congratulations on your first PR to AtoM! I'm know you are getting your development environment set up still, but the Prettier test failed. Everything else passed!

apps/qubit/modules/admin/templates/error404Success.php
```diff
@@ -12,9 +12,7 @@
       </p>
 
       <p class="mb-0">
-        <a href="javascript:history.go(-1)">
-          <?php echo __('Back to previous page.'); ?>
-        </a><br>
+        <a href="#" class="js-back-link"  data-fallback-url="<?php echo url_for('@homepage'); ?>"><?php echo __('Back to previous page'); ?></a><br>
```
melaniekung: We should use a data-action like `data-action="back"` instead of a class to call the js function for better readability. (Same for all the other php classes)

js/historyNavigation.js
```diff
@@ -0,0 +1,26 @@
+(($) => {
+  "use strict";
+
+  $(() => {
+    historyNavigation();
+  });
+
+  function historyNavigation() {
+    document.querySelectorAll('.js-back-link').forEach(function (el) {
```
melaniekung: Using a data-action will require updating the querySelectorAll here, but should work with the rest of the function as is.

plugins/arDominionB5Plugin/js/main.js
```diff
@@ -39,3 +39,4 @@ import "/js/fullWidthTreeView";
 import "/js/deletePhysicalStorage";
 import "/js/settingsFindingAid";
 import "/js/refreshJobs";
+import "/js/historyNavigation";
```
melaniekung: Daniel has mentioned to us that you are having troubles setting up prettier checks so I thought it may be useful to add this - I believe the Prettier test may be failing due to the missing empty line here.

melaniekung (changes requested):
Hi @Eashana0104 , thanks for the contribution! Just added some small feedback, but otherwise looks great!

apps/qubit/modules/admin/templates/error404Success.php
Eashana0104: This does increase the readability. I have resolved this in my commits and formatted the closing tags properly to increase readability further!

js/historyNavigation.js
Eashana0104: This makes sense in accordance to the change on selector. Resolved by https://github.com/artefactual/atom/pull/2233/commits/52deaf9106a70f8c0480c1113be5ee0a650f597e

plugins/arDominionB5Plugin/js/main.js
Eashana0104: The Prettier check was failing due to some formatting requirement failing in `historyNavigation.js`. i have fixed that using the command  `prettier --write js/historyNavigation.js`. Now all the .js files are formatted according to Prettier rules. Resolved by https://github.com/artefactual/atom/pull/2233/commits/5f7db106db0b7f8c0bded2fa4ee43a58479003f7

melaniekung (approved):
Thanks for the updates @Eashana0104 , this looks good to me! Congrats on completing your first PR to AtoM!

js/historyNavigation.js
```diff
@@ -0,0 +1,25 @@
+(($) => {
+  "use strict";
+
+  $(() => {
+    historyNavigation();
```
anvit: I would suggest putting the body of the entire function here instead of a single named function call

js/historyNavigation.js
```diff
@@ -0,0 +1,25 @@
...
+  });
+
+  function historyNavigation() {
+    document.querySelectorAll('[data-action="back"]').forEach(function (el) {
+      el.addEventListener("click", function (e) {
+        e.preventDefault();
+
+        if (window.history.length > 1) {
+          window.history.back();
+        } else {
```
anvit: It would be cleaner to just `return;` after the `window.history.back();` and skip the else entirely.

anvit (changes requested):
Added a couple small comments

js/historyNavigation.js
Eashana0104: Resolved by https://github.com/artefactual/atom/pull/2233/commits/b84114c6a70f7fc61a67de3fba5056a8375ac5c6

js/historyNavigation.js
Eashana0104: Resolved by https://github.com/artefactual/atom/pull/2233/commits/04767247d038df0bbfeafb8d3d1b951a0a9f7617

anvit (approved):
Thanks @Eashana0104 ! This looks good!

PR #2238, opened 2025-11-24
Add max_result_window setting to ES config (#2237)

### LINKED ISSUE #2237: Problem: Fullwidth treeview doesn't load when `treeview_items_per_page_max` is increased
### Current Behavior

1. As per [AtoM documentations](https://accesstomemory.org/en/docs/2.10/user-manual/administer/settings/#fullwidth-treeview-settings:~:text=treeview_items_per_page_max%3A%2010000), for sites with very large hierarchies that wants to set the initial page load value higher than 10,000 records, a system administrator can modify this max value in the configuration file found at `config/app.yml`. Look for the following line in the file:
`treeview_items_per_page_max: 10000`
2. Go to Admin > Settings > Treeview, under Full-width, increase the Items per page to match the value updated in `config/app.yml`
3. Clear application cache, populate index, and restart PHP-fpm
4. Go to a description
5. Note that the treeview will not load, irregardless of the number of items in the treeview

### Expected Behavior

Treeview should load

### Version used

AtoM 2.10

melaniekung:
Allow users to increase index.max_result_window to support large fullwidth treeview hierarchies. This addresses an issue where treeview fails to load when treeview_items_per_page_max is set above ES's default 10000 result limit.

### DIFF
--- a/plugins/arElasticSearchPlugin/config/search.yml
+++ b/plugins/arElasticSearchPlugin/config/search.yml
@@ -31,6 +31,7 @@ all:
         number_of_shards: 4
         number_of_replicas: 1
         mapping.total_fields.limit: 3000
+        max_result_window: 10000
 
         analysis:
 


anvit (approved):
Looks good!

PR #2242, opened 2025-11-25
Fix CSP violation with inline JS (#2239)

### LINKED ISSUE #2239: Problem: Refresh page links violate CSP
### Current Behavior

1. **As a non-authenticated user**, attach records to your clipboard.
2. Export them as a CSV.
3. In the notification that appears when you click Export, click the link that says "Refresh the page"
4. Notice that nothing happens.
5. Open the developer console in your browser and notice the error message.

Image

### Expected Behavior

Clicking the link should refresh the page.

### Possible Solution

Like in #2233, the inline JS should be removed, a `data-action="refresh"` attribute should be added on these links, and non-inline Javascript should be used to query for elements with that attribute. When elements matching that query are clicked, then reload the window.

There is a file that can do this with non-inline Javascript, see `/js/refreshJobs.js`. I think this file could be made more generic to provide window refreshing on any element that has `data-action="refresh"`.

```javascript
document.querySelectorAll('[data-action="refresh"]').forEach(function (el) {
  el.addEventListener("click", function (e) {
    e.preventDefault();
    window.location.reload();
  }
});
```

It would make sense to rename this file from `refreshJobs.js` to `windowReload.js`.

### Context and Notes

This issue is similar to #2232.

### Version used

AtoM 2.10 (latest qa/2.x)

Docker (`php:8.3-fpm-alpine`)

en

### PHP version

PHP 8.3

### Contact details

_No response_

melaniekung:


js/windowReload.js
```diff
@@ -0,0 +1,20 @@
+(($) => {
+  "use strict";
+
+  $(() => {
+    refresh();
```
anvit: I would suggest remove this single function call and just put the body of the function itself here, like in the previous implementation in refreshJobs

js/windowReload.js
```diff
@@ -0,0 +1,20 @@
...
+  });
+
+  function refresh() {
+    // Use event delegation on document to catch dynamically added elements
+    document.addEventListener("click", function (e) {
```
anvit: I would not suggest doing a blanket listener for clicks on the entire document. We could either use a jQuery solution similar to what we had before (something like `$('[data-action="refresh"]');`) or if we were sticking to native js, using a selector like `document.querySelectorAll('[data-action="refresh"]')` and only listening for clicks on those.

anvit (changes requested):
Added a couple of small flags

js/windowReload.js
melaniekung: `querySelectorAll` doesn't work when we define html elements using `$this->context->i18n` in the Action classes, so using jQuery.

js/windowReload.js
```diff
@@ -0,0 +1,10 @@
+(($) => {
+  "use strict";
+
+  $(() => {
+    $(document).on("click", '[data-action="refresh"]', function (e) {
```
anvit: This still puts the selector on all of the document, and then filters it. It would be better to use `$('[data-action="refresh"]).on("click", function …`

anvit (approved):

PR #2244, opened 2025-11-25
Trim spaces for CSV lang/script imports (#2243)

### LINKED ISSUE #2243: Problem: CSV import not trimming spaces for language, languageOfDescription, script, scriptOfDescription, eventStartDates and eventEndDates fields
### Current Behavior

1. Import CSV where language/script fields have multiple entries separated by `|` have spaces (ex. `en | fr` or `Arab | Cyrl`)
2. Import fails as values with spaces are not valid

### Expected Behavior

CSV import should remove spaces for these fields

### Context and Notes

CSV validator checks these fields, and they pass validation since we're trimming these fields during validation but not during import.

### Version used

AtoM 2.10

melaniekung:
for language, script, languageOfDescription and scriptOfDescription fields

### DIFF
--- a/lib/QubitFlatfileImport.class.php
+++ b/lib/QubitFlatfileImport.class.php
@@ -1672,12 +1672,14 @@ class QubitFlatfileImport
 
             // Add language properties
             if (isset($self->languageMap, $self->languageMap[$columnName]) && $value) {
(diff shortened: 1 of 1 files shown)


anvit (approved):
Looks good!

PR #2249, opened 2025-11-27
Clarify public AD generation settings (#2248)

### LINKED ISSUE #2248: Unclear setting for `Generate archival description reports as public user`
Following [Unclear Finding aid setting](https://github.com/artefactual/atom/issues/2145), `Generate archival description as public user` setting should also be updated for clarification. When set to 'yes' generated archival descriptions will exclude Drafts.

melaniekung:
Related to [Unclear Finding aid setting](https://github.com/artefactual/atom/issues/2145)


anvit (approved):
Looks good

PR #2254, opened 2025-12-03
Add env option for activating the OIDC plugin

sbreker:
Adds the ability to activate the OIDC plugin using an environment variable. Setting the env var 'ATOM_ACTIVATE_OIDC_PLUGIN' to "true" before starting AtoM will activate the OIDC plugin.


anvit (approved):
Looks good! Thanks @sbreker

PR #2255, opened 2025-12-05
Fix accession type option in clipboard export

### LINKED ISSUE #2221: Problem: Clipboard export page cannot change type to Accessions
### Current Behavior

1. Log in as a user with permissions to export Accessions
2. Go to any record and add to clipboard
3. Go to Clipboard > Go to clipboard
4. Change entity type to whichever type you've added record to clipboard
5. Click Export
6. In this Clipboard export page, regardless which Type you're currently on, attempting to change Type to Accession, it redirects you to Archival Descriptions, even if you have Accessions on your clipboard.

### Expected Behavior

Selecting Type Accession should allow you to export Accessions on your clipboard and not redirect you to the Archival Descriptions export page.

### Context and Notes

Note: Accessions export still works by selecting Accession in for Entity type from the Go to clipboard page

### Version used

AtoM 2.10

### LINKED ISSUE #2222: Problem: Clipboard export type Accession option is visible to all users
### Current Behavior

1. As a public user or any user with no permissions to view Accessions, add any type of record to clipboard
2. Go to Clipboard > Go to clipboard
3. Change entity type to whichever type you've added record to clipboard
4. Click Export
5. In this Clipboard export page, all users are given the option to change Type to Accession

### Expected Behavior

Only users with permissions to view Accessions should be given the option to choose Accession
### Context and Notes

Related to [Clipboard export page cannot change type to Accessions #2221](Clipboard export page cannot change type to Accessions #2221)

### Version used

AtoM 2.10.0

danloveg:
Closes #2221 
Closes #2222 

These issues are closely related so I addressed them both in this PR.

There is some JS that watches for changes in the Type selector on the clipboard export page, and "accession" was not one of the expected types, so it always fell back to "informationObject". This is why it wasn't possible to select the Accession option.

And now I'm only showing the Accession option in that dropdown when the user is an editor or admin.

### DIFF
--- a/apps/qubit/modules/clipboard/actions/exportAction.class.php
+++ b/apps/qubit/modules/clipboard/actions/exportAction.class.php
@@ -238,8 +238,11 @@ class ClipboardExportAction extends DefaultEditAction
             'informationObject' => sfConfig::get('app_ui_label_informationobject'),
             'actor' => sfConfig::get('app_ui_label_actor'),
             'repository' => sfConfig::get('app_ui_label_repository'),
-            'accession' => sfConfig::get('app_ui_label_accession'),
         ];
+
+        if ($this->context->user->hasCredential(['editor', 'administrator'], false)) {
+            $this->typeChoices['accession'] = sfConfig::get('app_ui_label_accession');
+        }
 
         $this->form->getValidatorSchema()->setOption('allow_extra_fields', true);
     }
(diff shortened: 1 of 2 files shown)


danloveg:
Looks like a network issue cause the Electron integration tests to fail. I don't have permission to retry the job but I'd expect it to succeed if it was retried

anvit (approved):
Looks good! Thanks @danloveg

PR #2262, opened 2025-12-22
Fix js files in root directory ignored by prettier

### LINKED ISSUE #2234: Problem: JS files in root /js/ are ignored by Prettier
### Current Behavior

1. Run the following command: `npm run check-format -- --loglevel debug`
2. Note the only files checked by prettier are `webpack.config.js` and Javascript + Sass files in `plugins/arDominionB5Plugin`. None of the files in the root `/js/` directory are picked up by prettier.

If you go back to a previous release, e.g., 2.9, and run this same command you'll see all the files in `arDominionB5Plugin/js/` get picked up, but it looks like those were moved to the root of the repository when BS2 themes were deprecated. The prettier ignore file wasn't updated to reflect the files being moved.

### Expected Behavior

Javascript files in the root `/js/` directory should be checked when running Prettier.

My reasoning is that since `plugins/arDominionB5Plugin/js/main.js` imports the code from the root `/js/` directory, those files should be checked for formatting.

### Possible Solution

Add this line to `.prettierignore`:

```
!/js
```

And fix Prettier issues in `/js/`.

### Context and Notes

The [syntax-checks](https://github.com/artefactual/atom/actions/workflows/syntax-checks.yml) workflow is allowing non-formatted Javascript code in `/js/`.

### Version used

AtoM 2.10 (latest qa/2.x)

Docker (`php:8.3-fpm-alpine`)

en

### PHP version

PHP 8.3

### Contact details

_No response_

Eashana0104:
Closes https://github.com/artefactual/atom/issues/2234

Change notes-

1. Add !/js to `.prettierignore` file so it does not ignore files in root js directory.
2. Run `prettier --write .` command to fix all the formatting errors in `/js/`


js/aclDialog.js
```diff
@@ -1,60 +1,69 @@
...
+      this.label +
+      "</div>" +
       '<div class="bd">' +
-        '<form action="" method="post" style="border: none"></form>' +
-      '</div>' +
-    '</div>' +
-  '</div>');
+      '<form action="" method="post" style="border: none"></form>' +
+      "</div>" +
+      "</div>" +
```
anvit: @danloveg I think prettier's auto fix might have removed the indentation here. Is that something you have the capacity to fix? It's a minor nitpick so it's not an issue if you can't but fixing the indentation would make this easier to read.

danloveg: Hi @anvit I agree that the indentation on those lines were necessary to understand the HTML. Since I can't make any changes to this branch, what I could do is open a new pull request to address this issue. Does that work or does it make sense to accept this PR and then open a new issue to address that?

anvit: @danloveg never mind! I was able to use my maintainer privileges to push the fix here, and figured that would be the easiest way to do it!

danloveg: That works!

melaniekung (approved):
LGTM, thanks @Eashana0104  and @anvit !

PR #2266, opened 2026-01-07
Fix updated_at for desc pub status updates (#2259)

### LINKED ISSUE #2259: Problem: Updating publicationStatus for top-level record and its descendants does not update the `updated_at` column in the object table for child items
### Current Behavior

When updating publicationStatus for a top-level item, including its descendants, only the top-level item's `updated_at` entry is updated. 

### Expected Behavior

When updating publicationStatus for a top-level item, including its descendants, the `updated_at` entry for all items updated should be updated.

### Context and Notes

This was brought to our attention with an OAI harvesting issue where updating the publicationStatus for a top-level record, including its descendants, when trying to harvest metadata, the child level items are not outputted in the results.

melaniekung:
Update `updated_at` entry in object table for descendants when updating publication status including descendants.


anvit (approved):
Looks good, thanks @melaniekung !

PR #2267, opened 2026-01-07
Set differing msgs for pub/auth users (#2257)

### LINKED ISSUE #2257: Problem: Draft record disclosed through different access denial messages
Currently, public users trying to access a draft record get's a "Sorry, you do not have permission to access that page" which confirms the page exists and is likely in draft.

If feasible, authenticated users without permission can continue to get the "Sorry, you do not have permission to access that page", but for anonymous/public users should get the generic "Page does not exist" error. Otherwise, all users should get the "Page does not exist" error page.

melaniekung:


anvit (approved):
Thanks @melaniekung

PR #2273, opened 2026-01-16
Improve nested set build performance

### LINKED ISSUE #2272: Performance issue: Nested set takes a long time to build for many information objects
For context, I manage an AtoM database with 3.3 million information objects. I have been noticing when importing new information objects, that re-building the nested set takes a very long time to complete. I've measured the nested set build taking upwards of twenty minutes.

For small imports of new records, this is an annoying length of time to have to wait for an import to complete.

# Potential Solution

I've been looking into the code that rebuilds the nested set and have identified an area that could be optimized to reduce the nested set build time.

In `lib/task/propel/propelBuildNestedSetTask.class.php`, there is a SQL update call for every single information_object in the database that looks something like this:

```sql
UPDATE information_object SET lft=?, rgt=? WHERE id=?;
```

This results in 3.3 million separate SQL statements in my database which is a performance issue.

My proposed solution is, instead of executing an update for every object, execute a batched update for multiple information objects at a time. For a batch size of three information objects, that would look something like this:

```sql
UPDATE
    information_object
SET
    lft = CASE id WHEN ? THEN ? WHEN ? THEN ? WHEN ? THEN ?,
    rgt = CASE id WHEN ? THEN ? WHEN ? THEN ? WHEN ? THEN ?
WHERE
    id IN (?, ?, ?);
```

The `WHEN ?` selects for the given information object by id. If the case matches, the lft value in `THEN ?` is that object's new lft value. The same thing happens with `rgt`.

# Testing

I tested out whether this would have an actual performance impact on my local copy of AtoM. My local instance has **656,000** information objects.

For each run, I cleared the cache and then ran the command `php symfony propel:build-nested --exclude-tables=term,menu`.

Over three runs, this is how long the pre-optimized nested set build took:

1. 00:01:19.503 (1 minute, 19 seconds)
2. 00:01:26.504
3. 00:01:21.157

The average nested set build time run was 82.388 seconds.

And in my optimized version, I played around with the batch size and found the sweet spot for performance to be a batch size of **64**. This is how long the nested set build took in my updated version:

1. 00:00:09.069 (0 minutes, 9 seconds)
2. 00:00:09.248
3. 00:00:09.412

The average build time was 9.243 seconds.

# Conclusion

In my local environment, **the optimized nested set build speed is 8.9x faster**! This is a direct result of reducing the number of SQL statements by a factor of 64. This change should make the nested set build faster for everyone, not just administrators who manage huge million-record databases.

I'll open a pull request outlining my proposed solution.

danloveg:
Closes #2272 

Reduce the execution time of `propel:build-nested-set` by around 90% by applying a couple of performance optimizations:

- Batch-update lft and rgt values instead of writing them one-by-one
- Use `[]` to create child-parent array instead of `array_push()`

### DIFF
--- a/lib/task/propel/propelBuildNestedSetTask.class.php
+++ b/lib/task/propel/propelBuildNestedSetTask.class.php
@@ -179,38 +179,27 @@ EOF;
             return;
         }
 
-        $lftParams = [];
-        $rgtParams = [];
-        $idParams = [];
+        $lftCases = [];
+        $rgtCases = [];
+        $ids = [];
 
         foreach ($this->pendingUpdates as $node) {
-            // WHEN id THEN lft
-            $lftParams[] = $node['id'];
-            $lftParams[] = $node['lft'];
-
-            // WHEN id THEN rgt
-            $rgtParams[] = $node['id'];
-            $rgtParams[] = $node['rgt'];
-
-            $idParams[] = $node['id'];
+            $lftCases[] = sprintf('WHEN %d THEN %d', $node['id'], $node['lft']);
+            $rgtCases[] = sprintf('WHEN %d THEN %d', $node['id'], $node['rgt']);
+            $ids[] = $node['id'];
         }
-
-        $numUpdates = count($this->pendingUpdates);
-
-        $casePlaceholders = implode(' ', array_fill(0, $numUpdates, 'WHEN ? THEN ?'));
-        $idPlaceholders = implode(',', array_fill(0, $numUpdates, '?'));
 
         $sql = sprintf(
             'UPDATE %s SET %s = CASE id %s END, %s = CASE id %s END WHERE id IN (%s);',
             $classname::TABLE_NAME,
             $classname::LFT,
-            $casePlaceholders,  // <- $lftParams maps to these
+            implode(' ', $lftCases),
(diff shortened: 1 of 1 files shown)


lib/task/propel/propelBuildNestedSetTask.class.php
```diff
@@ -153,22 +155,55 @@ protected function recursivelyUpdateTree($node, $classname)
...
+        $rgtCases = [];
+        $ids = [];
+
+        foreach ($this->pendingUpdates as $node) {
+            $lftCases[] = sprintf('WHEN %d THEN %d', $node['id'], $node['lft']);
+            $rgtCases[] = sprintf('WHEN %d THEN %d', $node['id'], $node['rgt']);
+            $ids[] = $node['id'];
+        }
+
+        $sql = sprintf(
```
anvit, line 192: Neat! Batching these up as a single big query should be a big improvement

anvit (approved):
Thanks @danloveg !

lib/task/propel/propelBuildNestedSetTask.class.php
danloveg: For sure, it was a huge performance improvement from my testing!

PR #2274, opened 2026-01-19
Add physical object type and location to box label reports

### LINKED ISSUE #2241: Enhancement: Include container types and locations in Box Label reports
Box label reports include the **physical object name** but are missing **container types** and **locations**. I have heard feedback that it might be useful to include these two pieces of information in this report as well. They are currently available in the Physical Storage report. Further discussion regarding the use cases for these two reports, and if/why they should include different pieces of information, might be beneficial.

danloveg:
Closes #2241 

Adds the physical storage type and physical storage location to generated box label reports. Additionally, I've added a hyperlink to the physical storage object to match the physical storage location report.

Before my changes:

before-change

After my changes:

after-change


anvit (approved):
Looks great! Thanks @danloveg !

PR #2279, opened 2026-01-27
Use getLabel when displaying a physical object

### LINKED ISSUE #2240: Enhancement: Always display container names together with container types
### Issue

When creating containers, it’s possible to assign the same name/number to different types of boxes (e.g., Folder 664 and Hollinger Box 664), even if they are stored in separate locations. Throughout AtoM, container names are not always listed together with their container types. This makes it difficult to select the correct the select container when linking physical storage or generating reports. 

### Enhancement suggestion

Always display the container names together with container types to disambiguate. 

For example:

1. Navigate to an archival description
2. Click More > Link physical storage
3. Under "Add container links," open the drop down menu
4. Container names are listed **without** their containers types. If you have 3 containers named 664, they will simply appear in the drop down list as:

- 664
- 664 
- 644

It might be more useful if they appeared as:

- Folder: 664
- Hollinger box: 664
- Box: 664

Etc.

ammopt:
Closes #2240 

The name composition logic already exists and resides in QubitPhysicalObject->getLabel, utilize that instead of falling back to __toString.

### DIFF
--- a/apps/qubit/modules/physicalobject/templates/autocompleteSuccess.php
+++ b/apps/qubit/modules/physicalobject/templates/autocompleteSuccess.php
@@ -3,7 +3,7 @@
     <?php foreach ($physicalObjects as $item) { ?>
       <tr class="<?php echo 0 == @++$row % 2 ? 'even' : 'odd'; ?>">
         <td>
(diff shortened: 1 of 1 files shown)


anvit (approved):
This change looks good to me! Thanks so much for contributing this fix @ammopt !

PR #2286, opened 2026-02-02
Replace CLI import/export progress display

sbreker:
- Remove per-row dot output. renderProgressDescription() now writes a single update line to STDERR every ~5s.
- Print a final summary to STDOUT.

### DIFF
--- a/lib/task/import/importBulkTask.class.php
+++ b/lib/task/import/importBulkTask.class.php
@@ -127,17 +127,11 @@ EOF;
                 rename($file, $move_destination);
             }
 
-            if (!$options['verbose']) {
-                echo '.';
-            }
-
             if ($importer->hasErrors()) {
                 foreach ($importer->getErrors() as $message) {
                     $this->log('('.$file.'): '.$message);
                 }
             }
-
-            echo '.';
 
             // Try to free up memory
             unset($importer);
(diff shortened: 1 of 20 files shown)


lib/task/import/importBulkTask.class.php
```diff
@@ -176,7 +170,16 @@ protected function execute($arguments = [], $options = [])
             QubitSearch::getInstance()->optimize();
         }
 
-        $this->log("\nImported ".$count.' XML/CSV files in '.$timer->elapsed().' s. '.memory_get_peak_usage().' bytes used.');
+        if (!empty($options['verbose'])) {
+            $elapsed = $timer->elapsed();
+            $peakMb = round(memory_get_peak_usage() / 1048576, 2);
```
anvit: Can we use a constant instead 1048576? Or use 1024 * 1024 for easier comprehension and consistency like QubitFlatFileImport does?

anvit (commented):
Thanks @sbreker ! I had a minor nitpick but looks good otherwise!

anvit:
Also wanted to note that the docs will need some updates. I noticed at a quick glance that `items-until-update` has been documented in our [CLI tools docs](https://www.accesstomemory.org/en/docs/2.10/admin-manual/maintenance/cli-tools/)

lib/task/import/importBulkTask.class.php
sbreker: Hi @anvit! I have added a commit that addresses this item and updates a couple return statements.

sbreker:
> Also wanted to note that the docs will need some updates. I noticed at a quick glance that `items-until-update` has been documented in our [CLI tools docs](https://www.accesstomemory.org/en/docs/2.10/admin-manual/maintenance/cli-tools/)

Hi @anvit - Yes I dropped this option as it seemed unnecessary in light of changing to periodic updates. Let me know if you agree with this or would like a different behaviour. I can update the atom-docs if we agree this should be removed.

anvit (approved):
Thanks @sbreker , this looks good! And yeah, I think it would be good to get rid of the `items-until-update` option

PR #2287, opened 2026-02-03
Bump composer dependencies

anvit:
Updating composer dependencies. This will also address the update for phpunit to v9.6.33+ which resolves a dependabot alert


melaniekung (approved):
Thanks @anvit

PR #2290, opened 2026-02-06
Fail when CSV export file cannot be opened

### LINKED ISSUE #2289: Problem: CSV exports silently fail when path is not writable
### Current Behavior

This is difficult to reproduce using the Docker container since the user is root and has access to write to any file. But for example, on a real AtoM host where the AtoM user is `nginx` or `www-data` this can be reproduced by: 

1. Create a directory *not* owned by the AtoM user (e.g., with your own user account). Make its permissions `755` so that only the owner can write to that directory.
2. Run a `csv:export` command targeting the directory or a file inside that directory.
3. AtoM will look like it's running the export, but when it finishes, there will be no file.

Some of the export tasks do check that the path is writable before trying to write to it, but this check is not applied across the board. For example, the `csv:export` task does not check whether a path can be written to.

### Expected Behavior

An error should be returned so that the user knows that they can't create an export at the chosen location.

### Possible Solution

Check that the file is opened successfully when opening it for the first time, here:

https://github.com/artefactual/atom/blob/78444afa60e9df2426a325d0e7c3011934d3d3f6/lib/flatfile/QubitFlatfileExport.class.php#L531

This will return `false` when the file can't be opened, but there is no check here to validate whether the file was opened or not.

Putting the check here will ensure that checking that a path can be written to happens by default. Currently, a developer needs to remember to add a check that the path can be written before starting an export; this solution takes having to remember that step out of the equation.

### Context and Notes

I can open a PR for this issue.

### Version used

AtoM 2.10

Docker (`php:8.3-fpm-alpine`)

en

### PHP version

PHP 8.3

### Contact details

_No response_

danloveg:
Closes #2289 

Checks that the CSV export file is opened successfully, and throws an exception if not. This will ensure that the output file handle is valid before trying to write to it.

### DIFF
--- a/lib/flatfile/QubitFlatfileExport.class.php
+++ b/lib/flatfile/QubitFlatfileExport.class.php
@@ -529,6 +529,10 @@ class QubitFlatfileExport
     {
         if (!isset($this->currentFileHandle)) {
             $this->currentFileHandle = fopen($filePath, 'a');
+
+            if (false === $this->currentFileHandle) {
+                throw new sfException("Failed to open the file at the path '{$filePath}'");
+            }
         }
 
         fputcsv($this->currentFileHandle, $row);


anvit (approved):
Thanks @danloveg !

PR #2304, opened 2026-02-25
Swap identifier and title in output

### LINKED ISSUE #2303: Problem: identifier and title are in wrong order in update output
### Current Behavior

1. Run a `match-and-update` or `delete-and-replace` in the latest version of AtoM with the `csv:import` tool.
2. Note the output that looks like this:

```
Row 30: Matching description found, updating in place; row (identifier: No title, title: 01a-c000004-d0019-001, culture: en)
```

The identifier and the title should be swapped.

### Expected Behavior

In the above output, I would expect to see "identifier: 01a-c000004-d0019-001", and "title: No title"

### Possible Solution

Swap the order in the `sprintf()` calls where these messages are constructed.

### Context and Notes

_No response_

### Version used

AtoM 2.10.1 (latest qa/2.x)

`php:8.3-fpm-alpine` (Docker)

en

### PHP version

8.3

### Contact details

_No response_

danloveg:
Closes #2303

### DIFF
--- a/lib/QubitFlatfileImport.class.php
+++ b/lib/QubitFlatfileImport.class.php
@@ -1863,8 +1863,8 @@ class QubitFlatfileImport
                 if (1 < count($ids)) {
                     $msg = sprintf(
                         'Multiple matching descriptions found for row (identifier: %s, title: %s, culture: %s). Skipping record.',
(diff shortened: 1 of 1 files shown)


anvit (approved):
Ah good catch! Thanks @danloveg

PR #2307, opened 2026-03-02
Fix language variable in various cultures

### LINKED ISSUE #2012: Problem: Untranslated string in certain languages in the translation interface
### Current Behavior

1. Change language to French (on an instance where sf_culture is not French)
2. Open the Translations Interface
3. Notice that the right column says `%Langue% de traduction`, instead of `Françaises de traduction`

### Expected Behavior

`%Langue%` should be properly translated

### Notes

Also noticed in pt, pt_BR language

danloveg:
Closes #2012

Fixes the mistakenly translated language variable back to "%language%" for all languages.

As noted in the issue, `fr` and `pt_BR` were two cultures where this was an issue, but there were a few more as well.

### DIFF
--- a/plugins/sfTranslatePlugin/i18n/fr/messages.xml
+++ b/plugins/sfTranslatePlugin/i18n/fr/messages.xml
@@ -18,7 +18,7 @@
       </trans-unit>
       <trans-unit id="4">
         <source>%language% translation</source>
-        <target>%Langue% de traduction</target>
+        <target>%language% de traduction</target>
       </trans-unit>
       <trans-unit id="5">
         <source>Save translation</source>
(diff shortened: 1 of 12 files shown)


anvit:
@danloveg these strings come from weblate, and I think we'd created this task as a flag for ourselves to review weblate translations for these kinds of errors. I've just gone and fixed these instances in weblate translations so they will automatically be fixed by translations pull

danloveg:
> @danloveg these strings come from weblate, and I think we'd created this task as a flag for ourselves to review weblate translations for these kinds of errors. I've just gone and fixed these instances in weblate translations so they will automatically be fixed by translations pull

That makes sense, didn't realize these came from elsewhere!

PR #2313, opened 2026-03-10
Fix JS challenge endpoint exceptions (#2312)

### LINKED ISSUE #2312: Built-in JS challenge ignores endpoint exceptions when query parameters are present
This issue affects AtoM’s built-in JavaScript challenge (configured via `config/appChallenge.yml` and implemented in `lib/challenge/filter.php`). It does not involve any external reverse-proxy challenge. [appChallenge.yml.tmpl#L25](https://github.com/artefactual/atom/blob/qa/2.x/config/appChallenge.yml.tmpl#L25)

When the built-in challenge is enabled and endpoint exceptions are configured such as:

```
endpoint_exceptions:
  - /api
  - /qtSwordPlugin
  - /;oai
```

Requests to the OAI-PMH endpoint are still challenged whenever they include query parameters, for example:

* `/;oai?verb=Identify`
* `/;oai?verb=ListRecords&metadataPrefix=oai_dc`

Instead of bypassing the challenge and returning the expected OAI-PMH response, AtoM redirects the client to `/challenge`.This issue affects AtoM’s built-in JavaScript challenge (configured via config/appChallenge.yml and implemented in lib/challenge/filter.php). It does not involve any external reverse-proxy challenge. [appChallenge.yml.tmpl#L25](https://github.com/artefactual/atom/blob/qa/2.x/config/appChallenge.yml.tmpl#L25)
When the built-in challenge is enabled and endpoint exceptions are configured such as:

endpoint_exceptions:
  - /api
  - /qtSwordPlugin
  - /;oai

Requests to the OAI-PMH endpoint are still challenged whenever they include query parameters, for example:

        
      /;oai?verb=Identify

        
      /;oai?verb=ListRecords&metadataPrefix=oai_dc

Instead of bypassing the challenge and returning the expected OAI-PMH response, AtoM redirects the client to /challenge.

## **Impact**

The OAI-PMH endpoint is primarily accessed by harvesters and automated clients. These systems cannot complete the built-in browser challenge, causing harvesting to fail entirely. This breaks interoperability with aggregators and external discovery systems.

---

### **Actual vs. Expected Behavior**

* **Expected:** When `/;oai` is listed in exceptions, all requests targeting that path should bypass the challenge, including query strings (`?verb=...`).
* **Actual:** Only the exact path `/;oai` bypasses. Any request containing query parameters triggers the challenge.

### **Root Cause**

In `lib/challenge/filter.php`, exception patterns are matched against `$_SERVER['REQUEST_URI']`. This value includes both the **path** and the **query string**. Because the exceptions are path-style entries, they fail to match when parameters are appended.

### **Scope**

This affects all path-based exceptions, including `/api` and `/qtSwordPlugin`, whenever query parameters are present.

---

## **Proposed Fix**

Modify the built-in challenge filter to match exceptions against the **URI path component only**:

1. Extract the path using `parse_url($_SERVER['REQUEST_URI'], PHP_URL_PATH)`.
2. Perform exception regex checks against that extracted path.

melaniekung:


sbreker (approved):
Looks good @melaniekung! Thanks for this!

PR #2318, opened 2026-03-17
Fix GA key to use strict string comparison (#2317)

### LINKED ISSUE #2317: Problem: GA key not loaded from sfConfig when setting is unset or null
### Current Behavior

1. Add `google_analytics_api_key` to `config/app.yml` or `apps/qubit/config/app.yml` [as documented](https://accesstomemory.org/en/docs/2.10/admin-manual/customization/config-files/#config-app-yml)
2. Go to Admin > Settings > Web Analytics
3. Notice the note from AtoM: `Google analytics is currently set in the app.yml.` However, if you dev tools inspect, you'll notice that the tag is missing and Google will say that there's no tag detected for the site.

### Expected Behavior

AtoM should be able to take a GA key from the Web Analytics GUI or from app.yml files and render it correctly.

### Context and Notes

When `QubitSetting::getByName('google_analytics')` returns null (or 0, false, '0') — i.e. the setting doesn't exist — the empty() check evaluated to true and skipped the reassignment from sfConfig. As a result, `$gaKey `stayed null and Google Analytics was never initialized, even when `app_google_analytics_api_key` was configured.

### Version used

AtoM 2.10

melaniekung:


anvit (approved):
Thanks @melaniekung

PR #2321, opened 2026-03-20
Improve slugify() safely and performance; Make slug truncation UTF‑8 safe

sbreker:
### Migrated slugify() tests to PHPUnit test group

Migrated old test/unit/slugifyTest.php to a new PHP Unit test and
enhanced to cover additional cases and sanity checks.

### Make slug truncation UTF‑8 safe before insert

- Use mb_substr(..., 'UTF-8') to cap slugs at 235 Unicode characters,
  falling back to substr() if mbstring is unavailable.
- Prevents splitting multibyte code points (e.g., Arabic) during truncation,
  which could lead to invalid UTF‑8 and failures/hangs at save time.
- Keeps existing uniqueness logic (numeric suffixing) intact while avoiding
  extended collision loops for very long non‑ASCII slugs.

### Improve slugify() performance in permissive mode using Unicode properties.

- Replace permissive-mode regex with a property-based pattern:
  [^\p{L}\p{N}-_:;,=*@~]+ (with /u)
  allowing Unicode letters/digits plus selected ASCII: - _ ~ : ; , = * @.
- Removes reliance on a very large explicit Unicode range, significantly
  reducing PCRE compile/match cost and memory usage during large CSV imports.
- Behavior remains aligned with unit tests; disallowed characters are still
  collapsed to dashes. Note: some Unicode punctuation formerly permitted by
  the broad range will now be replaced.

Here are common examples that were previously allowed but are now replaced with “-” in
permissive mode:
    
- Dash punctuation: en dash “–” (U+2013), em dash “—” (U+2014), minus sign “−” (U+2212),
  non‑breaking hyphen “‑” (U+2011).
- Curly quotes: “ ” (U+201C/U+201D), ‘ ’ (U+2018/U+2019).
- Ellipsis and bullets: “…” (U+2026), “•” (U+2022), middle dot “·” (U+00B7).
- Currency and symbols: “€” (U+20AC), “£” (U+00A3), “¥” (U+00A5), “©” (U+00A9), “®” (U+00AE),
  “™” (U+2122), “°” (U+00B0), “§” (U+00A7), “†” (U+2020).
- Mathematical operators: “±” (U+00B1), “×” (U+00D7), “÷” (U+00F7), plus many others.
- Arrows and other pictographs: “← → ↑ ↓” (U+2190–U+2193), etc.
- No‑break space: NBSP (U+00A0) and other non‑ASCII spaces.
- Zero‑width/format controls: ZWJ/​ZWNJ (U+200D/U+200C) and similar.
- Emoji and emoji components: e.g., “😀” (U+1F600) and related sequences.

What still passes unchanged:

- All letters and digits from any script (Arabic, Cyrillic, CJK, etc.).
- ASCII “safe” punctuation we explicitly allow: “- _ ~ : ; , = * @”.
- Regular ASCII spaces are turned into “-” (unchanged behavior).


anvit (commented):
@sbreker thanks! The change looks good to me, but I was wondering if there was a reason why we were replacing control flow, formatting, and line separator chars with ` ` instead of `-`?

sbreker:
> @sbreker thanks! The change looks good to me, but I was wondering if there was a reason why we were replacing control flow, formatting, and line separator chars with ` ` instead of `-`?

Since the last step is to convert unaccepted chars (including spaces) into dash/minus it sorts out in the end but I think what you are describing is more clear. I will add a commit that replaces ' ' with '-' in that spot.

sbreker:
@anvit I have squashed a couple of the commits together including a change to convert flow control chars etc to '-' instead of  space char. This is ready for review.

anvit (approved):
Thanks @sbreker

PR #2323, opened 2026-03-23
Fix EAD XML export (#2322)

### LINKED ISSUE #2322: Problem: EAD XML Export fails due to unescaped special characters
### Current Behavior

1. Create a repository where the identifier has special characters (ex. "T&T")
2. Create a description which is part of the new repository
3. Try to export EAD 2002 XML, the EAD export fails and returns invalid XML output:
<?xml version="1.0"?>

### Expected Behavior

Special characters should be escaped properly so EAD XML exports do not fail

### Possible Solution

**Further investigation showed the generated XML contains:**

`mainagencycode="UN L&A"`

The `&` character is not escaped, which results in invalid XML.

melaniekung:
- escape special characters for $mainAgencyCode


anvit (approved):
Looks good! Thanks @melaniekung

PR #2324, opened 2026-03-25
Pin trivy version in SBOM generation workflow

replaceafill:


sbreker (approved):
Thanks @replaceafill!

PR #2328, opened 2026-03-30
Restrict file types for imports. (#2327)

### LINKED ISSUE #2327: File import input does not restrict accepted file types
### Description:
File import input field accepts any file type, which can lead to users uploading unsupported formats, and AtoM saves the import file candidate in AtoM's uploads folder when the import fails.

### Expected behavior:
The file picker should only show/accept supported formats

melaniekung:


anvit (approved):
Thanks @melaniekung

PR #2329, opened 2026-04-01
Enhance JS Challenge endpoint exception matching

sbreker:
Match endpoint exceptions against the request path instead of the full REQUEST_URI, so configured exceptions like /;oai still match when the request includes query parameters.

Add regression tests for /;oai, /api, and /qtSwordPlugin cases.


mamedin (approved):
LGTM!

anvit (approved):
Looks great! Thanks for the refactor and adding the tests @sbreker !

PR #2331, opened 2026-04-01
Add option to highlight search result matches

danloveg:
Adds an option to highlight search results for information objects. I've been working with researchers using AtoM recently and they mentioned that being able to see why certain descriptions are returned in the search results at a glance would be a useful feature. This PR adds subtle highlighting to search results to show what text matched the search query. If the title, scope and content, or creator is matched on, then those highlights are added inline to the existing rendered fields. If other fields that are not typically included in the search results are matched on, they are added in a new block. See here:

highlight

Elasticsearch already has this capability, it was just a matter of turning highlighting on and rendering its results. It returns matches wrapped in `<mark></mark>` tags.

Since administrators might not want this active, you can disable it in the settings. I've made it enabled by default.

setting

I understand this feature probably won't make it for 2.11 but thought I'd open this pull request in case this would be a useful feature for others, as well.

### DIFF
--- a/apps/qubit/modules/informationobject/actions/browseAction.class.php
+++ b/apps/qubit/modules/informationobject/actions/browseAction.class.php
@@ -244,9 +244,14 @@ class InformationObjectBrowseAction extends DefaultBrowseAction
 
         $this->setView($request);
 
+        // Determine whether search highlighting is required
+        $hasQuery = 1 !== preg_match('/^[\s\t\r\n]*$/', $request->query) || 1 !== preg_match('/^[\s\t\r\n]*$/', $request->sq0);
+        $highlightSetting = QubitSetting::getByName('highlight_search_results');
+        $highlightValue = null === $highlightSetting ? 0 : intval($highlightSetting->getValue());
+        $highlightEnabled = 1 === $highlightValue;
+
         // Add search term highlighting when any search criteria are present
-        if (1 !== preg_match('/^[\s\t\r\n]*$/', $request->query)
-            || 1 !== preg_match('/^[\s\t\r\n]*$/', $request->sq0)) {
+        if ($hasQuery && $highlightEnabled) {
             $this->search->query->setHighlight([
                 'pre_tags' => ['<mark>'],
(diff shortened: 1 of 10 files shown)


anvit:
Hi @danloveg , we'd been testing this feature out, and it looks great! We had a couple of minor things we noticed during testing that we wanted to flag:

- Currently the contrast ratio for the highlight colour against the background colour is too low. I would recommend picking a darker/more saturated shade. You could use [Web AIM's contrast checker](https://webaim.org/resources/contrastchecker/) to test a chosen colour and confirm that it meets the accessibility standards.
-  When a language is picked to narrow down search results, the language code itself shows up as a match. Not sure if this is something that's easy to fix since I haven't dived into this PR yet. 
image

danloveg:
> Hi @danloveg , we'd been testing this feature out, and it looks great! We had a couple of minor things we noticed during testing that we wanted to flag:
> 
>  * Currently the contrast ratio for the highlight colour against the background colour is too low. I would recommend picking a darker/more saturated shade. You could use [Web AIM's contrast checker](https://webaim.org/resources/contrastchecker/) to test a chosen colour and confirm that it meets the accessibility standards.
> 
>  * When a language is picked to narrow down search results, the language code itself shows up as a match. Not sure if this is something that's easy to fix since I haven't dived into this PR yet.

Thanks for taking a look! I've upped the contrast, there's no warning about contrast issues in my browser's accessibility tab any more. This is what I came up with for the new styling.

improved contrast

And regarding the language, thanks for flagging that. I am ignoring `i18n.languages` as a highlighted field which fixes the language being shown in the highlighting, and there was some extra culture checking I needed to do to handle multi-lingual descriptions. It appears to be working correctly now.

JBosse-Artefactual:
@danloveg Looks great!! @anvit This looks good to me!!

danloveg:
FYI, I added some extra context to the search highlighting results because I realized useful information is hidden about other highlights that are not shown. I added a bit of text that shows how many other fields the search term matched that might not be shown:

extra-highlighting-context

PR #2335, opened 2026-04-06
Clean up tmp files after import. (#2334)

### LINKED ISSUE #2334: Clean up temporary files after import (xml/csv)
On import, a file is created in `atom/src/uploads/tmp`. Files should be deleted/cleaned up after import, whether the import failed or succeeded.

melaniekung:
- this will delete all tmp files for most jobs


anvit (approved):
Thanks @melaniekung

PR #2351, opened 2026-05-11
Add Image carousel user interface label. (#2337)

### LINKED ISSUE #2337: Enhancement: Add "Image Carousel" as a customizable string to user interface label settings
## Problem Summary

Currently, when lower levels of an archival unit include digital objects, the DO carousel will be shown - there is an example in the public demo site [here](https://demo.accesstomemory.org/kantokoski-koski-koivula-korpela-family-3). 

This carousel currently comes with a hard-coded section header label that says "Image Carousel." However: 

* This is inconsistent with the fact that elsewhere we regularly use "Digital object" as a more generic label
* Digital object is a customizable string in the user interface label settings, but this label is not
* Not all DOs in the carousel are necessarily images

## Proposed enhancement

This issue proposes that "Image carousel" be added as its own user interface label setting. That way people can label it as appropriate for their content. 

## Alternatives considered

Using "Digital object carousel" as the default, and then making the "Digital object" part of that linked to the user interface label settings would be an acceptable, if slightly less customizable, enhancement. 

Simply relabeling it as "Digital object carousel" would address some of the original inconsistencies but be far less useful, as it could not be changed. This would also add to the problems outlined in recently-filed issue #2336 however, and is therefore not recommended. The same is true for any other hard-coded labels.

melaniekung:
WIP - must be merged after arMigration198!!!

### DIFF
--- a/apps/qubit/modules/digitalobject/templates/_imageflow.php
+++ b/apps/qubit/modules/digitalobject/templates/_imageflow.php
@@ -10,7 +10,7 @@
   <div class="accordion-item border-0">
     <h2 class="accordion-header rounded-0 rounded-top border border-bottom-0" id="heading-carousel">
       <button class="accordion-button rounded-0 rounded-top text-primary" type="button" data-bs-toggle="collapse" data-bs-target="#collapse-carousel" aria-expanded="true" aria-controls="collapse-carousel">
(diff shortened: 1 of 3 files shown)


anvit (dismissed):
Thanks @melaniekung !

PR #2352, opened 2026-05-13
Speed up install fixture loading

sevein:
I've been finding the installer a bit slow in my local development environment, likely made more noticeable by shared-volume filesystem overhead. While profiling `tools:purge --demo`, I found a couple of small fixture-loading fixes that significantly reduce the time spent loading the initial data during install/purge.

Measured `tools:purge --demo` improving from about 54s before this optimization pass to about 10s after it. This is more noticeable on slower shared filesystems, but it still avoids repeated i18n work on faster environments too.


sbreker (approved):
Thanks for this performance improvement @sevein  :+1:

PR #2354, opened 2026-05-15
Fix REST API digital object downloads

sevein:
Return the digital object download marker from the action and stream the file contents directly. The previous code passed readfile() into setContent(), which mixed immediate output with Symfony response content handling and stored the byte count returned by readfile() as response content.

Mark the Symfony response as header-only after readfile() so the final rendering filter does not send response content after the file body has already been emitted.


sbreker (approved):
Thanks for this fix @sevein 👍

PR #2356, opened 2026-05-15
Fix digital object load and summary messages (#2345)

### LINKED ISSUE #2345: Problem: log of digitalobject:load does not count imported records
### Current Behavior

1. Go to shell and start `php symfony digitalobject:load import_file.csv` (valid form, records pointing to existing JPG files)
2. Watch the log: the count in front of "remaining" stays on max value

### Expected Behavior

The log should count the remaining number of records down in each line according to progress.

### Possible Solution

https://github.com/artefactual/atom/blob/648b36612c057ec4d4d4bcedaa935491f6517353/lib/task/digitalobject/digitalObjectLoadTask.class.php#L146

- move the local variable to $this->

### Context and Notes

Annoying information lag while doing loads of lot images.
Changed it in my local installation after verifying there the bug.

### Version used

AtoM 2.10.1

Ubuntu 26.04

de

### PHP version

PHP 8.4

### Contact details

hans-arno@mielsch.de

ShrishtyJainTech:


lib/task/digitalobject/digitalObjectLoadTask.class.php
```diff
@@ -374,7 +374,7 @@ protected function addDigitalObject($objectId, $path, $options = [])
 
         $remainingImportCount = $this->totalObjCount - $this->skippedCount - $importedCount;
```
anvit: Remove `$importedCount` here since it's always going to be zero and is unused.

lib/task/digitalobject/digitalObjectLoadTask.class.php
```diff
@@ -228,7 +228,7 @@ public function execute($arguments = [], $options = [])
             Qubit::clearClassCaches();
         }
 
-        $this->logSection('digital-object', 'Successfully Loaded '.self::$count.' digital objects.');
+        $this->logSection('digital-object', 'Successfully Loaded '.self::$count.' digital objects and skipped '.$this->skippedCount.'.');
```
anvit: I missed the last bit of the discussion in the morning, but I thought we'd talked about displaying the total count as well? So something along the lines of `Successfully loaded 10 out of 15 digital objects, and skipped 5`? Feel free to ignore this if you and Jay decided that we don't need to display that.

anvit (changes requested):

lib/task/digitalobject/digitalObjectLoadTask.class.php
ShrishtyJainTech: Updated as suggested. Please take another look when convenient.

lib/task/digitalobject/digitalObjectLoadTask.class.php
ShrishtyJainTech: Updated!!

anvit (approved):

PR #2360, opened 2026-05-28
Fix analysis config path with settings key (#2359)

### LINKED ISSUE #2359: Problem: Populating index fails with diacritics
### Current Behavior

1. Go to Admin > Settings > Diacritics
2. Enable and upload diacritics mapping
3. Populate search index from CLI
4. Note the error: `unknown key [analysis] for create index`

### Expected Behavior

Search index should be populated without error

melaniekung:
- Fix diacritic index population by adding missing ['settings'] key to analysis config path for ES7 compatibility


anvit (approved):
Thanks @melaniekung !

PR #2361, opened 2026-05-28
Limit XML clipboard exports

danloveg:
Sets a limit for XML exports from the clipboard. See the screenshot below for the alert shown when this limit is exceeded. This applies only to information object object XML exports from the clipboard (XML exports from the backend are not limited).

xml_export_screenshot

### DIFF
--- a/apps/qubit/modules/clipboard/actions/exportAction.class.php
+++ b/apps/qubit/modules/clipboard/actions/exportAction.class.php
@@ -19,6 +19,9 @@
 
 class ClipboardExportAction extends DefaultEditAction
 {
+    // Limit how many XML information object a user can export
+    private const XML_EXPORT_LIMIT_DESCENDANTS = 1000;
+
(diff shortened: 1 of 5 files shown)


anvit:
@danloveg I took a look through the PR and it mostly looks good! We were wondering if you'd be able to make export limit a variable that could be configured through the settings (and access it via `sfConfig::get`) rather than just making it a private variable so that it's possible to customize/change it.

danloveg:
> @danloveg I took a look through the PR and it mostly looks good! We were wondering if you'd be able to make export limit a variable that could be configured through the settings (and access it via `sfConfig::get`) rather than just making it a private variable so that it's possible to customize/change it.

Hi @anvit, thanks for taking a look. That makes sense to me, I've added a new setting that can be accessed from the "Clipboard" section of the settings. The default value is 1000

clipboard-setting

anvit (approved):

PR #2363, opened 2026-06-09
Add docker configuration for keycloak

melaniekung:


docker/bootstrap.php
```diff
@@ -113,7 +115,7 @@ function get_host_and_port($value, $default_port)
   read_only: false
   htmlpurifier_enabled: false
   csp:
-    response_header: Content-Security-Policy
+    response_header: Content-Security-Policy-Report-Only
```
sbreker, line 118: This change should perhaps be in its own commit as it seems unrelated to adding Keycloak to the dev Docker env.

jraddaoui: If you want to avoid using `-Report-Only`, you should be able to add the Keycloak host to the CSP header/meta `connect-src`. It must be the host without path, so it works with all redirects.

plugins/arOidcPlugin/modules/oidc/actions/loginAction.class.php
```diff
@@ -34,8 +34,8 @@ public function execute($request)
             $this->context->user->setAttribute('atom-login-referer', $request->getReferer());
         }
 
-        if ($request->isMethod('post') || isset($_REQUEST['code'])) {
-            if (null !== $providerId = $this->context->user->parseProviderIdFromUrl($this->context->user->getAttribute('atom-login-referer', null))) {
+        if (($request->isMethod('post') || isset($_REQUEST['code'])) && null !== $this->context->user->getAttribute('atom-login-referer')) {
+            if (null !== $providerId = $this->context->user->parseProviderIdFromUrl($this->context->user->getAttribute('atom-login-referer'))) {
```
sbreker, line 38: Why are the checks duplicated?

docker/bootstrap.php
```diff
@@ -124,8 +126,64 @@ function get_host_and_port($value, $default_port)
...
+        authorization_endpoint: '{$keycloakBaseUrl}/realms/artefactual/protocol/openid-connect/auth'
+        token_endpoint: '{$keycloakBaseUrl}/realms/artefactual/protocol/openid-connect/token'
+        userinfo_endpoint: '{$keycloakBaseUrl}/realms/artefactual/protocol/openid-connect/userinfo'
+        jwks_uri: '{$keycloakBaseUrl}/realms/artefactual/protocol/openid-connect/certs'
+
+        send_oidc_logout: true
+
+        enable_refresh_token_use: true
+
+        server_cert: 'docker/certs/cert.pem'
```
sbreker, line 153: This will require generating certs and adding them to the certs folder

docker/etc/environment
```diff
@@ -20,3 +20,5 @@ ATOM_MYSQL_DSN=mysql:host=percona;port=3306;dbname=atom;charset=utf8mb4
 ATOM_MYSQL_USERNAME=atom
 ATOM_MYSQL_PASSWORD=atom_12345
 NODE_ENV=development
+ATOM_KEYCLOAK_HOST=host.docker.internal
+ATOM_KEYCLOAK_PORT=9000
```
sbreker, line 24: The OIDC plugin will need activation. [This commit](https://github.com/artefactual/atom/commit/15308df2d86258092aeff99aece8a18beda66571) allows the plugin to be activated using an env var.

sbreker:
Will a dev still be able to select local logins for authenticate vs using OIDC/Keycloak if their testing requires?

plugins/arOidcPlugin/lib/oidcUser.class.php
```diff
@@ -455,7 +455,22 @@ protected function setOidcProviderDetails(string $providerId = ''): bool
...
+                        $endpointOverrides[$endpointName] = $provider[$endpointName];
+                    }
+                }
+
+                if (!empty($endpointOverrides)) {
+                    $this->oidcClient->providerConfigParam($endpointOverrides);
+                }
+            } else {
+                $this->oidcClient->setProviderUrl($provider['url']);
+            }
```
sbreker, line 473: I noticed a new OIDC param 'issuer' - what is this for?

PR #2375, opened 2026-06-27
Use Imagick methods instead of mogrify() for repo theme images

sbreker:
Refactored class arRepositoryThemeCropValidatedFile to use Imagick methods instead of mogrify command.

Replace the previous crop-only behavior with resize-and-crop processing for repository logo and banner uploads. Images are now scaled to cover the target dimensions while preserving aspect ratio, then center-cropped to the final logo or banner size. This keeps uploaded assets small after processing and avoids squashing non-square source images. Also increased the allowed upload size from 256K to 500K to give users more room to upload larger source images before resize and crop.

Set Imagick memory, map, and pixel-area limits before reading repository logo and banner uploads.

### DIFF
--- a/lib/form/arRepositoryThemeCropValidatedFile.class.php
+++ b/lib/form/arRepositoryThemeCropValidatedFile.class.php
@@ -29,6 +29,9 @@ class arRepositoryThemeCropValidatedFile extends sfValidatedFile
     public const LOGO_MAX_HEIGHT = 270;
     public const BANNER_MAX_WIDTH = 800;
     public const BANNER_MAX_HEIGHT = 300;
+    public const IMAGICK_MEMORY_LIMIT_MB = 64;
+    public const IMAGICK_MAP_LIMIT_MB = 64;
+    public const IMAGICK_AREA_LIMIT_PIXELS = 16000000;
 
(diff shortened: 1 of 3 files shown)


lib/form/arRepositoryThemeCropValidatedFile.class.php
```diff
@@ -29,48 +29,113 @@ class arRepositoryThemeCropValidatedFile extends sfValidatedFile
...
+            'resizeHeight' => $resizeHeight,
+            'cropX' => max(0, (int) floor(($resizeWidth - $targetWidth) / 2)),
+            'cropY' => max(0, (int) floor(($resizeHeight - $targetHeight) / 2)),
+        ];
+    }
+
+    protected function shouldCropImages()
+    {
+        return extension_loaded('imagick');
+    }
```
anvit, line 100: Instead of defining this method, could we just re-use `QubitDigitalObject::imagickExtensionLoaded` since that is a static method?

sbreker: Good idea - I have updated this to use the static method :+1:

anvit (approved):

PR #2378, opened 2026-06-29
Update slugify() default permissive set

sbreker:


sevein (commented):
LGTM

lib/model/QubitSlug.php
```diff
@@ -103,7 +103,7 @@ public static function slugify($slug, $creationType = null)
             case QubitSlug::SLUG_PERMISSIVE:
                 if ($isAscii) {
                     // ASCII allowlist: letters/digits and safe ASCII - _ ~ : ; , = * @; others become "-"
```
sevein: This is still listing `;` as allowed!

lib/model/QubitSlug.php
sevein: This can be removed.

sevein, line 119: Not related to this pull request, but I've found this issue with `gpt-5.6-sol max` and reproduced locally:

> These regular expressions use the `u` modifier, so PHP expects the input to be valid UTF-8. With malformed input, such as `"A\xFFB"`, `preg_replace()` returns `null`. The next replacement receives that `null`, emits a deprecation on PHP 8.3, and the slug ultimately becomes empty. Could we handle invalid UTF-8 at the start of `slugify()` and add a regression test?

lib/model/QubitSlug.php
sbreker: Fixed

lib/model/QubitSlug.php
sbreker: Done

sbreker: I have added a commit to this PR that addresses this. I will keep it as a separate commit and not squash it at merge time.

anvit (approved):
Looks good!

PR #2380, opened 2026-06-30
Refactor watermark generation to use Imagick

sbreker:


anvit (approved):
It might worth logging the exception, but the rest looks good

PR #2381, opened 2026-07-01
Refactor CSV Import command

sbreker:


lib/QubitCsvImport.class.php
```diff
@@ -110,7 +110,7 @@ public function import($csvFile, $type = null, $csvOrigFileName = null)
                 case 'delete-and-replace':
                     $commandUpdate = ('match-and-update' == $this->updateType) ? '--update="match-and-update"' : '--update="delete-and-replace"';
                     $commandSkipUnmatched = ($this->skipUnmatched) ? '--skip-unmatched' : '';
-                    $commandLimit = ('' !== $this->limit) ? "--limit=\"{$this->limit}\"" : '';
+                    $commandLimit = ('' !== $this->limit) ? sprintf('--limit=%s', $this->limit) : '';
```
anvit: Do we not need these string arguments to be in quotes?

sbreker: You are correct. I missed adding escapeshellargs() here which returns a quoted string. See latest commit which addresses this.

anvit (approved):

PR #2384, opened 2026-07-06
Refactor HTTP_X_ATOM_CULTURE handling

sbreker:
Move HTTP_X_ATOM_CULTURE handling from myUser initialization into a dedicated filter that runs after settings are loaded.

Add PHPUnit coverage for the culture allowlist logic in the new filter.


anvit (approved):
Looks good! Thanks @sbreker

PR #2385, opened 2026-07-07
Add digital object type params and sf_format check

anvit:
Add a check for digital object type params and sf_format and assume safe defaults in case on invalid values.

### DIFF
--- a/apps/qubit/modules/informationobject/actions/browseAction.class.php
+++ b/apps/qubit/modules/informationobject/actions/browseAction.class.php
@@ -139,10 +139,11 @@ class InformationObjectBrowseAction extends DefaultBrowseAction
                 'image' => QubitTerm::IMAGE_ID,
                 'text' => QubitTerm::TEXT_ID,
                 'video' => QubitTerm::VIDEO_ID,
+                'other' => QubitTerm::OTHER_ID,
             ];
 
             if (false == array_search($this->getParameters['mediatypes'], $validMediaTypes)) {
-                $this->getParameters['mediatypes'] = QubitTerm::IMAGE_ID;
+                $this->getParameters['mediatypes'] = QubitTerm::OTHER_ID;
             }
         }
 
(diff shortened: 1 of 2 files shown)


sbreker (approved):
Looks great @anvit! 👍

apps/qubit/modules/informationobject/actions/browseAction.class.php
```diff
@@ -133,6 +133,19 @@ public function execute($request)
             $this->collection = QubitInformationObject::getById($this->getParameters['collection']);
         }
 
+        if (isset($this->getParameters['mediatypes'])) {
+            $validMediaTypes = [
+                'audio' => QubitTerm::AUDIO_ID,
+                'image' => QubitTerm::IMAGE_ID,
+                'text' => QubitTerm::TEXT_ID,
+                'video' => QubitTerm::VIDEO_ID,
+            ];
```
sevein, line 143: Could we add `'other' => QubitTerm::OTHER_ID` to this list? When a user selects "Other" the request sends `mediatypes=139`. This list does not include 139, so the code changes it to Image and shows the wrong results, right?

vendor/symfony/lib/request/sfWebRequest.class.php
```diff
@@ -740,6 +740,11 @@ public function setFormat($format, $mimeTypes)
    */
   public function setRequestFormat($format)
   {
+    $validFormats = ['xml', 'ead', 'dc', 'csv', 'html', 'json'];
+
+    if (false === array_search($format, $validFormats)) {
+        $format = 'html';
+    }
```
sevein: This is blocking formats already registered in `$this->formats`. For example, `txt` is registered by Symfony, but this code changes it to `html`.  Is this intentional? Maybe we can also accept the registered formats and use `html` only for values that are not known?

apps/qubit/modules/informationobject/actions/browseAction.class.php
anvit: Ah, good catch! Defaulting to `Other` does make more sense than defaulting to text. Done!

vendor/symfony/lib/request/sfWebRequest.class.php
anvit: Updated!

PR #2386, opened 2026-07-09
Add shared redirect target validator

sbreker:
Use the configured site base URL to validate redirect targets.

Apply the filter to user login, OIDC login return, actor edit, and physical object edit/delete and cancel-link flows.


anvit (approved):
Looks good! Thanks @sbreker

PR #2387, opened 2026-07-10
Remove bullet list from thematic areas in table view

### LINKED ISSUE #2371: Problem: Archival Institution browse in table view Thematic area should not be lists
### Current Behavior

1. Go to Browse > Archival Institutions > Table view
2. Note that the Thematic area looks wonky with the list bullets

### Expected Behavior

Thematic area should look like all other columns without list bullets

ShrishtyJainTech:


apps/qubit/modules/repository/templates/_browseTableView.php
```diff
@@ -72,7 +72,7 @@
         <td>
           <?php if (isset($doc['thematicAreas'])) { ?>
             <?php foreach ($doc['thematicAreas'] as $areaTerm) { ?>
-              <li><?php echo render_value_inline(QubitTerm::getById($areaTerm)); ?></li>
+             <?php echo render_value_inline(QubitTerm::getById($areaTerm)); ?><br>
```
anvit: We should use a block element (such as `<p>` or `<div>`) here. Using `<br />` for styling purposes should always be avoided

anvit (changes requested):

apps/qubit/modules/repository/templates/_browseTableView.php
ShrishtyJainTech: @anvit  I've made the changes.

anvit (approved):
Looks good @ShrishtyJainTech !

PR #2389, opened 2026-07-15
Update XML import parser configuration

sbreker:
Update XML import and OAI parsing to disable external entity resolution.

Add a small DOMDocument subclass for parser metadata fields used by AtoM, avoiding dynamic property deprecations on newer PHP versions. Add focused parser tests for entity handling.


anvit (approved):
Looks good! Thanks @sbreker

PR #2390, opened 2026-07-15
Restrict physical object API creation

sbreker:
Align REST API physical object creation with existing UI permissions. Users require appropriate ACL role before creating physical objects through the API.


sevein (commented):
LGTM

anvit (approved):

PR #2393, opened 2026-07-16
Normalize stored serialized data handling

sbreker:
Add a shared helper for reading serialized application values and update runtime call sites to use it with explicit defaults.

### DIFF
--- a/apps/qubit/modules/informationobject/actions/inventoryAction.class.php
+++ b/apps/qubit/modules/informationobject/actions/inventoryAction.class.php
@@ -89,7 +89,7 @@ class InformationObjectInventoryAction extends DefaultBrowseAction
         }
 
         $setting = QubitSetting::getByName('inventory_levels');
-        if (null === $setting || false === $value = unserialize($setting->getValue())) {
+        if (null === $setting || false === $value = Qubit::safeUnserialize($setting->getValue(), false)) {
             return;
         }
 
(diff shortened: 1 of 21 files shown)


sevein (commented):
`gpt-5.6-sol max` is still reporting one issue:

> containsObject() needs cycle detection. PHP preserves serialized references, so the valid payload a:1:{i:0;R:1;} produces a self-referential array. The recursive call then exhausts memory and terminates the request.

Regression test:

```php
  public function testSafeUnserializeRejectsCircularArrays()
  {
      $serialized = 'a:1:{i:0;R:1;}';

      $this->assertSame(
          [],
          Qubit::safeUnserialize($serialized, [])
      );
  }
```

Full report here: https://gist.github.com/sevein/0aa4b44e21ec352f9cbdac5faf073c29.

sbreker:
> `gpt-5.6-sol max` is still reporting one issue:
> 
> > containsObject() needs cycle detection. PHP preserves serialized references, so the valid payload a:1:{i:0;R:1;} produces a self-referential array. The recursive call then exhausts memory and terminates the request.
> 
> Regression test:
> 
> ```
>   public function testSafeUnserializeRejectsCircularArrays()
>   {
>       $serialized = 'a:1:{i:0;R:1;}';
> 
>       $this->assertSame(
>           [],
>           Qubit::safeUnserialize($serialized, [])
>       );
>   }
> ```
> 
> Full report here: https://gist.github.com/sevein/0aa4b44e21ec352f9cbdac5faf073c29.

Good catch! I have added a new commit that addresses this.

anvit:
@sbreker this has a merge conflict due to the same test updated in #2408 . Let us know if you have the capacity to rebase and fix this, but if not, we can rebase and merge this change

PR #2395, opened 2026-07-21
Remove unused digital object module action

sbreker:


sevein (commented):
Oopsie! Looks like this has been unused since at least the 2012 SVN import. The form has always posted to digitalobject/edit! LGTM.

PR #2396, opened 2026-07-22
Fix delete language and script notes field (#2392)

### LINKED ISSUE #2392: Problem: Emptying the textbox for Language and script notes does not save.
### Current Behavior

1. Go to an archival description
2. Click Edit
3. Open "Conditions of access and use area"
4. Remove text from "Language and script notes"
5. Click Save
The text originally in the "Language and script notes" box will still be there. I don't see any errors.

### Expected Behavior

The text in "Language and script notes" should be gone.

### Possible Solution

I wondered whether #2061 might be related?

### Context and Notes

_No response_

### Version used

Atom 2.10.1

Ubuntu 24.04

en

### PHP version

PHP 8.3

### Contact details

_No response_

melaniekung:


plugins/sfIsadPlugin/lib/sfIsadPlugin.class.php
```diff
@@ -80,11 +80,9 @@ public function __set($name, $value)
                     $note = $notes->offsetGet(0);
                 }
 
-                if (0 == strlen($value)) {
+                if (0 == strlen($value) || (!$missingNote && is_countable($note))) {
                     // Delete note if it's available
-                    if (!$missingNote && is_countable($note)) {
-                        $note->delete();
-                    }
+                    $note->delete();
```
sevein: I was testing this block from a new test file ([`test/phpunit/sfIsadPlugin/sfIsadPluginTest.php`](https://gist.github.com/sevein/312c983195e64fd4a13c2cffd3e7682d) - feel free to use it!) and the last case listed in the table fails with this change, i.e. when the submitted value is empty and no note exists, `$note` contains the empty `ArrayObject`, so the call to the delete method produces:

    Call to undefined method ArrayObject::delete()

Here's an alternative that passes in all four test cases. `$note` is consistently either a `QubitNote` or `null`, which I find a little easier to read:

```php
case 'languageNotes':
    $notes = $this->resource->getMemoryNotesByType(['noteTypeId' => QubitTerm::LANGUAGE_NOTE_ID]);
    $note = $notes[0] ?? null;

    // Clear the stored note when the field is empty; otherwise create the
    // note if necessary and update its content.
    if (0 === strlen($value)) {
        if (null !== $note) {
            $note->delete();
        }

        break;
    }

    if (null === $note) {
        $note = new QubitNote();
        $note->typeId = QubitTerm::LANGUAGE_NOTE_ID;
        $note->userId = sfContext::getInstance()
            ->user
            ->getAttribute('user_id');

        $this->resource->notes[] = $note;
    }

    $note->content = $value;

    return $this;
```

melaniekung: thank you @sevein!

anvit (approved):

PR #2397, opened 2026-07-23
Fix worker context leak in CSV exports

anvit:
CSV export jobs were creating a new qubit/prod Symfony context inside the long-running worker.

That leaked prod into later jobs and made arUpdateEsIoDocumentsJob enqueue descendant updates asynchronously instead of handling them inline.

Keep the worker context intact and scope only the export culture changes needed for translated CSV output.

### DIFF
--- a/lib/job/arAccessionExportJob.class.php
+++ b/lib/job/arAccessionExportJob.class.php
@@ -105,8 +105,7 @@ class arAccessionExportJob extends arExportJob
      */
     protected function csvActionExport($resource, $writer)
     {
-        $configuration = ProjectConfiguration::getApplicationConfiguration('qubit', 'prod', false);
-        $this->context = sfContext::createInstance($configuration);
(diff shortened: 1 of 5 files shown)


sbreker (approved):
Looks good @anvit 👍 

A couple easy unit tests could exercise withUserCulture(): assert that withUserCulture() restores culture after success and after an exception

anvit:
> Looks good @anvit 👍
> 
> A couple easy unit tests could exercise withUserCulture(): assert that withUserCulture() restores culture after success and after an exception

I just added a couple of test cases

sbreker:
> > Looks good @anvit 👍
> > A couple easy unit tests could exercise withUserCulture(): assert that withUserCulture() restores culture after success and after an exception
> 
> I just added a couple of test cases

Looks great @anvit! Thanks!

PR #2398, opened 2026-07-24
Improve generated salt entropy

sbreker:


sevein (commented):
I would suggest a small description in the commit, e.g.:

> Generate salts from 16 cryptographically secure random bytes and encode them as hexadecimal to preserve the existing 32-character storage format. The salted SHA-1 pre-hash remains so migrated and newly created credentials share the same verification path, requiring no migration for existing users.

anvit (approved):
Looks good @sbreker

PR #2403, opened 2026-08-06
Allow zero month dates in CSV validator

### LINKED ISSUE #2402: Problem: CSV Validation fails when event dates have `00` for month or date.
### Current Behavior

1. Create a CSV where eventStartDate/eventEndDate is in the format of `YYYY-00-00`
2. Run CSV Validation for this CSV
3. Note that it fails CSV Validation with "Invalid event dates"

### Expected Behavior

If AtoM allows the format of `YYYY-00-00` to be manually added to event start/end dates via the GUI, CSV Validation should accept this type of format as well.

danloveg:
Closes #2402

Add YYYY-00-00, YYYY-00, and YYYY-MM-00 as valid patterns to the event date validator.

### DIFF
--- a/.github/workflows/integration-tests.yml
+++ b/.github/workflows/integration-tests.yml
@@ -48,6 +48,8 @@ jobs:
         key: npm-${{ hashFiles('package-lock.json') }}
     - name: Install NPM dependencies
       run: sudo npm install -g npm && npm ci
+    - name: Install Cypress binary
+      run: npx cypress install
     - name: Modify Gearman config
       run: |
         echo -e "all:\n  servers:\n    default: 127.0.0.1:4730" \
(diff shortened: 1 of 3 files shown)


anvit:
Hi @danloveg , thanks so much for this fix! This was unfortunately not captured in the original issue, but AtoM also treats YYYY-MM-00 as valid. I think the change looks good but would you be able to update the regex pattern to also account for this format? 

Thanks again!

danloveg:
> Hi @danloveg , thanks so much for this fix! This was unfortunately not captured in the original issue, but AtoM also treats YYYY-MM-00 as valid. I think the change looks good but would you be able to update the regex pattern to also account for this format?
> 
> Thanks again!

No problem, I've added YYYY-MM-00 as a valid pattern to these changes.

anvit (approved):
Thanks @danloveg !

PR #2407, opened 2026-08-07
Add check to autocomplete actions

anvit:
Check if user is permitted to perform for the following autocomplete actions:
- user
- informationobject
- function
- repository
- term
- taxonomy


melaniekung (approved):
LGTM, thanks @anvit !

PR #2408, opened 2026-08-10
Use tempnam() to create temp files

sbreker:
Use tempnam() when saving temporary files. Adds logic to sanitize extensions.


anvit (approved):
Looks good!

PR #2415, opened 2026-08-19
Align finding aid action permissions

sbreker:
Keep endpoint behavior consistent with the finding aid UI controls and existing upload handling.


anvit (approved):
Thanks @sbreker

PR #2416, opened 2026-08-21
Fix error in tools:unlink-creators

### LINKED ISSUE #2414: Problem: Unlinking creators can fail sometimes in AtoM 2.9+
### Current Behavior

In AtoM 2.9 and higher, which support PHP 8.3 (recommended in the install documentation, at least), the [tools:unlink-creators](https://www.accesstomemory.org/en/docs/2.9/admin-manual/maintenance/cli-tools/#unlink-creators-from-child-descriptions-and-reapply-inheritance-to-hierarchy) can fail.

Lines 83 and 84 of [/plugins/arElasticSearchPlugin/lib/model/arElasticSearchModelBase.class.php](https://github.com/artefactual/atom/blob/qa/2.x/plugins/arElasticSearchPlugin/lib/model/arElasticSearchModelBase.class.php) read as follows:

```
 if (!in_array($row['culture'], self::$allowedLanguages)) {
                    continue;
                }
```

In PHP 7.x, if self::$allowedLanguages was null (which it apparently can be, though I don't know exactly why), in_array would bail out of the function call and return **NULL**, which was a falsy value. In PHP 8.0+, if self::$allowedLanguages is null, in_array returns with a type error, which halts execution.

### Expected Behavior

The tools:unlink-creators task should run and unlink creators.

### Possible Solution

I think we can fix this by coalescing self::$allowedLanguages to an empty array, which is semantically similar to passing NULL, and is also a falsy value.

```
 if (!in_array($row['culture'], self::$allowedLanguages ?? [] )) {
                    continue;
                }
```

### Context and Notes

This issue isn't fixed in 2.10.x either.

### Version used

AtoM 2.9.1

Ubuntu 24.04

en

### PHP version

PHP 8.3.6

### Contact details

_No response_

danloveg:
Closes #2414 

The `tools:unlink-creators` command was failing because the `app_i18n_languages` setting was not set. This is a setting that comes from the database, and must explicitly be loaded when a CLI task runs. Switching the inherited class from `sfBaseTask` to `arBaseTask` and calling that class's `execute()` function loads the settings and fixes the bug.

### DIFF
--- a/lib/task/tools/unlinkCreatorTask.class.php
+++ b/lib/task/tools/unlinkCreatorTask.class.php
@@ -22,7 +22,7 @@
  *
  * @author     Steve Breker <sbreker@artefactual.com>
  */
-class unlinkCreatorTask extends sfBaseTask
+class unlinkCreatorTask extends arBaseTask
 {
     protected $actor;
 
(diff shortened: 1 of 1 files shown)


anvit (approved):
Looks good! Thanks @danloveg

PR #2429, opened 2026-09-10
Convert non-BS & slick deps to @use in main.scss

### LINKED ISSUE #2427: Problem: Custom themes extending arDominionB5Plugin missing Bootstrap CSS variables in 2.10.2
### Current Behavior

1. Create a custom theme extending arDominionB5Plugin
2. Run `rm -rf dist/ node_modules && npm i && npm run build` against a checkout after [Bump postcss from 8.5.13 to 8.5.26](https://github.com/artefactual/atom/commit/a9a45e450b0bc67358d98231963ec6f840537081)
3. Inspect the compiled CSS output for the custom theme (or the :root selector specifically)

### Expected Behavior

The full set of Bootstrap default CSS custom properties should be defined in `:root `(as inherited from `bootstrap/scss/_root.scss`), regardless of which theme — default or custom — is being compiled.

### Possible Solution

_No response_

### Context and Notes

Custom themes built on top of arDominionB5Plugin are compiling with many default Bootstrap CSS custom properties (`--bs-*`) missing from the generated `:root` block. We've bisected this to commit [Bump postcss from 8.5.13 to 8.5.26](https://github.com/artefactual/atom/commit/a9a45e450b0bc67358d98231963ec6f840537081), which bumped postcss from 8.5.13 to 8.5.26.

postcss is not a direct dependency of AtoM — it's pulled in transitively via bootstrap's build tooling.

### Version used

AtoM 2.10.2

_No response_

_No response_

### PHP version

_No response_

### Contact details

_No response_

melaniekung:
Migrate remaining Sass dependencies in main.scss to the module system
(@use/@forward), replacing legacy @import statements.

Bootstrap and Slick are excluded because their upstream repos have not
yet migrated to the Sass module system themselves. Until they ship @use/
@forward-compatible releases, their styles will continue to be pulled in via
@import.

### DIFF
--- a/plugins/arDominionB5Plugin/scss/main.scss
+++ b/plugins/arDominionB5Plugin/scss/main.scss
@@ -1,3 +1,14 @@
+// Font Awesome
+@use "@fortawesome/fontawesome-free/scss/fontawesome";
+@use "@fortawesome/fontawesome-free/scss/solid";
+
+// Other components (CSS files)
+@use "@uppy/core/dist/style" as *;
+@use "@uppy/dashboard/dist/style" as *;
+@use "jquery-ui-dist/jquery-ui";
+@use "jstree/dist/themes/default/style";
+@use "mediaelement/build/mediaelementplayer";
+
 // Default variable overrides
 @import "variables";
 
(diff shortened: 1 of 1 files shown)


anvit (approved):
Looks good! Thanks @melaniekung

PR #2432, opened 2026-09-21
Fix SVG support in the Imagick adapter

sevein:
SVG uploads are accepted, but then we get no thumbnail or reference image. Imagick reports image/svg+xml and the adapter only knows image/svg, so it rejects the file even when Imagick can read it. A bit confusing because from the upload side everything looks fine.

The adapter now accepts image/svg+xml too, so the derivatives can be created. There is also a small SVG fixture and a regression test for the JPEG thumbnail generation.


anvit (approved):
This looks all good to me @sevein ! Though it does make me wonder what the use case for svg upload in AtoM is

sevein:
> Though it does make me wonder what the use case for svg upload in AtoM is

Maybe a logo contest that an organisation is archiving? With the original SVG submissions. JPEG is perhaps not ideal for the preview, but that is what AtoM generates today.

